PR #7, opened 2023-04-06
Add middleware

gpsanant:
Added a simple little middleware that allows validators to sign off on on a msg to confirm it. A dispute can be launched that hashes the message 1000 times and if the message doesnt start with a threshold amount of zeros, then all validators that signed are slashed.

I also added a ECDSA registry.

Planning to comment a bit soon, but mostly should be self explanatory for reviewers.

This will help middlewares get a better idea of how to integrate as we realized we didn't have a service manager implementation in the repo and the BLS registry contains a bit of logic that is unneccesary for peeps not using BLS.

TODO:
- Add tests for ECDSA registry
- ~~Add Natspec and more inline comments~~

### DIFF
--- a/src/contracts/middleware/RegistryBase.sol
+++ b/src/contracts/middleware/RegistryBase.sol
@@ -71,7 +71,9 @@ abstract contract RegistryBase is VoteWeigherBase, IQuorumRegistry {
         uint8 _NUMBER_OF_QUORUMS
     ) VoteWeigherBase(_strategyManager, _serviceManager, _NUMBER_OF_QUORUMS)
     // solhint-disable-next-line no-empty-blocks
-    {}
+    {
+        require(_NUMBER_OF_QUORUMS <= 2, "RegistryBase.constructor: NUMBER_OF_QUORUMS must be less than or equal to 2");
+    }
 
     /**
      * @notice Adds empty first entries to the dynamic arrays `totalStakeHistory` and `totalOperatorsHistory`,
@@ -81,8 +83,7 @@ abstract contract RegistryBase is VoteWeigherBase, IQuorumRegistry {
      */
     function _initialize(
         uint256[] memory _quorumBips,
-        StrategyAndWeightingMultiplier[] memory _firstQuorumStrategiesConsideredAndMultipliers,
-        StrategyAndWeightingMultiplier[] memory _secondQuorumStrategiesConsideredAndMultipliers
+        StrategyAndWeightingMultiplier[][] memory _quorumStrategiesConsideredAndMultipliers
     ) internal virtual onlyInitializing {
         VoteWeigherBase._initialize(_quorumBips);
 
@@ -94,8 +95,13 @@ abstract contract RegistryBase is VoteWeigherBase, IQuorumRegistry {
         OperatorIndex memory _totalOperators;
         totalOperatorsHistory.push(_totalOperators);
 
-        _addStrategiesConsideredAndMultipliers(0, _firstQuorumStrategiesConsideredAndMultipliers);
-        _addStrategiesConsideredAndMultipliers(1, _secondQuorumStrategiesConsideredAndMultipliers);
+        // add the strategies and multipliers to each quorum
+        for (uint i = 0; i < _quorumStrategiesConsideredAndMultipliers.length;) {
+            _addStrategiesConsideredAndMultipliers(i, _quorumStrategiesConsideredAndMultipliers[i]);
+            unchecked {
+                i++;
+            }
+        }
     }
 
     /**
--- a/src/contracts/middleware/RegistryBase.sol
+++ b/src/contracts/middleware/RegistryBase.sol
@@ -72,7 +72,7 @@ abstract contract RegistryBase is VoteWeigherBase, IQuorumRegistry {
     ) VoteWeigherBase(_strategyManager, _serviceManager, _NUMBER_OF_QUORUMS)
     // solhint-disable-next-line no-empty-blocks
     {
-        require(_NUMBER_OF_QUORUMS <= 2, "RegistryBase.constructor: NUMBER_OF_QUORUMS must be less than or equal to 2");
+        require(_NUMBER_OF_QUORUMS <= 2, "RegistryBase: NUMBER_OF_QUORUMS must be less than or equal to 2");
     }
 
     /**
@@ -83,7 +83,8 @@ abstract contract RegistryBase is VoteWeigherBase, IQuorumRegistry {
      */
     function _initialize(
         uint256[] memory _quorumBips,
-        StrategyAndWeightingMultiplier[][] memory _quorumStrategiesConsideredAndMultipliers
+        StrategyAndWeightingMultiplier[] memory _firstQuorumStrategiesConsideredAndMultipliers,
+        StrategyAndWeightingMultiplier[] memory _secondQuorumStrategiesConsideredAndMultipliers
     ) internal virtual onlyInitializing {
         VoteWeigherBase._initialize(_quorumBips);
 
@@ -95,13 +96,8 @@ abstract contract RegistryBase is VoteWeigherBase, IQuorumRegistry {
         OperatorIndex memory _totalOperators;
         totalOperatorsHistory.push(_totalOperators);
 
-        // add the strategies and multipliers to each quorum
-        for (uint i = 0; i < _quorumStrategiesConsideredAndMultipliers.length;) {
-            _addStrategiesConsideredAndMultipliers(i, _quorumStrategiesConsideredAndMultipliers[i]);
-            unchecked {
-                i++;
-            }
-        }
+        _addStrategiesConsideredAndMultipliers(0, _firstQuorumStrategiesConsideredAndMultipliers);
+        _addStrategiesConsideredAndMultipliers(1, _secondQuorumStrategiesConsideredAndMultipliers);
     }
 
     /**
--- a/src/contracts/middleware/RegistryBase.sol
+++ b/src/contracts/middleware/RegistryBase.sol
@@ -72,7 +72,7 @@ abstract contract RegistryBase is VoteWeigherBase, IQuorumRegistry {
     ) VoteWeigherBase(_strategyManager, _serviceManager, _NUMBER_OF_QUORUMS)
     // solhint-disable-next-line no-empty-blocks
     {
-        require(_NUMBER_OF_QUORUMS <= 2, "RegistryBase: NUMBER_OF_QUORUMS must be less than or equal to 2");
+        require(_NUMBER_OF_QUORUMS <= 2 && _NUMBER_OF_QUORUMS > 0, "RegistryBase: NUMBER_OF_QUORUMS must be less than or equal to 2 and greater than 0");
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/contracts/middleware/RegistryBase.sol
```diff
@@ -71,7 +71,9 @@ abstract contract RegistryBase is VoteWeigherBase, IQuorumRegistry {
         uint8 _NUMBER_OF_QUORUMS
     ) VoteWeigherBase(_strategyManager, _serviceManager, _NUMBER_OF_QUORUMS)
     // solhint-disable-next-line no-empty-blocks
-    {}
+    {
+        require(_NUMBER_OF_QUORUMS <= 2, "RegistryBase: NUMBER_OF_QUORUMS must be less than or equal to 2");
```
ChaoticWalrus, line 75: if we're bothering with this, probably also want to check it's nonzero

src/contracts/middleware/example/ECDSARegistry.sol
```diff
@@ -0,0 +1,260 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
+
+import "../RegistryBase.sol";
+
+/**
+ * @title A Registry-type contract using aggregate BLS signatures.
```
ChaoticWalrus, line 7: incorrect comment. there are similarly copy-pasted revert messages which still refer to "BLSRegistry"

src/contracts/middleware/example/ECDSARegistry.sol
```diff
@@ -0,0 +1,260 @@
...
+
+    /**
+     * @notice Called by the whitelister, removes a list of operators to the whitelist
+     * @param operators the operators to remove from the whitelist
+     */
+    function removeFromWhitelist(address[] calldata operators) external onlyWhitelister {
+        for (uint i = 0; i < operators.length; i++) {
+            whitelisted[operators[i]] = false;
+        }
+    }
```
ChaoticWalrus, line 105: might be nice to abstract this away some of this copy-pasted stuff into a mix-in kind of contract

src/contracts/middleware/example/ECDSARegistry.sol
```diff
@@ -0,0 +1,260 @@
...
+
+    constructor(
+        IStrategyManager _strategyManager,
+        IServiceManager _serviceManager
+    )
+        RegistryBase(
+            _strategyManager,
+            _serviceManager,
+            1
```
ChaoticWalrus, line 51: note: this is where we are setting the number of quorums to `1`

src/contracts/middleware/example/ECDSARegistry.sol
```diff
@@ -0,0 +1,260 @@
...
+            unchecked {
+                ++i;
+            }
+        }
+
+        // update storage of total stake
+        _recordTotalStakeUpdate(_totalStake);
+    }
+
+    //TODO: The subgraph doesnt like uint256[4][] argument here. Figure this out laters
```
ChaoticWalrus, line 198: we should probably indeed sort this out. the copy-pasted part doesn't make sense for ECDSA signatures though. perhaps just delete this function for now?

src/contracts/middleware/example/HashThreshold.sol
```diff
@@ -0,0 +1,110 @@
...
+pragma solidity =0.8.12;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+
+import "../../interfaces/IServiceManager.sol";
+import "./ECDSARegistry.sol";
+
+/**
+ * @title An EigenLayer middleware example service manager that slashes validators that sign a message that, when hashed 1000 types starts with less than 5 0s.
```
ChaoticWalrus, line 11: please change 1000 hashes to something lighter-weight for testing. recommending 10x, can rename `kilohash` => `decahash`

src/contracts/middleware/example/HashThreshold.sol
```diff
@@ -0,0 +1,110 @@
...
+        }
+        (uint96 totalStake,) = registry.totalStake();
+        require(stakeSigned >= 666667 * uint256(totalStake) / 1000000, "Need more than 2/3 of stake to sign");
+
+        uint32 newLatestTime = uint32(block.number + disputePeriodBlocks);
+
+        certifiedMessageMetadatas[message] = CertifiedMessageMetadata({
+            validAfterBlock: newLatestTime,
+            signaturesHash: keccak256(signatures)
+        });
```
ChaoticWalrus, line 75: nothing in the code is checking the status of `certifiedMessageMetadatas[message]` before this, so someone can just submit the same signatures over-and-over for the same message. should probably change this, even if it's just meant as an example

ChaoticWalrus (changes requested):
Seems reasonable. The exact example middleware is not a good idea in practice, but I think it could be a helpful reference.
Would mostly like to see more quality on the comments; adding some test coverage sounds good as well (especially since tests can demo the product working / show a bit more how it would be used in practice).

src/contracts/middleware/example/HashThreshold.sol
```diff
@@ -0,0 +1,110 @@
...
+        require(msg.sender == tx.origin, "EOA must call this function");
+        uint128 stakeSigned = 0;
+        for(uint256 i = 0; i < signatures.length; i += 65) {
+            // we fetch all the signers and check their signatures and their stake
+            address signer = ECDSA.recover(message, signatures[i:i+65]);
+            require(registry.isActiveOperator(signer), "Signer is not an active operator");
+            stakeSigned += registry.firstQuorumStakedByOperator(signer);
+        }
+        (uint96 totalStake,) = registry.totalStake();
+        require(stakeSigned >= 666667 * uint256(totalStake) / 1000000, "Need more than 2/3 of stake to sign");
```
ChaoticWalrus, line 68: bit of magic number usage here. seems OK for demo/example tho 🙃

src/contracts/middleware/example/ECDSARegistry.sol
gpsanant: What do you think the best way to do this is? A new WhitelistableRegistryBase? sorta hard to move the `onlyServiceManagerOwner`s into their own contract without doing so

src/contracts/middleware/example/HashThreshold.sol
```diff
@@ -0,0 +1,110 @@
...
+        bytes32 hash = message;
+        for (uint256 i = 0; i < 1000; i++) {
+            hash = keccak256(abi.encodePacked(hash));
+        }
+        return hash;
+    }
+
+    function submitSignatures(bytes32 message, bytes calldata signatures) external {
+        // this makes it so that the signatures are viewable in calldata
+        require(msg.sender == tx.origin, "EOA must call this function");
```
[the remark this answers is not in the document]
gpsanant: stack trace is tough to get in my experience. everyone either emits events or puts it in calldata by an EOA

src/contracts/middleware/example/HashThreshold.sol
Sidu28: I recommend 69 hashes or septuagintahash

src/contracts/middleware/example/ECDSARegistry.sol
ChaoticWalrus: ah good point. that could make the inheritance tree (chain) pretty long then.
The mix-in could just define internal `_setWhitelister` and `_setWhitelistStatus` functions (we should probably have the latter, and add an event to it), and then the implementations could define external functions + attach auth to them? similar idea to what OZ does with their ERC20 design?

src/contracts/middleware/example/HashThreshold.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/7/commits/68f80a0c541b8ba06d37d8840a96fe9e4beee6e1 good catch

src/contracts/middleware/example/HashThreshold.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/7/commits/68f80a0c541b8ba06d37d8840a96fe9e4beee6e1

src/contracts/middleware/example/ECDSARegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/7/commits/68f80a0c541b8ba06d37d8840a96fe9e4beee6e1

src/contracts/middleware/example/ECDSARegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/7/commits/68f80a0c541b8ba06d37d8840a96fe9e4beee6e1

src/contracts/middleware/example/ECDSARegistry.sol
```diff
@@ -0,0 +1,260 @@
...
+        // make sure lengths are consistent
+        require(operatorsLength == prevElements.length, "BLSRegistry.updateStakes: prevElement is not the same length as operators");
+        // iterating over all the tuples that are to be updated
+        for (uint256 i = 0; i < operatorsLength;) {
+            // get operator's pubkeyHash
+            pubkeyHash = bytes32(uint256(uint160(operators[i])));
+            // fetch operator's existing stakes
+            currentStakes = pubkeyHashToStakeHistory[pubkeyHash][pubkeyHashToStakeHistory[pubkeyHash].length - 1];
+
+            // Note: we only edit the first quorum stake because this is a single quorum registry example
```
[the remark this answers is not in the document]
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/7/commits/68f80a0c541b8ba06d37d8840a96fe9e4beee6e1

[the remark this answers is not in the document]
gpsanant: added to top level contract comment

src/contracts/middleware/example/ECDSARegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/7/commits/68f80a0c541b8ba06d37d8840a96fe9e4beee6e1

src/contracts/middleware/RegistryBase.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/7/commits/68f80a0c541b8ba06d37d8840a96fe9e4beee6e1

src/contracts/middleware/example/HashThreshold.sol
gpsanant: ye sg :)

src/contracts/middleware/example/HashThreshold.sol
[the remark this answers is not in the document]
ChaoticWalrus: OK fair enough

Sidu28 (approved):

src/contracts/middleware/example/ECDSARegistry.sol
gpsanant: Will separate in another PR

ChaoticWalrus (approved):

PR #10, opened 2023-04-06
Restrict state mutability to pure to silence warnings.

adlerjohn:
Title. Silences the below warnings.

```console
warning[2018]: Warning: Function state mutability can be restricted to pure
  --> src/test/utils/Operators.sol:16:5:
   |
16 |     function operatorPrefix(uint256 index) public returns(string memory) {
   |     ^ (Relevant source part starts here and spans across multiple lines).



warning[2018]: Warning: Function state mutability can be restricted to pure
  --> src/test/utils/Owners.sol:16:5:
   |
16 |     function ownerPrefix(uint256 index) public returns(string memory) {
   |     ^ (Relevant source part starts here and spans across multiple lines).
```

### DIFF
--- a/src/test/utils/Operators.sol
+++ b/src/test/utils/Operators.sol
@@ -13,7 +13,7 @@ contract Operators is Test {
         operatorConfigJson = vm.readFile("./src/test/test-data/operators.json");
     }
 
-    function operatorPrefix(uint256 index) public returns(string memory) {
+    function operatorPrefix(uint256 index) public pure returns(string memory) {
         return string.concat(".operators[", string.concat(vm.toString(index), "]."));
     }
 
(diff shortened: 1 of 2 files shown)


ChaoticWalrus:
Thanks for the PR. I swear I tried changing this before but things broke when I made the change -- some kind of catch22 situation with foundry, I think. We'll see if the CI works with it now 🤞

ChaoticWalrus:
ahh, works locally for me but our CI is failing because [GitHub secrets are not passed to the runner when a workflow is triggered from a forked repository](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow). Going to look into this a touch more

ChaoticWalrus:
merging this since I've got offline confirmation that it works locally. we can sort out the GitHub actions bit later.

PR #11, opened 2023-04-06
switch from latestTime to latestServeUntilBlock

gpsanant:
We want to keep the updateBlock/latestServeUntil(block) logic consistent, so this switches the slasher to denominate entirely in blocks.


ChaoticWalrus (approved):
Changes all LGTM.
It's mostly semantics; think you covered everything that actually needs to change here.
I like the consistency (using blocks more widely over a mix of timestamps + blocks), and still agree that the upside of using blocks in worst-case scenarios (like a chain split event) outweighs the downside of less timing precision.

Issue #8, opened 2023-04-06
Add some sort of license to all files


adlerjohn:
A number of components are unlicensed, including, but not limited to, the below.

https://github.com/Layr-Labs/eigenlayer-contracts/blob/b1efe8b1ca6295ba124d6a46bb2cae531bdae3c7/src/contracts/libraries/BN254.sol#L20

https://github.com/Layr-Labs/eigenlayer-contracts/blob/b1efe8b1ca6295ba124d6a46bb2cae531bdae3c7/src/contracts/libraries/BeaconChainProofs.sol#L1

https://github.com/Layr-Labs/eigenlayer-contracts/blob/b1efe8b1ca6295ba124d6a46bb2cae531bdae3c7/src/contracts/libraries/Merkle.sol#L1

ChaoticWalrus:
thanks for noting these! seems like these mostly weren't caught because of spelling typos.

PR #12, opened 2023-04-07
correct licensing

ChaoticWalrus:
as noted in https://github.com/Layr-Labs/eigenlayer-contracts/issues/8, looks like we missed a few license notes/identifiers, probably mostly due to typos of 'UNLICENCED' instead of 'UNLICENSED'


gpsanant (approved):
:+1:

PR #15, opened 2023-04-11
Add new prover rule

ChaoticWalrus:
Inspired by SushiSwap router hack. Add a Prover rule that shows that we never transfer tokens out of an account other than the caller. (or conversely, if a user's token balance decreases as the result of a call to our contract(s), then that user must be the caller).
PR also comments out an old (failing) draft rule.


ChaoticWalrus:
seeing some time-outs specifically on the `completeQueuedWithdrawals` function, as well as a couple failures that appear to be not due to the preserved block not being properly observed(?). Deserves being investigated further but shouldn't be blocking for merging this.

PR #18, opened 2023-04-19
add beacon chain oracle initial write

gpsanant:
added an initial write of the beacon chain oracle

TODO:
- Discuss gas consumption (i.e. we are making 4 calls per root read here, should we "cache" roots in one of the contracts after read?)
- Unit tests


gpsanant:
closing due to redesign

PR #19, opened 2023-04-20
Added steth deployment address

Sidu28:
Adding the steth strategy contract address to the deployment output .


ChaoticWalrus (approved):
LGTM, merging now.

PR #23, opened 2023-04-24
BUGFIX: correct case-handling for new withdrawals following an operat…

ChaoticWalrus:
…or completely deregistering

This commit also adds a regression test (confirmed failing prior to fix + passing with fix) and a related test to verify that the fix is not overly strict (i.e. that it doesn't degrade our existing performance in other cases).

### DIFF
--- a/src/contracts/core/Slasher.sol
+++ b/src/contracts/core/Slasher.sol
@@ -275,6 +275,21 @@ contract Slasher is Initializable, OwnableUpgradeable, ISlasher, Pausable {
 
         // pull the MiddlewareTimes struct at the `middlewareTimesIndex`th position in `_operatorToMiddlewareTimes[operator]`
         MiddlewareTimes memory update = _operatorToMiddlewareTimes[operator][middlewareTimesIndex];
+
+        /**
+         * Case-handling for if the operator is not registered for any middlewares (i.e. they previously registered but are no longer registered for any),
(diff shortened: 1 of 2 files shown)


Sidu28 (approved):
lgtm

src/contracts/core/Slasher.sol
```diff
@@ -275,6 +275,21 @@ contract Slasher is Initializable, OwnableUpgradeable, ISlasher, Pausable {
 
         // pull the MiddlewareTimes struct at the `middlewareTimesIndex`th position in `_operatorToMiddlewareTimes[operator]`
         MiddlewareTimes memory update = _operatorToMiddlewareTimes[operator][middlewareTimesIndex];
+
+        /**
+         * Case-handling for if the operator is not registered for any middlewares (i.e. they previously registered but are no longer registered for any),
+         * AND the withdrawal was initiated after the 'stalestUpdateBlock' of the MiddlewareTimes struct specified by the provided `middlewareTimesIndex`.
+         * NOTE: we check the 2nd of these 2 conditions first for gas efficiency, to help avoid an extra SLOAD in all other cases.
+         */
+        if (withdrawalStartBlock >= update.stalestUpdateBlock && _operatorToWhitelistedContractsByUpdate[operator].size == 0) {
```
gpsanant, line 284: seems like this condition could be switched to save gas

gpsanant (approved):
LGTM

PR #27, opened 2023-05-03
Tvl limits

gpsanant:
Adds TVL limits to strategies and a limit on the number of pods deployed to EPM.

TODO:
- Write tests after @Sidu2's changes are merged
- Discuss how to handle numPods on testnet deployment


src/contracts/strategies/StrategyBaseTVLLimits.sol
```diff
@@ -0,0 +1,52 @@
...
+    constructor(IStrategyManager _strategyManager) StrategyBase(_strategyManager) {}
+
+    function initialize(uint256 _maxPerDeposit, uint256 _maxDeposits, IERC20 _underlyingToken, IPauserRegistry _pauserRegistry) public virtual initializer {
+        _setTVLLimits(_maxPerDeposit, _maxDeposits);
+        _initializeStrategyBase(_underlyingToken, _pauserRegistry);    }
+
+    /**
+     * @notice Sets the maximum deposits (in underlyingToken) that this strategy will hold and accept per deposit
+     * @param newMaxDeposits The new maximum deposits
+     * @dev Callable by the unpauser of this contract
```
ChaoticWalrus, line 25: comment is inaccurate. this is callable only by the pauser, not the unpauser

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -309,6 +312,9 @@ contract StrategyManagerUnitTests is Test {
     }
 
     function testDepositIntoStrategyWithSignatureSuccessfully(uint256 amount, uint256 expiry) public {
+        // min shares must be minted on strategy
+        cheats.assume(amount >= 1e9);
+
```
ChaoticWalrus, line 317: I'm pretty sure the reason we had StrategyWrapper in these tests instead of StrategyBase was to avoid doing this

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -1299,6 +1314,8 @@ contract StrategyManagerUnitTests is Test {
     }
 
     function testSlashSharesNotBeaconChainETHFuzzed(uint64 withdrawalAmount) external {
+        // cannot cause share value to increase too drastically
+        cheats.assume(withdrawalAmount <= 1e9 || withdrawalAmount == 1e18);
```
ChaoticWalrus, line 1318: this is some weird filtering -- does the test just make really unusual assumptions about the starting state of the storage?

src/contracts/strategies/StrategyBaseTVLLimits.sol
```diff
@@ -0,0 +1,52 @@
...
+
+import "./StrategyBase.sol";
+
+/**
+ * @title A Strategy implementation inheriting from `StrategyBase` that limits the total amount of deposits it will accept.
+ * @author Layr Labs, Inc.
+ */
+contract StrategyBaseTVLLimits is StrategyBase {
+    /// The maximum deposit (in underlyingToken) that this strategy will accept per deposit
+    uint256 public maxPerDeposit;
```
ChaoticWalrus, line 12: seems a bit like overkill to also have this, but we can discuss

src/contracts/strategies/StrategyWrapper.sol
```diff
@@ -1,152 +0,0 @@
-// SPDX-License-Identifier: BUSL-1.1
-pragma solidity =0.8.12;
```
ChaoticWalrus, line 2: any big reason for deleting this file? I could see situations in which it is preferable to use

src/contracts/pods/EigenPodManager.sol
```diff
@@ -54,6 +54,12 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
     /// @notice Pod owner to deployed EigenPod address
     mapping(address => IEigenPod) public ownerToPod;
 
+    /// @notice The number of pods that have been deployed
+    uint256 public numPods;
+
+    /// @notice The maximum number of pods that can be deployed
+    uint256 public maxPods;
+
```
ChaoticWalrus, line 62: this seems like the least impactful change. we should discuss if existing pausability is sufficient instead of this.

gpsanant: I agree with @ChaoticWalrus. We should discuss with @cpdcpd the efficacy of this limit as it has also be pointed out that users could create pods via the site (at a moderate price) and not even use them. Perhaps we should make the UI interaction a bit less available.

src/contracts/strategies/StrategyBaseTVLLimits.sol
gpsanant: A max per deposit makes it so that it is annoying for a whale to stake entirely to the maxDeposits for the strategy. This allows small stakers to get in by making it anoying for whales and eating into their costst with gas.

src/test/unit/StrategyBaseTVLLimitsUnit.sol
```diff
@@ -0,0 +1,114 @@
...
+        strategy.setTVLLimits(maxPerDeposit, maxDeposits);
+        cheats.stopPrank();
+
+        cheats.startPrank(address(strategyManager));
+        cheats.expectRevert(bytes("StrategyBaseTVLLimits: max per deposit exceeded"));
+        strategy.deposit(underlyingToken, amount);
+        cheats.stopPrank();
+    }
+
+    function testDepositMorthanMaxDeposits() public {
```
gpsanant, line 91: mispelled

src/test/unit/StrategyBaseTVLLimitsUnit.sol
```diff
@@ -0,0 +1,114 @@
...
+        cheats.startPrank(address(strategyManager));
+        for (uint256 i = 0; i < numDeposits-1; i++) {
+            underlyingToken.transfer(address(strategy), maxPerDeposit);
+            strategy.deposit(underlyingToken, maxPerDeposit);
+        }
+        underlyingToken.transfer(address(strategy), maxPerDeposit);
+        cheats.expectRevert(bytes("StrategyBaseTVLLimits: max deposits exceeded"));
+        strategy.deposit(underlyingToken, maxPerDeposit);
+        cheats.stopPrank();
+    }
```
gpsanant, line 111: prob we should add a test the a deposit below or equal to max per deposit works? and that a list of amounts less than max per deposit and sum to less than max deposits works?

gpsanant:
Note: This may not be correct on testnet since we were not keeping track of the amount since deployment. We have a couple options.

1. Allow new 1 time callable function to set the number of eigenpods deployed. Then we would pause the EPM, get the number of eigenpods, call the function with this number, then unpause. We would take this out of the code for mainnet.
2. Be fine with having incorrect pod count on testnet. the limit would be since upgrading, not since inception.

I prefer (1) to keep the environments as similar as possible, although it is more work.

src/test/unit/StrategyBaseTVLLimitsUnit.sol
Sidu28: fixed

src/contracts/strategies/StrategyBaseTVLLimits.sol
ChaoticWalrus: fixed

src/test/unit/StrategyManagerUnit.t.sol
ChaoticWalrus: fixed

src/test/unit/StrategyManagerUnit.t.sol
ChaoticWalrus: I still don't know what's happening here

src/contracts/strategies/StrategyBaseTVLLimits.sol
ChaoticWalrus: Agreed offline it's OK to add

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: Agreed offline it's OK to add

src/contracts/strategies/StrategyWrapper.sol
ChaoticWalrus: Agreed offline it's OK to delete

src/contracts/pods/EigenPodManager.sol
```diff
@@ -98,6 +106,8 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
      */
     function createPod() external {
         require(!hasPod(msg.sender), "EigenPodManager.createPod: Sender already has a pod");
+        require(numPods + 1 <= maxPods, "EigenPodManager.createPod: Max pods exceeded");
+        numPods++;
```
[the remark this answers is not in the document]
ChaoticWalrus: I fixed this but we should add test coverage

src/test/unit/StrategyManagerUnit.t.sol
gpsanant: thanks

src/contracts/pods/EigenPodManager.sol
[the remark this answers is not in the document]
gpsanant: agreed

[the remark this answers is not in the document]
ChaoticWalrus: coverage added in https://github.com/Layr-Labs/eigenlayer-contracts/pull/27/commits/04f0151ebf80e1dfaf9d7ad82927f0e6b4e44869
please confirm coverage is good

ChaoticWalrus:
> I prefer (1) to keep the environments as similar as possible, although it is more work.

I am OK with either approach, but the focus here should be on mainnet code. If we go with route (1) then it should be implemented on a separate branch after the mainnet code is complete.

ChaoticWalrus (approved):
This LGTM now. Got offline approval from @gpsanant on my changes that are part of this as well.

PR #32, opened 2023-05-10
update naming

ChaoticWalrus:
makes the following changes as appropriate, to reflect the updated governance architecture:
- 'communityMultisig' => 'executorMultisig' (for situations in which the executor multisig will now hold roles instead of the community multisig)
- add executorMultisig and timelock to config files


gpsanant (approved):
:+1:

PR #33, opened 2023-05-11
Update specs in /base to CVL2

teryanarmen:
separate specs and run scripts are kept in certora/spec22


ChaoticWalrus (approved):
LGTM, merging now. Am planning to try things out, make any necessary changes + deprecate/delete our old CVL1 specs

PR #34, opened 2023-05-11
Code4Rena Fixes

Sidu28:


src/contracts/core/DelegationManager.sol
```diff
@@ -36,6 +36,14 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         _;
     }
 
+    modifier onlyNotFrozen(address staker) {
+        require(
+            !slasher.isFrozen(staker),
+            "StrategyManager.onlyNotFrozen: staker has been frozen and may be subject to slashing"
```
ChaoticWalrus, line 42: revert message has typo, refers to wrong contract

src/contracts/core/DelegationManager.sol
```diff
@@ -137,7 +145,7 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
      * @notice Callable only by the StrategyManager
      * @dev Should only ever be called in the event that the `staker` has no active deposits in EigenLayer.
      */
-    function undelegate(address staker) external onlyStrategyManager {
+    function undelegate(address staker) external onlyStrategyManager onlyNotFrozen(staker){
```
ChaoticWalrus, line 148: why aren't we applying this at the StrategyManager level instead?

src/contracts/libraries/Merkle.sol
```diff
@@ -123,8 +125,6 @@ library Merkle {
     /**
      @notice this function returns the merkle root of a tree created from a set of leaves using sha256 as its hash function
      @param leaves the leaves of the merkle tree
-
-     @notice requires the leaves.length is a power of 2
```
ChaoticWalrus, line 127: I thought we wanted to *clarify* this wording, not delete it entirely.
The usage of 'require' here was apparently to signal a pre-condition of calling this.
Purely deleting the comment seems unhelpful and undesirable.

src/contracts/pods/DelayedWithdrawalRouter.sol
```diff
@@ -107,14 +107,40 @@ contract DelayedWithdrawalRouter is Initializable, OwnableUpgradeable, Reentranc
...
         uint256 delayedWithdrawalsLength = _userWithdrawals[user].delayedWithdrawals.length;
         uint256 claimableDelayedWithdrawalsLength = delayedWithdrawalsLength - delayedWithdrawalsCompleted;
-        DelayedWithdrawal[] memory claimableDelayedWithdrawals = new DelayedWithdrawal[](claimableDelayedWithdrawalsLength);
+
+        uint256 count = 0;
         for (uint256 i = 0; i < claimableDelayedWithdrawalsLength; i++) {
-            claimableDelayedWithdrawals[i] = _userWithdrawals[user].delayedWithdrawals[delayedWithdrawalsCompleted + i];
+            DelayedWithdrawal memory delayedWithdrawal = _userWithdrawals[user].delayedWithdrawals[delayedWithdrawalsCompleted + i];
+            //emit log_named_uint("block.number", block.number);
+            //emit log_named_uint("delayedWithdrawal.blockCreated + withdrawalDelayBlocks", delayedWithdrawal.blockCreated + withdrawalDelayBlocks);
```
ChaoticWalrus, line 132: clean up logging

src/test/unit/DelegationUnit.t.sol
```diff
@@ -76,6 +76,18 @@ contract DelegationUnitTests is EigenLayerTestHelper {
...
+        delegationContract.registerAsOperator(IDelegationTerms(address(this)));
+        cheats.stopPrank();
+
+        slasherMock.setOperatorFrozenStatus(address(this), true);
+        cheats.expectRevert(bytes("StrategyManager.onlyNotFrozen: staker has been frozen and may be subject to slashing"));
+        cheats.startPrank(address(strategyManagerMock));
+        delegationContract.undelegate(address(this));
+        cheats.stopPrank();
+    }
+
```
ChaoticWalrus, line 90: move this test to the StrategyManager tests along with moving the modifier to the StrategyManager

src/contracts/core/DelegationManager.sol
Sidu28: yeah good point, fixed

src/test/unit/DelegationUnit.t.sol
Sidu28: done

src/contracts/libraries/Merkle.sol
Sidu28: yeah but as far as the library is concerned, the number of leaves is irrelevant.  So I thought its better to remove it to keep the library as general as possible.   We make the check and comment on it in the BeaconChainProofs, where this condition matters

Sidu28: As far as the Merkle library is, it doesn't matter what the length of the leaves is.  It feels like making comments about preconditions seems inappropriate for a general library.  There are comments in the relevant  protocol code

ChaoticWalrus: I completely disagree. This code is only intended for having an input whose length is a power of 2. It will cut off other inputs improperly.
Noting this behavior in a comment _purely helps protect other people who might use this code_ which was the intention of the original comment. Just removing the comment means making it easier for people to make mistakes in usage.

src/contracts/interfaces/IDelayedWithdrawalRouter.sol
```diff
@@ -39,8 +39,11 @@ interface IDelayedWithdrawalRouter {
     /// @notice Getter function for the mapping `_userWithdrawals`
     function userWithdrawals(address user) external view returns (UserDelayedWithdrawals memory);
 
+    /// @notice Getter function to get all delayedWithdrawals of the `user`
+    function getUserDelayedWithdrawals(address user) external view returns (DelayedWithdrawal[] memory);
+
     /// @notice Getter function to get all delayedWithdrawals that are currently claimable by the `user`
-    function claimableUserDelayedWithdrawals(address user) external view returns (DelayedWithdrawal[] memory);
+    function getClaimableUserDelayedWithdrawals(address user) external view returns (DelayedWithdrawal[] memory);
```
ChaoticWalrus, line 46: @gpsanant any input on the naming change here?

src/contracts/core/StrategyManager.sol
```diff
@@ -571,10 +571,10 @@ contract StrategyManager is
                     // tell the strategy to send the appropriate amount of funds to the recipient
                     queuedWithdrawal.strategies[i].withdraw(recipient, tokens[i], queuedWithdrawal.shares[i]);
                 }
-                unchecked {
+            }
+            unchecked {
                     ++i;
```
ChaoticWalrus, line 576: can we get a regression test for this? could be added to the unit tests

ChaoticWalrus (changes requested):
I think this is nearly there. New tests are looking good, just needs a few last changes I think (see comments).

src/contracts/libraries/Merkle.sol
Sidu28: shit for some reason I thought I remembered this thing used zero hashes to pad but it totally doesn't.  Honestly I think a require statement might be the right thing here.

gpsanant: the runtime of checking whether a number is a power of two (without witnesses) is logarithmic. i think it's better we expect users of this function to know the exact expected length and make a constant time check before feeding into this function?

ChaoticWalrus: I'm OK with log-cost checking, seems reasonable to add. An extra check for correctness is likely worth the gas. We can always benchmark it and see the cost.
It's totally infeasible to use this to build a big tree, so number of steps shouldn't be too high, right?

src/contracts/interfaces/IDelayedWithdrawalRouter.sol
gpsanant: This will break the UI, can you please change back? @Sidu28

src/contracts/pods/DelayedWithdrawalRouter.sol
```diff
@@ -107,14 +107,38 @@ contract DelayedWithdrawalRouter is Initializable, OwnableUpgradeable, Reentranc
...
+        DelayedWithdrawal[] memory userDelayedWithdrawals = new DelayedWithdrawal[](userDelayedWithdrawalsLength);
+        for (uint256 i = 0; i < userDelayedWithdrawalsLength; i++) {
+            userDelayedWithdrawals[i] = _userWithdrawals[user].delayedWithdrawals[delayedWithdrawalsCompleted + i];
+        }
+        return userDelayedWithdrawals;
+    }
+
     /// @notice Getter function to get all delayedWithdrawals that are currently claimable by the `user`
-    function claimableUserDelayedWithdrawals(address user) external view returns (DelayedWithdrawal[] memory) {
+    function getClaimableUserDelayedWithdrawals(address user) external view returns (DelayedWithdrawal[] memory) {
```
gpsanant, line 123: let's keep this function exactly the same to not break the UI. We can create a new function that does this. Prob would be good to just add a comment that this function is slightly misnamed.. :)

src/contracts/core/StrategyManager.sol
Sidu28: added

src/contracts/pods/DelayedWithdrawalRouter.sol
```diff
@@ -107,14 +107,39 @@ contract DelayedWithdrawalRouter is Initializable, OwnableUpgradeable, Reentranc
...
+            DelayedWithdrawal memory delayedWithdrawal = _userWithdrawals[user].delayedWithdrawals[delayedWithdrawalsCompleted + i];
+            if (block.number > delayedWithdrawal.blockCreated + withdrawalDelayBlocks) {
+                count++;
+            }
+        }
+        
+        DelayedWithdrawal[] memory claimableDelayedWithdrawals = new DelayedWithdrawal[](count);
+        for (uint256 i = 0; i < claimableDelayedWithdrawalsLength; i++) {
+            DelayedWithdrawal memory delayedWithdrawal = _userWithdrawals[user].delayedWithdrawals[delayedWithdrawalsCompleted + i];
+            if (block.number > delayedWithdrawal.blockCreated + withdrawalDelayBlocks) {
```
ChaoticWalrus, line 140: is there a reason we check this condition in both loops?
this isn't technically wrong + it is 'only a view function' so I'm OK with it, just seems a bit weird / inefficient

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -227,10 +230,14 @@ contract StrategyManagerUnitTests is Test {
...
-    function testDepositIntoStrategySuccessfully(address staker, uint256 amount) public filterFuzzedAddressInputs(staker) {
-        IStrategy strategy = dummyStrat;
+    function testDepositIntoStrategySuccessfully(IStrategy strategy, address staker, uint256 amount) public filterFuzzedAddressInputs(staker) {
         IERC20 token = dummyToken;
 
+        /// IMPORTANT: ensure that when using this test as a part of another test, that the input strategy is already whitelisted!
+        if (!strategyManager.strategyIsWhitelistedForDeposit(strategy)){
+            strategy = dummyStrat;
+        }
+
```
ChaoticWalrus, line 240: this is some questionable test design right here. we shouldn't be modifying inputs like this -- maybe make an internal version of this test that takes a strategy as input and a public version that hard-codes the strategy as `dummyStrat` instead?
This feels like it is a massive footgun that doesn't need to exist

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -1181,6 +1234,14 @@ contract StrategyManagerUnitTests is Test {
         require(balanceAfter == balanceBefore, "balanceAfter != balanceBefore");
     }
 
+    function testUndelegateWithFrozenStaker(address undelegator) public {
```
ChaoticWalrus, line 1237: why does this have a fuzzed input if we don't use it at all?

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -2167,12 +2245,46 @@ contract StrategyManagerUnitTests is Test {
         return (queuedWithdrawal, tokensArray, withdrawalRoot);
     }
 
+    function _setUpQueuedWithdrawalStructSingleStratMultipleStrategies(
```
ChaoticWalrus, line 2248: what is this naming?
is this supposed to be `_setUpQueuedWithdrawalStruct_MultipleStrategies` or something?

src/test/unit/StrategyManagerUnit.t.sol
Sidu28: yeah I made a comment on it too.  Didn't want to add more duplicate code to an already bloated file, but its not really great.

src/contracts/pods/DelayedWithdrawalRouter.sol
Sidu28: I think we have to check the if statement in both loops?  Because both loops iterate through the whole list of non completed withdrawals, the first one just finds the length of the array

src/test/unit/StrategyManagerUnit.t.sol
ChaoticWalrus: I feel am not comfortable ever merging code that looks like this. Code bloat is 100% preferable to booby-trapped code.

Sidu28: fixed it

src/contracts/interfaces/IDelayedWithdrawalRouter.sol
ChaoticWalrus, line 46: needs correct comments on the interface too

src/contracts/pods/DelayedWithdrawalRouter.sol
ChaoticWalrus: ah I missed this. why aren't we just going through the correct indices in the second loop? this is even more inefficient than I thought.

src/contracts/interfaces/IDelayedWithdrawalRouter.sol
Sidu28: seems like the comments on the interface are correct?

ChaoticWalrus: ahh mb it's just missing the one note about naming

Sidu28: oh oh I forgot to mention, we're going to change the function name on the UI.  Chris Moller confirmed it wasn't a big change.

ChaoticWalrus: OK, sounds like @gpsanant should be the final approver on this then if you've agreed on something offline

src/contracts/pods/DelayedWithdrawalRouter.sol
Sidu28: I felt like using a fixed array dynamically wasn't clean but its definitely the more efficient solution, which is more important.  Fixed.

src/contracts/pods/DelayedWithdrawalRouter.sol
```diff
@@ -107,14 +107,39 @@ contract DelayedWithdrawalRouter is Initializable, OwnableUpgradeable, Reentranc
...
+                claimableDelayedWithdrawalIndices[count] = i;
+                count++;
+            }
+        }
+        
+        DelayedWithdrawal[] memory claimableDelayedWithdrawals = new DelayedWithdrawal[](count);
+        for (uint256 i = 0; i < count; i++) {
+            uint256 index = claimableDelayedWithdrawalIndices[i];
+            DelayedWithdrawal memory delayedWithdrawal = _userWithdrawals[user].delayedWithdrawals[delayedWithdrawalsCompleted + index];
+            claimableDelayedWithdrawals[i] = delayedWithdrawal;
```
ChaoticWalrus, line 142: OK so I think maybe there was some confusion here, now looking at this updated code. all withdrawals are strictly in order. we shouldn't need to store an array of `claimableDelayedWithdrawalIndices` -- this is always just going to be a list starting with the index of the first non-claimed withdrawal, and containing all subsequent indices in order before the first unclaimable one

ChaoticWalrus:
Just resolved the merge conflicts this had.
This PR LGTM now.

ChaoticWalrus (approved):

PR #35, opened 2023-05-12
Multiquorums

gpsanant:
For a broad overview of the technical changes we have added the architechture detailed [here](https://github.com/Layr-Labs/eigenlayer-contracts/tree/avs-docs/docs/middleware).

Each contract has its own individual technical documentation as well.

I would also appreciate making PRs to the [docs branch](https://github.com/Layr-Labs/eigenlayer-contracts/tree/avs-docs) for filling gaps and clarifying as you go through your review.

Resources:
- [The docs](https://github.com/Layr-Labs/eigenlayer-contracts/tree/avs-docs/docs/middleware)
- [Design review notes](https://docs.google.com/document/d/1Hyk2fa73yHQzIXLiz-QWl-6h_gbpTsHInZ7C1ksFuuA/edit#heading=h.s3f776rv3imd)

Things to review:
- src/middleware (Start here, and in the order of the AVS docs recommend)
- src/libraries
- src/interfaces
- Scripts
- Tests

Areas for particular scrutiny:
- Consistent (non absurd) state between all registries and the coordinator
- Operators cannot cause damage by registering/deregistering from and to certain quorums repeatedly (i.e. register for 0, 1, deregister 0, register 0, 3, ...)
- ChurnApprover implementation is correct
- Any functions that are permissionless to call :)

Things to not review:
- BLSSignatureCheckerOld.sol
- src/core

Potential changes:
- Allowing operators to specify a maximum stake in the stake registry which their stake will be capped by
- Roles/Governance (i.e. servicemanager owner everywhere might be changed)


src/contracts/core/StrategyManager.sol
```diff
@@ -90,27 +90,23 @@ contract StrategyManager is
...
-    modifier onlyEigenPodManager {
+    function _onlyEigenPodManager() internal {
         require(address(eigenPodManager) == msg.sender, "StrategyManager.onlyEigenPodManager: not the eigenPodManager");
-        _;
     }
 
-    modifier onlyStrategyWhitelister {
+    function _onlyStrategyWhitelister() internal {
         require(msg.sender == strategyWhitelister, "StrategyManager.onlyStrategyWhitelister: not the strategyWhitelister");
-        _;
```
ChaoticWalrus, line 113: @gpsanant we previously discussed reverting these changes as part of this merge. you might want to pull from master, resolve merge conflicts and make sure this gets properly handled

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+ *      3) an `IndexRegistry` that keeps track of an ordered list of operators for each quorum
+ * 
+ * @author Layr Labs, Inc.
+ */
+contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistryCoordinatorWithIndices, ISocketUpdater, Pausable {
+    using BN254 for BN254.G1Point;
+
+    /// @notice The EIP-712 typehash for the `DelegationApproval` struct used by the contract
+    bytes32 public constant OPERATOR_CHURN_APPROVAL_TYPEHASH =
+        keccak256("OperatorChurnApproval(bytes32 registeringOperatorId, OperatorKickParam[] operatorKickParams)OperatorKickParam(address operator, BN254.G1Point pubkey, bytes32[] operatorIdsToSwap)BN254.G1Point(uint256 x, uint256 y)");
```
ChaoticWalrus, line 35: the spacing looks slightly different than what I'm used to in typehash definitions. I think it's supposed to be
```
        keccak256("OperatorChurnApproval(bytes32 registeringOperatorId,OperatorKickParam[] operatorKickParams)OperatorKickParam(address operator,BN254.G1Point pubkey,bytes32[] operatorIdsToSwap)BN254.G1Point(uint256 x,uint256 y)");
```
I get this from looking at the EIP itself https://eips.ethereum.org/EIPS/eip-712 and in particular the example: "An example encoding is `Transaction(Person from,Person to,Asset tx)Asset(address token,uint256 amount)Person(address wallet,string name)`."

src/contracts/core/StrategyManager.sol
```diff
@@ -183,9 +179,9 @@ contract StrategyManager is
      */
     function recordBeaconChainETHBalanceUpdate(address podOwner, uint256 beaconChainETHStrategyIndex, int256 sharesDelta)
```
Sidu28, line 180: Hmm this stuff is no longer there in SM, I think master needs to be merged into this branch

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+    /// @notice the Service Manager for the service that this contract is coordinating
+    IServiceManager public immutable serviceManager;
+    /// @notice the BLS Pubkey Registry contract that will keep track of operators' BLS public keys
+    IBLSPubkeyRegistry public immutable blsPubkeyRegistry;
+    /// @notice the Stake Registry contract that will keep track of operators' stakes
+    IStakeRegistry public immutable stakeRegistry;
+    /// @notice the Index Registry contract that will keep track of operators' indexes
+    IIndexRegistry public immutable indexRegistry;
+
+    /// @notice the mapping from quorum number to the maximum number of operators that can be registered for that quorum
```
ChaoticWalrus, line 55: "maximum number of operators that can be registered for that quorum" -- there's more info stored in this struct, right?

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+    /// @notice the mapping from operator's address to the operator struct
+    mapping(address => Operator) internal _operators;
+    /// @notice the dynamic-length array of the registries this coordinator is coordinating
+    address[] public registries;
+    /// @notice the address of the entity allowed to sign off on operators getting kicked out of the AVS during registration
+    address public churnApprover;
+    /// @notice whether the salt has been used for an operator churn approval
+    mapping(bytes32 => bool) public isChurnApproverSaltUsed;
+    /// @notice the address of the entity allowed to eject operators from the AVS
+    address public ejector;
```
ChaoticWalrus, line 68: nit: typically storage is organized by type

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+            for (uint32 j = 0; j < length; j++) {
+                if (_operatorIdToQuorumBitmapHistory[operatorIds[i]][length - j - 1].updateBlockNumber <= blockNumber) {
+                    require(
+                        _operatorIdToQuorumBitmapHistory[operatorIds[i]][length - j - 1].nextUpdateBlockNumber == 0 ||
+                        _operatorIdToQuorumBitmapHistory[operatorIds[i]][length - j - 1].nextUpdateBlockNumber > blockNumber,
+                        "BLSRegistryCoordinatorWithIndices.getQuorumBitmapIndicesByOperatorIdsAtBlockNumber: operatorId has no quorumBitmaps at blockNumber"
+                    );
+                    indices[i] = length - j - 1;
+                    break;
+                }
```
ChaoticWalrus, line 155: making a note here -- we should make sure that our docs explain that in the general pattern of storing a history, the choices has been made that:

- if there are multiple updates in a block, then the last update for the block is canonical
- an update in a block is also treated as being "active" _in that block_, as opposed to being "active" starting from the next block

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+    function getQuorumBitmapUpdateByOperatorIdByIndex(bytes32 operatorId, uint256 index) external view returns (QuorumBitmapUpdate memory) {
+        return _operatorIdToQuorumBitmapHistory[operatorId][index];
+    }
+
+    /// @notice Returns the current quorum bitmap for the given `operatorId` or 0 if the operator is not registered for any quorum
+    function getCurrentQuorumBitmapByOperatorId(bytes32 operatorId) external view returns (uint192) {
+        uint256 quorumBitmapHistoryLength = _operatorIdToQuorumBitmapHistory[operatorId].length;
+        if (quorumBitmapHistoryLength == 0 || _operatorIdToQuorumBitmapHistory[operatorId][quorumBitmapHistoryLength - 1].nextUpdateBlockNumber != 0) {
+            return 0;
+        }
```
ChaoticWalrus, line 190: might help to add a comment that the second part of this condition is when the operator has deregistered. was about to ask if we needed this at all

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+        bytes calldata quorumNumbers,
+        BN254.G1Point memory pubkey,
+        string calldata socket
+    ) external onlyWhenNotPaused(PAUSED_REGISTER_OPERATOR) {
+        _registerOperatorWithCoordinatorAndNoOverfilledQuorums(msg.sender, quorumNumbers, pubkey, socket);
+    }
+
+    /**
+     * @notice Registers msg.sender as an operator with the middleware when the quorum operator limit is full. To register 
+     * while maintaining the limit, the operator chooses another registered opeerator with lower stake to kick.
```
ChaoticWalrus, line 283: typo: 'opeerator'

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+                // if the number of operators for the quorum is less than or equal to the max operator count, 
+                // then the quorum has not reached the max operator count
+                if(numOperatorsPerQuorum[i] <= operatorSetParam.maxOperatorCount) {
+                    continue;
+                }
+
+                require(
+                    operatorKickParams[i].quorumNumber == quorumNumber, 
+                    "BLSRegistryCoordinatorWithIndices.registerOperatorWithCoordinator: quorumNumber not the same as signed"
+                );
```
ChaoticWalrus, line 323: am I reading this right that the `operatorKickParams` array needs to be the same length as the number of quorums the operator is registering for?
seems like this check would fail, otherwise.
if this is a requirement, we should probably just have a simple length check higher up.
if we don't want it to be a requirement, then we need something like a second index here, but I think the logic could get messy.

**possible security concern:** if these have to be the same length, then the `churnApprover` may need to sign arrays containing some empty entries -- I'm concerned that we handle these correctly such signed arrays under all circumstances.

ChaoticWalrus: to elaborate on the potential security concern, with an example:

- operator wants to register for quorums [1,3] and currently only quorum 3 is full
- churnApprover signs array with [empty other than quorum number, empty other than quorum number, full details of operatorToKick] entries
- new operator registers for quorum 1, making it full
- original operator calls this function, and we end up hitting a case where we try to kick an operator based on the signed entry that's empty other than quorum number

I guess this might be more of a race condition, actually? If we are enforcing behavior well on deregistrations then we should at least just hit a revert in this scenario

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+    /**
+     * @notice Deregisters the msg.sender as an operator from the middleware
+     * @param quorumNumbers are the bytes representing the quorum numbers that the operator is registered for
+     * @param deregistrationData is the the data that is decoded to get the operator's deregistration information
+     * @dev `deregistrationData` should be a tuple of the operator's BLS public key, the list of operator ids to swap
+     */
+    function deregisterOperatorWithCoordinator(
+        bytes calldata quorumNumbers,
+        bytes calldata deregistrationData
+    ) external onlyWhenNotPaused(PAUSED_DEREGISTER_OPERATOR) {
```
ChaoticWalrus, line 363: might help to point out or explain somewhere that these overloaded versions of functions with just 'bytes' input data are part of the more general interface / (seemingly) exist so developers can override the function to add more parameters, without changing the function signature?
also curious on thoughts about marking some or all functions in these contracts as 'virtual' to enable easier overriding -- personally feels like it might be overkill / I wouldn't want to encourage this too much, as there would be a lot of potential footguns for developers who somewhat blindly override a function.

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: will, e.g., the call to `getCurrentOperatorStakeForQuorum` below revert? if not then I think there could be an issue here

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+    }
+
+    /**
+     * @notice Deregisters the msg.sender as an operator from the middleware
+     * @param quorumNumbers are the bytes representing the quorum numbers that the operator is registered for
+     * @param pubkey is the BLS public key of the operator
+     * @param operatorIdsToSwap is the list of the operator ids to swap the index of the operator with in each 
+     * quorum when removing the operator from the quorum's ordered list. The provided operator ids should be the 
+     * those of the operator's with the largest index in each quorum that the operator is deregistering from, in
+     * ascending order of quorum number.
```
ChaoticWalrus, line 378: do we not keep track of the total number of operators for each quorum?
if not, makes sense. otherwise I'd assume we could look this info up.
I'll try to circle back on this.

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+            operatorId: operatorId,
+            status: OperatorStatus.REGISTERED
+        });
+
+        // record a stake update not bonding the operator at all (unbonded at 0), because they haven't served anything yet
+        // serviceManager.recordFirstStakeUpdate(operator, 0);
+
+        emit OperatorRegistered(operator, operatorId);
+
+        emit OperatorSocketUpdate(operatorId, socket);
```
ChaoticWalrus, line 479: can this function (`_registerOperatorWithCoordinator`) be called when the operator is already registered for some quorums?
seems to me like it can be (and I think it's supposed to be)?
documentation should make intended behavior here clear -- TODO for myself @ChaoticWalrus  to check the docs PR for this.
I'm somewhat concerned about the highlighted code being run when the operator is already registered. here's a few questions:

- we aren't updating the previous entry in the "quorum bitmap history" -- is this wrong? seems like we could end up in a case where multiple entries have `nextUpdateBlockNumber = 0`, which would be a problem AFAIK
- we're setting the operator struct here -- presumably the state being set to REGISTERED is fine, am pretty sure the operatorId can't change here, just want to be certain (seems like potentially an unstated but used property of this system)
- do we want the events here? I guess the socket update is happening, so it makes sense, and the `OperatorRegistered` event could trigger logic to check what (new) quorums the operator is registered for?

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+        // get the quorumNumbers of the operator
+        uint256 quorumsToRemoveBitmap = BitmapUtils.orderedBytesArrayToBitmap(quorumNumbers);
+        require(quorumsToRemoveBitmap <= MiddlewareUtils.MAX_QUORUM_BITMAP, "BLSRegistryCoordinatorWithIndices._deregisterOperatorWithCoordinator: quorumsToRemoveBitmap exceeds of max bitmap size");
+        uint256 operatorQuorumBitmapHistoryLengthMinusOne = _operatorIdToQuorumBitmapHistory[operatorId].length - 1;
+        uint192 quorumBitmapBeforeUpdate = _operatorIdToQuorumBitmapHistory[operatorId][operatorQuorumBitmapHistoryLengthMinusOne].quorumBitmap;
+        // check that the quorumNumbers of the operator matches the quorumNumbers passed in
+        require(
+            quorumBitmapBeforeUpdate & quorumsToRemoveBitmap == quorumsToRemoveBitmap,
+            "BLSRegistryCoordinatorWithIndices._deregisterOperatorWithCoordinator: cannot deregister operator for quorums that it is not a part of"
+        );
```
ChaoticWalrus, line 510: going to point out a potential race condition here -- if an operator deregisters from a single quorum, they can potentially block themselves from being "ejected" from other quorums (or at least cause a call to eject them to revert). doesn't seem like a major issue, but there's definitely a number of similar race conditions here.

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+                nextUpdateBlockNumber: 0,
+                quorumBitmap: quorumBitmapBeforeUpdate & ~uint192(quorumsToRemoveBitmap) // this removes the quorumsToRemoveBitmap from the quorumBitmapBeforeUpdate
+            }));
+        } else {
+            // @notice Registrant must continue to serve until the latest block at which an active task expires. this info is used in challenges
+            uint32 latestServeUntilBlock = serviceManager.latestServeUntilBlock();
+
+            // record a stake update unbonding the operator after `latestServeUntilBlock`
+            // serviceManager.recordLastStakeUpdateAndRevokeSlashingAbility(operator, latestServeUntilBlock);
+            // set the status of the operator to DEREGISTERED
```
ChaoticWalrus, line 539: this variable is unused due to the other commented-out lines. should presumably be commented out as well.

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -0,0 +1,555 @@
...
+    }
+
+    /// @notice verifies churnApprover's signature on operator churn approval and increments the churnApprover nonce
+    function _verifyChurnApproverSignatureOnOperatorChurnApproval(bytes32 registeringOperatorId, OperatorKickParam[] memory operatorKickParams, SignatureWithSaltAndExpiry memory signatureWithSaltAndExpiry) internal {
+        // make sure the salt hasn't been used already
+        require(!isChurnApproverSaltUsed[signatureWithSaltAndExpiry.salt], "BLSRegistryCoordinatorWithIndices._verifyChurnApproverSignatureOnOperatorChurnApproval: churnApprover salt already used");
+        require(signatureWithSaltAndExpiry.expiry >= block.timestamp, "BLSRegistryCoordinatorWithIndices._verifyChurnApproverSignatureOnOperatorChurnApproval: churnApprover signature expired");        
+        EIP1271SignatureUtils.checkSignature_EIP1271(churnApprover, calculateOperatorChurnApprovalDigestHash(registeringOperatorId, operatorKickParams, signatureWithSaltAndExpiry.salt, signatureWithSaltAndExpiry.expiry), signatureWithSaltAndExpiry.signature);
+        // set salt used to true
+        isChurnApproverSaltUsed[signatureWithSaltAndExpiry.salt] = true;
```
ChaoticWalrus, line 553: this is some CIE design, rather than CEI.
although reentrancy by a trusted role is less important to protect against, we should have the order of operations flipped here (i.e. mark salt as spent, then potentially call to external contract)

src/contracts/middleware/BLSPublicKeyCompendium.sol
```diff
@@ -20,49 +19,44 @@ contract BLSPublicKeyCompendium is IBLSPublicKeyCompendium {
...
-                    )
-                )
-            );
+    function registerBLSPublicKey(BN254.G1Point memory signedMessageHash, BN254.G1Point memory pubkeyG1, BN254.G2Point memory pubkeyG2) external {
+        // H(m) 
+        BN254.G1Point memory messageHash = BN254.hashToG1(keccak256(abi.encodePacked(
+            msg.sender, 
+            block.chainid, 
+            "EigenLayer_BN254_Pubkey_Registration"
+        )));
```
ChaoticWalrus, line 36: IMO the message hash should also include `address(this)`.
I also highly recommend adding a public getter function for the message hash, for convenience.

src/contracts/middleware/BLSPublicKeyCompendium.sol
```diff
@@ -20,49 +19,44 @@ contract BLSPublicKeyCompendium is IBLSPublicKeyCompendium {
...
         ), "BLSPublicKeyCompendium.registerBLSPublicKey: G1 and G2 private key do not match");
 
-        // getting pubkey hash
         bytes32 pubkeyHash = BN254.hashG1Point(pubkeyG1);
 
-        require(pubkeyHash != ZERO_PK_HASH, "BLSPublicKeyCompendium.registerBLSPublicKey: operator attempting to register the zero public key");
-
         require(
             operatorToPubkeyHash[msg.sender] == bytes32(0),
             "BLSPublicKeyCompendium.registerBLSPublicKey: operator already registered pubkey"
```
ChaoticWalrus, line 62: we should revert earlier, if possible (move these require statements up)

src/contracts/middleware/StakeRegistryStorage.sol
```diff
@@ -0,0 +1,41 @@
...
+
+    /// @notice In order to register for a quorum i, an operator must have at least `minimumStakeForQuorum[i]`
+    /// evaluated by this contract's 'VoteWeigher' logic.
+    uint96[256] public minimumStakeForQuorum;
+
+    /// @notice array of the history of the total stakes for each quorum -- marked as internal since getTotalStakeFromIndex is a getter for this
+    OperatorStakeUpdate[][256] internal _totalStakeHistory;
+
+    /// @notice mapping from operator's operatorId to the history of their stake updates
+    mapping(bytes32 => mapping(uint8 => OperatorStakeUpdate[])) public operatorIdToStakeHistory;
```
ChaoticWalrus, line 27: might want this to be marked 'internal' with a custom getter instead, so that the return type can be a `OperatorStakeUpdate memory` object, instead of a tuple

src/contracts/middleware/StakeRegistryStorage.sol
```diff
@@ -0,0 +1,41 @@
...
+        IStrategyManager _strategyManager,
+        IServiceManager _serviceManager
+    ) VoteWeigherBase(_strategyManager, _serviceManager)
+    // solhint-disable-next-line no-empty-blocks
+    {
+        registryCoordinator = _registryCoordinator;
+    }
+
+    // storage gap
+    uint256[50] private __GAP;
```
ChaoticWalrus, line 40: not too sure on the proper size for this gap, but typically the idea is the total storage for a contract is a round number of slots.
looks like we're using 513 slots already -- maybe a gap size of 37 makes sense, so the total storage is a round 550? other obvious option is 87 slot gap to get 600 total slots

src/contracts/middleware/StakeRegistry.sol
```diff
@@ -0,0 +1,461 @@
...
+            uint32 length = uint32(_totalStakeHistory[quorumNumber].length);
+            for (uint32 j = 0; j < length; j++) {
+                if (_totalStakeHistory[quorumNumber][length - j - 1].updateBlockNumber <= blockNumber) {
+                    indices[i] = length - j - 1;
+                    break;
+                }
+            }
+        }
+        return indices;
+    }
```
ChaoticWalrus, line 112: what's the intended failure mode of this function? e.g. if I input a block number that's too-low (or too-high)?

ChaoticWalrus: seems like this will fail with underflow rn (?) on this line:
`indices[i] = length - j - 1;`

src/contracts/middleware/StakeRegistry.sol
```diff
@@ -0,0 +1,461 @@
...
+            return operatorStakeUpdate;
+        } else {
+            operatorStakeUpdate = operatorIdToStakeHistory[operatorId][quorumNumber][historyLength - 1];
+            return operatorStakeUpdate;
+        }
+    }
+
+    function getStakeHistoryLengthForQuorumNumber(bytes32 operatorId, uint8 quorumNumber) external view returns (uint256) {
+        return operatorIdToStakeHistory[operatorId][quorumNumber].length;
+    }
```
ChaoticWalrus, line 165: this function appears to be a duplicate of `getLengthOfOperatorIdStakeHistoryForQuorum` -- recommend deleting it.

src/contracts/middleware/StakeRegistry.sol
```diff
@@ -0,0 +1,461 @@
...
+    /**
+     * @notice Used for updating information on deposits of nodes.
+     * @param operators are the addresses of the operators whose stake information is getting updated
+     * @param operatorIds are the ids of the operators whose stake information is getting updated
+     * @param prevElements are the elements before this middleware in the operator's linked list within the slasher
+     * @dev Precondition:
+     *          1) `quorumBitmaps[i]` should be the bitmap that represents the quorums that `operators[i]` registered for
+     * @dev reverts if there are no operators registered with index out of bounds
+     */
+    function updateStakes(address[] calldata operators, bytes32[] calldata operatorIds, uint256[] calldata prevElements) external {
```
ChaoticWalrus, line 250: This precondition seems improperly copy-pasted onto this function, and should presumably just be deleted

src/contracts/middleware/StakeRegistry.sol
```diff
@@ -0,0 +1,461 @@
...
+        // for each quorum, evaluate stake and add to total stake
+        for (uint8 quorumNumbersIndex = 0; quorumNumbersIndex < quorumNumbers.length;) {
+            // get the next quorumNumber
+            uint8 quorumNumber = uint8(quorumNumbers[quorumNumbersIndex]);
+            // evaluate the stake for the operator
+            // since we don't use the first output, this will use 1 extra sload when deregistered operator's register again
+            (, uint96 stake) = _updateOperatorStake(operator, operatorId, quorumNumber);
+            // @JEFF: This reverts pretty late, but i think that's fine. wdyt?
+            // check if minimum requirement has been met, will be 0 if not
+            require(stake != 0, "StakeRegistry._registerOperator: Operator does not meet minimum stake requirement for quorum");
```
ChaoticWalrus, line 335: seems hard to avoid a fairly late revert here; sometimes these things just are what they are. I think it's OK, but something to note when explaining some of the race conditions with this design.

src/contracts/middleware/StakeRegistry.sol
```diff
@@ -0,0 +1,461 @@
...
+        }
+    }
+
+    /**
+     * @notice Removes the stakes of the operator with `operatorId` from the quorums specified in `quorumNumbers`
+     * the total stake of the quorums specified in `quorumNumbers` will be updated and so will the operator's individual
+     * stake updates. These operator's individual stake updates will have a 0 stake value for the latest update.
+     */
+    function _deregisterOperator(bytes32 operatorId, bytes memory quorumNumbers) internal {
+        // check the operator is deregistering from only valid quorums
```
ChaoticWalrus, line 360: the comment here is erroneous / no such check is occurring -- I'm assuming the check has moved higher up the call stack?

src/contracts/middleware/StakeRegistry.sol
```diff
@@ -0,0 +1,461 @@
...
+            operatorId,
+            quorumNumber,
+            operatorStakeUpdate.stake
+        );
+
+        return (stakeBeforeUpdate, operatorStakeUpdate.stake);
+    }
+
+    /// @notice Records that `operatorId`'s current stake is now param @operatorStakeUpdate
+    function _recordOperatorStakeUpdate(bytes32 operatorId, uint8 quorumNumber, OperatorStakeUpdate memory operatorStakeUpdate) internal returns(uint96) {
```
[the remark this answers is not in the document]
ChaoticWalrus: I have a similar note about documenting the return values (and order thereof) for `_updateOperatorStake`

src/contracts/middleware/VoteWeigherBase.sol
```diff
@@ -16,86 +16,95 @@ import "./VoteWeigherBaseStorage.sol";
...
+contract VoteWeigherBase is VoteWeigherBaseStorage {
+    /// @notice emitted when a new quorum is created
+    event QuorumCreated(uint8 indexed quorumNumber);
+    /// @notice emitted when `strategy` has been added to the array at `strategiesConsideredAndMultipliers[quorumNumber]` with the `multiplier`
+    event StrategyAddedToQuorum(uint8 indexed quorumNumber, IStrategy strategy, uint96 multiplier);
     /// @notice emitted when `strategy` has removed from the array at `strategiesConsideredAndMultipliers[quorumNumber]`
-    event StrategyRemovedFromQuorum(uint256 indexed quorumNumber, IStrategy strategy);
+    event StrategyRemovedFromQuorum(uint8 indexed quorumNumber, IStrategy strategy);
+    /// @notice emitted when `strategy` has its `multiplier` updated in the array at `strategiesConsideredAndMultipliers[quorumNumber]`
+    event StrategyMultiplierUpdated(uint8 indexed quorumNumber, IStrategy strategy, uint256 multiplier);
```
ChaoticWalrus, line 27: noting that the events `StrategyAddedToQuorum` and `StrategyMultiplierUpdated` have different uint types for the `multiplier` field (uint96 vs. uint256). I presume we are emitting _both_ events when a new strategy is added, so think deleting the field from the `StrategyAddedToQuorum` might make sense?
ping to @gpsanant as I think he's most knowledgeable about how we're indexing this

ChaoticWalrus: adding: it looks like we are *not* emitting a `StrategyMultiplierUpdated` event inside of the `_addStrategiesConsideredAndMultipliers` function. IMO we should do this and remove the `multiplier` field from the `StrategyAddedToQuorum` event, as I suggested above.
I'm open to other ideas, just feel like this is the most logical.

src/contracts/middleware/VoteWeigherBaseStorage.sol
```diff
@@ -44,34 +35,33 @@ abstract contract VoteWeigherBaseStorage is Initializable, IVoteWeigher {
...
      */
-    mapping(uint256 => StrategyAndWeightingMultiplier[]) public strategiesConsideredAndMultipliers;
+    mapping(uint8 => StrategyAndWeightingMultiplier[]) public strategiesConsideredAndMultipliers;
 
     /**
      * @notice This defines the earnings split between different quorums. Mapping is quorumNumber => BIPS which the quorum earns, out of the total earnings.
      * @dev The sum of all entries, i.e. sum(quorumBips[0] through quorumBips[NUMBER_OF_QUORUMS - 1]) should *always* be 10,000!
      */
-    mapping(uint256 => uint256) public quorumBips;
+    mapping(uint8 => uint256) public quorumBips;
```
ChaoticWalrus, line 51: looks like we aren't using nor setting this right now.
that's fine, but the associated comment is definitely a bit of a scare here.
perhaps just something to track as a TODO going forward?

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -0,0 +1,263 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
+
+
+import "../interfaces/IIndexRegistry.sol";
+import "../interfaces/IRegistryCoordinator.sol";
+import "../libraries/BN254.sol";
+
+contract IndexRegistry is IIndexRegistry {
```
ChaoticWalrus, line 9: looks like this contract is missing top-level context / comments -- please add them

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -0,0 +1,263 @@
...
+     * @dev Preconditions (these are assumed, not validated in this contract):
+     *         1) `quorumNumbers` has no duplicates
+     *         2) `quorumNumbers.length` != 0
+     *         3) `quorumNumbers` is ordered in ascending order
+     *         4) the operator is not already registered
+     */
+    function registerOperator(bytes32 operatorId, bytes calldata quorumNumbers) external onlyRegistryCoordinator returns(uint32[] memory) {
+        uint32[] memory numOperatorsPerQuorum = new uint32[](quorumNumbers.length);
+        //add operator to operatorList
+        globalOperatorList.push(operatorId);
```
ChaoticWalrus, line 50: pointing out that we will push an operator repeatedly to this list when they register for new quorums. seems fine though, given we've already said the list 'may include duplicates' ?

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -0,0 +1,263 @@
...
+    function getTotalOperatorsForQuorumAtBlockNumberByIndex(uint8 quorumNumber, uint32 blockNumber, uint32 index) external view returns (uint32){
+        OperatorIndexUpdate memory operatorIndexToCheck = _totalOperatorsHistory[quorumNumber][index];
+
+        // blocknumber must be at or after the "index'th" entry's fromBlockNumber
+        require(blockNumber >= operatorIndexToCheck.fromBlockNumber, "IndexRegistry.getTotalOperatorsForQuorumAtBlockNumberByIndex: provided index is too far in the past for provided block number");
+        
+        // if there is an index update after the "index'th" update, the blocknumber must be before the next entry's fromBlockNumber
+        if (index != _totalOperatorsHistory[quorumNumber].length - 1){
+            OperatorIndexUpdate memory nextOperatorIndex = _totalOperatorsHistory[quorumNumber][index + 1];
+            require(blockNumber < nextOperatorIndex.fromBlockNumber, "IndexRegistry.getTotalOperatorsForQuorumAtBlockNumberByIndex: provided index is too far in the future for provided block number");
```
ChaoticWalrus, line 144: there's a number of lines here that are way, way too long + need to be broken up (especially the quoted 'require' statements

ChaoticWalrus: I think `getTotalOperatorsForQuorumAtBlockNumberByIndex` should also have a similar note about reversion behavior to the dev tag on `getOperatorIndexForQuorumAtBlockNumberByIndex`

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -0,0 +1,263 @@
...
+        for (uint i = 0; i < globalOperatorList.length; i++) {
+            bytes32 operatorId = globalOperatorList[i];
+            uint32 index = _getIndexOfOperatorForQuorumAtBlockNumber(operatorId, quorumNumber, blockNumber);
+            // if the operator was not in the quorum at the given block number, skip it
+            if (index == OPERATOR_DEREGISTERED_INDEX)
+                continue;
+            quorumOperatorList[index] = operatorId;
+        }
+        return quorumOperatorList;
+    }
```
ChaoticWalrus, line 161: any concerns about the unbounded nature of this loop?
I guess the function can always be simulated with eth_call and an excessively high gas budget?

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -0,0 +1,263 @@
...
+
+    function totalOperatorsForQuorum(uint8 quorumNumber) external view returns (uint32){
+        return _totalOperatorsHistory[quorumNumber][_totalOperatorsHistory[quorumNumber].length - 1].index;
+    }
```
ChaoticWalrus, line 165: function is lacking documentation, could use a simple 'notice' tag

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -0,0 +1,263 @@
...
+        }        
+        return _totalOperatorsHistory[quorumNumber][0].index;
+    }
+    
+
+    /// @notice Returns the index of the `operatorId` at the given `blockNumber` for the given `quorumNumber`, or `OPERATOR_DEREGISTERED_INDEX` if the operator was not registered for the `quorumNumber` at blockNumber
+    function _getIndexOfOperatorForQuorumAtBlockNumber(bytes32 operatorId, uint8 quorumNumber, uint32 blockNumber) internal view returns(uint32) {
+        uint256 operatorIndexHistoryLength = _operatorIdToIndexHistory[operatorId][quorumNumber].length;
+        // loop backward through index history to find the index of the operator at the given block number
+        for (uint i = 0; i < _operatorIdToIndexHistory[operatorId][quorumNumber].length; i++) {
```
ChaoticWalrus, line 250: we've just stored `_operatorIdToIndexHistory[operatorId][quorumNumber].length` in memory and should use it in the loop condition.
also style nit: generally prefer 'uint256' over the alias 'uint'

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -0,0 +1,263 @@
...
+                // one special case is that this will be OPERATOR_DEREGISTERED_INDEX if the operator was deregistered from the quorum
+                return operatorIndexUpdate.index;
+            }
+        }
+
+        // the operator is still active or not in the quorum, so we return the latest index or the default max uint32
+        // this will be hit if `blockNumber` is before when the operator registered or the operator has never registered for the given quorum
+        return OPERATOR_DEREGISTERED_INDEX;
+    }
+}
```
ChaoticWalrus, line 263: noticed there's no storage gap variable for this contract. looks like it's non-upgradeable right now, but I believe it's supposed to be upgradeable

ChaoticWalrus: checked with @gpsanant -- sounds like the intention was to make this contract (IndexRegistry) upgradeable, so we should add that functionality (inherit from Initializable, add initializer as necessary, add storage gap, disable initializers in constructor)

src/contracts/middleware/BLSSignatureChecker.sol
```diff
@@ -1,508 +1,218 @@
...
 /**
  * @title Used for checking BLS aggregate signatures from the operators of a `BLSRegistry`.
  * @author Layr Labs, Inc.
  * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
  * @notice This is the contract for checking the validity of aggregate operator signatures.
  */
-abstract contract BLSSignatureChecker {
+contract BLSSignatureChecker {
+    using BN254 for BN254.G1Point;    
+
```
ChaoticWalrus, line 17: it looks like we have an interface defined for this contract, but the contract isn't inheriting from the interface. This means things like the struct definitions are duplicated in both -- we should define the structs in the interface and have this contract inherit from the interface.

src/contracts/middleware/BLSSignatureChecker.sol
```diff
@@ -1,508 +1,218 @@
...
+    struct NonSignerStakesAndSignature {
+        uint32[] nonSignerQuorumBitmapIndices;
+        BN254.G1Point[] nonSignerPubkeys;
+        BN254.G1Point[] quorumApks;
+        BN254.G2Point apkG2;
+        BN254.G1Point sigma;
+        uint32[] quorumApkIndices;
+        uint32[] totalStakeIndices;  
+        uint32[][] nonSignerStakeIndices; // nonSignerStakeIndices[quorumNumberIndex][nonSignerIndex]
     }
```
ChaoticWalrus, line 29: Can we document/define the purpose of this struct and all of these fields?
It's challenging to keep track of everything; especially things like `nonSignerQuorumBitmapIndices` don't necessarily immediately have clear meaning

src/contracts/middleware/BLSSignatureChecker.sol
```diff
@@ -1,508 +1,218 @@
...
-    // specifying a G2 public key requires 4 32-byte slots worth of data
-    uint256 internal constant BYTE_LENGTH_G1_POINT = 64;
-    uint256 internal constant BYTE_LENGTH_G2_POINT = 128;
-    uint256 internal constant BYTE_LENGTH_stakeIndex = 4;
-    // uint256 internal constant BYTE_LENGTH_NON_SIGNER_INFO = BYTE_LENGTH_G1_POINT + BYTE_LENGTH_stakeIndex;
-    uint256 internal constant BYTE_LENGTH_NON_SIGNER_INFO = 68;
-    uint256 internal constant BYTE_LENGTH_apkIndex = 4;
+    // gas cost of multiplying 2 pairings
+    // TODO: verify this
+    uint256 constant PAIRING_EQUALITY_CHECK_GAS = 120000;
```
ChaoticWalrus, line 47: have we to-done this?

src/contracts/middleware/BLSSignatureChecker.sol
```diff
@@ -1,508 +1,218 @@
...
-     * {uint256[2] pubkeyG1, uint32 stakeIndex}[numberOfNonSigners] the G1 public key and the index to query of `pubkeyHashToStakeHistory` for each nonsigner,
-     * uint32 apkIndex, the index in the `apkUpdates` array at which we want to load the aggregate public key
-     * uint256[2] apkG1 (G1 aggregate public key, including nonSigners),
-     * uint256[4] apkG2 (G2 aggregate public key, not including nonSigners),
-     * uint256[2] sigma, the aggregate signature itself
-     * >
      * 
      * @dev Before signature verification, the function verifies operator stake information.  This includes ensuring that the provided `referenceBlockNumber`
      * is correct, i.e., ensure that the stake returned from the specified block number is recent enough and that the stake is either the most recent update
      * for the total stake (or the operator) or latest before the referenceBlockNumber.
```
ChaoticWalrus, line 73: looks like a typo: I think `or the operator` should be `of the operator`

src/contracts/middleware/BLSSignatureChecker.sol
```diff
@@ -1,508 +1,218 @@
...
-     * Finally  the siganture is verified by computing the elliptic curve pairing.
      */
-    function checkSignatures(bytes calldata data)
-        public
+    function checkSignatures(
+        bytes32 msgHash, 
+        bytes calldata quorumNumbers,
+        uint32 referenceBlockNumber, 
+        NonSignerStakesAndSignature memory nonSignerStakesAndSignature
+    )
```
ChaoticWalrus, line 80: seems like the NatSpec explaining params got deleted but not replaced.
the purposes are much clearer from the improved naming + structuring, but we should clarify with new NatSpec comments

src/contracts/middleware/BLSSignatureChecker.sol
```diff
@@ -1,508 +1,218 @@
...
+                            referenceBlockNumber, 
+                            nonSignerStakesAndSignature.nonSignerQuorumBitmapIndices[i]
+                        );
+                    
+                    // subtract the nonSignerPubkey from the running apk to get the apk of all signers
+                    apk = apk.plus(
+                        nonSignerStakesAndSignature.nonSignerPubkeys[i]
+                            .negate()
+                            .scalar_mul_tiny(
+                                BitmapUtils.countNumOnes(nonSignerQuorumBitmaps[i] & signingQuorumBitmap) // we subtract the nonSignerPubkey from each quorum that they are a part of, TODO:
```
ChaoticWalrus, line 137: there's an empty 'TODO' here, not sure what for.
In general the procedure makes sense, but do we have a doc that abstracts away from the code, so that e.g. someone who is just good at cryptography could easily review?

src/contracts/middleware/BLSSignatureChecker.sol
```diff
@@ -1,508 +1,218 @@
...
+                    stakeRegistry.getTotalStakeAtBlockNumberFromIndex(quorumNumber, referenceBlockNumber, nonSignerStakesAndSignature.totalStakeIndices[quorumNumberIndex]);
+                // copy total stake to signed stake
+                quorumStakeTotals.signedStakeForQuorum[quorumNumberIndex] = quorumStakeTotals.totalStakeForQuorum[quorumNumberIndex];
+                // loop through all nonSigners, checking that they are a part of the quorum via their quorumBitmap
+                // if so, load their stake at referenceBlockNumber and subtract it from running stake signed
+                for (uint32 i = 0; i < nonSignerStakesAndSignature.nonSignerPubkeys.length; i++) {
+                    // keep track of the nonSigners index in the quorum
+                    uint32 nonSignerForQuorumIndex = 0;
+                    // if the nonSigner is a part of the quorum, subtract their stake from the running total
+                    if (nonSignerQuorumBitmaps[i] >> quorumNumber & 1 == 1) {
```
ChaoticWalrus, line 157: I see this kind of operation a lot, including -- I think -- with the operations "reversed", i.e. something like
```
if (nonSignerQuorumBitmaps[i] && (1 << quorumNumber) != 0) {
```
Feels like it would be worth it to define a utility and add it to the BitmapUtils library for this. am imagining something like
`function numberIsInBitmap(uint256 bitmap, uint8 numberToCheckForInclusion) returns (bool)`

src/contracts/middleware/BLSSignatureChecker.sol
```diff
@@ -1,508 +1,218 @@
...
-        // set compressedSignatoryRecord variable used for fraudproofs
-        compressedSignatoryRecord = MiddlewareUtils.computeSignatoryRecordHash(
-            taskNumberToConfirm,
-            pubkeyHashes,
-            signedTotals.signedStakeFirstQuorum,
-            signedTotals.signedStakeSecondQuorum
+            // verify the signature
+            (bool pairingSuccessful, bool sigantureIsValid) = trySignatureAndApkVerification(msgHash, apk, nonSignerStakesAndSignature.apkG2, nonSignerStakesAndSignature.sigma);
+            require(pairingSuccessful, "BLSSignatureChecker.checkSignatures: pairing precompile call failed");
+            require(sigantureIsValid, "BLSSignatureChecker.checkSignatures: signature is invalid");
```
ChaoticWalrus, line 180: typo: `siganture` instead of `signature`
also this line seems too long and should be broken up

src/contracts/middleware/BLSSignatureChecker.sol
ChaoticWalrus: this should also have explicit visibility marked. 'internal' seems fine

src/contracts/libraries/BN254.sol
```diff
@@ -127,6 +131,49 @@ library BN254 {
...
+        // if s is 1 return p
+        if(s == 1) {
+            return p;
+        }
+
+        // the accumulated product to return
+        BN254.G1Point memory acc = BN254.G1Point(0, 0);
+        // the 2^n*p to add to the accumulated product in each iteration
+        BN254.G1Point memory p2n = p;
+        // value of most signifigant bit
```
ChaoticWalrus, line 152: typo: `signifigant` instead of `significant`. this typo is repeated (should search-and-replace)

src/contracts/libraries/BN254.sol
```diff
@@ -127,6 +131,49 @@ library BN254 {
         require(success, "ec-add-failed");
     }
 
+    /**
+     * @notice an optimized ecMul implementation that takes O(log_2(s)) ecAdds
+     * @param p the point to multiply
+     * @param s the scalar to multiply by
+     * @dev this function is only safe to use if the scalar is 9 bits or less
+     */ 
+    function scalar_mul_tiny(BN254.G1Point memory p, uint16 s) internal view returns (BN254.G1Point memory) {
```
ChaoticWalrus, line 140: function makes sense -- do we have test coverage ensuring equivalence with ecMul ?
if not, can you please file an issue for adding this test coverage?

src/contracts/libraries/BN254.sol
```diff
@@ -127,6 +131,49 @@ library BN254 {
...
+        // value of most signifigant bit
+        uint16 m = 1;
+        // index of most signifigant bit
+        uint8 i = 0;
+
+        //loop until we reach the most signifigant bit
+        while(s > m){
+            unchecked {
+                // if the  current bit is 1, add the 2^n*p to the accumulated product
+                if (s >> i & 1 == 1) {
```
ChaoticWalrus, line 161: suggestion: add parentheses to clarify order-of-operations:
`if (s >> i & 1 == 1) {` => `if ((s >> i) & 1 == 1) {`

src/contracts/libraries/BN254.sol
```diff
@@ -127,6 +131,49 @@ library BN254 {
...
+                    acc = plus(acc, p2n);
+                }
+                // double the 2^n*p for the next iteration
+                p2n = plus(p2n, p2n);
+
+                // increment the index and double the value of the most signifigant bit
+                m <<= 1;
+                ++i;
```
ChaoticWalrus, line 169: so `m` is always equal to `1 << i`, right?
seems like a potential simplification, presumably  just need to make sure i is a larger integer type to avoid overflowing the uint8 type when left-shifting.

src/contracts/libraries/BN254.sol
```diff
@@ -260,11 +307,18 @@ library BN254 {
...
+        BN254.G2Point memory pk
+    ) internal pure returns (bytes32) {
+        return keccak256(abi.encodePacked(pk.X[0], pk.X[1], pk.Y[0], pk.Y[1]));
+    }
 
     /**
      * @notice adapted from https://github.com/HarryR/solcrypto/blob/master/contracts/altbn128.sol
      */
-    function hashToG1(bytes32 _x) internal view returns (uint256, uint256) {
+    function hashToG1(bytes32 _x) internal view returns (G1Point memory) {
```
ChaoticWalrus, line 321: discussed offline with @gpsanant -- we can remove the comment below (`// XXX: Gen Order (n) or Field Order (p) ?`) carried over from the code this function is adapted from; the use of the field order is correct here.

src/contracts/libraries/MiddlewareUtils.sol
```diff
@@ -3,20 +3,20 @@
...
-        uint256 signedStakeSecondQuorum
+        uint32 referenceBlockNumber,
+        bytes32[] memory nonSignerPubkeyHashes
     ) internal pure returns (bytes32) {
         return keccak256(
-            abi.encodePacked(globalDataStoreId, nonSignerPubkeyHashes, signedStakeFirstQuorum, signedStakeSecondQuorum)
+            abi.encodePacked(referenceBlockNumber, nonSignerPubkeyHashes)
         );
     }
 }
```
ChaoticWalrus, line 22: big question:
*What is this library actually supposed to be?*
Feels like it contains a single constant that belongs probably in the VoteWeigher, and a function that belongs in the BLSSignatureChecker.
I recommend moving these functions to their proper locations and deleting this library.

ChaoticWalrus (changes requested):
I believe I've looked at all the new / relevant contracts, and have added a bunch of comments. Things are looking good especially for the size of this PR.
So far I haven't really focused on scripts or tests; would appreciate more guidance if you want significant review on these.
Note: I also gave only light review to the BLSOperatorStateRetriever and the ServiceManagerBase contracts, since I understood the BLSOperatorStateRetriever to be easily redeployable / significantly lower-risk due to being essentially a utility for reading info off-chain, and the ServiceManagerBase to be merely an 'example'  -- please inform me if either of these deserve the same level of scrutiny as the other contracts.

src/contracts/middleware/StakeRegistryStorage.sol
ChaoticWalrus: looks like the `minimumStakeForQuorum` array actually gets packed into 128 slots, since it's using uint96s instead of uint256s. will make a commit to fix the gap size here.

src/contracts/middleware/StakeRegistry.sol
stevennevins, line 250: We need additional checks on the input arrays here. We need to guarantee that each element at an index across each of the lists is associated and can't be swapped by the caller

src/contracts/middleware/StakeRegistry.sol
ChaoticWalrus: > seems like this will fail with underflow rn (?) on this line:

I'm wrong, it'll do `(length - (length - 1) - 1) = 0`

so I guess the behavior here is returning index 0 if the block number is from before the quorum existed. seems fine, but intention should still be documented

src/contracts/middleware/StakeRegistry.sol
ChaoticWalrus: good point, right now there's no guarantee that the `operators` and `operatorIds` inputs are even associated

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,185 @@
...
+        bytes32 pubkeyHash = BN254.hashG1Point(pubkey);
+
+        require(pubkeyCompendium.pubkeyHashToOperator(pubkeyHash) == operator,"BLSPubkeyRegistry.registerOperator: operator does not own pubkey");
+
+        // update each quorum's aggregate pubkey
+        _processQuorumApkUpdate(quorumNumbers, pubkey.negate());
+        
+        emit OperatorRemovedFromQuorums(operator, quorumNumbers);
+        return pubkeyHash;
+    }
```
ChaoticWalrus, line 89: in the only existing usage of this function (`BLSPubkeyRegistry. deregisterOperator`) -- inside of `BLSRegistryCoordinatorWithIndices._deregisterOperatorWithCoordinator` -- we are not using nor checking the return value of this function. Should this function maybe just not return anything?

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,185 @@
...
+
+    constructor(
+        IRegistryCoordinator _registryCoordinator,
+        IBLSPublicKeyCompendium _pubkeyCompendium
+    ){
+        registryCoordinator = _registryCoordinator;
+        pubkeyCompendium = _pubkeyCompendium;
+    }
+
```
ChaoticWalrus, line 38: is this contract supposed to be upgradeable? seems like yes? @gpsanant
if so, we should add:

- inherit from Initializable
- disable initializers in constructor
- add a storage gap

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -0,0 +1,263 @@
...
+        require(msg.sender == address(registryCoordinator), "IndexRegistry.onlyRegistryCoordinator: caller is not the registry coordinator");
+        _;
+    }
+
+    constructor(
+        IRegistryCoordinator _registryCoordinator
+    ){
+        registryCoordinator = _registryCoordinator;
+    }
+
```
ChaoticWalrus, line 34: similar question to BLSPubkeyRegistry -- is this contract supposed to be upgradeable or not?

ChaoticWalrus: confirmed: this should be upgradeable.

src/contracts/middleware/BLSPubkeyRegistry.sol
ChaoticWalrus: confirmed: this should be upgradeable.

src/contracts/libraries/BitmapUtils.sol
```diff
@@ -271,4 +265,14 @@ library BytesArrayBitmaps {
         }
         return bytesArray;
     }
+
+    /// @return count number of ones in binary representation of `n`
+    function countNumOnes(uint256 n) public pure returns (uint16) {
```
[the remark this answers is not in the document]
ChaoticWalrus: added in commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/722e9c67d5d7eb78c3fac5b6acc050e9c055308d

src/contracts/middleware/BLSSignatureChecker.sol
ChaoticWalrus: resolved https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/7dec57d13c4154092cffb3c92bebf98e0e0199ad

src/contracts/libraries/BN254.sol
ChaoticWalrus: resolved https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/7dec57d13c4154092cffb3c92bebf98e0e0199ad

src/contracts/libraries/BN254.sol
ChaoticWalrus: resolved https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/056c33ebe34279a5c4a7e66e81554d289674994d

src/contracts/middleware/StakeRegistry.sol
ChaoticWalrus: resolved https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/147fb31086a64a2ecf150c7283ecbd4cf64d2d37

src/contracts/middleware/BLSSignatureChecker.sol
ChaoticWalrus: resolved https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/53db6100c35ef1e3b3b7f411ccf81749f45dd2f2

src/contracts/middleware/BLSSignatureChecker.sol
ChaoticWalrus: opened a small PR for this https://github.com/Layr-Labs/eigenlayer-contracts/pull/213

src/contracts/middleware/StakeRegistry.sol
```diff
@@ -0,0 +1,561 @@
...
+
+        _afterDeregisterOperator(operatorId, quorumNumbers);
+    }
+
+    /**
+     * @notice Used for updating information on deposits of nodes.
+     * @param operators are the addresses of the operators whose stake information is getting updated
+     * @dev reverts if there are no operators registered with index out of bounds
+     */
+    function updateStakes(address[] calldata operators) external {
```
stevennevins, line 284: need a test to verify the happy path for this function

src/contracts/middleware/StakeRegistryStorage.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
+
+import "@openzeppelin/contracts/utils/math/Math.sol";
```
stevennevins, line 4: This can be removed. It doesn't appear to be used anywhere

src/test/Delegation.t.sol
```diff
@@ -4,23 +4,28 @@ pragma solidity =0.8.12;
 import "@openzeppelin/contracts/utils/math/Math.sol";
 import "@openzeppelin/contracts/utils/Address.sol";
```
stevennevins, line 5: Address and Math imports can be removed

src/contracts/middleware/StakeRegistry.sol
stevennevins: bumping this one @0x0aa0

ChaoticWalrus:
@ChaoticWalrus and @gpsanant to do another pass of review on this and try to move towards finalizing merging.

src/contracts/middleware/BLSSignatureChecker.sol
ChaoticWalrus: PR was merged, resolving this now.

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: resolved here https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/163c0189fbcf3a1a6bf3b9175d68df085cb61bec

gpsanant: yes you're right! @0x0aa0 make sure you see this

src/contracts/middleware/BLSPubkeyRegistry.sol
ChaoticWalrus: this was resolved with PR https://github.com/Layr-Labs/eigenlayer-contracts/pull/216

src/contracts/middleware/IndexRegistry.sol
ChaoticWalrus: this was resolved with PR https://github.com/Layr-Labs/eigenlayer-contracts/pull/216

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: we store the total number of operators, but we don't store a list of operators

we store (address => index) not (index => address)

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: > the arrays don't need to be the same length? it just checks that the entry in the array matches with the quorum that has met the cap?

But it checks the `i`-th entry in the array of `operatorKickParams`. So like, if I'm registering for 12 quorums and need to kick someone only in the 12th quorum, I need to submit `operatorKickParams` of length 12, because this code will look at the 12th entry in the array.
I think this is "OK", just perhaps a tad counterintuitive(?)

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: ahhh, gotcha! OK makes sense. We know the max indices but not the corresponding operator IDs 👍

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: @gpsanant see the above comment -- I think the concern about ending up with multiple entries in the bitmap history that have `nextUpdateBlockNumber = 0` is the most serious here.

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: seems like this particular race condition could be fixed by setting `quorumsToRemoveBitmap = quorumBitmapBeforeUpdate & quorumsToRemoveBitmap` (i.e. filtering out quorums that the ejected operator isn't in), rather than the strict require statement here.
Going to recommend that change, since I think the race condition is ~low-medium bad as-is.
@0x0aa0 friendly ping on this

src/contracts/middleware/VoteWeigherBase.sol
ChaoticWalrus: @gpsanant any input on the proper event behavior here? I gave my suggestion above.

gpsanant: your suggestion sounds good

ChaoticWalrus: working on this now.

ChaoticWalrus: @gpsanant please check out the state VoteWeigher events as of https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/f26ee275f77d12bf82e2b27d331e396de052c039 and resolve this thread if you're satisfied

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
0x0aa0: @gpsanant

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: `// the first part of this if statement is met if the operator has never registered. the second part is met if the operator has previously registered, but is currently deregistered`

src/contracts/middleware/StakeRegistry.sol
ChaoticWalrus: talked about this offline. seems like the desired behavior here is _reversion_. We should document and implement this.

ChaoticWalrus: I'll look and see if there are other similar instances.

src/contracts/middleware/VoteWeigherBaseStorage.sol
ChaoticWalrus: conclusion: remove this storage

src/contracts/middleware/IndexRegistry.sol
ChaoticWalrus: talked offline, this is technically acceptable.

src/contracts/interfaces/ISocketUpdater.sol
```diff
@@ -0,0 +1,25 @@
...
+import "./IIndexRegistry.sol";
+
+/**
+ * @title Interface for an `ISocketUpdater` where operators can update their sockets.
+ * @author Layr Labs, Inc.
+ */
+interface ISocketUpdater {
+    // EVENTS
+
+    event OperatorSocketUpdate(bytes32 operatorId, string socket);
```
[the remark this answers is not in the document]
ChaoticWalrus: agreed offline this makes sense to be tagged as 'indexed'

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: I think my suggestion might be broken, since we pass on the input from which `quorumsToRemoveBitmap` is derived -- might need to do something like
```
if( quorumBitmapBeforeUpdate & quorumsToRemoveBitmap != quorumsToRemoveBitmap ) {
 use the BitmapUtils library to take the quorumBitmapBeforeUpdate and derive the appropriate 'bytes' object, then pass that on.
}
```

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: great point. updates here: https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/cf6e343b6542825ece78995d7a3e47284f5ef3b5

src/contracts/middleware/StakeRegistry.sol
```diff
@@ -0,0 +1,461 @@
...
+        return operatorIdToStakeHistory[operatorId][quorumNumber][_getStakeUpdateIndexForOperatorIdForQuorumAtBlockNumber(operatorId, quorumNumber, blockNumber)].stake;
+    }
+    /**
+     * @notice Returns the stake weight from the latest entry in `_totalStakeHistory` for quorum `quorumNumber`.
+     * @dev Will revert if `_totalStakeHistory[quorumNumber]` is empty.
+     */
+    function getCurrentTotalStakeForQuorum(uint8 quorumNumber) external view returns (uint96) {
+        // no chance of underflow / error in next line, since an empty entry is pushed in the constructor
+        return _totalStakeHistory[quorumNumber][_totalStakeHistory[quorumNumber].length - 1].stake;
+    }
```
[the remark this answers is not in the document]
ChaoticWalrus: fixed here https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/3e8498797062cfebb869ba83092bad9e0bf890ec

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: Great find again. Here: https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/b9549ac23785fde8d3abca15702f2ee73fbfddd7

ChaoticWalrus: LGTM, thanks.

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: LGTM, ty ty. And you're happy with the events emitted here?

src/contracts/middleware/StakeRegistry.sol
ChaoticWalrus: > I'll look and see if there are other similar instances.

AFAICT there aren't other instances of similar quirky edge-case behavior in this PR. Similar functions seem to all either revert in the same kind of scenario, or return a special index + explain (in comments) that they will return a special index in these edge cases.

src/contracts/middleware/StakeRegistryStorage.sol
```diff
@@ -0,0 +1,40 @@
...
+        IStrategyManager _strategyManager,
+        IServiceManager _serviceManager
+    ) VoteWeigherBase(_strategyManager, _serviceManager)
+    // solhint-disable-next-line no-empty-blocks
+    {
+        registryCoordinator = _registryCoordinator;
+    }
+
+    // storage gap for upgradeability
+    uint256[65] private __GAP;
```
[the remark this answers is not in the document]
ChaoticWalrus: going to add a comment that disables the detector for this particular instance, because this gap variable is working as intended here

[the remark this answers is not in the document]
ChaoticWalrus: should be resolved by https://github.com/Layr-Labs/eigenlayer-contracts/pull/35/commits/1be4928c43a22d8190f0621c32d39b5a153399a6

src/contracts/middleware/StakeRegistry.sol
0x0aa0: 👀

gpsanant:
@ChaoticWalrus could you check this one failing delegation test? seems like a frozen check was taken out of the DM?

ChaoticWalrus:
> @ChaoticWalrus could you check this one failing delegation test? seems like a frozen check was taken out of the DM?

Test should be deleted, since the call was removed. See PR https://github.com/Layr-Labs/eigenlayer-contracts/pull/243

PR #39, opened 2023-05-16
remove SonarCloud

ChaoticWalrus:
We aren't getting value out of this, it is purely causing CI to fail.


Sidu28 (approved):
This is a much needed change

PR #40, opened 2023-05-17
Small gas optimization

haythemsellami:
- Use `token` passed as arg instead of reading from storage (One less SLOAD)
- Check if staker is delegator before calling to increase delegate shares

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -150,23 +150,20 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         external
         onlyStrategyManager
     {
-        //if the staker is delegated to an operator
-        if (isDelegated(staker)) {
-            address operator = delegatedTo[staker];
+        address operator = delegatedTo[staker];
 
-            // add strategy shares to delegate's shares
-            operatorShares[operator][strategy] += shares;
+        // add strategy shares to delegate's shares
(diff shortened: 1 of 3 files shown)


haythemsellami:
Added another commit, to check if staker delegated to an operator before calling the `increaseDelegatedShares()`

ChaoticWalrus:
First commit seems like a decent change, thanks for the PR. Timing isn't the best with our planned audit/release schedule but we can consider for future inclusion 👍 
I think the latter change you added is redundant, but we could clarify the behavior in the future, or perhaps rename the function something like `increaseDelegatedSharesIfDelegated`, or add this change as a redundant 'safety' measure.

haythemsellami:
> First commit seems like a decent change, thanks for the PR. Timing isn't the best with our planned audit/release schedule but we can consider for future inclusion 👍 I think the latter change you added is redundant, but we could clarify the behavior in the future, or perhaps rename the function something like `increaseDelegatedSharesIfDelegated`, or add this change as a redundant 'safety' measure.

Yeah that make sense. For the `increaseDelegatedShares()`, I think checking if staker is a delegator before calling the functions is a little bit gas efficient, but yeah maybe it is not that super important. Other than that, I guess it is better to either rename the function or only call it when it will for sure update the contract state (update the operator shares).

ChaoticWalrus:
Closing this out due to staleness. Still really appreciate you submitting the PR and discussing it though!

PR #44, opened 2023-05-19
Add event checks to unit tests

ChaoticWalrus:
This PR is still a bit of a WIP.
It is OK to merge (all tests currently passing, pure improvement), but it still needs coverage added to the StrategyBase tests, at least.

### DIFF
--- a/src/test/EigenPod.t.sol
+++ b/src/test/EigenPod.t.sol
@@ -243,7 +243,8 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
 
     function testStaking() public {
         cheats.startPrank(podOwner);
-        cheats.expectEmit(true, false, false, false);
+        IEigenPod newPod = eigenPodManager.getPod(podOwner);
+        cheats.expectEmit(true, true, true, true, address(newPod));
         emit EigenPodStaked(pubkey);
         eigenPodManager.stake{value: stakeAmount}(pubkey, signature, depositDataRoot);
         cheats.stopPrank();
@@ -254,7 +255,7 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
         IEigenPod pod = eigenPodManager.getPod(podOwner);
         require(pod.hasRestaked() == false, "Pod should not be restaked");
 
-        //simulate a withdrawal
+        // simulate a withdrawal
         cheats.deal(address(pod), stakeAmount);
         cheats.startPrank(podOwner);
         cheats.expectEmit(true, true, true, true, address(delayedWithdrawalRouter));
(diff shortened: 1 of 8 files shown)


src/test/EigenPod.t.sol
```diff
@@ -58,10 +58,49 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
     bytes32[] withdrawalFields;
     bytes32[] validatorFields;
 
+
+    // EIGENPODMANAGER EVENTS
+    /// @notice Emitted to notify the update of the beaconChainOracle address
+    event BeaconOracleUpdated(address indexed newOracleAddress);
```
gpsanant, line 64: Is it possible to import the events like in the examples here: https://book.getfoundry.sh/cheatcodes/expect-emit

src/test/unit/DelayedWithdrawalRouterUnit.t.sol
```diff
@@ -131,7 +143,13 @@ contract DelayedWithdrawalRouterUnitTests is Test {
         delayedWithdrawalRouter.createDelayedWithdrawal{value: delayedWithdrawalAmount}(podOwner, address(0));
     }
 
-    function testClaimDelayedWithdrawals(uint8 delayedWithdrawalsToCreate, uint8 maxNumberOfDelayedWithdrawalsToClaim, uint256 pseudorandomNumber_, address recipient, bool useOverloadedFunction)
+    function testClaimDelayedWithdrawals(
+        uint8 delayedWithdrawalsToCreate,
+        uint8 maxNumberOfDelayedWithdrawalsToClaim,
+        uint256 pseudorandomNumber_,
```
gpsanant, line 149: looked into the file and doesn't seem like we use this param?

gpsanant: looks like we use the _getPseudorandomNumber function though

src/test/EigenPod.t.sol
```diff
@@ -614,6 +664,8 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
         eigenPodManager.stake{value: 12 ether}(_pubkey, _signature, _depositDataRoot);
         
         // successful call
+        cheats.expectEmit(true, false, false, false);
```
gpsanant, line 667: no address check?

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -463,6 +525,10 @@ contract StrategyManagerUnitTests is Test, Utils {
         cheats.startPrank(strategyManager.owner());
         IStrategy[] memory _strategy = new IStrategy[](1);
         _strategy[0] = IStrategy(address(reenterer));
+        for (uint256 i = 0; i < _strategy.length; ++i) {
```
gpsanant, line 528: these 5 lines are used a lot, feel like we could make them another function

src/test/EigenPod.t.sol
ChaoticWalrus: pretty sure I just copy-pasted the existing call to `cheats.expectEmit`. I'll fix it everywhere

src/test/unit/DelayedWithdrawalRouterUnit.t.sol
ChaoticWalrus: we are using indeed using this input. we set the storage `_pseudorandomNumber` equal to the `pseudorandomNumber_`. The storage value is then used inside the `_getPseudorandomNumber` function.

src/test/EigenPod.t.sol
ChaoticWalrus: can't sort this out, not sure it's possible how you're picturing -- the foundry docs are unclear to me.
if you can sort this out then go right ahead.

src/test/EigenPod.t.sol
ChaoticWalrus: fixed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/44/commits/b9be2206e32fd06a43963152bc5e128eb6138a17

src/test/unit/StrategyManagerUnit.t.sol
ChaoticWalrus: we use this 4 times. I think it is OK to not make a function for now.

src/test/EigenPod.t.sol
gpsanant: not a big deal, just will error out if we forget to propagate.

src/test/unit/StrategyManagerUnit.t.sol
gpsanant: sounds good

gpsanant (approved):
thanks for adding this, huge for our subgraph correctness

ChaoticWalrus:
Leaving the branch open for the moment, so I can add more coverage and open another PR.

PR #47, opened 2023-05-23
adding BLSPubkeyRegistry

Sidu28:
Adds a separate registry for managing middleware operators' BLS public keys.


src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+import "../interfaces/IBLSPublicKeyCompendium.sol";
+
+import "../libraries/BN254.sol";
+
+import "forge-std/Test.sol";
+
+
+contract BLSPubkeyRegistry is IBLSPubkeyRegistry, Test {
+
+    bytes32 internal constant ZERO_PK_HASH = hex"ad3228b676f7d3cd4284a5443f17f1962b36e491b30a40b2405849e597ba5fb5";
```
gpsanant, line 16: comments on these variables would be nice. a good amount can be replicated from the interface

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+    IBLSPublicKeyCompendium public immutable pubkeyCompendium;
+
+
+    ApkUpdate[] public globalApkUpdateList;
+
+    mapping(uint8 => ApkUpdate[]) public quorumToApkUpdates;
+    
+    mapping(uint8 => BN254.G1Point) public quorumToApk;
+
+    event RegistrationEvent(
```
gpsanant, line 30: So I think the events should be:
```solidity
event PubkeyAdded(
     address operator,
     BN254.G1Point pubkey
)
```
and the RegistryCoordinator will emit whatever the quorumBitmap is so indexers can keep track of offchain state.

gpsanant: And an analagous event for 
```solidity
event PubkeyRemoved(
    address operator,
    BN254.G1Point pubkey
)
```

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+        IBLSPublicKeyCompendium _pubkeyCompendium
+    ){
+        registryCoordinator = _registryCoordinator;
+        pubkeyCompendium = _pubkeyCompendium;
+
+        _initializeApkUpdates();
+    }
+
+
+    function registerOperator(address operator, uint8[] memory quorumNumbers, BN254.G1Point memory pubkey) external onlyRegistryCoordinator returns(bytes32){
```
gpsanant, line 59: comments from interfaces need to be brought over

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+        bytes32 pubkeyHash = BN254.hashG1Point(pubkey);
+        require(pubkeyHash != ZERO_PK_HASH, "BLSRegistry._registerOperator: cannot register zero pubkey");
+        require(quorumNumbers.length > 0, "BLSRegistry._registerOperator: must register for at least one quorum");
+        require(pubkeyCompendium.pubkeyHashToOperator(pubkeyHash) == operator,"BLSRegistry._registerOperator: operator does not own pubkey");
+        _processQuorumApkUpdate(quorumNumbers, pubkey, true);
```
gpsanant, line 64: comments please

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+        bytes32 globalApkHash = _processGlobalApkUpdate(BN254.plus(_globalApk, pubkey));
```
gpsanant, line 65: If you add [this line](https://github.com/Layr-Labs/eigenlayer-contracts/blob/7f96a3a057e53004bf5c85dca447432c846934dc/src/contracts/middleware/BLSSignatureChecker.sol#L14) to the top of the file, you can get [this sexy syntax](https://github.com/Layr-Labs/eigenlayer-contracts/blob/7f96a3a057e53004bf5c85dca447432c846934dc/src/contracts/middleware/BLSSignatureChecker.sol#L103)

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+    function getApkUpdateForQuorumByIndex(uint8 quorumNumber, uint256 index) external view returns (ApkUpdate memory){
+        return quorumToApkUpdates[quorumNumber][index];
+    }
+
+    /// @notice returns the `ApkUpdate` struct at `index` in the list of APK updates that keep track of the sum of the APK amonng all quorums
+    function globalApkUpdates(uint256 index) external view returns (ApkUpdate memory){
+        return globalApkUpdateList[index];
+    }
+
+    function quorumApk(uint8 quorumNumber) external view returns (BN254.G1Point memory){
```
gpsanant, line 98: you could just name the map this if you want, this is fine too though

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+        emit DeregistrationEvent(operator, pubkeyHash, globalApkHash);
+    }
+
+    /// @notice returns the `ApkUpdate` struct at `index` in the list of APK updates for the `quorumNumber`
+    function getApkUpdateForQuorumByIndex(uint8 quorumNumber, uint256 index) external view returns (ApkUpdate memory){
+        return quorumToApkUpdates[quorumNumber][index];
+    }
+
+    /// @notice returns the `ApkUpdate` struct at `index` in the list of APK updates that keep track of the sum of the APK amonng all quorums
+    function globalApkUpdates(uint256 index) external view returns (ApkUpdate memory){
```
gpsanant, line 94: we should just name the list this and make it public

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+        return _globalApk;
+    }
+
+
+    /**
+     * @notice get hash of the apk of `quorumNumber` at `blockNumber` using the provided `index`;
+     * called by checkSignatures in BLSSignatureChecker.sol.
+     */
+    function getApkHashForQuorumAtBlockNumberFromIndex(uint8 quorumNumber, uint32 blockNumber, uint256 index) external view returns (bytes32){
+        ApkUpdate memory quorumApkUpdate = quorumToApkUpdates[quorumNumber][index];
```
gpsanant, line 112: Could you use the pattern shown [here](https://github.com/Layr-Labs/eigenlayer-contracts/blob/7f96a3a057e53004bf5c85dca447432c846934dc/src/contracts/middleware/RegistryBase.sol#L211) and [here](https://github.com/Layr-Labs/eigenlayer-contracts/blob/7f96a3a057e53004bf5c85dca447432c846934dc/src/contracts/middleware/RegistryBase.sol#L683). Way cleaner IMO.

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+        ApkUpdate memory latestGlobalApkUpdate;
+        latestGlobalApkUpdate.apkHash = globalApkHash;
+        latestGlobalApkUpdate.updateBlockNumber = uint32(block.number);
+        globalApkUpdateList.push(latestGlobalApkUpdate);
+
+        return globalApkHash;
+    }
+
+    function _processQuorumApkUpdate(uint8[] memory quorumNumbers, BN254.G1Point memory pubkey, bool isRegistration) internal {
+        BN254.G1Point memory latestApk;
```
gpsanant, line 161: `apkBeforeUpdate` and `apkAfterUpdate` are clearer names

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+            } else {
+                latestApk = BN254.plus(currentApk, BN254.negate(pubkey));
+            }
+
+            //update aggregate public key for this quorum
+            quorumToApk[quorumNumber] = latestApk;
+            //update nextUpdateBlockNumber of the current latest ApkUpdate
+            quorumToApkUpdates[quorumNumber][quorumToApkUpdates[quorumNumber].length - 1].nextUpdateBlockNumber = uint32(block.number);
+            //create new ApkUpdate to add to the mapping
+            ApkUpdate memory latestApkUpdate;
```
gpsanant, line 182: nit: you could declare this in one line

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+
+        ApkUpdate memory latestGlobalApkUpdate;
+        latestGlobalApkUpdate.apkHash = globalApkHash;
+        latestGlobalApkUpdate.updateBlockNumber = uint32(block.number);
+        globalApkUpdateList.push(latestGlobalApkUpdate);
+
+        return globalApkHash;
+    }
+
+    function _processQuorumApkUpdate(uint8[] memory quorumNumbers, BN254.G1Point memory pubkey, bool isRegistration) internal {
```
gpsanant, line 160: i feel like a cleaner method of doing this is renaming `pubkey` to `point` and just negating the pubkey and inputting it during deregistration and keeping it as the pubkey during registration, and removing the `isRegistration` param. This abstracts the most logic and makes this unit simpler.

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+
+    function getQuorumApkHistoryLength(uint8 quorumNumber) external view returns(uint32){
+        return uint32(quorumToApkUpdates[quorumNumber].length);
+    }
+
+    function getGlobalApkHistoryLength() external view returns(uint32){
+        return uint32(globalApkUpdateList.length);
+    }
+
+    function _processGlobalApkUpdate(BN254.G1Point memory newGlobalApk) internal returns(bytes32){
```
gpsanant, line 145: i find it a bit weird that this param is the new apk becuase we don't do that in the quorum function? inputting the point to add to the globalApkBeforeUpdate is cleaner. See https://github.com/Layr-Labs/eigenlayer-contracts/pull/47/files#r1207099006

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+        bytes32 globalApkHash = BN254.hashG1Point(newGlobalApk);
+        if(globalApkUpdateList.length > 0){
```
gpsanant, line 147: you don't do this check below because of the initialization function. I'm pro keeping this check, getting rid of the intiialization, and adding the check to the quorum specific function as well

Sidu28: ah yeah, I added the initialization after and didn't get rid of this.  I'm fine either way, so no initialization it is

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,211 @@
...
+import "../interfaces/IBLSPubkeyRegistry.sol";
+import "../interfaces/IRegistryCoordinator.sol";
+import "../interfaces/IBLSPublicKeyCompendium.sol";
+
+import "../libraries/BN254.sol";
+
+import "forge-std/Test.sol";
+
+
+contract BLSPubkeyRegistry is IBLSPubkeyRegistry, Test {
```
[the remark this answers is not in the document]
Sidu28: probably will keep until we implement unit tests and merge the PR

src/contracts/middleware/BLSPubkeyRegistry.sol
Sidu28: done

src/contracts/middleware/BLSPubkeyRegistry.sol
Sidu28: added

src/contracts/middleware/BLSPubkeyRegistry.sol
Sidu28: yup

src/test/unit/BLSPubkeyRegistryUnit.t.sol
```diff
@@ -0,0 +1,211 @@
...
+
+    function testDeregisteringWithNoQuorums() public {
+        cheats.startPrank(address(registryCoordinator));
+        cheats.expectRevert(bytes("BLSRegistry._deregisterOperator: must register for at least one quorum"));
+        blsPubkeyRegistry.deregisterOperator(defaultOperator, new uint8[](0), BN254.G1Point(1, 0));
+        cheats.stopPrank();
+    }
+
+    function testRegisterOperatorBLSPubkey(address operator) public {
+        bytes32 pkHash = BN254.hashG1Point(defaultPubKey);
```
gpsanant, line 92: how feel safer if we had a random uint256 input which we hashed to a g1 pubkey as the public key

src/test/unit/BLSPubkeyRegistryUnit.t.sol
```diff
@@ -0,0 +1,211 @@
...
+        cheats.startPrank(address(registryCoordinator));
+        bytes32 registeredpkHash = blsPubkeyRegistry.registerOperator(operator, quorumNumbers, negatedQuorumApk);
+        cheats.stopPrank();
+
+        BN254.G1Point memory zeroPk = BN254.G1Point(0,0);
+        require(BN254.hashG1Point(blsPubkeyRegistry.quorumApk(defaulQuorumNumber)) == ZERO_PK_HASH, "quorumApk not set correctly");
+    }
+    
+    function testQuorumApkUpdatesDeregistration() external {
+        uint8[] memory quorumNumbers = new uint8[](2);
```
gpsanant, line 190: do we need to fuzz over quorum numbers? maybe for one happy case test?

src/test/unit/BLSPubkeyRegistryUnit.t.sol
Sidu28: do you mean fuzz the actual G1Point, as in x and y?

gpsanant: yep

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,213 @@
...
+     * @dev access restricted to the RegistryCoordinator
+     * @dev Preconditions:
+     *         1) `quorumNumbers` has no duplicates
+     *         2) `quorumNumbers.length` != 0
+     *         3) `quorumNumbers` is ordered in ascending order
+     *         4) the operator is not already deregistered
+     *         5) `quorumNumbers` is the same as the parameter use when registering
+     *         6) `pubkey` is the same as the parameter used when registering
+     */   
+    function deregisterOperator(address operator, uint8[] memory quorumNumbers, BN254.G1Point memory pubkey) external onlyRegistryCoordinator returns(bytes32){
```
gpsanant, line 97: The quorum numbers will be formatted as such: https://github.com/Layr-Labs/eigenlayer-contracts/blob/1fb084c35454e8fb4a453ab6a2e529667bd36584/src/contracts/middleware/StakeRegistry.sol#L299

Sidu28: done

src/contracts/middleware/BLSPubkeyRegistry.sol
gpsanant: way more expensive for not that much more ugliness?

src/test/unit/BLSPubkeyRegistryUnit.t.sol
gpsanant: fair enough

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,210 @@
...
+        require(pubkeyHash != ZERO_PK_HASH, "BLSRegistry._registerOperator: cannot register zero pubkey");
+        require(quorumNumbers.length > 0, "BLSRegistry._registerOperator: must register for at least one quorum");
+        //ensure that the operator owns their public key by referencing the BLSPubkeyCompendium
+        require(pubkeyCompendium.pubkeyHashToOperator(pubkeyHash) == operator,"BLSRegistry._registerOperator: operator does not own pubkey");
+        // update each quorum's aggregate pubkey
+        _processQuorumApkUpdate(quorumNumbers, pubkey);
+        // update the global aggregate pubkey
+        _processGlobalApkUpdate(pubkey);
+        // emit event so offchain actors can update their state
+        emit PubkeyAdded(operator, pubkey);
```
gpsanant, line 80: actually, we likely need the quorum numbers as well

gpsanant: oops just realized we do not

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,210 @@
...
+     *         1) `quorumNumbers` has no duplicates
+     *         2) `quorumNumbers.length` != 0
+     *         3) `quorumNumbers` is ordered in ascending order
+     *         4) the operator is not already registered
+     */
+    function registerOperator(address operator, bytes memory quorumNumbers, BN254.G1Point memory pubkey) external onlyRegistryCoordinator returns(bytes32){
+        //calculate hash of the operator's pubkey
+        bytes32 pubkeyHash = BN254.hashG1Point(pubkey);
+
+        require(pubkeyHash != ZERO_PK_HASH, "BLSRegistry._registerOperator: cannot register zero pubkey");
```
gpsanant, line 71: these error messages are wrong

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,210 @@
...
+        require(quorumNumbers.length > 0, "BLSRegistry._registerOperator: must register for at least one quorum");
```
gpsanant, line 72: this is already a precondition, doesn't need to be checked

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,210 @@
...
+        BN254.G1Point memory apkBeforeUpdate;
+        BN254.G1Point memory apkAfterUpdate;
+
+        for (uint8 i = 0; i < uint8(quorumNumbers.length);) {
+            uint8 quorumNumber = uint8(quorumNumbers[i]);
+
+            uint256 quorumApkUpdatesLength = quorumApkUpdates[quorumNumber].length;
+            if (quorumApkUpdatesLength > 0) {
+                // update nextUpdateBlockNumber of the current latest ApkUpdate
+                quorumApkUpdates[quorumNumber][quorumApkUpdates[quorumNumber].length - 1].nextUpdateBlockNumber = uint32(block.number);
```
gpsanant, line 179: can use `quorumApkUpdatesLength` here

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,209 @@
...
+     *         1) `quorumNumbers` has no duplicates
+     *         2) `quorumNumbers.length` != 0
+     *         3) `quorumNumbers` is ordered in ascending order
+     *         4) the operator is not already registered
+     */
+    function registerOperator(address operator, bytes memory quorumNumbers, BN254.G1Point memory pubkey) external onlyRegistryCoordinator returns(bytes32){
+        //calculate hash of the operator's pubkey
+        bytes32 pubkeyHash = BN254.hashG1Point(pubkey);
+
+        require(pubkeyHash != ZERO_PK_HASH, "BLSRegistry.registerOperator: cannot register zero pubkey");
```
gpsanant, line 71: these error messages are wrong

src/test/unit/BLSPubkeyRegistryUnit.t.sol
```diff
@@ -0,0 +1,384 @@
...
+        cheats.assume(blockGap < 100);
+
+        BN254.G1Point memory quorumApk = BN254.G1Point(0,0);
+        bytes32 quorumPkHash;
+        for (uint256 i = 0; i < numRegistrants; i++) {
+            bytes32 pk = _getRandomPk(i);
+            testRegisterOperatorBLSPubkey(defaultOperator, pk);
+            quorumApk = quorumApk.plus(BN254.hashToG1(pk));
+            quorumPkHash = BN254.hashG1Point(quorumApk);
+            require(quorumPkHash == blsPubkeyRegistry.getApkHashForQuorumAtBlockNumberFromIndex(defaultQuorumNumber, uint32(block.number + blockGap) , i), "incorrect quorum aok updates");
```
gpsanant, line 264: *apk

src/contracts/interfaces/IBLSPubkeyRegistry.sol
```diff
@@ -20,44 +20,61 @@ interface IBLSPubkeyRegistry is IRegistry {
...
+     * @param operator The address of the operator to deregister.
+     * @param quorumNumbers The quourm numbers the operator is deregistering from, where each byte is an 8 bit integer quorumNumber.
+     * @param pubkey The public key of the operator.
+     * @dev access restricted to the RegistryCoordinator
+     * @dev Preconditions:
+     *         1) `quorumNumbers` has no duplicates
+     *         2) `quorumNumbers.length` != 0
+     *         3) `quorumNumbers` is ordered in ascending order
+     *         4) the operator is not already deregistered
+     *         5) `quorumNumbers` is the same as the parameter use when registering
```
ChaoticWalrus, line 47: does this mean there is no way to register for additional quorums later?

src/contracts/interfaces/IBLSPubkeyRegistry.sol
```diff
@@ -20,44 +20,61 @@ interface IBLSPubkeyRegistry is IRegistry {
...
-
     /**
      * @notice get hash of the apk of `quorumNumber` at `blockNumber` using the provided `index`;
      * called by checkSignatures in BLSSignatureChecker.sol.
+     * @param quorumNumber is the quorum whose ApkHash is being retrieved
+     * @param blockNumber is the number of the block for which the latest ApkHash muust be retrieved
+     * @param index is the provided witness of the onchain index calculated offchain
      */
-    function getApkHashForQuorumAtBlockNumberFromIndex(uint8 quorumNumber, uint32 blockNumber, uint256 index) external view returns (bytes32);
+    function getApkHashForQuorumAtBlockNumberFromIndex(uint8 quorumNumber, uint32 blockNumber, uint256 index) external returns (bytes32);
```
ChaoticWalrus, line 65: I'm assuming the change from 'view' to non-view is transient / for logging purposes?

src/contracts/interfaces/IBLSPubkeyRegistry.sol
```diff
@@ -20,44 +20,61 @@ interface IBLSPubkeyRegistry is IRegistry {
...
 
 	/**
-     * @notice get hash of the apk among all quourms at `blockNumber` using the provided `index`;
+     * @notice get hash of the apk among all quourums at `blockNumber` using the provided `index`;
```
ChaoticWalrus, line 68: this is still misspelled 🙃  should be 'quorums'

src/contracts/interfaces/IBLSPubkeyRegistry.sol
```diff
@@ -20,44 +20,61 @@ interface IBLSPubkeyRegistry is IRegistry {
...
      * called by checkSignatures in BLSSignatureChecker.sol.
+     * @param blockNumber is the number of the block for which the latest ApkHash muust be retrieved
```
ChaoticWalrus, line 70: another typo: 'muust' => 'must'
I would suggest 'will' or similar though; 'must' seems very imperative

src/contracts/interfaces/IIndexRegistry.sol
```diff
@@ -31,10 +31,11 @@ interface IIndexRegistry is IRegistry {
      * @notice Deregisters the operator with the specified `operatorId` for the quorums specified by `quorumBitmap`.
      * @param operatorId is the id of the operator that is being deregistered
      * @param quorumNumbers is the quorum numbers the operator is deregistered for
-     * @param indexes is an array of indexes for each quorum as witnesses for the last operators to swap for each quorum
+     * @param quorumToOperatorListIndexes is an array of indexes for each quorum as witnesses for the last operators to swap for each quorum
```
ChaoticWalrus, line 34: I might need this to be worded differently. I realize it's complicated but I'm finding this confusing even knowing what it's supposed to do.

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,208 @@
...
+        _;
+    }
+
+    constructor(
+        IRegistryCoordinator _registryCoordinator,
+        IBLSPublicKeyCompendium _pubkeyCompendium
+    ){
+        registryCoordinator = _registryCoordinator;
+        pubkeyCompendium = _pubkeyCompendium;
+    }
```
ChaoticWalrus, line 53: should perhaps have a wider discussion on if this + other contracts should be upgradeable (or support both upgradeable and non-upgradeable uses?)

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,208 @@
...
+     *         1) `quorumNumbers` has no duplicates
+     *         2) `quorumNumbers.length` != 0
+     *         3) `quorumNumbers` is ordered in ascending order
+     *         4) the operator is not already deregistered
+     *         5) `quorumNumbers` is the same as the parameter use when registering
+     *         6) `pubkey` is the same as the parameter used when registering
+     */   
+    function deregisterOperator(address operator, bytes memory quorumNumbers, BN254.G1Point memory pubkey) external onlyRegistryCoordinator returns(bytes32){
+        bytes32 pubkeyHash = BN254.hashG1Point(pubkey);
+
```
ChaoticWalrus, line 99: am I crazy or is the same check (as compared to the `registerOperator` function) that the operator owns the pubkey missing here?
Really just trying to understand why the `registerOperator` and `deregisterOperator` functions do different checks and whether this is a potential issue or not

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,208 @@
...
+     * @param blockNumber is the number of the block for which the latest ApkHash muust be retrieved
+     * @param index is the provided witness of the onchain index calculated offchain
+     */
+    function getApkHashForQuorumAtBlockNumberFromIndex(uint8 quorumNumber, uint32 blockNumber, uint256 index) external returns (bytes32){
+        ApkUpdate memory quorumApkUpdate = quorumApkUpdates[quorumNumber][index];
+        _validateApkHashForQuorumAtBlockNumber(quorumApkUpdate, blockNumber);
+        return quorumApkUpdate.apkHash;
+    }
+
+	/**
```
ChaoticWalrus, line 132: formatting mistake? (too much indent)

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,208 @@
...
+     * @notice get hash of the apk among all quourums at `blockNumber` using the provided `index`;
```
ChaoticWalrus, line 133: typo again: 'quourums' => 'quorums'
I would also recommend changing the wording to 'the global apk' to be consistent, although I think we should also define the 'global apk' more clearly, perhaps in the top-level contract description?

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,208 @@
...
+    function getQuorumApkHistoryLength(uint8 quorumNumber) external view returns(uint32){
+        return uint32(quorumApkUpdates[quorumNumber].length);
+    }
+    /// @notice Returns the length of ApkUpdates for the global APK
+    function getGlobalApkHistoryLength() external view returns(uint32){
+        return uint32(globalApkUpdates.length);
+    }
+
+    function _processGlobalApkUpdate(BN254.G1Point memory point) internal {
+        // load and store in memory in common case we need to access the length again
```
ChaoticWalrus, line 151: errant word? ('common'=>'')

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,208 @@
...
+        uint256 globalApkUpdatesLength = globalApkUpdates.length;
+        // update the nextUpdateBlockNumber of the previous update
+        if (globalApkUpdates.length > 0) {
```
ChaoticWalrus, line 154: immediately not using the thing we just stored in memory, and instead SLOADing again

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,208 @@
...
+        latestGlobalApkUpdate.apkHash = BN254.hashG1Point(globalApk);
+        latestGlobalApkUpdate.updateBlockNumber = uint32(block.number);
+        globalApkUpdates.push(latestGlobalApkUpdate);
+    }
+
+    function _processQuorumApkUpdate(bytes memory quorumNumbers, BN254.G1Point memory point) internal {
+        BN254.G1Point memory apkBeforeUpdate;
+        BN254.G1Point memory apkAfterUpdate;
+
+        for (uint8 i = 0; i < uint8(quorumNumbers.length);) {
```
ChaoticWalrus, line 171: is there a reason the loop incrementor `i` here is a uint8? seems like it increases the risk/potential of an overflow bug without delivering any benefit

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,208 @@
...
+
+            uint256 quorumApkUpdatesLength = quorumApkUpdates[quorumNumber].length;
+            if (quorumApkUpdatesLength > 0) {
+                // update nextUpdateBlockNumber of the current latest ApkUpdate
+                quorumApkUpdates[quorumNumber][quorumApkUpdatesLength - 1].nextUpdateBlockNumber = uint32(block.number);
+            }
+            
+            apkBeforeUpdate = quorumApk[quorumNumber];
+
+            apkAfterUpdate = apkBeforeUpdate.plus(point);
```
ChaoticWalrus, line 182: in this function this step is broken up into multiple lines, while in `_processGlobalApkUpdate`, this operation is combined into a single line.
Single line version seems more efficient (no unnecessary copying to memory) and IMO is at least as readable.

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -0,0 +1,208 @@
...
+
+    function _validateApkHashForQuorumAtBlockNumber(ApkUpdate memory apkUpdate, uint32 blockNumber) internal pure {
+        require(
+            blockNumber >= apkUpdate.updateBlockNumber, 
+            "BLSPubkeyRegistry._validateApkHashForQuorumAtBlockNumber: index too recent"
+        );
+        require(
+            apkUpdate.nextUpdateBlockNumber == 0 || blockNumber < apkUpdate.nextUpdateBlockNumber, 
+            "BLSPubkeyRegistry._validateApkHashForQuorumAtBlockNumber: not latest apk update"
+        );
```
ChaoticWalrus, line 206: seems worth having an in-context comment noting that `nextUpdateBlockNumber` = 0 indicates that the update is the latest one

src/contracts/interfaces/IBLSPubkeyRegistry.sol
Sidu28: yup

src/contracts/interfaces/IBLSPubkeyRegistry.sol
Sidu28: lmao

Sidu28: This is interface, I blame gpsanant

src/contracts/middleware/BLSPubkeyRegistry.sol
Sidu28: whoops

src/contracts/middleware/BLSPubkeyRegistry.sol
Sidu28: I believe the preconditions for the deregister function ensure that the operator provided pubkey will be the same as the one that was registered with.

src/contracts/middleware/BLSPubkeyRegistry.sol
Sidu28: definitely

gpsanant: agreed. I think the pubkeyCompendium should be non-upgradeable because we anticipate many middlewares will use it and it is siloed logic. for the other contracts, we should have more discussion

src/contracts/middleware/BLSPubkeyRegistry.sol
gpsanant: what @Sidu28 is detailing is what i intended as well. would you prefer we did the check here as well? how do you feel about this @ChaoticWalrus ?

src/contracts/interfaces/IBLSPubkeyRegistry.sol
ChaoticWalrus: OK.

src/contracts/middleware/BLSPubkeyRegistry.sol
ChaoticWalrus: I think having the check here as well would be safer. IMO a slightly increased gas cost is worth it for the peace of mind here (also makes it easier to review).

src/contracts/middleware/BLSPubkeyRegistry.sol
ChaoticWalrus: I agree about the PubkeyCompendium
I think the other contracts should be designed + built to support upgradeable and non-upgradeable versions, but I'm OK merging this and making that change later.

ChaoticWalrus (approved):
This looks OK to merge into the multiquorums branch. I'd recommend adding the one check we discussed first.

PR #53, opened 2023-05-31
Unit test vote weigher

gpsanant:
Started a very first draft of VoteWeigher unit tests.

Definitely not in the best shape to be merged. Looking for feedback on how to speed it up (not everything needs to be fuzzed), make it cleaner, more comprehensive, and more understandable.


src/contracts/middleware/VoteWeigherBase.sol
```diff
@@ -60,12 +69,7 @@ contract VoteWeigherBase is VoteWeigherBaseStorage {
 
                 // add the weight from the shares for this strategy to the total weight
                 if (sharesAmount > 0) {
-                    weight += uint96(
-                        (
-                            (strategyAndMultiplier.strategy).sharesToUnderlying(sharesAmount)
-                                * strategyAndMultiplier.multiplier
-                        ) / WEIGHTING_DIVISOR
-                    );
+                    weight += uint96(sharesAmount * strategyAndMultiplier.multiplier / WEIGHTING_DIVISOR);
```
ChaoticWalrus, line 72: discussed this offline and agreed it was an acceptable change.
I also think this _could_ be abstracted into a 'hook' or similar

src/contracts/middleware/VoteWeigherBase.sol
```diff
@@ -100,25 +104,15 @@ contract VoteWeigherBase is VoteWeigherBaseStorage {
...
-        for (uint256 i = 0; i < numStrats;) {
-            // check that the provided index is correct
-            require(
-                strategiesConsideredAndMultipliers[quorumNumber][indicesToRemove[i]].strategy == _strategiesToRemove[i],
-                "VoteWeigherBase.removeStrategiesConsideredAndWeights: index incorrect"
-            );
-
+        require(quorumNumber < quorumCount, "VoteWeigherBase.removeStrategiesConsideredAndMultipliers: quorumNumber is greater than or equal to quorumCount");
+        for (uint256 i = 0; i < indicesToRemove.length;) {
+            emit StrategyRemovedFromQuorum(quorumNumber, strategiesConsideredAndMultipliers[quorumNumber][indicesToRemove[i]].strategy);
```
ChaoticWalrus, line 111: although this is a simplification it also feels someone more 'unsafe'
I think I'm OK with it but not really feeling like it's better to me.

src/contracts/middleware/VoteWeigherBase.sol
```diff
@@ -171,6 +166,7 @@ contract VoteWeigherBase is VoteWeigherBaseStorage {
         StrategyAndWeightingMultiplier[] memory _strategiesConsideredAndMultipliers
     ) internal {
         uint8 quorumNumber = quorumCount;
+        require(quorumNumber < 255, "VoteWeigherBase._createQuorum: number of quorums cannot 256");
```
ChaoticWalrus, line 169: typo: "cannot 256" => "cannot exceed 256" ?

Also shouldn't this be 256 instead of 255? This was the max you'll hit is 255 -- maybe you were thinking of using non-strict inequality in the check when using 255?

Also probably want to define a constant like `MAX_QUORUMS = 256` and then use it instead of a "magic number"

src/contracts/middleware/VoteWeigherBase.sol
```diff
@@ -191,7 +187,8 @@ contract VoteWeigherBase is VoteWeigherBaseStorage {
         uint8 quorumNumber,
         StrategyAndWeightingMultiplier[] memory _newStrategiesConsideredAndMultipliers
     ) internal {
-        require(quorumNumber <= quorumCount, "VoteWeigherBase._addStrategiesConsideredAndMultipliers: quorumNumber exceeds quorumCount");
+        require(quorumNumber < quorumCount, "VoteWeigherBase._addStrategiesConsideredAndMultipliers: quorumNumber exceeds quorumCount");
```
ChaoticWalrus, line 190: maybe make this a modifier? I see it used a couple places

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+            delegationMock,
+            IEigenPodManager(address(uint160(uint256(keccak256(abi.encodePacked("eigenPodManager")))))),
+            ISlasher(address(uint160(uint256(keccak256(abi.encodePacked("slasher"))))))
+        );
+
+        strategyManager = IStrategyManager(address(strategyManagerMock));
+
+        // make the serviceManagerOwner the owner of the serviceManager contract
+        cheats.prank(serviceManagerOwner);
+        serviceManager = IServiceManager(address(new OwnableMock()));
```
ChaoticWalrus, line 71: probably want to `cheats.stopPrank()` after this call?

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+            assertEq(address(strategyAndWeightingMultiplier.strategy), address(strategiesAndWeightingMultipliers[i].strategy));
+            assertEq(strategyAndWeightingMultiplier.multiplier, strategiesAndWeightingMultipliers[i].multiplier);
+        }
+    }
+
+    function testCreateQuorum_FromNotServiceManagerOwner_Reverts(
+        address notServiceManagerOwner,
+        IVoteWeigher.StrategyAndWeightingMultiplier[] memory strategiesAndWeightingMultipliers
+    ) public {
+        cheats.assume(notServiceManagerOwner != serviceManagerOwner);
```
ChaoticWalrus, line 114: you'll also need to assume the `notServiceManagerOwner` is not the `proxyAdmin` address or else you'll get flaky test failures (because . Our other tests define a mapping of addresses to exclude from fuzzed addresses and then define a `fuzzedAddress` or similarly-named modifier, to reuse for this kind of filtering.

src/test/harnesses/VoteWeigherBaseHarness.sol
```diff
@@ -0,0 +1,20 @@
...
+        IServiceManager _serviceManager
+    ) VoteWeigherBase(_strategyManager, _serviceManager) {}
+
+    function getMaxWeighingFunctionLength() public pure returns (uint8) {
+        return MAX_WEIGHING_FUNCTION_LENGTH;
+    }
+
+    function getWeightingDivisor() public pure returns (uint256) {
+        return WEIGHTING_DIVISOR;
+    }
```
ChaoticWalrus, line 19: should these constants maybe just be defined as `public` in the first place?
I'm OK with needing a harness for this, but don't think there's really a very significant downside to having the public default getters exist (technically there is a small function dispatching gas cost increase, but IMO it's fine).

src/test/unit/VoteWeigherBaseUnit.t.sol
ChaoticWalrus: oops nevermind, I'm used to using `startPrank` and `stopPrank` instead of just `prank`.

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+    ) public {
+        strategiesAndWeightingMultipliers = _convertToValidStrategiesAndWeightingMultipliers(strategiesAndWeightingMultipliers);
+            
+        cheats.startPrank(serviceManagerOwner);
+        for (uint i = 0; i < 255; i++) {
+            voteWeigher.createQuorum(strategiesAndWeightingMultipliers);
+        }
+        cheats.expectRevert("VoteWeigherBase._createQuorum: number of quorums cannot 256");
+        voteWeigher.createQuorum(strategiesAndWeightingMultipliers); 
+    }
```
ChaoticWalrus, line 195: this test needs to be corrected to actually check for 256 quorums, rather than 255 (incorrect test matches incorrect implementation).
maybe add a line that checks that the number of quorums *is actually 256* before the `expectRevert`?

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+
+    function testAddStrategiesConsideredAndMultipliers_Valid(
+        uint256 randomSplit,
+        IVoteWeigher.StrategyAndWeightingMultiplier[] memory strategiesAndWeightingMultipliers
+    ) public {
```
ChaoticWalrus, line 200: really interesting test, I like it. smells like an equivalence property that we could prove (equivalence between initializing the quorum and adding the strategies)

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+        for (uint256 i = 0; i < strategiesAndWeightingMultipliers.length - 1; i++) {
+            strategiesAndWeightingMultipliers1[i] = strategiesAndWeightingMultipliers[i];
+        }
+        strategiesAndWeightingMultipliers2[0] = strategiesAndWeightingMultipliers[strategiesAndWeightingMultipliers.length - 1];
+
+        // create quorum with all but the last element
+        uint8 quorumNumber = voteWeigher.quorumCount();
+        cheats.prank(serviceManagerOwner);
+        voteWeigher.createQuorum(strategiesAndWeightingMultipliers1);
+
```
ChaoticWalrus, line 257: I _suspect_ that you don't actually need all of this setup, based on the order in which reverts will occur. That being said, this seems good since it's less brittle, because changing the order of reverts shouldn't break it.

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+        // add the last element
+        cheats.prank(notServiceManagerOwner);
+        cheats.expectRevert("VoteWeigherBase.onlyServiceManagerOwner: caller is not the owner of the serviceManager");
+        voteWeigher.addStrategiesConsideredAndMultipliers(quorumNumber, strategiesAndWeightingMultipliers2);
+    }
+
+    function testAddStrategiesConsideredAndMultipliers_ForNonExistantQuorum_Reverts(
```
ChaoticWalrus, line 264: typo: 'existant' => 'existent'

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+        voteWeigher.createQuorum(strategiesAndWeightingMultipliers1);
+
+        // add the last element
+        cheats.expectRevert("VoteWeigherBase._addStrategiesConsideredAndMultipliers: quorumNumber exceeds quorumCount");
+        voteWeigher.addStrategiesConsideredAndMultipliers(quorumNumber+1, strategiesAndWeightingMultipliers2);        
+    }
+
+    function testRemoveStrategiesConsideredAndMultipliers_Valid(
+        IVoteWeigher.StrategyAndWeightingMultiplier[] memory strategiesAndWeightingMultipliers,
+        uint256[] memory indicesToRemove
```
ChaoticWalrus, line 291: this is a complex (at least in terms of implementation) test. probably merits some top-level comment explaining it a bit more?

ChaoticWalrus: overall, I feel like there *has* to be a much more efficient way to implement this, but I feel like I can follow your reasoning

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+            for (uint256 k = 0; k < indicesToRemove.length; k++) {
+                if (i == indicesToRemove[k]) {
+                    break;
+                }
+                if (k == indicesToRemove.length - 1) {
+                    strategiesAndWeightingMultipliersLocal[j] = strategiesAndWeightingMultipliers[i];
+                    j++;
+                }
+            }
+        }
```
ChaoticWalrus, line 338: this nested loop seems particularly inefficient, and I suspect it's avoidable
tests are hanging for me and I believe this test is the (or at least _a_ culprit)

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+
+        // sort both arrays by strategy so that they are easily comaparable
+        strategiesAndWeightingMultipliersFromContract = _sortStrategiesAndWeightingMultipliersByStrategy(strategiesAndWeightingMultipliersFromContract);
+        strategiesAndWeightingMultipliersLocal = _sortStrategiesAndWeightingMultipliersByStrategy(strategiesAndWeightingMultipliersLocal);
```
ChaoticWalrus, line 342: maybe sort earlier if it will make this test run in a reasonable time?

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+        uint8 quorumNumber = voteWeigher.quorumCount();
+        cheats.startPrank(serviceManagerOwner);
+        voteWeigher.createQuorum(strategiesAndWeightingMultipliers);
+
+        // remove strategies from a non-existant quorum
+        cheats.expectRevert("VoteWeigherBase.removeStrategiesConsideredAndMultipliers: quorumNumber is greater than or equal to quorumCount");
+        voteWeigher.removeStrategiesConsideredAndMultipliers(quorumNumber + 1, indicesToRemove);
+    }
+
+    function testRemoveStrategiesConsideredAndMultipliers_EmptyIndicesToRemove_HasNoEffect(
```
ChaoticWalrus, line 388: should we maybe revert on an empty array instead?
seems acceptable either way, but this is a case where reverting might actually make more sense -- if you are providing an empty array then you've presumably messed something up, and getting a revert message would make it easier to spot that you've made a mistake + figure this out.

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+        // remove no strategies
+        voteWeigher.removeStrategiesConsideredAndMultipliers(quorumNumber, new uint256[](0));
+        // make sure the quorum strategies and weights haven't changed
+        for (uint i = 0; i < strategiesAndWeightingMultipliers.length; i++) {
+            IVoteWeigher.StrategyAndWeightingMultiplier memory strategyAndWeightingMultiplier = voteWeigher.strategyAndWeightingMultiplierForQuorumByIndex(quorumNumber, i);
+            assertEq(address(strategyAndWeightingMultiplier.strategy), address(strategiesAndWeightingMultipliers[i].strategy));
+            assertEq(strategyAndWeightingMultiplier.multiplier, strategiesAndWeightingMultipliers[i].multiplier);
+        }
+    }
+
```
ChaoticWalrus, line 407: should we also have a test for trying to remove invalid indices? I'm presuming this would just break with an out-of-bounds access error?

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+        // trim the provided weights to the length of the strategyIndices
+        uint96[] memory newWeightsTrim = new uint96[](strategyIndices.length);
+        for (uint256 i = 0; i < strategyIndices.length; i++) {
+            if(i < newWeights.length) {
+                newWeightsTrim[i] = newWeights[i];
+            } else {
+                newWeightsTrim[i] = strategiesAndWeightingMultipliers[strategyIndices[i]].multiplier - 1;
+            }
+        }
+        newWeights = newWeightsTrim;
```
ChaoticWalrus, line 452: so this is just making the lengths match. seems OK, but I might describe it more accurately (it's trimming in some cases and extending in others)

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+        uint8 quorumNumber = voteWeigher.quorumCount();
+        cheats.startPrank(serviceManagerOwner);
+        voteWeigher.createQuorum(strategiesAndWeightingMultipliers);
+
+        // modify certain strategies
+        cheats.expectRevert("VoteWeigherBase.modifyStrategyWeights: input length mismatch");
+        voteWeigher.modifyStrategyWeights(quorumNumber, strategyIndices, newWeights);
+    }
+
+    function testModifyStrategyWeights_EmptyStrategyIndicesAndWeights_HasNoEffect(
```
ChaoticWalrus, line 515: I again think this is a case where reverting probably makes more sense than having no effect. current behavior seems acceptable though and this test looks alright.

ChaoticWalrus (commented):
Looks really good overall.
Tests seem super heavy with all the fuzzing + nested loops for sorting.
Some of this is pretty certainly overkill, so maybe we can find the worst offenders and improve a bit.

src/contracts/middleware/VoteWeigherBase.sol
gpsanant: as long as we have state diffs, I think this should be fine. A race condition could only triggered by transactions are also from this address.

src/test/unit/VoteWeigherBaseUnit.t.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/d4be2b7f8585ad8ef9171fba9171db253b572e0b

src/test/harnesses/VoteWeigherBaseHarness.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/d4be2b7f8585ad8ef9171fba9171db253b572e0b

src/test/unit/VoteWeigherBaseUnit.t.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/d4be2b7f8585ad8ef9171fba9171db253b572e0b

src/contracts/middleware/VoteWeigherBase.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/d4be2b7f8585ad8ef9171fba9171db253b572e0b

src/contracts/middleware/VoteWeigherBase.sol
gpsanant: great catch: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/d4be2b7f8585ad8ef9171fba9171db253b572e0b

src/test/unit/VoteWeigherBaseUnit.t.sol
gpsanant: agree https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/7b14be2012f508f84ff205a7f82cfab9a12d9eb9

src/test/unit/VoteWeigherBaseUnit.t.sol
gpsanant: fixed

src/test/unit/VoteWeigherBaseUnit.t.sol
gpsanant: fixed

src/test/unit/VoteWeigherBaseUnit.t.sol
gpsanant: agreeed, fixed

ChaoticWalrus:
For some context on current test load:
running the existing foundry tests takes ~90 seconds on master
running the foundry tests took ~80 minutes on this branch

src/test/unit/VoteWeigherBaseUnit.t.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/2c3df9ab884f24281df9c3ef514b72b7edb086b3

src/test/unit/VoteWeigherBaseUnit.t.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/7992f656918f7852cf5c9b5e7c383fb2241a1313

gpsanant:
Spent some time working on your feedback (thanks!). Tests are now ~6s without compromising too much on fuzzing.

src/contracts/interfaces/IVoteWeigher.sol
```diff
@@ -34,17 +34,20 @@ interface IVoteWeigher {
...
+    function quorumCount() external view returns (uint16);
 
     /**
      * @notice This defines the earnings split between different quorums. Mapping is quorumNumber => BIPS which the quorum earns, out of the total earnings.
      * @dev The sum of all entries, i.e. sum(quorumBips[0] through quorumBips[NUMBER_OF_QUORUMS - 1]) should *always* be 10,000!
      */
     function quorumBips(uint8 quorumNumber) external view returns (uint256);
 
+    /// @notice Returns the strategy and weight multiplier for the `index`'th strategy in the quorum `quorumNumber`
+    function strategyAndWeightingMultiplierForQuorumByIndex(uint8 quorumNumber, uint256 index) external view returns (StrategyAndWeightingMultiplier memory);
```
ChaoticWalrus, line 49: we should add a getter for the length of these if we don't have it already

src/contracts/middleware/VoteWeigherBase.sol
```diff
@@ -210,8 +208,12 @@ contract VoteWeigherBase is VoteWeigherBaseStorage {
                     ++j;
                 }
             }
+            require(
+                _newStrategiesConsideredAndMultipliers[i].multiplier > 0,
+                "VoteWeigherBase._addStrategiesConsideredAndMultipliers: cannot add strategy with zero weight"
+            );
```
ChaoticWalrus, line 214: we are requiring the weights be nonzero here but not in `modifyStrategyWeights`
IMO a weight of zero is not invalid, but I feel like we should be consistent about this behavior and clear about what we desire. Probably deserves a dedicated test or two.

src/contracts/middleware/VoteWeigherBase.sol
```diff
@@ -100,25 +110,16 @@ contract VoteWeigherBase is VoteWeigherBaseStorage {
...
-            require(
-                strategiesConsideredAndMultipliers[quorumNumber][indicesToRemove[i]].strategy == _strategiesToRemove[i],
-                "VoteWeigherBase.removeStrategiesConsideredAndWeights: index incorrect"
-            );
-
+    ) external virtual onlyServiceManagerOwner validQuorumNumber(quorumNumber) {
+        uint256 indicesToRemoveLength = indicesToRemove.length;
+        require(indicesToRemoveLength > 0, "VoteWeigherBase.removeStrategiesConsideredAndMultipliers: no indices to remove provided");
+        for (uint256 i = 0; i < indicesToRemoveLength;) {
+            emit StrategyRemovedFromQuorum(quorumNumber, strategiesConsideredAndMultipliers[quorumNumber][indicesToRemove[i]].strategy);
```
ChaoticWalrus, line 118: we enforce elsewhere (when creating new quorums) that each quorum contains at least one strategy.
seems like this is a property we _might_ want to ensure
IMO it's better to not have this property, but if we agree it's OK then we should have a test that specifically removes all strategies from a quorum and then tests the behavior (I guess the desired behavior would be just return 0 weight for everyone and not revert?)

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,596 @@
...
+    function testCreateQuorum_StrategiesAndWeightingMultipliers_LengthGreaterThanMaxAllowed_Reverts(
+        IVoteWeigher.StrategyAndWeightingMultiplier[] memory strategiesAndWeightingMultipliers
+    ) public {
+        strategiesAndWeightingMultipliers = _removeDuplicates(strategiesAndWeightingMultipliers);
+        strategiesAndWeightingMultipliers = _replaceZeroWeights(strategiesAndWeightingMultipliers);
+
+        cheats.assume(strategiesAndWeightingMultipliers.length > voteWeigher.MAX_WEIGHING_FUNCTION_LENGTH());
+        cheats.prank(serviceManagerOwner);
+        cheats.expectRevert("VoteWeigherBase._addStrategiesConsideredAndMultipliers: exceed MAX_WEIGHING_FUNCTION_LENGTH");
+        voteWeigher.createQuorum(strategiesAndWeightingMultipliers);
```
ChaoticWalrus, line 142: I realize the way this is implemented is that the check is in the internal function. still might be good to also make a call to `addStrategies...` and check that this reverts appropriately as well (you might need to initialize the quorum with the default list first)

ChaoticWalrus:
> Spent some time working on your feedback (thanks!). Tests are now ~6s without compromising too much on fuzzing.

This is looking really good! Just dropped a couple more comments on the PR, but I'd feel comfortable merging this once all the open items are addressed.

src/contracts/middleware/VoteWeigherBase.sol
ChaoticWalrus: going to note here that I think adding an `@dev` tag to this function that explains the side-effects/risk of this function ignoring fluctations in the conversion rate of shares to underlying

src/test/unit/VoteWeigherBaseUnit.t.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/43b1830872bd61cc16f74b9db576ca4bd0eec68a

src/test/unit/VoteWeigherBaseUnit.t.sol
```diff
@@ -0,0 +1,664 @@
...
+        // create a valid quorum
+        uint8 quorumNumber = voteWeigher.quorumCount();
+        cheats.startPrank(serviceManagerOwner);
+        voteWeigher.createQuorum(strategiesAndWeightingMultipliers);
+
+        // modify certain strategies of a non-existant quorum
+        cheats.expectRevert("VoteWeigherBase.modifyStrategyWeights: quorumNumber is greater than or equal to quorumCount");
+        voteWeigher.modifyStrategyWeights(quorumNumber + 1, strategyIndices, newWeights);
+    }
+
```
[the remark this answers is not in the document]
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/43b1830872bd61cc16f74b9db576ca4bd0eec68a

src/contracts/interfaces/IVoteWeigher.sol
gpsanant: already got one

src/contracts/middleware/VoteWeigherBase.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/8ee2fc0de766656c46ac2d92ae089703959d8f60 agree the functionality is not needed.

src/contracts/middleware/VoteWeigherBase.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/c35fa349a17d5bf871aeac38b9df81bb805f9d68 i said a 0 weight is invalid and changed revert messages and added tests. Remove the strategy if you want a zero weight imo.

src/test/unit/VoteWeigherBaseUnit.t.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/263ce1d0bdbe1ed73daf27d600f280981d4bc262

src/contracts/middleware/VoteWeigherBase.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/53/commits/07031237ae72de11a7507e3ff3bb2f1e08ea9eb4

src/contracts/middleware/VoteWeigherBase.sol
ChaoticWalrus: OK, works for me. Great opportunity for a Certora Prover invariant IMO (to verify there's no way to get a 0 multiplier strategy).

ChaoticWalrus:
LGTM now.

PR #54, opened 2023-05-31
changed tvl limits to deposit per address

Sidu28:


ChaoticWalrus:
Reached agreement offline that this change is not needed.

PR #55, opened 2023-06-01
Deploy config

Sidu28:
This PR is to address issue #851.  Specifically, it sets the specific configs for strategies, including contract addresses of the decided list of strategies and other parameters.  We also add checks for the initializations


Sidu28:
> Refer



> @Sidu28 please provide a description detailing this PR a bit more looks like as-is `init_withdrawal_delay_blocks` is being set to an inappropriate value for mainnet deployment, unless we want to start with this very low and then adjust up (which would be news to me, as adjustments are permissioned to only the `StrategyManager.owner` role) would definitely appreciate you adding additional checks to the deploy script as possible as part of this PR, to ensure that we avoid any operational mistakes

@ChaoticWalrus  sorry this is still a draft, requested you by mistake.  Trying to sort out a compile error from JSON atm, not idea wtf is happening

PR #57, opened 2023-06-04
Update docgen to use `forge doc`

Sabnock01:
This PR updates the docgen located at `docs/docgen` to use `forge doc` instead of `npx hardhat doc` and removes Hardhat from the repo as it is now no longer needed. 

The `README.md` mentions using Hardhat for the docgen as `forge doc` was not yet available. This is old information as of [this PR](https://github.com/foundry-rs/foundry/pull/2701)

This PR uses the same `docgen` folder to house the documentation generated by `forge doc` and updates the `README.md` to reflect this and adds instructions for building and serving it locally. 

If you all desire an alternative file structure for the docgen or want a change to the language added to the `README.md`, please let me know. 

Additionally, I'll add that when generating the docgen locally with `npx hardhat doc`, it produced different output from what is currently in the repo. Making the current documentation deprecated nonetheless.


Sabnock01:
It would probably be helpful, by the way, in reviewing this PR to look run the book locally with `mdbook serve` which will allow you to view most of the new files more easily.

Sabnock01:
It would probably be helpful, by the way, in reviewing this PR to run the book locally with `mdbook serve` which will allow you to view most of the new files more easily.

ChaoticWalrus:
Thanks for this PR. I outlined my thinking + main questions in another similar PR.
For now, I'm hesitant to make this change without any clear signal on the output actually being preferable.

PR #58, opened 2023-06-06
payment coordinator

Sidu28:
Creating a preliminary implementation of the PaymentCoordinator contract.

### DIFF
--- a/src/contracts/core/PaymentCoordinator.sol
+++ b/src/contracts/core/PaymentCoordinator.sol
@@ -37,7 +37,7 @@ contract PaymentCoordinator is
     uint256 public constant MAX_EIGENLAYER_SHARE_BIPS = 1500;
 
     /// @notice Array of roots of posted Merkle trees, as well as associated data like tree height
-    MerkleRootPost[] internal _merkleRoots;
+    MerkleRootPost[] internal _merkleRootPosts;
 
     /// @notice Mapping token => recipient => cumulative amount *claimed*
     mapping(IERC20 => mapping(address => uint256)) public cumulativeTokenAmountClaimedByRecipient;
@@ -107,14 +107,14 @@ contract PaymentCoordinator is
     // @notice Permissioned function which allows posting a new Merkle root
     function postMerkleRoot(bytes32 newRoot, uint256 height, uint256 calculatedUpToBlockNumber) external onlyRootPublisher {
         MerkleRootPost memory newMerkleRoot = MerkleRootPost(newRoot, height, block.number + merkleRootActivationDelayBlocks, calculatedUpToBlockNumber);
-        _merkleRoots.push(newMerkleRoot);
+        _merkleRootPosts.push(newMerkleRoot);
         emit NewMerkleRootPosted(newMerkleRoot);
     }
 
     /// @notice Permissioned function which allows rootPublisher to nullify a Merkle root
     function nullifyMerkleRoot(uint256 rootIndex) external onlyOwner {
-        require(block.number <= _merkleRoots[rootIndex].confirmedAtBlockNumber, "PaymentCoordinator.nullifyMerkleRoot: Merkle root already confirmed");
-        delete _merkleRoots[rootIndex];
+        require(block.number <= _merkleRootPosts[rootIndex].confirmedAtBlockNumber, "PaymentCoordinator.nullifyMerkleRoot: Merkle root already confirmed");
+        delete _merkleRootPosts[rootIndex];
     }
 
     // @notice Permissioned function which allows withdrawal of EigenLayer's share of `token` from all received payments
@@ -137,9 +137,9 @@ contract PaymentCoordinator is
         uint256 leafIndex
     ) external {
         require(leaf.amounts.length == leaf.tokens.length, "PaymentCoordinator.proveAndClaimEarnings: leaf amounts and tokens must be same length");
-        require(_merkleRoots[rootIndex].confirmedAtBlockNumber < block.number, "PaymentCoordinator.proveAndClaimEarnings: Merkle root not yet confirmed");
+        require(_merkleRootPosts[rootIndex].confirmedAtBlockNumber < block.number, "PaymentCoordinator.proveAndClaimEarnings: Merkle root not yet confirmed");
 
-        bytes32 root = _merkleRoots[rootIndex].root;
+        bytes32 root = _merkleRootPosts[rootIndex].root;
         require(root != bytes32(0), "PaymentCoordinator.proveAndClaimEarnings: Merkle root is null");
 
         bytes32 leafHash = _computeLeafHash(leaf);
@@ -168,13 +168,13 @@ contract PaymentCoordinator is
 
 
     /// @notice Getter function for the length of the `merkleRoots` array
-    function merkleRootsLength() external view returns (uint256) {
-        return _merkleRoots.length;
+    function merkleRootPostsLength() external view returns (uint256) {
+        return _merkleRootPosts.length;
     }
 
     /// @notice Getter function for a merkleRoot at a given index
-    function merkleRoots(uint256 index) external view returns (MerkleRootPost memory) {
-        return _merkleRoots[index];
+    function merkleRootPosts(uint256 index) external view returns (MerkleRootPost memory) {
+        return _merkleRootPosts[index];
     }
 
     function _setRootPublisher(address _rootPublisher) internal {
--- a/src/contracts/core/PaymentCoordinator.sol
+++ b/src/contracts/core/PaymentCoordinator.sol
@@ -8,9 +8,6 @@ import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
 import "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
-import "forge-std/Test.sol";
-
-
 
 /**
  * @title Contract used to coordinate payments from AVSs to operators and in particular the subsequent splitting of earnings from operators to stakers
@@ -19,8 +16,7 @@ import "forge-std/Test.sol";
 contract PaymentCoordinator is 
     IPaymentCoordinator,
     Initializable,
-    OwnableUpgradeable,
-    Test
+    OwnableUpgradeable
 {
     using SafeERC20 for IERC20;
 
--- a/src/contracts/core/PaymentCoordinator.sol
+++ b/src/contracts/core/PaymentCoordinator.sol
@@ -79,8 +79,7 @@ contract PaymentCoordinator is
         }
         payment.token.safeTransferFrom(msg.sender, address(this), sumAmounts);
 
-        uint256 eigenLayerShare = sumAmounts * eigenLayerShareBIPs / MAX_BIPS;
-        cumulativeEigenLayerTokeEarnings[payment.token] += eigenLayerShare;
+        cumulativeEigenLayerTokeEarnings[payment.token] += sumAmounts * eigenLayerShareBIPs / MAX_BIPS;
 
         emit PaymentReceived(msg.sender, payment);
     }
--- a/src/contracts/core/PaymentCoordinator.sol
+++ b/src/contracts/core/PaymentCoordinator.sol
@@ -112,7 +112,7 @@ contract PaymentCoordinator is
     }
 
     /// @notice Permissioned function which allows rootPublisher to nullify a Merkle root
-    function nullifyMerkleRoot(uint256 rootIndex) external onlyRootPublisher {
+    function nullifyMerkleRoot(uint256 rootIndex) external onlyOwner {
         require(block.number <= _merkleRoots[rootIndex].confirmedAtBlockNumber, "PaymentCoordinator.nullifyMerkleRoot: Merkle root already confirmed");
         delete _merkleRoots[rootIndex];
     }
@@ -133,7 +133,8 @@ contract PaymentCoordinator is
     function proveAndClaimEarnings(
         bytes memory proof,
         uint256 rootIndex,
-        MerkleLeaf memory leaf
+        MerkleLeaf memory leaf,
+        uint256 leafIndex
     ) external {
         require(leaf.amounts.length == leaf.tokens.length, "PaymentCoordinator.proveAndClaimEarnings: leaf amounts and tokens must be same length");
         require(_merkleRoots[rootIndex].confirmedAtBlockNumber < block.number, "PaymentCoordinator.proveAndClaimEarnings: Merkle root not yet confirmed");
@@ -142,11 +143,13 @@ contract PaymentCoordinator is
         require(root != bytes32(0), "PaymentCoordinator.proveAndClaimEarnings: Merkle root is null");
 
         bytes32 leafHash = _computeLeafHash(leaf);
-        require(Merkle.verifyInclusionKeccak(proof, root, leafHash, leaf.index), "PaymentCoordinator.proveAndClaimEarnings: Invalid proof");
+        require(Merkle.verifyInclusionKeccak(proof, root, leafHash, leafIndex), "PaymentCoordinator.proveAndClaimEarnings: Invalid proof");
 
         for(uint256 i = 0; i < leaf.amounts.length; i++) {
-            leaf.tokens[i].safeTransfer(leaf.recipient, leaf.amounts[i]);
-            cumulativeTokenAmountClaimedByRecipient[leaf.tokens[i]][leaf.recipient] += leaf.amounts[i];
+            uint256 amount = leaf.amounts[i] - cumulativeTokenAmountClaimedByRecipient[leaf.tokens[i]][leaf.recipient];
+            cumulativeTokenAmountClaimedByRecipient[leaf.tokens[i]][leaf.recipient] = leaf.amounts[i];
+            leaf.tokens[i].safeTransfer(leaf.recipient, amount);
+            
         }
         emit PaymentClaimed(leaf);
     }
@@ -193,6 +196,6 @@ contract PaymentCoordinator is
     }
 
     function _computeLeafHash(MerkleLeaf memory leaf) internal pure returns (bytes32) {
-        return keccak256(abi.encodePacked(leaf.recipient, keccak256(abi.encodePacked(leaf.tokens)), keccak256(abi.encodePacked(leaf.amounts)), leaf.index));
+        return keccak256(abi.encodePacked(leaf.recipient, keccak256(abi.encodePacked(leaf.tokens)), keccak256(abi.encodePacked(leaf.amounts))));
     }
 }
--- a/src/contracts/core/PaymentCoordinator.sol
+++ b/src/contracts/core/PaymentCoordinator.sol
@@ -33,7 +33,7 @@ contract PaymentCoordinator is
     uint256 public constant MAX_BIPS = 10000;
 
     /// @notice Array of roots of posted Merkle trees, as well as associated data like tree height
-    MerkleRootPost[] public merkleRootPosts;
+    MerkleRootPost[] public merkleRoots;
 
     /// @notice Mapping token => recipient => cumulative amount *claimed*
     mapping(IERC20 => mapping(address => uint256)) public cumulativeTokenAmountClaimedByRecipient;
@@ -89,15 +89,20 @@ contract PaymentCoordinator is
     // @notice Permissioned function which allows posting a new Merkle root
     function postMerkleRoot(bytes32 newRoot, uint256 height, uint256 calculatedUpToBlockNumber) external onlyRootPublisher{
         MerkleRootPost memory newMerkleRoot = MerkleRootPost(newRoot, height, block.number + merkleRootActivationDelay, calculatedUpToBlockNumber);
-        merkleRootPosts.push(newMerkleRoot);
+        merkleRoots.push(newMerkleRoot);
         emit NewMerkleRootPosted(newMerkleRoot);
+    }
+
+    /// @notice Permissioned function which allows rootPublisher to nullify a Merkle root
+    function nullifyMerkleRoot(uint256 rootIndex) external onlyRootPublisher{
+        require(block.number <= merkleRoots[rootIndex].confirmedAtBlockNumber, "PaymentCoordinator.nullifyMerkleRoot: Merkle root already confirmed");
+        merkleRoots[rootIndex].root = bytes32(0);
     }
 
     // @notice Permissioned function which allows withdrawal of EigenLayer's share of `token` from all received payments
     function withdrawEigenlayerShare(IERC20 token, address recipient) external onlyRootPublisher{
         uint256 amount = cumulativeEigenLayerTokeEarnings[token];
         token.safeTransfer(recipient, amount);
-
         cumulativeEigenLayerTokeEarnings[token] = 0;
     }
 
@@ -113,10 +118,10 @@ contract PaymentCoordinator is
         MerkleLeaf memory leaf
     ) external{
         require(leaf.amounts.length == leaf.tokens.length, "PaymentCoordinator.proveAndClaimEarnings: leaf amounts and tokens must be same length");
-        require(merkleRootPosts[rootIndex].confirmedAtBlockNumber > block.number, "PaymentCoordinator.proveAndClaimEarnings: Merkle root not yet confirmed");
-        bytes32 leafHash = keccak256(abi.encodePacked(leaf.recipient, keccak256(abi.encodePacked(leaf.tokens)), keccak256(abi.encodePacked(leaf.amounts)), leaf.index));
-        bytes32 root = merkleRootPosts[rootIndex].root;
+        require(merkleRoots[rootIndex].confirmedAtBlockNumber < block.number, "PaymentCoordinator.proveAndClaimEarnings: Merkle root not yet confirmed");
+        bytes32 root = merkleRoots[rootIndex].root;
         require(root != bytes32(0), "PaymentCoordinator.proveAndClaimEarnings: Merkle root is null");
+        bytes32 leafHash = keccak256(abi.encodePacked(leaf.recipient, keccak256(abi.encodePacked(leaf.tokens)), keccak256(abi.encodePacked(leaf.amounts)), leaf.index));
         require(Merkle.verifyInclusionKeccak(proof, root, leafHash, leaf.index), "PaymentCoordinator.proveAndClaimEarnings: Invalid proof");
 
         for(uint i = 0; i < leaf.amounts.length; i++) {
@@ -124,11 +129,6 @@ contract PaymentCoordinator is
             cumulativeTokenAmountClaimedByRecipient[leaf.tokens[i]][leaf.recipient] += leaf.amounts[i];
         }
         emit PaymentClaimed(leaf);
-    }
-
-    function nullifyMerkleRoot(uint256 rootIndex) external onlyRootPublisher{
-        require(block.number <= merkleRootPosts[rootIndex].confirmedAtBlockNumber, "PaymentCoordinator.nullifyMerkleRoot: Merkle root already confirmed");
-        merkleRootPosts[rootIndex].root = bytes32(0);
     }
 
     function setRootPublisher(address _rootPublisher) external onlyOwner{
@@ -144,9 +144,9 @@ contract PaymentCoordinator is
     }
 
 
-    /// @notice Getter function for the length of the `merkleRootPosts` array
-    function merkleRootPostsLength() external view returns (uint256){
-        return merkleRootPosts.length;
+    /// @notice Getter function for the length of the `merkleRoots` array
+    function merkleRootsLength() external view returns (uint256){
+        return merkleRoots.length;
     }
 
     function _setRootPublisher(address _rootPublisher) internal {
--- a/src/contracts/core/PaymentCoordinator.sol
+++ b/src/contracts/core/PaymentCoordinator.sol
@@ -6,6 +6,8 @@ import "../interfaces/IPaymentCoordinator.sol";
 import "../libraries/Merkle.sol";
 import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
 import "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
 
 
 /**
@@ -17,6 +19,7 @@ contract PaymentCoordinator is
     Initializable,
     OwnableUpgradeable
 {
+    using SafeERC20 for IERC20;
 
     /// @notice address approved to post new Merkle roots
     address public rootPublisher;
@@ -51,7 +54,7 @@ contract PaymentCoordinator is
         _;
     }
 
-    function initialize(address _initialOwner, uint256 _eigenlayerShareBips, address _rootPublisher, uint256 _merkleRootActivationDelay)
+    function initialize(address _initialOwner, address _rootPublisher, uint256 _eigenlayerShareBips, uint256 _merkleRootActivationDelay)
         external
         initializer
     {
@@ -110,10 +113,10 @@ contract PaymentCoordinator is
     ) external{
         require(leaf.amounts.length == leaf.tokens.length, "PaymentCoordinator.proveAndClaimEarnings: leaf amounts and tokens must be same length");
         require(merkleRootPosts[rootIndex].confirmedAtBlockNumber > block.number, "PaymentCoordinator.proveAndClaimEarnings: Merkle root not yet confirmed");
-        bytes32 leaf = keccak256(abi.encodePacked(leaf.recipient, keccak256(abi.encodePacked(leaf.tokens)), keccak256(abi.encodePacked(leaf.amounts)), leaf.index));
+        bytes32 leafHash = keccak256(abi.encodePacked(leaf.recipient, keccak256(abi.encodePacked(leaf.tokens)), keccak256(abi.encodePacked(leaf.amounts)), leaf.index));
         bytes32 root = merkleRootPosts[rootIndex].root;
         require(root != bytes32(0), "PaymentCoordinator.proveAndClaimEarnings: Merkle root is null");
-        require(Merkle.verifyInclusionKeccak(proof, root, leaf, leaf.index), "PaymentCoordinator.proveAndClaimEarnings: Invalid proof");
+        require(Merkle.verifyInclusionKeccak(proof, root, leafHash, leaf.index), "PaymentCoordinator.proveAndClaimEarnings: Invalid proof");
 
         for(uint i = 0; i < leaf.amounts.length; i++) {
             leaf.tokens[i].safeTransfer(leaf.recipient, leaf.amounts[i]);
--- a/src/contracts/core/PaymentCoordinator.sol
+++ b/src/contracts/core/PaymentCoordinator.sol
@@ -8,6 +8,8 @@ import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
 import "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
+import "forge-std/Test.sol";
+
 
 
 /**
@@ -17,7 +19,8 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 contract PaymentCoordinator is 
     IPaymentCoordinator,
     Initializable,
-    OwnableUpgradeable
+    OwnableUpgradeable,
+    Test
 {
     using SafeERC20 for IERC20;
 
@@ -80,7 +83,6 @@ contract PaymentCoordinator is
         payment.token.safeTransferFrom(msg.sender, address(this), sumAmounts);
 
         cumulativeEigenLayerTokeEarnings[payment.token] += sumAmounts * eigenLayerShareBIPs / MAX_BIPS;
-
         emit PaymentReceived(msg.sender, payment);
     }
 
--- a/src/contracts/core/PaymentCoordinator.sol
+++ b/src/contracts/core/PaymentCoordinator.sol
@@ -13,7 +13,7 @@ import "forge-std/Test.sol";
 
 
 /**
- * @title Contract used to coordinate payments from AVSs to operators and in particular the subsequency splitting of earnings from operators to stakers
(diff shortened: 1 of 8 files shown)


src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
+import "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import "forge-std/Test.sol";
+
+
+
+/**
+ * @title Contract used to coordinate payments from AVSs to operators and in particular the subsequency splitting of earnings from operators to stakers
```
ChaoticWalrus, line 16: typo: "subsequency" => "subsequent"

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+ * @author Layr Labs, Inc.
+ */
+contract PaymentCoordinator is 
+    IPaymentCoordinator,
+    Initializable,
+    OwnableUpgradeable,
+    Test
```
ChaoticWalrus, line 23: inheriting from Test

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+        require(Merkle.verifyInclusionKeccak(proof, root, leafHash, leaf.index), "PaymentCoordinator.proveAndClaimEarnings: Invalid proof");
+
+        for(uint i = 0; i < leaf.amounts.length; i++) {
+            leaf.tokens[i].safeTransfer(leaf.recipient, leaf.amounts[i]);
+            cumulativeTokenAmountClaimedByRecipient[leaf.tokens[i]][leaf.recipient] += leaf.amounts[i];
+        }
+        emit PaymentClaimed(leaf);
+    }
+
+    function setRootPublisher(address _rootPublisher) external onlyOwner{
```
ChaoticWalrus, line 134: should have a space before the curly braces (everywhere, just noting it in this one place)

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    event PaymentClaimed(MerkleLeaf merkleLeaf);
+
+    modifier onlyRootPublisher {
+        require(msg.sender == rootPublisher, "PaymentCoordinator: Only rootPublisher");
+        _;
+    }
+
+    function initialize(address _initialOwner, address _rootPublisher, uint256 _eigenlayerShareBips, uint256 _merkleRootActivationDelay)
+        external
+        initializer
```
ChaoticWalrus, line 62: probably want a constructor that disables initializers

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    uint256 public constant MAX_BIPS = 10000;
+
+    /// @notice Array of roots of posted Merkle trees, as well as associated data like tree height
+    MerkleRootPost[] public merkleRoots;
+
+    /// @notice Mapping token => recipient => cumulative amount *claimed*
+    mapping(IERC20 => mapping(address => uint256)) public cumulativeTokenAmountClaimedByRecipient;
+
+    /// @notice Mapping token => cumulative amount earned by EigenLayer
+    mapping(IERC20 => uint256) public cumulativeEigenLayerTokeEarnings;
```
ChaoticWalrus, line 42: typo: "Toke" => "Token"

ChaoticWalrus: also what is this?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    Initializable,
+    OwnableUpgradeable,
+    Test
+{
+    using SafeERC20 for IERC20;
+
+    /// @notice address approved to post new Merkle roots
+    address public rootPublisher;
+
+    uint256 public merkleRootActivationDelay;
```
ChaoticWalrus, line 30: this needs documentation and should probably have 'Blocks' or 'Seconds' on the end? (I prefer blocks, as we've chosen to use blocks elsewhere)

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+
+    /// @notice Getter function for the length of the `merkleRoots` array
+    function merkleRootsLength() external view returns (uint256){
+        return merkleRoots.length;
+    }
+
+    function _setRootPublisher(address _rootPublisher) internal {
+        rootPublisher = _rootPublisher;
```
ChaoticWalrus, line 153: there should probably be an associated `RootPublisherChanged` or similarly named event for this

ChaoticWalrus: the event should emit the current (past) and new (updated) addresses

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    function setRootPublisher(address _rootPublisher) external onlyOwner{
+        _setRootPublisher(_rootPublisher);
+    }
+
+    function setMerkleRootActivationDelay(uint256 _merkleRootActivationDelay) external onlyOwner{
+        _setMerkleRootActivationDelay(_merkleRootActivationDelay);
+    }
+
+    function setEigenLayerShareBIPS(uint256 _eigenlayerShareBips) external onlyOwner{
+        _setEigenLayerShareBIPS(_eigenlayerShareBips);
```
ChaoticWalrus, line 143: this should also have an event for when it changes

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+        }
+        emit PaymentClaimed(leaf);
+    }
+
+    function setRootPublisher(address _rootPublisher) external onlyOwner{
+        _setRootPublisher(_rootPublisher);
+    }
+
+    function setMerkleRootActivationDelay(uint256 _merkleRootActivationDelay) external onlyOwner{
+        _setMerkleRootActivationDelay(_merkleRootActivationDelay);
```
ChaoticWalrus, line 139: this should also have an event for when it changes

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+        return merkleRoots.length;
+    }
+
+    function _setRootPublisher(address _rootPublisher) internal {
+        rootPublisher = _rootPublisher;
+    }
+
+    function _setMerkleRootActivationDelay(uint256 _merkleRootActivationDelay) internal {
+        merkleRootActivationDelay = _merkleRootActivationDelay;
+    }
```
ChaoticWalrus, line 158: this should also have an event for when it changes

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+
+    function _setEigenLayerShareBIPS(uint256 _eigenlayerShareBips) internal {
+        require(_eigenlayerShareBips <= MAX_BIPS, "PaymentCoordinator: EigenLayer share cannot be greater than 100%");
+        eigenLayerShareBIPs = _eigenlayerShareBips;
+    }
```
ChaoticWalrus, line 163: this should also have an event for when it changes

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * @notice Makes a payment of sum(amounts) paid in `token`, for `operator`'s contributions to an AVS,
+     * between `startBlockNumber` (inclusive) and `endBlockNumber` (inclusive)
+     * @dev Transfers the total payment from the `msg.sender` to this contract, so the caller must have previously approved
+     * this contract to transfer at least sum(`amounts`) of `token`
+     * @notice Emits a `PaymentReceived` event
+     */
+    function makePayment(Payment calldata payment) external{
+        uint256 sumAmounts;
```
ChaoticWalrus, line 79: I think this function should also check that the lengths of `amounts` and `quorums` match. seems important to enforce this, and I think this is the proper place?

src/contracts/core/PaymentCoordinator.sol
ChaoticWalrus: to elaborate, we can't just store the cumulative earnings, we have to have some mechanism for these tokens to leave this contract

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    function withdrawEigenlayerShare(IERC20 token, address recipient) external onlyRootPublisher{
+        uint256 amount = cumulativeEigenLayerTokeEarnings[token];
+        token.safeTransfer(recipient, amount);
+        cumulativeEigenLayerTokeEarnings[token] = 0;
+    }
+
+    /**
+    * @notice Called by a staker or operator to prove the inclusion of their earnings in a posted Merkle root and claim them.
+    * @param proof Merkle proof showing that a leaf containing `(msg.sender, amount)` was included in the `rootIndex`-th
+    * Merkle root posted for the `token`
```
ChaoticWalrus, line 112: this comment seems inaccurate about the contents of the leaf

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    function proveAndClaimEarnings(
+        bytes memory proof,
+        uint256 rootIndex,
+        MerkleLeaf memory leaf
+    ) external{
+        require(leaf.amounts.length == leaf.tokens.length, "PaymentCoordinator.proveAndClaimEarnings: leaf amounts and tokens must be same length");
+        require(merkleRoots[rootIndex].confirmedAtBlockNumber < block.number, "PaymentCoordinator.proveAndClaimEarnings: Merkle root not yet confirmed");
+        bytes32 root = merkleRoots[rootIndex].root;
+        require(root != bytes32(0), "PaymentCoordinator.proveAndClaimEarnings: Merkle root is null");
+        bytes32 leafHash = keccak256(abi.encodePacked(leaf.recipient, keccak256(abi.encodePacked(leaf.tokens)), keccak256(abi.encodePacked(leaf.amounts)), leaf.index));
```
ChaoticWalrus, line 124: should we make this hash calculation a separate (pure & public) function?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+        require(Merkle.verifyInclusionKeccak(proof, root, leafHash, leaf.index), "PaymentCoordinator.proveAndClaimEarnings: Invalid proof");
+
+        for(uint i = 0; i < leaf.amounts.length; i++) {
```
ChaoticWalrus, line 127: please change usage of 'uint' to 'uint256' (explicit uint sizing)

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+            leaf.tokens[i].safeTransfer(leaf.recipient, leaf.amounts[i]);
+            cumulativeTokenAmountClaimedByRecipient[leaf.tokens[i]][leaf.recipient] += leaf.amounts[i];
```
ChaoticWalrus, line 129: this is incorrect, it should:

1. calculate the amount to send as (`leaf.amounts[i] - cumulativeTokenAmountClaimedByRecipient[leaf.tokens[i]][leaf.recipient]`)
2. set `cumulativeTokenAmountClaimedByRecipient[leaf.tokens[i]][leaf.recipient] = leaf.amounts[i]`
3. transfer the tokens

This is also not observing checks-effects-interactions ordering -- transfer after changing storage unless it's not possible.

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    */
+    function proveAndClaimEarnings(
+        bytes memory proof,
+        uint256 rootIndex,
+        MerkleLeaf memory leaf
+    ) external{
+        require(leaf.amounts.length == leaf.tokens.length, "PaymentCoordinator.proveAndClaimEarnings: leaf amounts and tokens must be same length");
+        require(merkleRoots[rootIndex].confirmedAtBlockNumber < block.number, "PaymentCoordinator.proveAndClaimEarnings: Merkle root not yet confirmed");
+        bytes32 root = merkleRoots[rootIndex].root;
+        require(root != bytes32(0), "PaymentCoordinator.proveAndClaimEarnings: Merkle root is null");
```
ChaoticWalrus, line 123: this check is insufficient. as-is, the code isn't checking the delay at all

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+
+    // @notice Permissioned function which allows posting a new Merkle root
+    function postMerkleRoot(bytes32 newRoot, uint256 height, uint256 calculatedUpToBlockNumber) external onlyRootPublisher{
+        MerkleRootPost memory newMerkleRoot = MerkleRootPost(newRoot, height, block.number + merkleRootActivationDelay, calculatedUpToBlockNumber);
+        merkleRoots.push(newMerkleRoot);
+        emit NewMerkleRootPosted(newMerkleRoot);
+    }
+
+    /// @notice Permissioned function which allows rootPublisher to nullify a Merkle root
+    function nullifyMerkleRoot(uint256 rootIndex) external onlyRootPublisher{
```
ChaoticWalrus, line 97: might want to make this owner-only instead of root publisher-only?
also I would say we should just delete the whole storage of the struct -- seems off to leave the storage 'dirty', as this could introduce some weird edge-case bugs

src/contracts/core/PaymentCoordinator.sol
ChaoticWalrus: oh, I see, it gets reset to 0 when `withdrawEigenlayerShare` is called. In that case it's not a 'cumulative' number, recommend removing this wording from the name.

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    /// @notice address approved to post new Merkle roots
+    address public rootPublisher;
+
+    uint256 public merkleRootActivationDelay;
+
+    /// @notice maximum BIPS
+    uint256 public constant MAX_BIPS = 10000;
+
+    /// @notice Array of roots of posted Merkle trees, as well as associated data like tree height
+    MerkleRootPost[] public merkleRoots;
```
ChaoticWalrus, line 36: IIRC we want this to be an internal array with a getter function that returns a struct in memory. otherwise the automatic getter function doesn't work how you'd expect 🙃

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    function _setRootPublisher(address _rootPublisher) internal {
+        rootPublisher = _rootPublisher;
+    }
+
+    function _setMerkleRootActivationDelay(uint256 _merkleRootActivationDelay) internal {
+        merkleRootActivationDelay = _merkleRootActivationDelay;
+    }
+
+    function _setEigenLayerShareBIPS(uint256 _eigenlayerShareBips) internal {
+        require(_eigenlayerShareBips <= MAX_BIPS, "PaymentCoordinator: EigenLayer share cannot be greater than 100%");
```
ChaoticWalrus, line 161: there should be a stronger guarantee than this. check the docs where we discussed this; I can't remember exactly what we agreed on

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -39,27 +40,12 @@ interface IPaymentCoordinator {
         IERC20[] tokens;
         // cumulative all-time earnings in each token
         uint256[] amounts;
+        //index in the merkle tree
+        uint256 index;
```
ChaoticWalrus, line 44: seems weird for the 'leaf' struct to have the index in it? In my mind that's part of the proof.

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    /// @notice Array of roots of posted Merkle trees, as well as associated data like tree height
+    MerkleRootPost[] public merkleRoots;
+
+    /// @notice Mapping token => recipient => cumulative amount *claimed*
+    mapping(IERC20 => mapping(address => uint256)) public cumulativeTokenAmountClaimedByRecipient;
+
+    /// @notice Mapping token => cumulative amount earned by EigenLayer
+    mapping(IERC20 => uint256) public cumulativeEigenLayerTokeEarnings;
+
+     /// @notice Constant that defines the share EigenLayer takes of all payments, in basis points
```
ChaoticWalrus, line 44: this documentation is wrong if this isn't going to be a constant.

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -39,27 +40,12 @@ interface IPaymentCoordinator {
...
-    /// @notice Array of roots of posted Merkle trees, as well as associated data like tree height
-    // MerkleRootPost[] public merkleRootPosts;
-    function merkleRootPosts(uint256 index) external view returns (MerkleRootPost memory);
-
     /// @notice Getter function for the length of the `merkleRootPosts` array
-    function merkleRootPostsLength() external view returns (uint256);
-
-    /// @notice Mapping token => recipient => cumulative amount *claimed*
-    // mapping(IERC20 => mapping(address => uint256)) public cumulativeTokenAmountClaimedByRecipient;
-    function cumulativeTokenAmountClaimedByRecipient(IERC20 token, address recipient) external view returns (uint256);
```
ChaoticWalrus, line 59: this should still be in the interface

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -39,27 +40,12 @@ interface IPaymentCoordinator {
...
-    // @notice Emitted when a new Merkle root is posted
-    event NewMerkleRootPosted(bytes32 root, uint256 height, uint256 confirmedAtBlockNumber, uint256 calculatedUpToBlockNumber);
-
-    /// @notice Array of roots of posted Merkle trees, as well as associated data like tree height
-    // MerkleRootPost[] public merkleRootPosts;
-    function merkleRootPosts(uint256 index) external view returns (MerkleRootPost memory);
-
     /// @notice Getter function for the length of the `merkleRootPosts` array
-    function merkleRootPostsLength() external view returns (uint256);
-
```
ChaoticWalrus, line 56: these should also still be in the interface

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -39,27 +40,12 @@ interface IPaymentCoordinator {
...
-    // TODO: better define this event?
-    event PaymentReceived(address indexed receivedFrom, Payment payment);
-
-    // @notice Emitted when a new Merkle root is posted
-    event NewMerkleRootPosted(bytes32 root, uint256 height, uint256 confirmedAtBlockNumber, uint256 calculatedUpToBlockNumber);
-
-    /// @notice Array of roots of posted Merkle trees, as well as associated data like tree height
-    // MerkleRootPost[] public merkleRootPosts;
-    function merkleRootPosts(uint256 index) external view returns (MerkleRootPost memory);
-
```
ChaoticWalrus, line 53: this should also still be in the interface

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -39,27 +40,12 @@ interface IPaymentCoordinator {
...
     /// @notice Getter function for the length of the `merkleRootPosts` array
-    function merkleRootPostsLength() external view returns (uint256);
-
-    /// @notice Mapping token => recipient => cumulative amount *claimed*
-    // mapping(IERC20 => mapping(address => uint256)) public cumulativeTokenAmountClaimedByRecipient;
-    function cumulativeTokenAmountClaimedByRecipient(IERC20 token, address recipient) external view returns (uint256);
-
-    // @notice Constant that defines the share EigenLayer takes of all payments, in basis points
-    function EIGENLAYER_SHARE_BIPS() external view returns (uint256);
+    function merkleRootsLength() external view returns (uint256);
```
ChaoticWalrus, line 48: was there a reason for renaming this? if we are calling the array `MerkleRootPosts`, I think `merkleRootPostsLength` is more technically accurate

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,158 @@
...
+    function testMakePayment(uint256[] memory amounts) public {
+        cheats.assume(amounts.length > 0);
+        uint256 sum = _sum(amounts);
+        PaymentCoordinator.Payment memory payment;
+        payment.token = dummyToken;
+        payment.amounts = amounts;
+
+        uint256 balanceBefore = dummyToken.balanceOf(address(paymentCoordinator));
+        uint256 cumumlativeEarningsBefore = paymentCoordinator.cumulativeEigenLayerTokeEarnings(dummyToken);
+        cheats.expectEmit(address(paymentCoordinator));
```
ChaoticWalrus, line 64: I'm pretty sure using `expectEmit` like this purely checks that the address emits _some_ event? 
Please correct me if I'm wrong but I've been using `expectEmit(true, true, true, true, address(...))` because as I understand it that's the strictest check.
Documentation here https://book.getfoundry.sh/cheatcodes/expect-emit

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,158 @@
...
+        merkleRootPost.height = height;
+        merkleRootPost.calculatedUpToBlockNumber = calculatedUpToBlockNumber;
+
+        uint256 numMerkleRootsBefore = paymentCoordinator.merkleRootsLength();
+
+        cheats.startPrank(paymentCoordinator.rootPublisher());
+        paymentCoordinator.postMerkleRoot(root, height, calculatedUpToBlockNumber);
+        cheats.stopPrank();
+
+        (bytes32 rootPost, uint256 heightPost, uint256 confirmedAtBlockNumber, uint256 calculatedUpToBlockNumberPost) = paymentCoordinator.merkleRoots(numMerkleRootsBefore);
```
ChaoticWalrus, line 91: this line seems excessively long, recommend breaking it up

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /// @notice Permissioned function which allows rootPublisher to nullify a Merkle root
+    function nullifyMerkleRoot(uint256 rootIndex) external onlyRootPublisher{
+        require(block.number <= merkleRoots[rootIndex].confirmedAtBlockNumber, "PaymentCoordinator.nullifyMerkleRoot: Merkle root already confirmed");
+        merkleRoots[rootIndex].root = bytes32(0);
+    }
+
+    // @notice Permissioned function which allows withdrawal of EigenLayer's share of `token` from all received payments
+    function withdrawEigenlayerShare(IERC20 token, address recipient) external onlyRootPublisher{
```
ChaoticWalrus, line 103: I 100% think this should be permissioned to a different address (at least the owner)

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,164 @@
...
+        uint256 amount = cumulativeEigenLayerTokeEarnings[token];
+        token.safeTransfer(recipient, amount);
+        cumulativeEigenLayerTokeEarnings[token] = 0;
```
ChaoticWalrus, line 106: this is anti-CEI. switch the line order.

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,158 @@
...
+
+    function testWithdrawEigenLayerShare(uint256[] memory amounts, address recipient) external {
+        cheats.assume(recipient != address(0));
+        testMakePayment(amounts);
+        uint256 amountToClaim = paymentCoordinator.cumulativeEigenLayerTokeEarnings(dummyToken);
+        cheats.startPrank(paymentCoordinator.rootPublisher());
+        paymentCoordinator.withdrawEigenlayerShare(dummyToken, recipient);
+        cheats.stopPrank();
+        require(paymentCoordinator.cumulativeEigenLayerTokeEarnings(dummyToken) == 0, "cumulativeEigenLayerTokeEarnings not updated correctly");
+        require(dummyToken.balanceOf(recipient) == amountToClaim, "incorrect token balance");
```
ChaoticWalrus, line 106: presumably want to check how the recipient's balance changes, rather than pure equivalence. also probably want to filter out the recipient being the paymentCoordinator itself or else this check will still break.

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,158 @@
...
+
+        emit log_named_uint("balance", dummyToken.balanceOf(recipient));
```
ChaoticWalrus, line 108: clean up logging before merging

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    function testNullifyMerkleRoot(bytes32 root, uint256 height, uint256 calculatedUpToBlockNumber) external {
+        testPostMerkleRoot(root, height, calculatedUpToBlockNumber);
+        paymentCoordinator.nullifyMerkleRoot(paymentCoordinator.merkleRootsLength() - 1);
```
ChaoticWalrus, line 113: seems worrying you aren't using any prank statement before this...
do we have a test that checks auth on the `nullifyMerkleRoot` function?

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,158 @@
...
+        (bytes32 rootPost, , , ) = paymentCoordinator.merkleRoots(paymentCoordinator.merkleRootsLength() - 1);
+        require(rootPost == bytes32(0), "root should be set");
```
ChaoticWalrus, line 115: see my comment about deleting the whole struct instead of just resetting one entry. I would also update this test to ensure the whole struct is deleted (just compare the returned struct to a new, empty struct instantiation -- this may require comparing hashes of the structs).

src/contracts/core/PaymentCoordinator.sol
ChaoticWalrus: nvm it's literally 2 lines up.

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,158 @@
...
+        leaf.tokens = new IERC20[](1);
+        cheats.expectRevert(bytes("PaymentCoordinator.proveAndClaimEarnings: Merkle root not yet confirmed"));
+        paymentCoordinator.proveAndClaimEarnings(new bytes(0), 0, leaf);
+    }
+
+    function _sum(uint[] memory numbers) internal returns (uint) {
+        uint total = 0;
+        
+        for (uint i = 0; i < numbers.length; i++) {
+            cheats.assume(numbers[i] < type(uint32).max && numbers[i] > 0);
```
ChaoticWalrus, line 150: kinda sus `assume` statement. very strict filtering you're doing here

ChaoticWalrus (changes requested):
some key functionality is wrong, otherwise just lots of more minor nits

src/contracts/core/PaymentCoordinator.sol
Sidu28: sorry that's copilot lol

src/contracts/core/PaymentCoordinator.sol
Sidu28: wait what's the difference in meaning - people in the office think they're the same?

src/contracts/core/PaymentCoordinator.sol
Sidu28: It is intended to be blocks

src/contracts/core/PaymentCoordinator.sol
Sidu28: added.

src/contracts/core/PaymentCoordinator.sol
Sidu28: Yeah.  Do we need quorums in the first place as far as the eigenlayer payments is concerned?  Seems like no?  Unless we do need it to be emitted for some offchain stuff

src/contracts/core/PaymentCoordinator.sol
Sidu28: ok so leaf amounts represents cumulativeAmount + newAmount?  I guess I've misunderstood the interface

src/contracts/core/PaymentCoordinator.sol
ChaoticWalrus: IMO:

- `cumulative` implies all-time (only increases)
- `accumulated` implies some kind of accumulation process (so in general it increases) but does not exclude the possibility of it decreasing.
- personally I don't think we should use either. if you need to attach a qualifier, `stored` might be more neutral

src/contracts/core/PaymentCoordinator.sol
ChaoticWalrus: yes, this is all formatted with off-chain calculations in mind

src/contracts/core/PaymentCoordinator.sol
ChaoticWalrus: IMO yes, making it public is nice for convenience.

src/contracts/core/PaymentCoordinator.sol
ChaoticWalrus: > ok so leaf amounts represents cumulativeAmount + newAmount? I guess I've misunderstood the interface

leaf amounts should only ever be greater for new trees. I would describe the amounts as "(all time) cumulative earnings". formatting things like this means users can wait and do single proof later, rather than being forced to prove against every root

src/contracts/core/PaymentCoordinator.sol
Sidu28: I think when we talked about it, the idea was that a rootPublisher, run by us, would be able to retract its own roots.  What's your thoughts about how the permissions should be?

src/contracts/core/PaymentCoordinator.sol
Sidu28: fuixed

src/contracts/core/PaymentCoordinator.sol
Sidu28: oh yeah, I think it was 15%

src/contracts/interfaces/IPaymentCoordinator.sol
Sidu28: This is for the parity of the leaf, which determines the order in which the proof elements are hashed with the leaf to verify it.  So the index is the index of the leaf in the entire tree

src/contracts/core/PaymentCoordinator.sol
Sidu28: whoops

src/contracts/interfaces/IPaymentCoordinator.sol
Sidu28: I've called the array merkleRoots

src/contracts/core/PaymentCoordinator.sol
Sidu28: yeah fair point

src/test/unit/PaymentCoordinatorUnit.t.sol
Sidu28: no I believe expectEmit checks that the provided addres emits the specified event below it.  The `expectEmit(true, true, true, true, address(...))` variation of the statement is for checking emissions of multiple events

src/test/unit/PaymentCoordinatorUnit.t.sol
Sidu28: fixed by returning a struct

src/test/unit/PaymentCoordinatorUnit.t.sol
Sidu28: yup good point

src/test/unit/PaymentCoordinatorUnit.t.sol
Sidu28: added test for this

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: yeah, I'm saying the index isn't _part of the leaf itself_ per se, it's the leaf's position in the tree. So it doesn't seem right for it to be part of the struct to me.

src/contracts/core/PaymentCoordinator.sol
Sidu28: done

src/contracts/interfaces/IPaymentCoordinator.sol
Sidu28: ah ok, I mean fair.  Felt it was cleaner, but we can make it another input.

ChaoticWalrus: I mean I'm OK with calling it LeafAndIndex or something, it just felt technically incorrect to call it a leaf / describe it as the contents of the leaf.

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: doesn't each entry in the array contain more than just a root though? feel like this is the same semantic debate as what we call a 'Leaf'

src/test/unit/PaymentCoordinatorUnit.t.sol
ChaoticWalrus: chatted about this offline, foundry docs seem unclear, `expectEmit(true, true, true, true, address(...))` seems safest/strictest.

src/contracts/interfaces/IPaymentCoordinator.sol
Sidu28: yeah in retrospect I agree with you on this as well.

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,198 @@
...
+    Initializable,
+    OwnableUpgradeable
+{
+    using SafeERC20 for IERC20;
+
+    /// @notice address approved to post new Merkle roots
+    address public rootPublisher;
+
+    /// @notice delay in blocks before a Merkle root can be activated
+    uint256 public merkleRootActivationDelayBlocks;
```
ChaoticWalrus, line 27: I feel like we might want this to be a constant, but am OK debating that later.

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,198 @@
...
+        }
+        payment.token.safeTransferFrom(msg.sender, address(this), sumAmounts);
+
+        accumulatedEigenLayerTokenEarnings[payment.token] += sumAmounts * eigenLayerShareBIPs / MAX_BIPS;
+        emit PaymentReceived(msg.sender, payment);
+    }
+
+    // @notice Permissioned function which allows posting a new Merkle root
+    function postMerkleRoot(bytes32 newRoot, uint256 height, uint256 calculatedUpToBlockNumber) external onlyRootPublisher {
+        MerkleRootPost memory newMerkleRoot = MerkleRootPost(newRoot, height, block.number + merkleRootActivationDelayBlocks, calculatedUpToBlockNumber);
```
ChaoticWalrus, line 106: I'd prefer the declaration explicitly list the struct fields, like:
```
MerkleRootPost memory newMerkleRoot = MerkleRootPost(
{
root: newRoot
...
}
);
```
I think it is clearer / easier to audit / harder to mess up.

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /// @notice Permissioned function which allows rootPublisher to nullify a Merkle root
+    function nullifyMerkleRoot(uint256 rootIndex) external onlyOwner {
+        require(block.number <= _merkleRootPosts[rootIndex].confirmedAtBlockNumber, "PaymentCoordinator.nullifyMerkleRoot: Merkle root already confirmed");
+        delete _merkleRootPosts[rootIndex];
+    }
+
+    // @notice Permissioned function which allows withdrawal of EigenLayer's share of `token` from all received payments
+    function withdrawEigenlayerShare(IERC20 token, address recipient) external onlyOwner {
```
ChaoticWalrus, line 118: inconsistent casing on EigenLayer: `Eigenlayer` => `EigenLayer`

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    function _setRootPublisher(address _rootPublisher) internal {
+        address currentRootPublisher = rootPublisher;
+        rootPublisher = _rootPublisher;
+        emit RootPublisherChanged(currentRootPublisher, rootPublisher);
+    }
+
+    function _setMerkleRootActivationDelayBlocks(uint256 _merkleRootActivationDelayBlocks) internal {
+        merkleRootActivationDelayBlocks = _merkleRootActivationDelayBlocks;
```
ChaoticWalrus, line 184: probably want some bounds on this check.
I am OK leaving it for the moment but I think we should have a `TODO` comment at least

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -78,18 +69,16 @@ interface IPaymentCoordinator {
 
     /**
     * @notice Called by a staker or operator to prove the inclusion of their earnings in a posted Merkle root and claim them.
-    * @param token ERC20 token to claim
-    * @param amount The `amount` contained in the leaf of the Merkle tree to be proved against the specified Merkle root
     * @param proof Merkle proof showing that a leaf containing `(msg.sender, amount)` was included in the `rootIndex`-th
     * Merkle root posted for the `token`
-    * @param nodeIndex Specifies the node inside the Merkle tree corresponding to the specified root, `merkleRoots[rootIndex].root`.
     * @param rootIndex Specifies the Merkle root to look up, using `merkleRootsByToken[token][rootIndex]`
+    * @param leaf The leaf to be inserted into the Merkle tree
```
ChaoticWalrus, line 75: ? the leaf is getting proven, not inserted, right?

ChaoticWalrus (approved):
Left a few minor changes that I'd like to see as comments.
LGTM overall; we can review more + refine a bit later.

PR #59, opened 2023-06-06
Unit test stake registry

gpsanant:
Added tests for the stake registry


src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -276,6 +276,15 @@ contract BLSRegistryCoordinatorWithIndices is Initializable, IBLSRegistryCoordin
...
 
+    /**
+     * @notice Updates the socket of the msg.sender given they are a registered operator
+     * @param socket is the new socket of the operator
+     */
+    function updateSocket(string memory socket) external {
+        require(_operators[msg.sender].status == OperatorStatus.REGISTERED, "BLSRegistryCoordinatorWithIndicies.updateSocket: operator is not registered");
+        emit OperatorSocketUpdate(_operators[msg.sender].operatorId, socket);
+    }
+
```
ChaoticWalrus, line 287: this should probably also be added to some interface, no?

please check for any other missing interface functions too!

gpsanant: done!

PR #61, opened 2023-06-07
Pauser changes

gpsanant:
- make pauser a mapping. role held by arbitrarily many addresses
- rename teamMultisig => operationsMultisig
- make TVL limits changeable by unpauser (previously this was permssioned to the pauser)


ChaoticWalrus (changes requested):
LGTM. This breaks the Certora Prover rule for the pauser but only because it is trying to access now non-existent functions. I am OK merging this and then merging a fix for that rule after.
Currently missing 'pauser'=>'unpauser' change for setting maxPods; once that's in then this is good to merge.

PR #67, opened 2023-06-07
Fix prover CI

ChaoticWalrus:
This PR is intended to fix the breakages in the Certora Prover aspects of our CI that appeared with the official introduction of CVL2 by Certora.
This is an interim solution; we still need to migrate to CVL2 in the next couple months, before CVL1 support is deprecated!


gpsanant (approved):
looks good, nice job on the prompt fix

PR #69, opened 2023-06-09
add mainnet config

gpsanant:
delploy script working


ChaoticWalrus (approved):
All looks correct to me.

PR #75, opened 2023-06-20
Adding a contract for viewing operator state

Sidu28:


Sidu28:
@gpsanant several things here.

 - We need to assume that the nodes querying will have the correct index in the history mappings for stake history, for eg.
 - We do not store a history of the actual G1 apk, just the hash of it.  "TotalOperators" struct in the da spec specifies a apk element.
 - Several of the structs, such as TotalOperators in the DA spec specify an "Index" parameter.  What is this in reference to?

PR #81, opened 2023-06-27
Edit bls sig checker

gpsanant:
- Renamed Library to BitmapUtils
- Added Registry System State View functions
- Moved events to interfaces
- Added stake registry tests (no need to review, these will be refactored in another PR)


src/contracts/interfaces/IStakeRegistry.sol
```diff
@@ -61,6 +69,15 @@ interface IStakeRegistry is IRegistry {
      */
     function getTotalStakeUpdateForQuorumFromIndex(uint8 quorumNumber, uint256 index) external view returns (OperatorStakeUpdate memory);
 
+    /// @notice Returns the indices of the operator stakes for the provided `quorumNumber` at the given `blockNumber`
+    function getStakeUpdateIndexForOperatorIdForQuorumAtBlockNumber(bytes32 operatorId, uint8 quorumNumber, uint32 blockNumber)
+        external
+        view
+        returns (uint32);
+
+    /// @notice Returns the indices of the total stakes for the provided `quorumNumbers` at the given `blockNumber`
```
ChaoticWalrus, line 78: this is actually the number of operators at the time, right?

src/contracts/libraries/BN254.sol
```diff
@@ -131,6 +131,29 @@ library BN254 {
...
+            // if the bit is 1, add the 2^n*p to the accumulated product
+            if (s >> i & 1 == 1) {
+                acc = plus(acc, p2n);
+            }
+            // double the 2^n*p for the next iteration
+            p2n = plus(p2n, p2n);
+        }
+        return acc;
+    }
+
```
ChaoticWalrus, line 156: what's the usage for this?
Should it explicitly enforce `s < (2**10)?

gpsanant: its a way to add without calling the ecMul precompile. @dev is a precondition and doesn't require to save gas. not sure whether this is actually more efficient because it needs the extra call gas to the precompile, but its a start.

ChaoticWalrus: OK, gotcha. IMO add the require statement for safety. The tiny bit of extra gas seems worth it to make the library more idiot-proof

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,123 @@
...
+        registryCoordinator = _registryCoordinator;
+
+        stakeRegistry = _registryCoordinator.stakeRegistry();
+        blsPubkeyRegistry = _registryCoordinator.blsPubkeyRegistry();
+        indexRegistry = _registryCoordinator.indexRegistry();
+    }
+
+    /**
+     * @notice returns the ordered list of operators (id and stake) for each quorum
+     * @param operatorId the id of the operator calling the function
```
ChaoticWalrus, line 40: this address isn't _calling the function_ per se

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,123 @@
...
+    constructor(IBLSRegistryCoordinatorWithIndices _registryCoordinator) {
+        registryCoordinator = _registryCoordinator;
+
+        stakeRegistry = _registryCoordinator.stakeRegistry();
+        blsPubkeyRegistry = _registryCoordinator.blsPubkeyRegistry();
+        indexRegistry = _registryCoordinator.indexRegistry();
+    }
+
+    /**
+     * @notice returns the ordered list of operators (id and stake) for each quorum
```
ChaoticWalrus, line 39: missing an apostrophe (and ticks)? `operators` => ```operator's```?

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,123 @@
...
+     * @param operatorId the id of the operator calling the function
+     * @return 2d array of operators. For each quorum the provided operaor is a part of, a ordered list of operators
```
ChaoticWalrus, line 41: these are `Operator` structs, not 'operators' -- either the struct name is confusing or this wording is confusing (or both!).
also typo: `operaor` => `operators`

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,123 @@
...
+pragma solidity =0.8.12;
+
+import "./BLSRegistryCoordinatorWithIndices.sol";
+
+import "../interfaces/IStakeRegistry.sol";
+import "../interfaces/IBLSPubkeyRegistry.sol";
+import "../interfaces/IIndexRegistry.sol";
+import "../interfaces/IBLSRegistryCoordinatorWithIndices.sol";
+
+
```
ChaoticWalrus, line 11: needs high-level explanation.

src/contracts/middleware/BLSOperatorStateRetriever.sol
ChaoticWalrus: nevermind, misunderstood the function. 
Specifically, this returns a 2D array, where each sub-array is an ordered list of the current operators for a single quorum, and the set of quorums whose data is returned matches the set of quorums that the provided `operatorId` is actively registered for.

src/contracts/middleware/BLSOperatorStateRetriever.sol
ChaoticWalrus: after better understanding this function, I think this is less bad than I originally considered, but I still think there's room for a bit of clarity here.

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,123 @@
...
+    function getOperatorState(bytes32 operatorId) external view returns (Operator[][] memory) {
+        bytes memory quorumNumbers = BitmapUtils.bitmapToBytesArray(registryCoordinator.getCurrentQuorumBitmapByOperatorId(operatorId));
+
+        return getOperatorState(quorumNumbers);
+    }
+
+    /**
+     * @notice returns the ordered list of operators (id and stake) for each quorum
+     * @param quorumNumbers are the ids of the quorums to get the operator state for
+     * @return 2d array of operators. For each quorum, a ordered list of operators
```
ChaoticWalrus, line 52: `a` => `an` (also in at least one other place)

src/contracts/interfaces/IStakeRegistry.sol
gpsanant: yessir

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /**
+     * @notice returns 
+     *          1) the indices of the quorumBitmaps for each of given operators at the given blocknumber
+     *          2) the indices of the total stakes entries for the given quorums at the given blocknumber
+     *          3) the indices of the stakes of each of the nonsigners in each of the quorums they are a 
+     *             part of (for each nonsigner, an array of length the number of quorums they a part of 
+     *             that are also part of the provided quorumNumbers) at the given blocknumber
+     *          4) the indices of the quourm apks for each of the provided quorums at the given blocknumber
```
ChaoticWalrus, line 79: typo in spelling of 'quorum'

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,123 @@
...
+        return operators;
+    }
+
+    /**
+     * @notice returns 
+     *          1) the indices of the quorumBitmaps for each of given operators at the given blocknumber
+     *          2) the indices of the total stakes entries for the given quorums at the given blocknumber
+     *          3) the indices of the stakes of each of the nonsigners in each of the quorums they are a 
+     *             part of (for each nonsigner, an array of length the number of quorums they a part of 
+     *             that are also part of the provided quorumNumbers) at the given blocknumber
```
ChaoticWalrus, line 78: it looks like this is not using _not_ the operators' current sets of quorums (as implied by the wording `quorums they a part of`), but rather the operators' past quorum-sets (i.e. "quorums that they were a part of at the `referenceBlockNumber`")

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,123 @@
...
+                    stake: stakeRegistry.getMostRecentStakeUpdateByOperatorId(operatorId, quorumNumber).stake
+                });
+            }
+        }
+            
+        return operators;
+    }
+
+    /**
+     * @notice returns
```
ChaoticWalrus, line 73: honestly a high-level explanation for "why" this function exists (and probably other functions in this contract) would help

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/4c961a135ce40d6e1c110441d7f8c62627ef2199

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/302f056cb5b572e4f107aaa92bc22cd9b3548f40

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/48df1fd6538b1ebfede3a4fdb39140d25dbd52d7

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,134 @@
...
+            for (uint8 j = 0; j < 192; j++) {
+                // if the operator is a part of the quorum and the quorum is a part of the provided quorums
+                if (nonSignerQuorumBitmap >> j & (quorumBitmap >> j & 1) == 1) {
+                    checkSignaturesIndices.nonSignerStakeIndices[i][j] = stakeRegistry.getStakeUpdateIndexForOperatorIdForQuorumAtBlockNumber(
+                        nonSignerOperatorIds[i],
+                        uint8(j),
+                        referenceBlockNumber
+                    );
+                }
+            }
```
ChaoticWalrus, line 127: So in this case we fill the entry with the proper stake update index, and otherwise (i.e. in all other cases), we leave the value as 0, right?
Possible bug: _what happens when the correct stake update index is zero_? How do we still differentiate?

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -91,6 +91,22 @@ contract BLSPubkeyRegistry is IBLSPubkeyRegistry, Test {
         return pubkeyHash;
     }
 
+    /// @notice Returns the index of the quorumApk index at `blockNumber` for the provided `quorumNumber`
```
ChaoticWalrus, line 94: this comment is written like it's for a single entry but this actually is for (and returns) an array

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -91,6 +91,22 @@ contract BLSPubkeyRegistry is IBLSPubkeyRegistry, Test {
...
+    function getApkIndicesForQuorumsAtBlockNumber(bytes calldata quourmNumbers, uint256 blockNumber) external view returns(uint32[] memory){
```
ChaoticWalrus, line 95: typo: mispelling of 'quorum'

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,134 @@
...
+        uint32[][] nonSignerStakeIndices; // nonSignerStakeIndices[quorumNumberIndex][nonSignerIndex]
+    }
+
+    IBLSRegistryCoordinatorWithIndices public registryCoordinator;
+    IStakeRegistry public stakeRegistry;
+    IBLSPubkeyRegistry public blsPubkeyRegistry;
+    IIndexRegistry public indexRegistry;
+
+    constructor(IBLSRegistryCoordinatorWithIndices _registryCoordinator) {
+        registryCoordinator = _registryCoordinator;
```
ChaoticWalrus, line 34: this shouldn't be an AVS-specific contract. we should just deploy one and use the registryCoordinator as an input to functions.

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -91,6 +91,22 @@ contract BLSPubkeyRegistry is IBLSPubkeyRegistry, Test {
...
+        uint256[] memory indices = new uint256[](quourmNumbers.length);
+        for (uint i = 0; i < quourmNumbers.length; i++) {
+            uint8 quorumNumber = uint8(quourmNumbers[i]);
+            uint32 length = uint32(quorumApkUpdates[quorumNumber].length);
```
ChaoticWalrus, line 99: I feel like this should be named something like `qorumApkUpdatesLength` instead of just `length`

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -96,6 +96,26 @@ contract BLSRegistryCoordinatorWithIndices is Initializable, IBLSRegistryCoordin
...
 
+    /// @notice Returns the indices of the quorumBitmaps for the provided `operatorIds` at the given `blockNumber`
+    function getQuorumBitmapIndicesByOperatorIdsAtBlockNumber(uint32 blockNumber, bytes32[] memory operatorIds) external view returns (uint32[] memory) {
+        uint32[] memory indices = new uint32[](operatorIds.length);
+        for (uint256 i = 0; i < operatorIds.length; i++) {
+            uint32 length = uint32(_operatorIdToQuorumBitmapHistory[operatorIds[i]].length);
+            for (uint32 j = 0; j < length; j++) {
+                if(_operatorIdToQuorumBitmapHistory[operatorIds[i]][length - j - 1].updateBlockNumber <= blockNumber) {
+                    require(
+                        _operatorIdToQuorumBitmapHistory[operatorIds[i]][length - j - 1].nextUpdateBlockNumber != 0 ||
```
ChaoticWalrus, line 107: possible bug: shouldn't this be `==0` instead of `!=0`?

src/contracts/middleware/BLSSignatureChecker.sol
```diff
@@ -154,6 +185,16 @@ abstract contract BLSSignatureChecker {
...
+    /// @return count number of ones in binary representation of `n`
+    function countNumOnes(uint256 n) public pure returns (uint16) {
+        uint16 count = 0;
+        while (n > 0) {
+            n &= (n - 1);
+            count++;
+        }
+        return count;
+    }
+
```
ChaoticWalrus, line 197: isn't this in the BitmapUtils library now? seems like code duplication.

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -141,6 +134,20 @@ contract IndexRegistry is IIndexRegistry {
...
+    /// @notice Returns an ordered list of operators of the services for the given `quorumNumber` at the given `blockNumber`
+    function getOperatorListForQuorumAtBlockNumber(uint8 quorumNumber, uint32 blockNumber) external view returns (bytes32[] memory){
+        bytes32[] memory quorumOperatorList = new bytes32[](_getTotalOperatorsForQuorumAtBlockNumber(quorumNumber, blockNumber));
+        for (uint i = 0; i < globalOperatorList.length; i++) {
+            bytes32 operatorId = globalOperatorList[i];
+            uint32 index = _getIndexOfOperatorForQuorumAtBlockNumber(operatorId, quorumNumber, blockNumber);
+            // if the operator was not in the quorum at the given block number, skip it
+            if(index == type(uint32).max)
+                continue;
+            quorumOperatorList[index] = operatorId;
```
ChaoticWalrus, line 146: we need to make sure that this behavior -- specifically, returning the max uint32 from `_getIndexOfOperatorForQuorumAtBlockNumber` when the operator was not in the quorum at the given block number -- is super clearly documented!

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -207,18 +214,44 @@ contract IndexRegistry is IIndexRegistry {
...
         }
-        globalOperatorList.pop();
+        
+        // if there is only one entry in the total operator history, return it
+        if (_totalOperatorsHistory[quorumNumber].length == 1) {
+            return _totalOperatorsHistory[quorumNumber][_totalOperatorsHistory[quorumNumber].length - 1].index;
+        }
+
+        // loop backwards through the total operator history to find the total number of operators at the given block number
+        for (uint i = _totalOperatorsHistory[quorumNumber].length - 2; i >= 0; i--) {
```
ChaoticWalrus, line 230: possible bug: shouldn't we still be starting at `_totalOperatorsHistory[quorumNumber].length - 1`? why skip the last entry? what if it is the correct one (and there are 2 or more entries in the history)?

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -207,18 +214,44 @@ contract IndexRegistry is IIndexRegistry {
...
+            OperatorIndexUpdate memory totalOperatorUpdate = _totalOperatorsHistory[quorumNumber][i];
+            if(totalOperatorUpdate.toBlockNumber <= blockNumber){
+                return totalOperatorUpdate.index;
+            }
+        }        
+        return 0;
+    }
+    
+
+    /// @notice Returns the index of the `operatorId` at the given `blockNumber` fro the given `quorumNumber`, or max uint32 if the operator is not active in the quorum
```
ChaoticWalrus, line 240: typo: `fro`=>`for`.
glad to see the note here in the documentation about returning max uint32 as a specific indicator 👍

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -207,18 +214,44 @@ contract IndexRegistry is IIndexRegistry {
...
+    function _getIndexOfOperatorForQuorumAtBlockNumber(bytes32 operatorId, uint8 quorumNumber, uint32 blockNumber) internal view returns(uint32) {
+        OperatorIndexUpdate memory operatorIndexUpdate;
+        // set to max uint32 value to indicate that the operator is not part of the quorum at all, until this is updated in the loop
+        operatorIndexUpdate.toBlockNumber = type(uint32).max; 
+        // loop forward through index history to find the index of the operator at the given block number
+        // this is less efficient than looping forwards, but is simpler logic and only called in view functions that aren't mined onchain
```
ChaoticWalrus, line 246: these seem to conflict. we are indeed looping forwards here, so this can't be "less efficient than looping forwards".

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -207,18 +214,44 @@ contract IndexRegistry is IIndexRegistry {
...
+        for (uint i = 0; i < _operatorIdToIndexHistory[operatorId][quorumNumber].length; i++) {
+            operatorIndexUpdate = _operatorIdToIndexHistory[operatorId][quorumNumber][i];
+            if(operatorIndexUpdate.toBlockNumber >= blockNumber){
```
ChaoticWalrus, line 249: formatting nit: `){` => `) {` (add a space) -- also `if(` => `if (` (again add a space)

src/contracts/middleware/StakeRegistry.sol
```diff
@@ -92,6 +85,32 @@ contract StakeRegistry is StakeRegistryStorage {
...
+        external
+        view
+        returns (uint32)
+    {
+        return _getStakeUpdateIndexForOperatorIdForQuorumAtBlockNumber(operatorId, quorumNumber, blockNumber);
+    }
+
+
+    /// @notice Returns the indices of the total stakes for the provided `quorumNumbers` at the given `blockNumber`
+    function getTotalStakeIndicesByQuorumNumbersAtBlockNumber(uint32 blockNumber, bytes calldata quourmNumbers) external view returns(uint32[] memory) {
```
ChaoticWalrus, line 99: typo again on spelling of 'quorum'

src/contracts/middleware/StakeRegistry.sol
```diff
@@ -265,6 +293,21 @@ contract StakeRegistry is StakeRegistryStorage {
 
     // INTERNAL FUNCTIONS
 
+    function _getStakeUpdateIndexForOperatorIdForQuorumAtBlockNumber(bytes32 operatorId, uint8 quorumNumber, uint32 blockNumber) internal view returns(uint32) {
+        uint32 length = uint32(operatorIdToStakeHistory[operatorId][quorumNumber].length);
+        for (uint32 i = 0; i < length; i++) {
+            if (operatorIdToStakeHistory[operatorId][quorumNumber][length - i - 1].updateBlockNumber <= blockNumber) {
+                require(
+                    operatorIdToStakeHistory[operatorId][quorumNumber][length - i - 1].nextUpdateBlockNumber != 0 ||
```
ChaoticWalrus, line 301: same comment as above (`!=0` vs. `==0`)

src/contracts/middleware/VoteWeigherBase.sol
```diff
@@ -59,29 +59,25 @@ contract VoteWeigherBase is VoteWeigherBaseStorage {
      * @notice This function computes the total weight of the @param operator in the quorum @param quorumNumber.
      * @dev returns zero in the case that `quorumNumber` is greater than or equal to `quorumCount`
      */
-    function weightOfOperator(uint8 quorumNumber, address operator) public virtual returns (uint96) {
+    function weightOfOperator(uint8 quorumNumber, address operator) public virtual validQuorumNumber(quorumNumber) returns (uint96) {
```
ChaoticWalrus, line 62: the dev tag is now outdated with this change. the function now *reverts* rather than returning 0

src/test/unit/StakeRegistryUnit.t.sol
```diff
@@ -0,0 +1,234 @@
...
+
+        stakeRegistry = StakeRegistryHarness(
+            address(
+                new TransparentUpgradeableProxy(
+                    address(stakeRegistryImplementation),
+                    address(proxyAdmin),
+                    abi.encodeWithSelector(
+                        StakeRegistry.initialize.selector,
+                        minimumStakeForQuorum,
+                        quorumStrategiesConsideredAndMultipliers // initialize with 0ed out 128 quorums
```
ChaoticWalrus, line 113: comment is incorrect. multipliers are 1-128

ChaoticWalrus (changes requested):
Think I've finished looking through all of this now. I've dropped a bunch of comments that I'd appreciate you addressing.
For the possible bugs, if any are real I think _we should implement tests that catch these bugs_, since clearly the existing tests are missing them.

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/85137533052f6db3db219880c3d77a7f1098fb58

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/85137533052f6db3db219880c3d77a7f1098fb58
https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/10877ff5ae1d916fee00d413f63a3ac1900b3439

src/contracts/middleware/StakeRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/802ff35f8f19ae61cc7ae882d1d55f7327d59687

src/contracts/middleware/BLSPubkeyRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/d4daf8507a5799729702562ebdd1e37b6f88caee

src/contracts/middleware/BLSPubkeyRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/d4daf8507a5799729702562ebdd1e37b6f88caee

src/contracts/middleware/BLSSignatureChecker.sol
gpsanant: yes!, thought we were going to merge these at different times. https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/4b9988c1fa57592bda1cab50540f32831fc09310

src/contracts/libraries/BN254.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/a4890ff2b3a0bedffa1bf33d71c8006aa97ea659

src/contracts/middleware/IndexRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/80c8beea00ed127e3101942a071860fb029f9262

src/contracts/middleware/IndexRegistry.sol
gpsanant: copilot :grimacing:

src/contracts/middleware/IndexRegistry.sol
gpsanant: agreed

src/contracts/middleware/IndexRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/7d71e880e8adaaeca4a5b8a919abcbb852610fa7 ah, i think this fixes?

src/contracts/middleware/VoteWeigherBase.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/7d71e880e8adaaeca4a5b8a919abcbb852610fa7

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: f1faa75d ah, this should fix it. actually was a bug in general compared to how the sigchecker uses it

src/contracts/middleware/StakeRegistry.sol
gpsanant: 06bca742 good find

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: 06bca742 good find

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/57d343906500c946be3c4f599c502cf08e9893a9

src/contracts/libraries/BN254.sol
```diff
@@ -131,6 +131,30 @@ library BN254 {
         require(success, "ec-add-failed");
     }
 
+    /**
+     * @notice an optimized ecMul implementation that takes O(log_2(s)) ecAdds
+     * @param p the point to multiply
+     * @param s the scalar to multiply by
+     * @dev this function is only safe to use if the scalar is 9 bits or less
```
ChaoticWalrus, line 138: going to suggest here as well that we just do an explicit check for this and revert if the condition isn't met.
I'd rather the library be ~0.01% less gas-efficient than there be a single bug related to this

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,134 @@
...
+
+    struct CheckSignaturesIndices {
+        uint32[] nonSignerQuorumBitmapIndices;
+        uint32[] quorumApkIndices;
+        uint32[] totalStakeIndices;  
+        uint32[][] nonSignerStakeIndices; // nonSignerStakeIndices[quorumNumberIndex][nonSignerIndex]
+    }
+
+    /**
+     * @notice This function is intended to by AVS nodes every time there is new triggered tasks. Since all of
```
ChaoticWalrus, line 29: `to by` => `to be called by`?
`there is new triggered tasks` => `a new task is created for the AVS`?
not sure what the intended meaning of this sentence is, exactly

also `of crucial` => `of the crucial`

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * @notice This function is intended to by AVS nodes every time there is new triggered tasks. Since all of 
+     * crucial information is kept onchain, nodes don't need to run indexers to fetch the data.
+     * @param registryCoordinator is the registry coordinator to fetch the AVS registry information from
+     * @param operatorId the id of the operator to fetch the quorums lists 
+     * @param blockNumber is the block number to get the operator state for
+     * @return 1) the quorumBitmap of the operator at the given blockNumber
+     *         2) 2d array of Operator tructs. For each quorum the provided operator
```
ChaoticWalrus, line 35: typo: `tructs` => `structs`

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,134 @@
...
+        uint256 quorumBitmap = registryCoordinator.getQuorumBitmapByOperatorIdAtBlockNumberByIndex(operatorId, blockNumber, index);
+
+        bytes memory quorumNumbers = BitmapUtils.bitmapToBytesArray(quorumBitmap);
+
+        return (quorumBitmap, getOperatorState(registryCoordinator, quorumNumbers, blockNumber));
+    }
+
+    /**
+     * @notice returns the ordered list of operators (id and stake) for each quorum. The AVS coordinator 
+     * may call this function directly to get the operator state for a given block number.s
```
ChaoticWalrus, line 51: typo: errant `.s`?

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * @notice this is called by the AVS operator to get the relevant indices for the checkSignatures function
+     * if they are not running an indexer    
+     * @param registryCoordinator is the registry coordinator to fetch the AVS registry information from
+     * @param referenceBlockNumber is the block number to get the indices for
+     * @param quorumNumbers are the ids of the quorums to get the operator state for
+     * @param nonSignerOperatorIds are the ids of the nonsigning operators
+     * @return 1) the indices of the quorumBitmaps for each of given operators at the given blocknumber
```
ChaoticWalrus, line 85: `given operators` => `the operators in the provided 'nonSignerOperatorIds' array`

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,134 @@
...
+     *         2) the indices of the total stakes entries for the given quorums at the given blocknumber
+     *         3) the indices of the stakes of each of the nonsigners in each of the quorums they were a 
+     *            part of (for each nonsigner, an array of length the number of quorums they were a part of
+     *            that are also part of the provided quorumNumbers) at the given blocknumber
```
ChaoticWalrus, line 89: question: if non-signers are/were only in a subset of quorums, how do we know which quorum each of the returned indices corresponds to?

ChaoticWalrus: or does that not matter and get looked up automatically?

src/contracts/middleware/IndexRegistry.sol
ChaoticWalrus: can you check (i.e. _write a test_) that the loop doesn't fail for trying to underflow when the correct entry is the 0-indexed one?
I previously had some interesting underflow behavior w/ Solidity when using decrementing loops.

Also, it seems like right now this will return `1` (the value of `_totalOperatorsHistory[quorumNumber][0].index`, right?) even if the provided `blockNumber` is before the first operator joined.

ChaoticWalrus: oh, also going to point out that this implementation would return `_totalOperatorsHistory[quorumNumber].length - 1` for an input value of `blockNumber` that is in the future, but I guess that's just a general behavior with these types of function / doesn't seem like a particular problem?
Maybe something to note with an `@dev` tag though.

gpsanant: i'm trying to push off writing tests to a future pr. this functionality more needs to pass out mental test to get merged into multiquourms, but needs to severely tested before master

ChaoticWalrus: OK, fair enough -- I'm OK putting off the serious tests until later.
I still think this is somewhat broken for at least one edge case, and am ~50% sure it will break with an underflow revert in the loop (and that `i > 0` is the correct condition).

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: gets looked up automatically. they are in ascending order of "1" bits in the operators quorum bitmap at the block number.

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/be2efe90a2630d953b4da4757b4dd40427df688f

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/be2efe90a2630d953b4da4757b4dd40427df688f

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/be2efe90a2630d953b4da4757b4dd40427df688f

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/be2efe90a2630d953b4da4757b4dd40427df688f does this help?

src/contracts/middleware/IndexRegistry.sol
gpsanant: ahh, i think you're right. here: 1e35bd8d

src/contracts/middleware/BLSOperatorStateRetriever.sol
ChaoticWalrus: yes

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -0,0 +1,135 @@
...
+        checkSignaturesIndices.nonSignerStakeIndices = new uint32[][](nonSignerOperatorIds.length);
+        for (uint i = 0; i < nonSignerOperatorIds.length; i++) {
+            uint192 nonSignerQuorumBitmap = 
+                registryCoordinator.getQuorumBitmapByOperatorIdAtBlockNumberByIndex(
+                    nonSignerOperatorIds[i], 
+                    referenceBlockNumber, 
+                    checkSignaturesIndices.nonSignerQuorumBitmapIndices[i]
+                );
+            // the number of quorums the operator was a part of that are also part of the provided quorumNumbers
+            checkSignaturesIndices.nonSignerStakeIndices[i] = new uint32[](BitmapUtils.countNumOnes(nonSignerQuorumBitmap & BitmapUtils.bytesArrayToBitmap(quorumNumbers)));
```
ChaoticWalrus, line 114: is there a reason for moving the `BitmapUtils.bytesArrayToBitmap(quorumNumbers)` operation inside the loop in this commit https://github.com/Layr-Labs/eigenlayer-contracts/commit/f1faa75d5e0ac150dbaa1e7a1dcd17298317f2c3 ?
seems both more inefficient (I realize that's not important here) and maybe a bit harder to parse? AFAICT it's the same calculation/output for every loop iteration.

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -199,26 +206,54 @@ contract IndexRegistry is IIndexRegistry {
...
+        // loop backwards through the total operator history to find the total number of operators at the given block number
+        uint256 totalOperatorsHistoryLength = _totalOperatorsHistory[quorumNumber].length;
+        for (uint i = 0; i <= totalOperatorsHistoryLength - 2; i++) {
+            uint256 index = totalOperatorsHistoryLength - 2 - i;
+            OperatorIndexUpdate memory totalOperatorUpdate = _totalOperatorsHistory[quorumNumber][index];
+            if (totalOperatorUpdate.toBlockNumber <= blockNumber) {
+                return _totalOperatorsHistory[quorumNumber][index + 1].index;
+            }
+        }        
+        return _totalOperatorsHistory[quorumNumber][0].index;
```
ChaoticWalrus, line 238: OK, am going to suggest a rewrite:
```
    /// @notice Returns the total number of operators of the service for the given `quorumNumber` at the given `blockNumber`
    /// @dev Returns zero if the @param blockNumber is from before the @param quorumNumber existed, and returns the current number of total operators if the @param blockNumber is in the future.
    function _getTotalOperatorsForQuorumAtBlockNumber(uint8 quorumNumber, uint32 blockNumber) internal view returns (uint32){
// store list length in memory
uint256 totalOperatorsHistoryLength = _totalOperatorsHistory[quorumNumber].length;
        // if there are no entries in the total operator history, return 0
        if (totalOperatorsHistoryLength == 0) {
            return 0;
        }
        
// if `blockNumber` is in future, return current number of operators
if (blockNumber > uint32(block.number)) {
return _totalOperatorsHistory[quorumNumber][totalOperatorsHistoryLength - 1].index;
// else if `blockNumber` is from before the `quorumNumber` existed, return `0`
} else if (`blockNumber` < _totalOperatorsHistory[quorumNumber][0].fromBlockNumber {
return 0;
}

        // loop backwards through the total operator history to find the total number of operators at the given block number
        for (uint256 i = 0; i <= totalOperatorsHistoryLength - 1; i++) {
            uint256 listIndex = (totalOperatorsHistoryLength - 1) - i;
            OperatorIndexUpdate memory totalOperatorUpdate = _totalOperatorsHistory[quorumNumber][listIndex];
            // look for the first update that began at or after `blockNumber`
            if (totalOperatorUpdate.fromBlockNumber >= blockNumber) {
                return _totalOperatorsHistory[quorumNumber][listIndex].index;
            }
        }        
        return _totalOperatorsHistory[quorumNumber][0].index;
```

I believe this has all the properties it should.

ChaoticWalrus:
Looks like a fuzzed test is failing at the moment too?

src/contracts/middleware/BLSOperatorStateRetriever.sol
gpsanant: stack too deep :(

src/contracts/middleware/IndexRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/57d8aca9b8c542f7781985eb7dd78b38e9069956

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -130,25 +123,33 @@ contract IndexRegistry is IIndexRegistry {
     function getTotalOperatorsForQuorumAtBlockNumberByIndex(uint8 quorumNumber, uint32 blockNumber, uint32 index) external view returns (uint32){
         OperatorIndexUpdate memory operatorIndexToCheck = _totalOperatorsHistory[quorumNumber][index];
 
-        // blocknumber must be before the "index'th" entry's toBlockNumber
-        require(blockNumber <= operatorIndexToCheck.toBlockNumber, "IndexRegistry.getTotalOperatorsForQuorumAtBlockNumberByIndex: provided index is too far in the future for provided block number");
+        // blocknumber must be before the "index'th" entry's fromBlockNumber
```
ChaoticWalrus, line 126: 'before' => 'at or after'

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -130,25 +123,33 @@ contract IndexRegistry is IIndexRegistry {
...
         
-        // if there is an index update before the "index'th" update, the blocknumber must be after the previous entry's toBlockNumber
-        if (index != 0){
-            OperatorIndexUpdate memory previousOperatorIndex = _totalOperatorsHistory[quorumNumber][index - 1];
-            require(blockNumber > previousOperatorIndex.toBlockNumber, "IndexRegistry.getTotalOperatorsForQuorumAtBlockNumberByIndex: provided index is too far in the past for provided block number");
+        // if there is an index update after the "index'th" update, the blocknumber must be before the previous entry's fromBlockNumber
+        if (index != _totalOperatorsHistory[quorumNumber].length - 1){
+            OperatorIndexUpdate memory previousOperatorIndex = _totalOperatorsHistory[quorumNumber][index + 1];
+            require(blockNumber < previousOperatorIndex.fromBlockNumber, "IndexRegistry.getTotalOperatorsForQuorumAtBlockNumberByIndex: provided index is too far in the future for provided block number");
```
ChaoticWalrus, line 132: OK, just so we're on the same page -- the canonical number of operators "at a block" is the number of operators at the last update for the block.
Makes sense to me; we should do what we can to make this crystal clear

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -130,25 +123,33 @@ contract IndexRegistry is IIndexRegistry {
...
+        // blocknumber must be before the "index'th" entry's fromBlockNumber
+        require(blockNumber >= operatorIndexToCheck.fromBlockNumber, "IndexRegistry.getTotalOperatorsForQuorumAtBlockNumberByIndex: provided index is too far in the past for provided block number");
         
-        // if there is an index update before the "index'th" update, the blocknumber must be after the previous entry's toBlockNumber
-        if (index != 0){
-            OperatorIndexUpdate memory previousOperatorIndex = _totalOperatorsHistory[quorumNumber][index - 1];
-            require(blockNumber > previousOperatorIndex.toBlockNumber, "IndexRegistry.getTotalOperatorsForQuorumAtBlockNumberByIndex: provided index is too far in the past for provided block number");
+        // if there is an index update after the "index'th" update, the blocknumber must be before the previous entry's fromBlockNumber
+        if (index != _totalOperatorsHistory[quorumNumber].length - 1){
+            OperatorIndexUpdate memory previousOperatorIndex = _totalOperatorsHistory[quorumNumber][index + 1];
```
ChaoticWalrus, line 131: `previousOperatorIndex` => `followingOperatorIndex`

ChaoticWalrus:
LGTM @gpsanant 👍 
Please do correct the couple typos/naming issues, and maybe add some documentation/notes about the canonical "number of operators at a block"

ChaoticWalrus (approved):

src/contracts/middleware/IndexRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/01f331ade2b5ca39933d03828ab4ba0cab624766

src/contracts/middleware/IndexRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/01f331ade2b5ca39933d03828ab4ba0cab624766

src/contracts/middleware/IndexRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/81/commits/01f331ade2b5ca39933d03828ab4ba0cab624766

PR #82, opened 2023-06-29
Update AVS-Guide.md

samlaf:
Some minor changes, which reflected my new understandings in light of today's discussions. Let me know if some of my updates are wrong.

Also this figure needs to be updated:
image
Middle interactions should be from operators, not stakers, I think?


Sidu28 (approved):
Made some small edits, LGTM thanks for this

PR #85, opened 2023-06-29
change apkupdate storage

gpsanant:
update storage to include a bytes24 instead of a bytes32 to save on storage. this doesn't effect security since ethereum has 20 byte addresses.


ChaoticWalrus (approved):
LGTM. Still safer than ETH address security, so should be fine 👍

PR #86, opened 2023-07-03
Unit Test registry coordinator

gpsanant:


src/contracts/interfaces/IStakeRegistry.sol
```diff
@@ -60,6 +60,9 @@ interface IStakeRegistry is IRegistry {
      */
     function deregisterOperator(bytes32 operatorId, bytes memory quorumNumbers) external;
 
+    /// @notice In order to register for a quorum i, an operator must have at least `minimumStakeForQuorum[i]`, as
```
ChaoticWalrus, line 63: seems like a copy-paste that's a bit incomplete?

gpsanant:
closed in favor of #88

PR #88, opened 2023-07-06
Unit test bls op state ret

gpsanant:
- added unit test for getting operator state after registration/deregistration

TODO
- write test for getting indices
- write test for operator stakes changing


src/contracts/interfaces/IBLSPubkeyRegistry.sol
```diff
@@ -66,7 +66,7 @@ interface IBLSPubkeyRegistry is IRegistry {
     function getApkForQuorum(uint8 quorumNumber) external view returns (BN254.G1Point memory);
 
     /// @notice Returns the index of the quorumApk index at `blockNumber` for the provided `quorumNumber`
-    function getApkIndicesForQuorumsAtBlockNumber(bytes calldata quorumNumbers, uint256 blockNumber) external view returns(uint32[] memory);
+    function getApkIndicesForQuorumsAtBlockNumber(bytes calldata quorumNumbers, uint256 blockNumber) external returns(uint32[] memory);
```
ChaoticWalrus, line 69: presumably want to restore this to 'view' before merging this PR

src/contracts/interfaces/IIndexRegistry.sol
```diff
@@ -86,5 +86,5 @@ interface IIndexRegistry is IRegistry {
     function totalOperatorsForQuorum(uint8 quorumNumber) external view returns (uint32);
 
     /// @notice Returns an ordered list of operators of the services for the given `quorumNumber` at the given `blockNumber`
-    function getOperatorListForQuorumAtBlockNumber(uint8 quorumNumber, uint32 blockNumber) external view returns (bytes32[] memory);
+    function getOperatorListForQuorumAtBlockNumber(uint8 quorumNumber, uint32 blockNumber) external returns (bytes32[] memory);
```
ChaoticWalrus, line 89: likewise, this should be restored to 'view' prior to merging

src/contracts/interfaces/IRegistryCoordinator.sol
```diff
@@ -30,8 +30,6 @@ interface IRegistryCoordinator {
     struct Operator {
         // the id of the operator, which is likely the keccak256 hash of the operator's public key if using BLSRegsitry
         bytes32 operatorId;
-        // start taskNumber from which the  operator has been registered
-        uint32 fromTaskNumber; // TODO: REMOVE
```
ChaoticWalrus, line 34: did we decide this isn't useful at all?
IIRC the main use was for payments, which I guess we can rely on the IndexRegistry (or another sub-registry) for, if needed?

src/contracts/interfaces/IStakeRegistry.sol
```diff
@@ -60,6 +60,9 @@ interface IStakeRegistry is IRegistry {
      */
     function deregisterOperator(bytes32 operatorId, bytes memory quorumNumbers) external;
 
+    /// @notice In order to register for a quorum i, an operator must have at least `minimumStakeForQuorum[i]`, as
+    function minimumStakeForQuorum(uint256 quorumNumber) external view returns (uint96);
```
ChaoticWalrus, line 64: comment needs updating / completion; it just kinda trails off right now

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -8,11 +8,13 @@ import "../interfaces/IBLSPubkeyRegistry.sol";
...
 import "../interfaces/IBLSRegistryCoordinatorWithIndices.sol";
 
+import "forge-std/Test.sol";
+
 /**
  * @title BLSOperatorStateRetriever with view functions that allow to retrieve the state of an AVSs registry system.
  * @author Layr Labs Inc.
  */
-contract BLSOperatorStateRetriever {
+contract BLSOperatorStateRetriever is Test {
```
ChaoticWalrus, line 17: inheriting from test + making a bunch of functions non-view (presumably want to revert these changes before merging)

src/contracts/middleware/BLSPubkeyRegistry.sol
```diff
@@ -92,19 +92,24 @@ contract BLSPubkeyRegistry is IBLSPubkeyRegistry, Test {
...
+
             for (uint32 j = 0; j < quorumApkUpdatesLength; j++) {
                 if (quorumApkUpdates[quorumNumber][quorumApkUpdatesLength - j - 1].updateBlockNumber <= blockNumber) {
                     indices[i] = quorumApkUpdatesLength - j - 1;
                     break;
                 }
             }
         }
-        revert("BLSPubkeyRegistry.getApkIndexForQuorumAtBlockNumber: no apk update found for quorum at block number");
+        return indices;
```
ChaoticWalrus, line 112: OK so this will return the current indices for a future `blockNumber`. Seems fine, please make sure the behavior is documented for this edge-case (i.e. we should be clear this is the intended/expected behavior for this case)

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -193,7 +202,6 @@ contract BLSRegistryCoordinatorWithIndices is Initializable, IBLSRegistryCoordin
         string calldata socket,
         OperatorKickParam[] calldata operatorKickParams
     ) external {
-        require(quorumNumbers.length == operatorKickParams.length, "BLSIndexRegistryCoordinator.registerOperatorWithCoordinator: quorumNumbers and operatorKickParams must be the same length");
```
ChaoticWalrus, line 196: is there a reason we no longer need this check? feels like a good sanity-check to me.
relatedly, do we have adequate checks on the `quorumNumbers` input in general? For example, do we need to check that it contains no duplicates?

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -207,10 +215,11 @@ contract BLSRegistryCoordinatorWithIndices is Initializable, IBLSRegistryCoordin
             OperatorSetParam memory operatorSetParam = _quorumOperatorSetParams[quorumNumber];
             {
                 uint32 numOperatorsForQuorum = indexRegistry.totalOperatorsForQuorum(quorumNumber);
-                require(
-                    numOperatorsForQuorum == operatorSetParam.maxOperatorCount + 1,
-                    "BLSIndexRegistryCoordinator.registerOperatorWithCoordinator: quorum has not reached max operator count"
-                );
+                // if the number of operators for the quorum is less than or equal to the max operator count, 
+                // then the quorum has not reached the max operator count
+                if(numOperatorsForQuorum <= operatorSetParam.maxOperatorCount) {
```
ChaoticWalrus, line 220: we are enforcing this max operator count elsewhere, right?
writing this as a bit of a note -- am assuming we have a test for this, or else a plan to add one, at least.

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -279,16 +283,17 @@ contract BLSRegistryCoordinatorWithIndices is Initializable, IBLSRegistryCoordin
...
 
     function _registerOperatorWithCoordinator(address operator, bytes calldata quorumNumbers, BN254.G1Point memory pubkey, string memory socket) internal {
-        require(
-            slasher.contractCanSlashOperatorUntilBlock(operator, address(serviceManager)) == type(uint32).max,
-            "StakeRegistry._registerOperator: operator must be opted into slashing by the serviceManager"
-        );
+        // require(
+        //     slasher.contractCanSlashOperatorUntilBlock(operator, address(serviceManager)) == type(uint32).max,
+        //     "StakeRegistry._registerOperator: operator must be opted into slashing by the serviceManager"
+        // );
```
ChaoticWalrus, line 289: dangerous commenting-out, to reverse prior to merging

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -211,12 +215,9 @@ contract IndexRegistry is IIndexRegistry {
...
-        // if `blockNumber` is in future, return current number of operators
-        if (blockNumber > uint32(block.number)) {
-            return _totalOperatorsHistory[quorumNumber][totalOperatorsHistoryLength - 1].index;
-        // else if `blockNumber` is from before the `quorumNumber` existed, return `0`
-        } else if (blockNumber < _totalOperatorsHistory[quorumNumber][0].fromBlockNumber) {
+
+        // if `blockNumber` is from before the `quorumNumber` existed, return `0`
+        if (blockNumber < _totalOperatorsHistory[quorumNumber][0].fromBlockNumber) {
             return 0;
         }
```
ChaoticWalrus, line 222: am I crazy or is this function missing the "meat" that actually calculates the correct value?

ChaoticWalrus: I was crazy, deleting lmao

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -211,12 +215,9 @@ contract IndexRegistry is IIndexRegistry {
         if (totalOperatorsHistoryLength == 0) {
             return 0;
         }
-                
-        // if `blockNumber` is in future, return current number of operators
-        if (blockNumber > uint32(block.number)) {
-            return _totalOperatorsHistory[quorumNumber][totalOperatorsHistoryLength - 1].index;
-        // else if `blockNumber` is from before the `quorumNumber` existed, return `0`
-        } else if (blockNumber < _totalOperatorsHistory[quorumNumber][0].fromBlockNumber) {
```
ChaoticWalrus, line 219: does removing this 'if' block make the comment about the function's behavior inaccurate? specifically `returns the current number of total operators if the @param blockNumber is in the future.`

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -225,7 +226,7 @@ contract IndexRegistry is IIndexRegistry {
             uint256 listIndex = (totalOperatorsHistoryLength - 1) - i;
             OperatorIndexUpdate memory totalOperatorUpdate = _totalOperatorsHistory[quorumNumber][listIndex];
             // look for the first update that began at or after `blockNumber`
-            if (totalOperatorUpdate.fromBlockNumber >= blockNumber) {
+            if (totalOperatorUpdate.fromBlockNumber <= blockNumber) {
```
ChaoticWalrus, line 229: is this a reversion? seems like this change is purely making the behavior wrong?

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -235,19 +236,18 @@ contract IndexRegistry is IIndexRegistry {
 
     /// @notice Returns the index of the `operatorId` at the given `blockNumber` for the given `quorumNumber`, or max uint32 if the operator is not active in the quorum
     function _getIndexOfOperatorForQuorumAtBlockNumber(bytes32 operatorId, uint8 quorumNumber, uint32 blockNumber) internal view returns(uint32) {
-        OperatorIndexUpdate memory operatorIndexUpdate;
-        // set to max uint32 value to indicate that the operator is not part of the quorum at all, until this is updated in the loop
-        operatorIndexUpdate.index = type(uint32).max; 
+        uint256 operatorIndexHistoryLength = _operatorIdToIndexHistory[operatorId][quorumNumber].length;
         // loop forward through index history to find the index of the operator at the given block number
         // this is less efficient than looping backwards, but is simpler logic and only called in view functions that aren't mined onchain
```
ChaoticWalrus, line 241: comment is inaccurate, this is looping backwards

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -235,19 +236,18 @@ contract IndexRegistry is IIndexRegistry {
...
+            uint256 listIndex = (operatorIndexHistoryLength - 1) - i;
+            OperatorIndexUpdate memory operatorIndexUpdate = _operatorIdToIndexHistory[operatorId][quorumNumber][listIndex];
             if (operatorIndexUpdate.fromBlockNumber <= blockNumber) {
                 return operatorIndexUpdate.index;
             }
         }
 
         // the operator is still active or not in the quorum, so we return the latest index or the default max uint32
-        return operatorIndexUpdate.index;
+        return type(uint32).max;
```
ChaoticWalrus, line 251: won't we now hit this only if the operator wasn't ever active before the blockNumber? something seems wrong here

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -235,19 +236,18 @@ contract IndexRegistry is IIndexRegistry {
...
+        uint256 operatorIndexHistoryLength = _operatorIdToIndexHistory[operatorId][quorumNumber].length;
         // loop forward through index history to find the index of the operator at the given block number
         // this is less efficient than looping backwards, but is simpler logic and only called in view functions that aren't mined onchain
         for (uint i = 0; i < _operatorIdToIndexHistory[operatorId][quorumNumber].length; i++) {
-            operatorIndexUpdate = _operatorIdToIndexHistory[operatorId][quorumNumber][i];
+            uint256 listIndex = (operatorIndexHistoryLength - 1) - i;
+            OperatorIndexUpdate memory operatorIndexUpdate = _operatorIdToIndexHistory[operatorId][quorumNumber][listIndex];
             if (operatorIndexUpdate.fromBlockNumber <= blockNumber) {
                 return operatorIndexUpdate.index;
             }
```
ChaoticWalrus, line 247: shouldn't we be checking something about the `toBlockNumber` as well, if we want to have the behavior of returning `max uint32 if the operator is not active in the quorum` ?

src/test/unit/BLSOperatorStateRetrieverUnit.t.sol
```diff
@@ -0,0 +1,239 @@
...
+
+            assertEq(operatorMetadatas[i].quorumBitmap, quorumBitmap);
+            bytes memory quorumNumbers = BitmapUtils.bitmapToBytesArray(quorumBitmap);
+            
+            _assertExpectedOperators(
+                quorumNumbers,
+                operators,
+                expectedOperatorOverallIndices,
+                operatorMetadatas
+            );
```
ChaoticWalrus, line 46: trying to understand this -- for _each operator_, we are getting just that operator's bitmap, converting that bitmap to a compressed list of quorum numbers, and then feeding that list of quorum numbers into the `_assertExpectedOperators` function.
seems like we should probably be checking something smaller once per operator, or doing one big check for all operators.
I'm also a bit concerned on the structure here -- clearly different operators should have different `quorumNumbers`, so what this tells me is that if quorumNumbers is empty/lacking (for whatever reason) then the checks in `_assertExpectedOperators` should vacuously pass, which doesn't seem desirable.

src/test/unit/BLSOperatorStateRetrieverUnit.t.sol
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        uint256 operatorIndexToDeregister = pseudoRandomNumber % maxOperatorsToRegister;
+        bytes memory quorumNumbersToDeregister = BitmapUtils.bitmapToBytesArray(operatorMetadatas[operatorIndexToDeregister].quorumBitmap);
+        BN254.G1Point memory pubkeyToDeregister = BN254.hashToG1(keccak256(abi.encodePacked("pubkey", pseudoRandomNumber, operatorIndexToDeregister)));
```
ChaoticWalrus, line 51: might want to store this somewhere, or at least have it in a function or something? feels brittle/easy to break to repeatedly calculate this.

src/contracts/interfaces/IRegistryCoordinator.sol
gpsanant: yes this can likely be taken out, I'll do it in another PR

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: its actually not the case!

the operator can register for more quorums than are at the cap. in this case they only kick operators out of the quorums that have reached the cap.

as for the other checks on `quorumNumbers`, this should be handled in the internal call to `_registerOperatorWithCoordinator`

src/contracts/middleware/IndexRegistry.sol
gpsanant: it's looping backwards now, will fix the comment

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: actually planning to keep this out for M2 i think...

src/contracts/middleware/IndexRegistry.sol
gpsanant: We will only hit this, if, after looping backwards through all of the updates, none of them have a fromBlockNumber before the given blockNumber

src/contracts/middleware/IndexRegistry.sol
gpsanant: there is no `toBlockNumber` for the given struct. when the operator deregisters, we push an update with `type(uint32).max` as the index

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: great catch, I don't think we are. I'll add logic for this

src/contracts/middleware/IndexRegistry.sol
gpsanant: i think looping backwards through the stake updates should achieve this behavior. am i mistaken?

src/test/unit/BLSOperatorStateRetrieverUnit.t.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/88/commits/8888f2a31ff9e13bca8370846c35f1c48b45d48f

src/contracts/middleware/IndexRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/88/commits/8888f2a31ff9e13bca8370846c35f1c48b45d48f

src/contracts/middleware/IndexRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/88/commits/8888f2a31ff9e13bca8370846c35f1c48b45d48f

src/contracts/middleware/BLSPubkeyRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/88/commits/8888f2a31ff9e13bca8370846c35f1c48b45d48f

src/contracts/interfaces/IStakeRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/88/commits/8888f2a31ff9e13bca8370846c35f1c48b45d48f

src/contracts/interfaces/IIndexRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/88/commits/8888f2a31ff9e13bca8370846c35f1c48b45d48f

src/contracts/interfaces/IBLSPubkeyRegistry.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/88/commits/8888f2a31ff9e13bca8370846c35f1c48b45d48f

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: also just going to add: I feel like this is a major decision if we're making it, not something to be done lightly without discussion of the ramifications

src/contracts/middleware/IndexRegistry.sol
ChaoticWalrus: looks like you switched the handedness of the check in the loop? will comment there

ChaoticWalrus: > looks like you switched the handedness of the check in the loop? will comment there

actually nevermind, I see that it was incorrect before and simply fixed now. resolving this thread.

src/contracts/middleware/IndexRegistry.sol
ChaoticWalrus: I would comment on the struct itself, where it's written, and especially anywhere we use the special value.
Probably also name a constant something like `OPERATOR_DEREGISTERED_INDICATOR` and use that instead of the "magic value" of max(uint32)

src/contracts/middleware/IndexRegistry.sol
ChaoticWalrus: OK. agreed the behavior is fine now, but you should update the comment. I believe this should get hit:
1) if the `blockNumber` is from before the operator registered
2) if the operator never registered (technically overlaps with above, but worth calling out separately)
Let me know if you disagree.

I'd also clarify this in the top-level comment for the function

src/contracts/middleware/IndexRegistry.sol
ChaoticWalrus: agreed this is correct now, was confused myself, thanks.

src/test/unit/BLSOperatorStateRetrieverUnit.t.sol
ChaoticWalrus: > as for speed, this file takes 6.37s for testing with 10 registering operators. tolerable imo

ah OK great, that's wayyy better than I feared. I agree that's acceptable.

ChaoticWalrus: > in the assertion, we loop through each of the quorum numbers and check that the returned operator lists and states were as expected. we do this for every operator the check that this function works over time as more operators register. i could also just run it 2-3 times at different points in the list.
> 
> as for different operators having different quorumNumbers, that's correct, but we're specifically checking the functionality from the POV of the registering operator, who will only request for their quorums.

Ahhh, OK, gotcha, so the check is basically:
I register a new operator for `quorumNumbers`, then I want to check that all of `operators` have the correct `expectedOperatorOverallIndices` and `operatorMetadatas` (specifically, the `operatorMetadatas.stakes` are correct)

Makes sense, I think 👍 
We could in theory check that things remain the same for other quorum numbers (that the operator isn't registering for), but that would just mean way more "hollow" loop iterations, so I'm OK not doing that for now.
I'd appreciate if you can maybe help me break this down into properties that we can try to prove?

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: OK, that makes sense but IMO the documentation here is lacking, particularly w/r/t requirements/explanation of `operatorKickParams`.

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -278,17 +281,19 @@ contract BLSRegistryCoordinatorWithIndices is Initializable, IBLSRegistryCoordin
...
         
         // check that the sender is not already registered
         require(_operators[operator].status != OperatorStatus.REGISTERED, "BLSIndexRegistryCoordinator._registerOperatorWithCoordinator: operator already registered");
 
         // get the quorum bitmap from the quorum numbers
-        uint256 quorumBitmap = BitmapUtils.orderedBytesArrayToBitmap_Yul(quorumNumbers);
+        uint256 quorumBitmap = BitmapUtils.orderedBytesArrayToBitmap(quorumNumbers);
+
         require(quorumBitmap != 0, "BLSIndexRegistryCoordinator._registerOperatorWithCoordinator: quorumBitmap cannot be 0");
         require(quorumBitmap <= type(uint192).max, "BLSIndexRegistryCoordinator._registerOperatorWithCoordinator: quorumBitmap cant have more than 192 set bits");
```
ChaoticWalrus, line 298: are these checks sufficient or should we also be checking something against the existing number of quorums? is that checked instead by the `stakeRegistry`? Just want to be confident we are adequately sanitizing the `quorumNumbers` input!

ChaoticWalrus: I'll add that we can + should probably try to prove that you can't register for a non-existent quorum

src/test/unit/BLSOperatorStateRetrieverUnit.t.sol
ChaoticWalrus: I'd suggest adding something like my explanation to the documentation for the check here. IMO this PR is really great work, it's decently "self documenting", but it's hard to parse things when they're so complex -- a few brief, written explanations in normal language could help a lot!

src/contracts/interfaces/IRegistryCoordinator.sol
ChaoticWalrus: > yes this can likely be taken out, I'll do it in another PR

We should realistically have a design meeting to focus on some issues like this and try to sort out clearer requirements. I like what you've done but want to try to get ahead of future design problems that will cause us to do a lot of revisions later.
I'm OK with this for now, but let's work on putting something together for a discussion?

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -235,19 +240,19 @@ contract IndexRegistry is IIndexRegistry {
 
     /// @notice Returns the index of the `operatorId` at the given `blockNumber` for the given `quorumNumber`, or max uint32 if the operator is not active in the quorum
```
ChaoticWalrus, line 241: `or max uint32 if the operator is not active in the quorum` => `or OPERATOR_DEREGISTERED_INDEX if the operator was not registered for the `quorumNumber` at `blockNumber`.

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
```diff
@@ -0,0 +1,649 @@
...
+    event PubkeyAddedToQuorums(
+        address operator,
+        bytes quorumNumbers
+    );
+
+    // Emitted when an operator pubkey is removed from a set of quorums
+    event PubkeyRemovedFromQuorums(
+        address operator, 
+        bytes quorumNumbers
+    );
```
ChaoticWalrus, line 30: should these be named 'PubkeyXYZ' if they emit the `operator` specifically, rather than the pubkey?
Also seems like the `operator` should be indexed

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
```diff
@@ -0,0 +1,649 @@
...
+
+import "../utils/MockAVSDeployer.sol";
+
+contract BLSRegistryCoordinatorWithIndicesUnit is MockAVSDeployer {
+    using BN254 for BN254.G1Point;
+
+    uint8 maxQuorumsToRegisterFor = 4;
+
+    event OperatorSocketUpdate(bytes32 operatorId, string socket);
+
```
ChaoticWalrus, line 12: index the `operatorId`

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
```diff
@@ -0,0 +1,649 @@
...
+        quorumNumbers[0] = bytes1(defaultQuorumNumber);
+        uint256 quorumBitmap = BitmapUtils.orderedBytesArrayToBitmap(quorumNumbers);
+
+        _registerOperatorWithCoordinator(defaultOperator, quorumBitmap, defaultPubKey);
+
+        quorumNumbers = new bytes(2);
+        quorumNumbers[0] = bytes1(defaultQuorumNumber);
+        quorumNumbers[1] = bytes1(uint8(192));
+
+        cheats.expectRevert("BLSIndexRegistryCoordinator._deregisterOperatorWithCoordinator: quorumsToRemoveBitmap cant have more than 192 set bits");
```
ChaoticWalrus, line 227: this revert message seems a tad...cryptic? maybe it should be something like 'quorum bitmap out of max bitmap size / range'? -- also maybe we should have a tiny internal function just for this, that we reuse the few places we do the check. any thoughts?

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
```diff
@@ -0,0 +1,649 @@
...
+
+        assertEq(
+            keccak256(abi.encode(registryCoordinator.getOperator(defaultOperator))), 
+            keccak256(abi.encode(IRegistryCoordinator.Operator({
+                operatorId: defaultOperatorId,
+                status: IRegistryCoordinator.OperatorStatus.DEREGISTERED
+            })))
+        );
+        cheats.expectRevert("BLSRegistryCoordinator.getCurrentQuorumBitmapByOperatorId: operator is not registered");
+        registryCoordinator.getCurrentQuorumBitmapByOperatorId(defaultOperatorId);
```
ChaoticWalrus, line 288: so desired behavior here is reverting? seems like returning an empty bitmap might be more sane / to-be-expected

src/test/unit/StakeRegistryUnit.t.sol
```diff
@@ -190,7 +191,226 @@ contract StakeRegistryUnitTests is Test {
...
+    function testRegisterFirstOperator_Valid(
+        uint256 quorumBitmap,
+        uint80[] memory stakesForQuorum
+    ) public {
+
+        uint96[] memory paddedStakesForQuorum = _registerOperatorValid(defaultOperator, defaultOperatorId, quorumBitmap, stakesForQuorum);
+
+        uint8 quorumNumberIndex = 0;
+        for (uint8 i = 0; i < 192; i++) {
+            if (quorumBitmap >> i & 1 == 1) {
```
ChaoticWalrus, line 203: should we make a utility getter function for this *on the stake registry* that is just something like
```
function isOperatorRegisteredForQuorum(address operator, uint8 quorumNumber) external view returns (uint8) {
...
}
```
?
I guess we could have a version that takes in a blockNumber or index too?
This could also be part of the BLSOperatorStateRetriever, I suppose; I'm just seeing you repeating the code block and realizing it could be useful!

ChaoticWalrus (changes requested):
This is looking really good.
I believe I've gone through everything & I've added a bunch of comments, but it is primarily minor stuff.
@gpsanant Do we have tracking of these unit test cases somewhere? I find it very hard to visualize or understand coverage when just reviewing code like this.

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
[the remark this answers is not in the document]
gpsanant: 7815293c

gpsanant: 7815293c, this is done by the stake registry when it gets there

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
gpsanant: 40533cf7

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
gpsanant: 40533cf7 chose to not do the internal function, what do u anticipate being recorded?

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
gpsanant: 40533cf7

src/contracts/middleware/IndexRegistry.sol
gpsanant: 40533cf7

src/contracts/middleware/IndexRegistry.sol
```diff
@@ -235,19 +240,19 @@ contract IndexRegistry is IIndexRegistry {
...
+            OperatorIndexUpdate memory operatorIndexUpdate = _operatorIdToIndexHistory[operatorId][quorumNumber][listIndex];
             if (operatorIndexUpdate.fromBlockNumber <= blockNumber) {
+                // one special case is that this will be OPERATOR_DEREGISTERED_INDEX if the operator was deregistered from the quorum
                 return operatorIndexUpdate.index;
             }
         }
 
         // the operator is still active or not in the quorum, so we return the latest index or the default max uint32
-        return operatorIndexUpdate.index;
+        // this will be hit if `blockNumber` is before when the operator registered or has not registered for the given quorum
```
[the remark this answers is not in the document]
gpsanant: 40533cf7

src/test/unit/StakeRegistryUnit.t.sol
gpsanant: i'm avoiding messing with the stake registry unit tests for now, planning for them to get done in another PR

src/test/unit/BLSOperatorStateRetrieverUnit.t.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/88/commits/7fae560531f2ad0ae1ff79289dd1c625b3368af5

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: Discussed offline, leaving out for now, but plan to have some form of opt-in to Slasher so we can support a smoother "migration" experience if we need to.

src/test/unit/StakeRegistryUnit.t.sol
ChaoticWalrus: OK, please add this as part of that PR then, as I think it would be a good addition

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/88/commits/c7204044c3cec47c02f205d069c2f29246496ab4 70e0c7da

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
gpsanant: any suggestions?

ChaoticWalrus: Literally just `OperatorAddedToQuorums` and `OperatorRemovedFromQuorums` instead?
I'm also unsure on our off-chain usage, maybe actually having the pubkey in these would be better?

gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/88/commits/f2f25d245b79e36f7fbb7551fd1da18f4f30f0ba

ChaoticWalrus (approved):

ChaoticWalrus:
LGTM now. I believe we've resolved all my questions; there's a couple cans that we're kicking down the road a little but we've at least aligned on these.

PR #92, opened 2023-07-13
Update Slasher.sol

samlaf:
I might be missing something here, but the only function I see relevant to slashing is slashOperator.
I believe on operators can be keys of the `frozenStatus` map right?


ChaoticWalrus:
Good catch, this appears to be a straggler comment from a while back. Thanks for clarifying the wording.

PR #94, opened 2023-07-14
Fix erc20mock not minting in transfer

samlaf:
moving _mint function to _transfer since it was currently only being applied in transferFrom and not transfer

I'm assuming this was a bug, but let me know if this was intended behavior.


samlaf:
meant to merge to multiquorums branch. closing.

PR #97, opened 2023-07-14
Delegation redesign

ChaoticWalrus:
I'm initially marking this as a draft because it is not yet ready for review. At minimum it still needs cleanup and more tests.

This PR implements significant changes to the delegation mechanisms in EigenLayer, as outlined in our design process meetings.

Notable major features:

1. Staker opt-out window as part of the operator-AVS registration process
2. Functionality for operators to 'force undelegate' stakers who delegate to them
3. Supporting a mode where an operator signature is required for a staker to delegate to the operator

Importantly, at the moment this PR does not finishing implementing item (1). We will have to decide if this PR should contain this entire feature addition or if the remainder of the implementation should be left for a later PR focused on the Slasher.


ChaoticWalrus:
Open question: should we renamed `StrategyManager.forceUndelegation` to something like `StrategyManager.forceTotalWithdrawal`? I am a bit torn -- I like that the first (existing) name clarifies when the function is used / the intention, but I think the latter name draws the parallel to a withdrawal better. I also dislike both the StrategyManager and the DelegationManager having identically-named functions. We could maybe name it something like `StrategyManager.processForcedUndelegation`?

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -13,37 +12,76 @@ import "../interfaces/ISlasher.sol";
...
-    bytes32 public constant DELEGATION_TYPEHASH =
-        keccak256("Delegation(address delegator,address operator,uint256 nonce,uint256 expiry)");
+    /// @notice The EIP-712 typehash for the StakerDelegation struct used by the contract
+    bytes32 public constant STAKER_DELEGATION_TYPEHASH =
+        keccak256("StakerDelegation(address staker,address operator,uint256 nonce,uint256 expiry)");
 
-    /// @notice EIP-712 Domain separator
-    bytes32 public DOMAIN_SEPARATOR;
+    /// @notice The EIP-712 typehash for the DelegationApproval struct used by the contract
+    bytes32 public constant DELEGATION_APPROVAL_TYPEHASH =
```
gpsanant, line 42: *DelegationApproval

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -13,37 +12,76 @@ import "../interfaces/ISlasher.sol";
...
+    /**
+     * @notice Mapping: staker => operator whom the staker is currently delegated to.
+     * @dev Note that returning address(0) indicates that the staker is not actively delegated to any operator.
+     */
     mapping(address => address) public delegatedTo;
 
-    /// @notice Mapping: delegator => number of signed delegation nonce (used in delegateToBySignature)
-    mapping(address => uint256) public nonces;
+    /// @notice Mapping: staker => number of signed messages (used in `delegateToBySignature`) from the staker that this contract has already checked.
+    mapping(address => uint256) public stakerNonce;
```
gpsanant, line 77: seems like it might be worth it to just set up a 
`mapping(address  => mapping(bytes32 => uint256)) nonces`
where `nonces[signer][typehash]` is their nonce for that struct they're signing?

src/contracts/core/DelegationManager.sol
```diff
@@ -1,35 +1,35 @@
...
 
     // chain id at the time of contract deployment
-    uint256 immutable ORIGINAL_CHAIN_ID;
+    uint256 internal immutable ORIGINAL_CHAIN_ID;
 
+    /**
+     * @notice Maximum value that `_operatorDetails[operator].stakerOptOutWindowBlocks` is allowed to take, for any operator.
+     * @dev This is 6 months (technically 180 days) in blocks.
+     */
+    uint256 public constant MAX_STAKER_OPT_OUT_WINDOW_BLOCKS = (180 * 24 * 60 * 60) / 12;
```
gpsanant, line 32: `180 days / 12`

src/contracts/core/DelegationManager.sol
```diff
@@ -193,149 +208,99 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        if (block.chainid == ORIGINAL_CHAIN_ID) {
+            return _DOMAIN_SEPARATOR;
+        }
+        else {
+            return _calculateDomainSeparator();
+        }
+    }
 
     /// @notice Returns 'true' if `staker` *is* actively delegated, and 'false' otherwise.
     function isDelegated(address staker) public view returns (bool) {
```
gpsanant, line 306: isDelegated and isNotDelegated, should have only 1 internal function?

src/contracts/core/DelegationManagerStorage.sol
ChaoticWalrus: is this meant to be a comment about a typo? If so, I'm not seeing the typo

src/contracts/core/DelegationManager.sol
ChaoticWalrus: Good point, I think we should just eliminate `isNotDelegated` entirely. IIRC there was initially some kind of weird, limbo-ish 3rd state but we moved away from that model long ago.

src/contracts/core/DelegationManagerStorage.sol
ChaoticWalrus: Interesting idea. Do you think this would be clearer for users/devs?
I feel like the existing design is "less clean" but the separation into different mappings/functions feels a little "safer"/nicer with respect to the dev UX? (slightly less footgunish, I think, with clearer differentation?)

src/contracts/core/DelegationManager.sol
ChaoticWalrus: https://github.com/Layr-Labs/eigenlayer-contracts/pull/97/commits/ccdfd071315668a7d1cb93706589bac167f3d8e4

src/contracts/core/DelegationManager.sol
ChaoticWalrus: https://github.com/Layr-Labs/eigenlayer-contracts/pull/97/commits/ccdfd071315668a7d1cb93706589bac167f3d8e4

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -13,37 +12,76 @@ import "../interfaces/ISlasher.sol";
  * @notice This storage contract is separate from the logic to simplify the upgrade process.
  */
 abstract contract DelegationManagerStorage is IDelegationManager {
-    /// @notice Gas budget provided in calls to DelegationTerms contracts
-    uint256 internal constant LOW_LEVEL_GAS_BUDGET = 1e5;
+    // TODO: documentation
+    struct StakerDelegation {
```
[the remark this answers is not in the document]
ChaoticWalrus: https://github.com/Layr-Labs/eigenlayer-contracts/pull/97/commits/ccdfd071315668a7d1cb93706589bac167f3d8e4 still needs documentation tho lol

gpsanant:
> Open question: should we renamed `StrategyManager.forceUndelegation`...

I thought about commenting this as well. If you feel this is the case as well, then I like `forceTotalWithdrawal`

src/contracts/core/DelegationManagerStorage.sol
gpsanant: the struct inside of the quotes, sorry wrong line

src/contracts/core/DelegationManagerStorage.sol
gpsanant: ah, i see. probably fine then. Another option would be do my mapping suggestion and have view functions for nonces. You can resolve.

ChaoticWalrus:
> > Open question: should we renamed `StrategyManager.forceUndelegation`...
> 
> I thought about commenting this as well. If you feel this is the case as well, then I like `forceTotalWithdrawal`

OK ty, will take your suggestion and push a commit changing the name.

src/contracts/core/DelegationManagerStorage.sol
ChaoticWalrus: OK yeah, I think it's effectively the same, going to opt to leave as-is but I still appreciate the suggestion + discussion

src/contracts/core/DelegationManagerStorage.sol
ChaoticWalrus: ohhhh good catch, just fixed this in https://github.com/Layr-Labs/eigenlayer-contracts/pull/97/commits/599b96de577503e67cc45dc0e29aa1b5ae218c0a
ty!

src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -1,81 +1,249 @@
...
- * - allowing new operators to provide a DelegationTerms-type contract, which may mediate their interactions with stakers who delegate to them
- * - enabling any staker to delegate its stake to the operator of its choice
- * - enabling a staker to undelegate its assets from an operator (performed as part of the withdrawal process, initiated through the StrategyManager)
+ * - allowing operators to specify parameters related to stakers who delegate to them
+ * - enabling any staker to delegate its stake to the operator of its choice (a given staker can only delegate to a single operator at a time)
+ * - enabling a staker to undelegate its assets from the operator it is delegated to (performed as part of the withdrawal process, initiated through the StrategyManager)
  */
 interface IDelegationManager {
+    // @notice Struct used for storing information about a single operator who has registered with EigenLayer
+    struct OperatorDetails {
```
gpsanant, line 18: can we add a metadata url to this. this will be where the operator's static metadata is hosted. we don't need to store this, we only need to index it. wdyt?

ChaoticWalrus: If we aren't storing it then doesn't it belong in an event rather than this struct used specifically for storage?

Thinking something like `OperatorMetadataURIUpdated(address indexed operator, bytes metadataURI)`

If we want it in storage then it could make sense here, but I do think it depends on how often we'd expect it to be updated (like maybe operators want more frequent updates to their metadata URI than these other addresses? right now there is a single setter function)

ChaoticWalrus:
I believe this is ready for review now.

src/contracts/interfaces/IDelegationManager.sol
gpsanant: metadataURI being a string is nicer, but other than that it works.

we discussed that adding a `string metadataURI` to the register operator and a `modifyMetadataURI` function

ChaoticWalrus:
Ah, going to call out that I still need to update documentation 😞

src/contracts/interfaces/IDelegationManager.sol
ChaoticWalrus: ah yeah, made it a string. just pushed relevant commits; please take a look and resolve this if you think it all will work.

gpsanant: works

src/contracts/core/DelegationManager.sol
```diff
@@ -45,107 +45,155 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
         require(
-            address(delegationTerms[msg.sender]) == address(0),
+            _operatorDetails[msg.sender].earningsReceiver == address(0),
             "DelegationManager.registerAsOperator: operator has already registered"
         );
-        // store the address of the delegation contract that the operator is providing.
-        delegationTerms[msg.sender] = dt;
-        _delegate(msg.sender, msg.sender);
-        emit RegisterAsOperator(msg.sender, dt);
+        _setOperatorDetails(msg.sender, registeringOperatorDetails);
```
gpsanant, line 71: should this be called `_modifyOperatorDetails` or the external function be called `_setOperatorDetails` for consistency?

src/contracts/core/DelegationManager.sol
```diff
@@ -45,107 +45,155 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+     * @dev Reverts if the `staker` is also an operator, since operators are not allowed to undelegate from themselves.
+     * @dev Does nothing (but should not revert) if the staker is already undelegated.
      */
     function undelegate(address staker) external onlyStrategyManager {
         require(!isOperator(staker), "DelegationManager.undelegate: operators cannot undelegate from themselves");
-        delegatedTo[staker] = address(0);
+        address _delegatedTo = delegatedTo[staker];
+        // only make storage changes + emit an event if the staker is actively delegated, otherwise do nothing
+        if (_delegatedTo != address(0)) {
+            emit StakerUndelegated(staker, delegatedTo[staker]);
```
gpsanant, line 168: _delegatedTo

gpsanant: more clearly, `operator`

src/contracts/core/DelegationManager.sol
```diff
@@ -193,162 +228,167 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        return _operatorDetails[operator].stakerOptOutWindowBlocks;
+    }
+
+    /**
+     * @notice External getter function that mirrors the staker signature hash calculation in the `delegateToBySignature` function
+     * @param staker The signing staker
+     * @param operator The operator who is being delegated
+     * @param expiry The desired expiry time of the staker's signature
+     */
+    function calculateStakerDigestHash(address staker, address operator, uint256 expiry) external view returns (bytes32) {
```
gpsanant, line 364: why not make public and reuse?

src/test/Slasher.t.sol
```diff
@@ -255,14 +287,26 @@ contract SlasherTests is EigenLayerTestHelper {
...
+        string memory emptyStringForMetadataURI;
+        delegation.registerAsOperator(operatorDetails, emptyStringForMetadataURI);
         slasher.optIntoSlashing(_slasher);
         cheats.stopPrank();
     }
 
     function testFreezeOperator() public {
         cheats.prank(operator);
-        delegation.registerAsOperator(delegationTerms);
+        IDelegationManager.OperatorDetails memory operatorDetails = IDelegationManager.OperatorDetails({
```
gpsanant, line 303: make this an internal function?

src/test/Delegation.t.sol
```diff
@@ -520,11 +583,16 @@ contract DelegationTests is EigenLayerTestHelper {
...
+                earningsReceiver: operator,
+                delegationApprover: address(0),
+                stakerOptOutWindowBlocks: 0
+            });
+            _testRegisterAsOperator(operator, operatorDetails);
         }
 
         //making additional deposits to the strategies
-        assertTrue(delegation.isNotDelegated(staker) == true, "testDelegation: staker is not delegate");
+        assertTrue(!delegation.isDelegated(staker) == true, "testDelegation: staker is not delegate");
```
gpsanant, line 595: pls add tests for 
- force undelegation
- calling functions that are permissioned to only the strategyManager
- changing opt out blocks
- chainging metadata uri

src/contracts/core/DelegationManager.sol
ChaoticWalrus: it's used both in `registerAsOperator` and `modifyOperatorDetails`, so I think giving the internal function a different name is more appropriate (since it _isn't_ just an alias for the single external function)

src/contracts/core/DelegationManager.sol
ChaoticWalrus: good call. changed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/97/commits/3af75132771abfead6143453ec75efc4e52a23cd

src/contracts/core/DelegationManager.sol
ChaoticWalrus: we discussed what I think is a good path forward on this offline. I just pushed a commit which you should review:
https://github.com/Layr-Labs/eigenlayer-contracts/pull/97/commits/787596a7f88dd4dbd057b588fefb132162467f2b

src/test/Slasher.t.sol
ChaoticWalrus: these are inside of the "Slasher" tests. I didn't want to mess with them since some blocks have additional logic (e.g. 'opting into slashing') but agree it's a good candidate for de-duplicating

src/test/Delegation.t.sol
ChaoticWalrus: >     * calling functions that are permissioned to only the strategyManager

This is covered by `testCannotCallUndelegateFromNonStrategyManagerAddress`, `testCannotCallIncreaseDelegatedSharesFromNonStrategyManagerAddress`, and `testCannotCallDecreaseDelegatedSharesFromNonStrategyManagerAddress`

>     * changing opt out blocks

This is covered by `testModifyOperatorParameters`

>     * chainging metadata uri

This is covered by `testUpdateOperatorMetadataURI`

>     * force undelegation

This appears to be missing coverage; I'll add a test!

ChaoticWalrus: > This appears to be missing coverage; I'll add a test!

I added the missing coverage (and refined the contract spec + behavior) in https://github.com/Layr-Labs/eigenlayer-contracts/pull/97/commits/e27672c12afcb8cbb3dd228a6bd6e0c9bf0aaddc
@gpsanant  plz review this commit!

gpsanant (approved):
nice work on this. lgtm

PR #99, opened 2023-07-15
fix double mint

samlaf:


ChaoticWalrus:
bug in a cursed contract used in tests; this is an insta-merge for me

PR #101, opened 2023-07-17
getMiddlewareTimesIndexBlock -> getMiddlewareTimesIndexStalestUpdateBlock

samlaf:
This name looked like a typo (compare to getMiddlewareTimesIndexServeUntilBlock just below), so I fixed it to follow same name convention


ChaoticWalrus:
Makes sense to me. @gpsanant does this break any dependencies or are we good-to-go with it?

gpsanant (approved):
this looks fine

PR #102, opened 2023-07-19
Test sig checker

gpsanant:
Added unit tests for the signature checker


src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -96,36 +96,44 @@ contract BLSOperatorStateRetriever {
         uint32 referenceBlockNumber, 
         bytes calldata quorumNumbers, 
         bytes32[] calldata nonSignerOperatorIds
-    ) external returns (CheckSignaturesIndices memory) {
+    ) external view returns (CheckSignaturesIndices memory) {
         IStakeRegistry stakeRegistry = registryCoordinator.stakeRegistry();
         CheckSignaturesIndices memory checkSignaturesIndices;
 
         checkSignaturesIndices.nonSignerQuorumBitmapIndices = registryCoordinator.getQuorumBitmapIndicesByOperatorIdsAtBlockNumber(referenceBlockNumber, nonSignerOperatorIds);
```
ChaoticWalrus, line 103: IMO this could use some comments given the complexity of the operations. IIRC this is getting the overlap of the set (`quorumNumbers`, the union of all quorums that the operators corresponding to `nonSignerOperatorIds` belonged to at the `referenceBlockNumber`).
If this is correct can we add a description to this effect? If it's not then please clarify for me!

ChaoticWalrus: Oh wait there's no overlap happening here, it's just getting the second of these set members.
should we be getting this overlap / do we perform this operation later?

ChaoticWalrus: am thinking just `checkSignaturesIndices.nonSignerQuorumBitmapIndices = quorumNumbers | checkSignaturesIndices.nonSignerQuorumBitmapIndices`

ChaoticWalrus: nevermind, `checkSignaturesIndices.nonSignerQuorumBitmapIndices` is itself an array, where each member is the index of the bitmap of the i-th operator in `nonSignerOperatorIds`, at `referenceBlockNumber`.
OK just going to say again this is complex enough to need documentation throughout

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -96,36 +96,44 @@ contract BLSOperatorStateRetriever {
...
+                    registryCoordinator.getQuorumBitmapByOperatorIdAtBlockNumberByIndex(
+                        nonSignerOperatorIds[i], 
+                        referenceBlockNumber, 
+                        checkSignaturesIndices.nonSignerQuorumBitmapIndices[i]
+                    );
+                
                 // if the operator was a part of the quorum and the quorum is a part of the provided quorumNumbers
-                if (nonSignerQuorumBitmap >> j & (BitmapUtils.bytesArrayToBitmap(quorumNumbers) >> j & 1) == 1) {
-                    checkSignaturesIndices.nonSignerStakeIndices[i][stakeIndexIndex] = stakeRegistry.getStakeUpdateIndexForOperatorIdForQuorumAtBlockNumber(
+                if (nonSignerQuorumBitmap >> uint8(quorumNumbers[quorumNumberIndex]) & 1 == 1) {
```
ChaoticWalrus, line 121: going to suggest parentheses for clarity here. I realize they're unnecessary but IMO it's easier to verify it's right if we write operators like this (multiple bitwise ops chained together) as:
` if ((nonSignerQuorumBitmap >> uint8(quorumNumbers[quorumNumberIndex])) & 1 == 1) {`

src/contracts/middleware/BLSOperatorStateRetriever.sol
```diff
@@ -96,36 +96,44 @@ contract BLSOperatorStateRetriever {
...
+                    checkSignaturesIndices.nonSignerStakeIndices[quorumNumberIndex][numNonSignersForQuorum] = stakeRegistry.getStakeUpdateIndexForOperatorIdForQuorumAtBlockNumber(
```
ChaoticWalrus, line 122: are operator arrays trimmed to the right length using this, or do arrays end up all the same length with zeros filling them? I'm feeling like we end up with zeroes here which would be problematic in the event that we can't differentiate from when the correct stake index is zero?
maybe I'm just missing the high- to mid-level documentation on how all this signature checking logic is supposed to work / exactly how the inputs are supposed to be structured

ChaoticWalrus: ahh I guess it all gets trimmed below. I think that works out then? again docs would help convince me that the design is right so I can then just check the design rather than trying to sort out the design from the code

src/contracts/middleware/BLSSignatureChecker.sol
```diff
@@ -115,26 +116,26 @@ abstract contract BLSSignatureChecker {
 
                 for (uint i = 0; i < nonSignerStakesAndSignature.nonSignerPubkeys.length; i++) {
                     nonSignerPubkeyHashes[i] = nonSignerStakesAndSignature.nonSignerPubkeys[i].hashG1Point();
-                    if (i != 0) {
-                        require(uint256(nonSignerPubkeyHashes[i]) > uint256(nonSignerPubkeyHashes[i - 1]), "BLSSignatureChecker.checkSignatures: nonSignerPubkeys not sorted");
-                    }
+                    // if (i != 0) {
+                    //     require(uint256(nonSignerPubkeyHashes[i]) > uint256(nonSignerPubkeyHashes[i - 1]), "BLSSignatureChecker.checkSignatures: nonSignerPubkeys not sorted");
+                    // }
```
ChaoticWalrus, line 121: is this just a temporary change to deal with ordering being hard?
IIRC this check was key for ensuring no duplicates (we should add a comment to this effect as well)

ChaoticWalrus:
Medium-level design question: what happens when none of the non-signers is part of one of the quorums? is the answer that we just don't optimize for this and can still handle it? because that is my guess based on review so far

src/test/harnesses/BLSRegistryCoordinatorWithIndicesHarness.sol
```diff
@@ -0,0 +1,31 @@
...
+        if (operatorQuorumBitmapHistoryLength != 0) {
+            _operatorIdToQuorumBitmapHistory[operatorId][operatorQuorumBitmapHistoryLength - 1].nextUpdateBlockNumber = uint32(block.number);
+        }
+
+        _operatorIdToQuorumBitmapHistory[operatorId].push(QuorumBitmapUpdate({
+            updateBlockNumber: uint32(block.number),
+            nextUpdateBlockNumber: 0,
+            quorumBitmap: quorumBitmap
+        }));
+    }
```
ChaoticWalrus, line 28: should there be a corresponding internal function in the `BLSRegistryCoordinatorWithIndices`? If there is already, why not use it? Am a bit concerned about any divergence between harnessed + non-harnessed versions

src/test/unit/BLSOperatorStateRetrieverUnit.t.sol
```diff
@@ -114,7 +101,7 @@ contract BLSOperatorStateRetrieverUnitTests is MockAVSDeployer {
         assertEq(checkSignaturesIndices.nonSignerQuorumBitmapIndices.length, 0);
         assertEq(checkSignaturesIndices.quorumApkIndices.length, allInclusiveQuorumNumbers.length);
         assertEq(checkSignaturesIndices.totalStakeIndices.length, allInclusiveQuorumNumbers.length);
-        assertEq(checkSignaturesIndices.nonSignerStakeIndices.length, 0);
+        assertEq(checkSignaturesIndices.nonSignerStakeIndices.length, allInclusiveQuorumNumbers.length);
```
ChaoticWalrus, line 104: could you add some revert messages to these? also helps document exactly what it's checking (although these seem ~fairly clear)

src/test/unit/BLSSignatureCheckerUnit.t.sol
```diff
@@ -0,0 +1,229 @@
...
+            quorumNumbers,
+            referenceBlockNumber, 
+            nonSignerStakesAndSignature
+        );
+        uint256 gasAfter = gasleft();
+        emit log_named_uint("gasUsed", gasBefore - gasAfter);
+
+        // 0 nonSigners: 159908
+        // 1 nonSigner: 178683
+        // 2 nonSigners: 197410
```
ChaoticWalrus, line 39: can we check anything here other than the call not reverting? it's a 'view' function so no state transition to check, but maybe we can check something about the returned values (looks like we're just tossing those out right now)

src/test/unit/BLSSignatureCheckerUnit.t.sol
```diff
@@ -0,0 +1,229 @@
...
+
+    BLSSignatureChecker blsSignatureChecker;
+
+    function setUp() virtual public {
+        _setUpBLSMockAVSDeployer();
+
+        blsSignatureChecker = new BLSSignatureChecker(registryCoordinator);
+    }
+
+    function testBLSSignatureChecker_SingleQuorum_Valid(uint256 pseudoRandomNumber) public {
```
ChaoticWalrus, line 18: it would be *awesome* if each test could have a few points like:
one sentence what this is supposed to do
things it should check
any assumed preconditions, or at least some of the important ones

some of these could also go inside the test; I just find it makes it much easier to understand

src/test/unit/BLSSignatureCheckerUnit.t.sol
```diff
@@ -0,0 +1,229 @@
...
+        nonSignerStakesAndSignature.sigma.X++;
+
+        cheats.expectRevert();
+        blsSignatureChecker.checkSignatures(
+            msgHash, 
+            quorumNumbers,
+            referenceBlockNumber, 
+            nonSignerStakesAndSignature
+        );
+    }
```
ChaoticWalrus, line 228: should this have a specific revert message or does this end up in some low-level call failure (like a failure in the precompile call?) ?

ChaoticWalrus (changes requested):
This is looking really good!
Definitely want more documentation on this, but I don't think I'm seeing any real problems; I'd like to take another look after documentation is added but am primarily expecting to understand better (but it should also help me identify bugs or holes in testing easier)

src/contracts/middleware/BLSSignatureChecker.sol
gpsanant: I actually think this is fine. if you subtract the same public key twice, then the amount of stake that has confirmed a fact will only be underestimated

src/test/unit/BLSSignatureCheckerUnit.t.sol
gpsanant: precompile call fails rip

src/test/unit/BLSSignatureCheckerUnit.t.sol
gpsanant: added a couple for the validity non trivial tests

src/test/unit/BLSSignatureCheckerUnit.t.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/102/commits/db7e47e6a6b44022e7f3c6095441a357bf1588f2

src/contracts/middleware/BLSSignatureChecker.sol
ChaoticWalrus: > I actually think this is fine. if you subtract the same public key twice, then the amount of stake that has confirmed a fact will only be underestimated

hmm OK I guess. I suppose you wouldn't have a valid signature for the resultant public key anyway, right (in the case where you subtract a signature twice)? Just want to be very sure before deleting anything that was intended as some kind of safety/correctness check.

src/test/unit/BLSSignatureCheckerUnit.t.sol
ChaoticWalrus: > [db7e47e](https://github.com/Layr-Labs/eigenlayer-contracts/commit/db7e47e6a6b44022e7f3c6095441a357bf1588f2)

looks like the commit you linked just puts the returned values into memory and then immediately dumps them? this is still short of checking anything about the returned values

src/test/unit/BLSSignatureCheckerUnit.t.sol
ChaoticWalrus: OK plz add a comment like
```
        // expect a non-specific low-level revert, since this call will ultimately fail as part of the precompile call
        cheats.expectRevert();
        blsSignatureChecker.checkSignatures(
            msgHash, 
            quorumNumbers,
            referenceBlockNumber, 
            nonSignerStakesAndSignature
        );
```

ChaoticWalrus (approved):

ChaoticWalrus:
Am approving the PR for now, assuming no further logical changes. Please add some more clarifying comments / documentation prior to merging.

src/contracts/middleware/BLSSignatureChecker.sol
gpsanant: correct, agreed.

PR #103, opened 2023-07-19
Gitignore docgen folder

ChaoticWalrus:
These auto-generated docs can be generated locally whenever someone wants them.
As-is, we are not adequately updating them as part of the development process, and when we do it just clogs up PRs and makes them harder to read/understand.
This seems like an improvement, at least.


ChaoticWalrus:
got offline approval on this, going to go ahead and merge

PR #109, opened 2023-07-22
add 'forge doc' and remove natspec

bowenli86:


ChaoticWalrus:
@bowenli86 Do you actually prefer this output to the existing one?

I have tried `forge doc` myself and found it to output a lot of extraneous files in a more inconvenient folder layout, as compared to docgen.

I don't personally use this auto-generated documentation at all, so if it is actually your preference, I'm OK with this change -- I hadn't merged a similar PR because as mentioned above, the output seemed worse.

If you do indeed prefer this format, IMO we should add the folder of generated docs to the `.gitignore` as part of this PR, because I don't expect us to be good about updating them + generating them locally is very simple.

ChaoticWalrus:
Closing out this stale PR; please reopen if you actually prefer the output here.

PR #111, opened 2023-07-22
add go bindings for a few contract and generation scripts

bowenli86:
add go bindings for a few contract and generation scripts, for purpose of supporting offchain calls


ChaoticWalrus (approved):
PR seems fine to me.
Will we need  to update these bindings when contract interfaces change?
Seems like the `compile.sh` file would accomplish this -- we could potentially add something to the CI to update these bindings automatically?

gpsanant:
Agreed we don't need this, projects that want to use go bindings can create them in their own repos

PR #115, opened 2023-07-25
initial write of AVSViewer

gpsanant:
This PR 
- Creates a base service manager for AVS teams to inherit and use
- Introduces an AVSViewer contract that all service managers must have that is used for fetching values needed on the eigenlayer frontend

### DIFF
--- a/src/contracts/interfaces/IAVSStateViewer.sol
+++ b/src/contracts/interfaces/IAVSStateViewer.sol
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity >=0.5.0;
+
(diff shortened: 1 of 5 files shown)


ChaoticWalrus:
@gpsanant is this PR really ready for review?
Looks like the ServiceManagerBase should probably have virtual/overridable functions, at least. Am also confused by the non-usage of the newly defined events.
I'm also finding it a bit weird that the AVSViewer is designed to fetch all of the strategies in each of an operator's current quorums. Is there reasoning behind fetching and returning this data specifically (and only this data)?

More context would be helpful; I'm happy to have a meeting about this -- or if this has already been considered/discussed, a link to a doc would be great.

ChaoticWalrus:
Closing out this stale PR -- we need more discussion on requirements here first.

PR #118, opened 2023-08-07
Permissioned operator churn

gpsanant:
Added churner role and tests for it

Tests summary/updates:
- Added utility for churner to sign on OperatorChurnApprovals
- Added tests for 
    - Expired signatures
    - Invalid signatures
    - Added churner signatures to existing happy path test cases
    - Abstracted internal logic from operator kicking tests

### DIFF
--- a/src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
+++ b/src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
@@ -56,8 +56,8 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
     address[] public registries;
     /// @notice the address of the entity allowed to sign off on operators getting kicked out of the AVS during registration
     address public churnApprover;
-    /// @notice the nonce of the churnApprover used in EIP-712 signatures
-    uint256 public churnApproverNonce;
+    /// @notice whether the salt has been used for an operator churn approval
+    mapping(bytes32 => bool) public isChurnApproverSaltUsed;
 
     modifier onlyServiceManagerOwner {
         require(msg.sender == serviceManager.owner(), "BLSRegistryCoordinatorWithIndices.onlyServiceManagerOwner: caller is not the service manager owner");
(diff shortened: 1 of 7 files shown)


src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -39,6 +45,14 @@ contract BLSRegistryCoordinatorWithIndices is Initializable, IBLSRegistryCoordin
     IStakeRegistry public immutable stakeRegistry;
     /// @notice the Index Registry contract that will keep track of operators' indexes
     IIndexRegistry public immutable indexRegistry;
+
+    /**
+     * @notice Original EIP-712 Domain separator for this contract.
+     * @dev The domain separator may change in the event of a fork that modifies the ChainID.
+     * Use the getter function `domainSeparator` to get the current domain separator for this contract.
+     */
+    bytes32 internal _DOMAIN_SEPARATOR;
```
ChaoticWalrus, line 54: doesn't the OZ library you're inheriting from define the equivalent of this? seems unneeded?

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -47,6 +61,10 @@ contract BLSRegistryCoordinatorWithIndices is Initializable, IBLSRegistryCoordin
     mapping(address => Operator) internal _operators;
     /// @notice the dynamic-length array of the registries this coordinator is coordinating
     address[] public registries;
+    /// @notice the address of the entity allowed to sign off on operators getting kicked out of the AVS during registration
+    address public churner;
+    /// @notice the nonce of the churner used in EIP-712 signatures
+    uint256 public churnerNonce;
```
ChaoticWalrus, line 67: for naming: the description suggests that this entity is really a `churnApprover` or similar.

am also thinking that 'salts' may make more sense than nonces here, but we can debate a bit -- the intended usage is important to consider

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -59,15 +77,17 @@ contract BLSRegistryCoordinatorWithIndices is Initializable, IBLSRegistryCoordin
...
         serviceManager = _serviceManager;
         stakeRegistry = _stakeRegistry;
         blsPubkeyRegistry = _blsPubkeyRegistry;
         indexRegistry = _indexRegistry;
     }
 
-    function initialize(OperatorSetParam[] memory _operatorSetParams) external initializer {
+    function initialize(address _churner, OperatorSetParam[] memory _operatorSetParams) external initializer {
+        // set the churner
+        churner = _churner;
```
ChaoticWalrus, line 90: should presumably have an event associated to this

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -398,4 +458,13 @@ contract BLSRegistryCoordinatorWithIndices is Initializable, IBLSRegistryCoordin
             _operators[operator].status = OperatorStatus.DEREGISTERED;
         }
     }
+
+    /// @notice verifies churner's signature on operator churn approval and increments the churner nonce
+    function _verifyChurnerSignatureOnOperatorChurnApproval(bytes32 registeringOperatorId, OperatorKickParam[] memory operatorKickParams, SignatureWithExpiry memory signatureWithExpiry) internal {
+        uint256 churnerNonceMem = churnerNonce;
+        require(signatureWithExpiry.expiry >= block.timestamp, "BLSRegistryCoordinatorWithIndices._verifyChurnerSignatureOnOperatorChurnApproval: churner signature expired");
```
ChaoticWalrus, line 465: geq/leq looks flipped here -- seems like you are requiring that the signature is expired?

ChaoticWalrus:
Gave some quick input on this PR. @gpsanant can you provide a brief summary of test changes / updates here?

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: Changed to churn approver, would prefer to keep the nonce logic for now since we expect this to not have many race conditions

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: we are checking that `expiry >= now`. seems right?

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: agreed offline that salt is more flexible here / nonce is not needed

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: oh yeah I'm wrong, ty

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: added

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: added

PR #123, opened 2023-08-09
update spec with change to function

ChaoticWalrus:
very minimal change, I haven't verified that rules are still correct, merely that they run


ChaoticWalrus:
this is a minimal change to fix a break in CI, just force-merging it.

PR #127, opened 2023-08-10
EigenPods Cleanup

Sidu28:
edit from @ChaoticWalrus:
This PR combines several changes to EigenPods. These include:

- Moving "beacon chain ETH shares" logic from the StrategyManager into the EigenPodManager (and associated changes to the DelegationManager)
- introduction of an "undelegation limbo" mode to provide stakers with a way to modify their delegation in EigenLayer without having to exit their validators from the beacon chain
- addition of a `receive` method to EigenPods + associated function for withdrawing "non-beacon chain ETH" and a method for sweeping out ERC20 tokens sent to EigenPods
- TODO @Sidu28 add other changes included in this PR; the description shouldn't be empty


src/contracts/core/StrategyManager.sol
```diff
@@ -941,8 +941,8 @@ contract StrategyManager is
         }   else {
                 uint256 shareAmount = uint256(sharesDelta);
                 //if change in shares is positive, add the shares
+                //note: _addShares calls increaseDelegatedShares while _removeShares does not.
```
gpsanant, line 944: Doesn't this mean queued withdrawals won't decrease delegated shares then?

gpsanant: can u add a space between `//`

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -44,6 +44,12 @@ library BeaconChainProofs {
     uint256 internal constant STATE_ROOTS_TREE_HEIGHT = 13;
     uint256 internal constant BLOCK_ROOTS_TREE_HEIGHT = 13;
 
+    //HISTORICAL_ROOTS_LIMIT = 2**24, so tree height is 24
```
gpsanant, line 47: add space pls

src/contracts/core/StrategyManager.sol
Sidu28: I think queuedWithdrawals makes a separate call to decreaseDelecatedSares

src/contracts/pods/EigenPod.sol
```diff
@@ -597,6 +604,26 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+    function receive() external payable {
+        nonBeaconChainETHBalanceWei += msg.value;
+        emit nonBeaconChainETHReceived(msg.value);
+    }
+
+    /// @notice Called by the pod owner to withdraw the nonBeaconChainETHBalanceWei
+    function withdrawnonBeaconChainETHBalanceWei(address recipient, uint256 amountToWithdraw) external onlyEigenPodOwner {
+        require(amountToWithdraw <= nonBeaconChainETHBalanceWei, "EigenPod.withdrawnonBeaconChainETHBalanceWei: amountToWithdraw is greater than nonBeaconChainETHBalanceWei");
+        AddressUpgradeable.sendValue(payable(recipient), amountToWithdraw);
+        nonBeaconChainETHBalanceWei -= amountToWithdraw;
```
ChaoticWalrus, line 617: this is CIE, not CEI. Just do the subtraction first.

src/contracts/pods/EigenPod.sol
```diff
@@ -107,6 +111,9 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
 
     /// @notice Emitted when podOwner enables restaking
     event restakingActivated(address indexed podOwner);
+
+    /// @notice Emitted when ETH is received via the receive fallback
+    event nonBeaconChainETHReceived(uint256 amountReceived);
```
ChaoticWalrus, line 116: might want an event for withdrawing? idk though

ChaoticWalrus:
Copying some of a comment from https://github.com/Layr-Labs/eigenlayer-contracts/pull/135 :
we've agreed that a PR addressing the following will be opened as soon as possible:

-     Outstanding PR comments in EigenPods
-     Full withdrawal proof needs to be fixed
-     Switch to use timestamps instead of slots in storage and events
-     Fix order of functions in EigenPods

src/contracts/core/DelegationManagerStorage.sol
wadealexc, line 1: Is the storage _gap variable right? Mainnet has [46], and it looks like we only have one additional storage variable in this contract (`eigenPodManager` is `immutable`, which shouldn't have a storage slot)

Sidu28: I just took a quick look, counting the storage in both master branch and whats on mainnet, seems like this gap should be 45 not 44?  Someone double check that

ChaoticWalrus: using `forge inspect` I count 6 storage slots used in the DelegationManager itself, so 44 seems correct

ChaoticWalrus: discussed this some offline -- the current deployment has an extra slot -- because it's at the end of the inheritance graph, this is something we should be able to fix in an upgrade

ChaoticWalrus:
copying another important comment:
[gpsanant](https://github.com/gpsanant) [Sep 11, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1322242122)

in verifyWithdrawalCredentials, we do not set 2 fields of the validatorInfo struct?
Member
@[ChaoticWalrus](https://github.com/ChaoticWalrus) ChaoticWalrus [Sep 13, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1325045048)

I see 'validatorIndex', 'restakedBalanceGwei', and 'status' getting updated, but not 'mostRecentBalanceUpdateSlot' -- this is just one field, but kind of awkward / doesn't seem desired.
@Sidu28 can you confirm if bug?

ChaoticWalrus:
A couple other open threads:
[ChaoticWalrus](https://github.com/ChaoticWalrus) [Sep 7, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1319207038)

what happens if I mark proofs.proveHistoricalRoot as 'false' but I supply a really old blockHeaderRoot?
where should the process fail?
do we have test coverage for this?

[Sidu28](https://github.com/Sidu28) [Sep 7, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1319228156)

doesnt exist, all the historical proof stuff is a wip

[ChaoticWalrus](https://github.com/ChaoticWalrus) [Sep 8, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1320256558)

I added a TODO + note in the code about this for now, at least.


AND

[gpsanant](https://github.com/gpsanant) [Sep 11, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1322239808)

in line 402, we are calling this the latestBlockHeaderRoot. seems like we should call this blockHeaderRoot
Member
@[ChaoticWalrus](https://github.com/ChaoticWalrus) ChaoticWalrus [Sep 12, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1323661594)

I believe I made the same comment elsewhere. Per @Sidu28 this is literally called the latestBlockHeaderRoot in the spec. I think at minimum we need to add some comments to clarify; if you have ideas where, please make a commit!
Member
@[gpsanant](https://github.com/gpsanant) gpsanant [Sep 13, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1324703139)

This was where the proof was broken
Member
@[ChaoticWalrus](https://github.com/ChaoticWalrus) ChaoticWalrus [Sep 13, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1325045483)

    This was where the proof was broken

please expand on this idea / I guess file an issue for @Sidu28 for this

Sidu28:
Additional changes made to this PR at commit [4eb1156](https://github.com/Layr-Labs/eigenlayer-contracts/pull/127/commits/4eb1156e73194cf8fcf86586a2985e5865e7f15e).  Fixed the "stateAgainstLatestBlockHeaderProof", the bug fix noted in this [issue](https://github.com/orgs/Layr-Labs/projects/7/views/2?filterQuery=assignee%3ASidu28&pane=issue&itemId=38512330)

Sidu28:
> A couple other open threads: [ChaoticWalrus](https://github.com/ChaoticWalrus) [Sep 7, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1319207038)
> 
> what happens if I mark proofs.proveHistoricalRoot as 'false' but I supply a really old blockHeaderRoot? where should the process fail? do we have test coverage for this?
> 
> [Sidu28](https://github.com/Sidu28) [Sep 7, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1319228156)
> 
> doesnt exist, all the historical proof stuff is a wip
> 
> [ChaoticWalrus](https://github.com/ChaoticWalrus) [Sep 8, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1320256558)
> 
> I added a TODO + note in the code about this for now, at least.
> 
> AND
> 
> [gpsanant](https://github.com/gpsanant) [Sep 11, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1322239808)
> 
> in line 402, we are calling this the latestBlockHeaderRoot. seems like we should call this blockHeaderRoot Member @[ChaoticWalrus](https://github.com/ChaoticWalrus) ChaoticWalrus [Sep 12, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1323661594)
> 
> I believe I made the same comment elsewhere. Per @Sidu28 this is literally called the latestBlockHeaderRoot in the spec. I think at minimum we need to add some comments to clarify; if you have ideas where, please make a commit! Member @[gpsanant](https://github.com/gpsanant) gpsanant [Sep 13, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1324703139)
> 
> This was where the proof was broken Member @[ChaoticWalrus](https://github.com/ChaoticWalrus) ChaoticWalrus [Sep 13, 2023](https://github.com/Layr-Labs/eigenlayer-contracts/pull/135#discussion_r1325045483)
> 
> ```
> This was where the proof was broken
> ```
> 
> please expand on this idea / I guess file an issue for @Sidu28 for this

All these have been addressed I believe.  Currently historical summary proofs dont exist yet, this PR is everything else (bug fix, delegation changes, etc."

src/contracts/pods/EigenPod.sol
```diff
@@ -456,6 +458,8 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         require(!provenWithdrawal[validatorPubkeyHash][Endian.fromLittleEndianUint64(withdrawalProofs.slotRoot)],
             "EigenPod._verifyAndProcessWithdrawal: withdrawal has already been proven for this slot");
 
+        emit log_named_uint("len", withdrawalProofs.latestBlockHeaderProof.length);
```
gpsanant: the oracle function needs to be renamed

src/contracts/pods/EigenPod.sol
gpsanant: Yep, this is a good idea

ChaoticWalrus: done in this commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/127/commits/fcfed5c7395a42faabb89e04a0ff374d0b0bbab6

ChaoticWalrus:
> Additional changes made to this PR at commit [4eb1156](https://github.com/Layr-Labs/eigenlayer-contracts/pull/127/commits/4eb1156e73194cf8fcf86586a2985e5865e7f15e). Fixed the "stateAgainstLatestBlockHeaderProof", the bug fix noted in this [issue](https://github.com/orgs/Layr-Labs/projects/7/views/2?filterQuery=assignee%3ASidu28&pane=issue&itemId=38512330)

I don't see this commit fixing the bug -- I believe you addressed this in some previous commit(s)?
I agree with G that the syntax we discussed would help with clarifying that your usage is now correct. I'm either not understanding your comment on the linked issue or the comment you've just made here, but the linked comment seems to be discussing how to fetch the state, not how you should be checking it / the fix itself.

Sidu28:
> > Additional changes made to this PR at commit [4eb1156](https://github.com/Layr-Labs/eigenlayer-contracts/pull/127/commits/4eb1156e73194cf8fcf86586a2985e5865e7f15e). Fixed the "stateAgainstLatestBlockHeaderProof", the bug fix noted in this [issue](https://github.com/orgs/Layr-Labs/projects/7/views/2?filterQuery=assignee%3ASidu28&pane=issue&itemId=38512330)
> 
> I don't see this commit fixing the bug -- I believe you addressed this in some previous commit(s)? I agree with G that the syntax we discussed would help with clarifying that your usage is now correct. I'm either not understanding your comment on the linked issue or the comment you've just made here, but the linked comment seems to be discussing how to fetch the state, not how you should be checking it / the fix itself.

Oh you're right, thats because the actual fix is one [line](https://github.com/Layr-Labs/eigenlayer-contracts/blob/a4ceb796a9e2f4e2aa32b1221ce0bfc1060755c7/src/contracts/libraries/BeaconChainProofs.sol#L245).  When you say linked comment, you mean linked issue?  The issue literally is "We are proving the block root against a provided state root rather than the state root against the oracle block root."  So this fix, aka the line I just sent you + the new tests (the commit i originall linked) fix this exact issue where we now are proving the state root against the oracle block root.

ChaoticWalrus:
> Oh you're right, thats because the actual fix is one [line](https://github.com/Layr-Labs/eigenlayer-contracts/blob/a4ceb796a9e2f4e2aa32b1221ce0bfc1060755c7/src/contracts/libraries/BeaconChainProofs.sol#L245). The commit I linked has all the test changes and updates which is the meat and potatoes of this fix. When you say linked comment, you mean linked issue? The issue literally states "We are proving the block root against a provided state root rather than the state root against the oracle block root." So this fix, aka the line I just sent you + the new tests (the commit i originall linked) fix this exact issue where we now are proving the state root against the oracle block root.

OK. This is the level of detail you should have provided without prompting and questioning. Now that you've provided it, I can ask some more relevant questions and provide comments:

- are you planning regression testing for this bug? if so, what tests / where are you tracking this? if not, why?
- is it possible there are similar bugs in the code? if not, why? if so, where?
- how will you prevent the same / similar bugs being reintroduced, and make it easier to make sure they aren't there?
- note that we already had a discussion in Slack about the above, which it seems you both didn't participate in and missed -- G already asked for the change we agreed made sense from that discussion, in [this comment above](https://github.com/Layr-Labs/eigenlayer-contracts/pull/127#discussion_r1329032758)

ChaoticWalrus:
> As far as details go, do you want me to regurgitate the entire issue again.......? The issue explains the bug, and the code diffs represent the fix (apart from the one line that wasn't in the commit, my bad). So I link to the issue in order to save myself time and the reviewer doesn't lose out because they can click on the link and read the issue.

You asked for review on an open PR with a bunch of existing comments, providing no context on what changes you had made. People asked for a comment, so you made a one-line, [incorrect comment](https://github.com/Layr-Labs/eigenlayer-contracts/pull/127#issuecomment-1723976284) simply pointing to a PR and saying that you fixed an issue, which is not what the commit did. You then linked an issue that remains open, where you *also* didn't provide any comment about fixing the issue. I asked for clarification and now you're apparently mad. You cold have just been clear from the start.

> As far as similar bugs go, I don't think so, if I did I'd fix them asap

I asked if it was possible, but your response seem to be "I don't think there are any", which is a response to a different question. I'm trying to understand if we have good practices here, not if you've fixed the one thing that was identified as wrong and done some minimal spot-checking.

> Yes I'm aware, he commented it as a response to my initial commit. I'm addressing that stuff plus some other small comments in my "HistoricalSummary" PR which I will merge into this one, as I discussed with G offline.

Thanks for filling me in.

src/contracts/pods/EigenPod.sol
```diff
@@ -616,6 +619,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         require(amountToWithdraw <= nonBeaconChainETHBalanceWei,
             "EigenPod.withdrawnonBeaconChainETHBalanceWei: amountToWithdraw is greater than nonBeaconChainETHBalanceWei");
         nonBeaconChainETHBalanceWei -= amountToWithdraw;
+        emit NonBeaconChainETHWithdrawn(recipient, amountToWithdraw);
         AddressUpgradeable.sendValue(payable(recipient), amountToWithdraw);
```
gpsanant, line 623: we are not sending this through DWR?

src/contracts/core/StrategyManagerStorage.sol
```diff
@@ -70,8 +70,6 @@ abstract contract StrategyManagerStorage is IStrategyManager {
     // slither-disable-next-line incorrect-shift-in-assembly
     uint256[1] internal _deprecatedStorage;
```
gpsanant, line 71: what's this? i seem to be forgetting...

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -4,6 +4,8 @@ pragma solidity >=0.5.0;
 import "../libraries/BeaconChainProofs.sol";
 import "./IEigenPodManager.sol";
 import "./IBeaconChainOracle.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
gpsanant, line 7: could we move events to interfaces for Pod stuff and delegation manager? I prefer it.

Wdyt @ChaoticWalrus

src/contracts/pods/EigenPod.sol
ChaoticWalrus: hmm, good callout. It might be "safer "to route it via the DWR, but I do feel like this is a fairly well-contained pair of functions. any particular concerns? If you are strongly opinionated and want to make a PR changing this behavior to route through the DWR instead, I'm fine with it.

src/contracts/core/StrategyManagerStorage.sol
ChaoticWalrus: there's a comment just above:
```
    /* 
     * Reserved space previously used by the deprecated mapping(address => uint256) beaconChainETHSharesToDecrementOnWithdrawal.
     * This mapping tracked beaconChainETH "debt" in case updates were made to shares retroactively.  However, this design was
     * replaced by a simpler design that prevents withdrawals from EigenLayer before withdrawals from the beacon chain, which
     * makes this tracking unnecessary.
     */
```

src/contracts/interfaces/IEigenPod.sol
ChaoticWalrus: yes please. I think this is "more correct" / better practice.

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -288,7 +287,7 @@ library BeaconChainProofs {
             * intermediate root indexes from the bottom of the sub trees (the block header container) to the top of the tree
             */
             uint256 blockHeaderIndex = BLOCK_ROOTS_INDEX << (BLOCK_ROOTS_TREE_HEIGHT)  | uint256(proofs.blockHeaderRootIndex);
-            // Verify the blockHeaderRoot against the beaconStateRoot
+            // Verify the blockHeaderRoot against the beaconStateRoot            
             require(Merkle.verifyInclusionSha256(proofs.blockHeaderProof, beaconStateRoot, proofs.blockHeaderRoot, blockHeaderIndex),
```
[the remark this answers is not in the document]
Sidu28: fixed

src/contracts/interfaces/IEigenPod.sol
Sidu28: will make a separate PR for this as well as other stuff mentioned: https://github.com/Layr-Labs/eignlayr-contracts/issues/962

src/contracts/pods/EigenPod.sol
gpsanant: I prefer DWR for monitoring and cohesiveness of invariants. Wdyt @Sidu28

src/contracts/core/StrategyManagerStorage.sol
gpsanant: oops thanks!

src/contracts/pods/EigenPod.sol
Sidu28: Sounds good, will implement in a separate PR associate with this ticket: https://github.com/Layr-Labs/eignlayr-contracts/issues/962

src/contracts/core/StrategyManagerStorage.sol
```diff
@@ -83,5 +81,5 @@ abstract contract StrategyManagerStorage is IStrategyManager {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[40] private __gap;
+    uint256[42] private __gap;
```
ChaoticWalrus, line 84: not sure why/when this gap size changed, but it looks to me like it should still be 40, not changed to 42?
I ran `forge inspect src/contracts/core/StrategyManager.sol:StrategyManager storage --pretty` and see the gap starting at slot 211, while this contract's storage starts at slot 201 (so to use a total of 50 slots, gap size of 40 seems appropriate).

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -272,29 +278,32 @@ library BeaconChainProofs {
...
+        require(proofs.historicalSummaryBlockRootProof.length == 32 * (BEACON_STATE_FIELD_TREE_HEIGHT + (HISTORICAL_SUMMARIES_TREE_HEIGHT + 1) + 1 + (BLOCK_ROOTS_TREE_HEIGHT)),
+        "BeaconChainProofs.verifyWithdrawalProofs: historicalSummaryBlockRootProof has incorrect length");
+        /**
+        * Note: Here, the "1" in "1 + (BLOCK_ROOTS_TREE_HEIGHT)" signifies that extra step of choosing the "block_root_summary" within the individual 
+        * "historical_summary". Everywhere else it signifies merkelize_with_mixin, where the length of an array is hashed with the root of the array,
+        * but not here.
+        */
+        uint256 historicalBlockHeaderIndex = HISTORICAL_SUMMARIES_INDEX << ((HISTORICAL_SUMMARIES_TREE_HEIGHT + 1) + 1 + (BLOCK_ROOTS_TREE_HEIGHT)) | 
+                                            uint256(proofs.historicalSummaryIndex) << 1 + (BLOCK_ROOTS_TREE_HEIGHT) |
+                                            BLOCK_SUMMARY_ROOT_INDEX << (BLOCK_ROOTS_TREE_HEIGHT) | uint256(proofs.blockHeaderRootIndex);
```
ChaoticWalrus, line 291: I'm trying to make sure I have the order of operations this is doing correct, and think some parentheses might be a good idea here.
Let me know if this is correct / here's a suggestion for what I personally think is clearer in terms of order-of-operations:
```
        uint256 historicalBlockHeaderIndex =
               (HISTORICAL_SUMMARIES_INDEX << ((HISTORICAL_SUMMARIES_TREE_HEIGHT + 1) + 1 + (BLOCK_ROOTS_TREE_HEIGHT))) | 
               (uint256(proofs.historicalSummaryIndex) << (1 + BLOCK_ROOTS_TREE_HEIGHT)) |
               (BLOCK_SUMMARY_ROOT_INDEX << BLOCK_ROOTS_TREE_HEIGHT) | 
               uint256(proofs.blockHeaderRootIndex);
```

src/contracts/pods/EigenPod.sol
```diff
@@ -73,19 +80,21 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
 
-    // STORAGE VARIABLES
     /// @notice the amount of execution layer ETH in this contract that is staked in EigenLayer (i.e. withdrawn from the Beacon Chain but not from EigenLayer), 
     uint64 public withdrawableRestakedExecutionLayerGwei;
 
     /// @notice an indicator of whether or not the podOwner has ever "fully restaked" by successfully calling `verifyCorrectWithdrawalCredentials`.
     bool public hasRestaked;
 
-    /// @notice This is a mapping of validatorPubkeyHash to slot to whether or not they have proven a withdrawal for that slot
+    /// @notice This is a mapping of validatorPubkeyHash to timestamp to whether or not they have proven a withdrawal for that slot
```
ChaoticWalrus, line 89: 'for that slot' should be 'for that timestamp', right?

src/contracts/pods/EigenPod.sol
```diff
@@ -94,20 +103,25 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-    event ValidatorBalanceUpdated(uint40 validatorIndex, uint64 newValidatorBalanceGwei);
+    event ValidatorBalanceUpdated(uint40 validatorIndex, uint64 timestamp, uint64 newValidatorBalanceGwei);
     
     /// @notice Emitted when an ETH validator is prove to have withdrawn from the beacon chain
-    event FullWithdrawalRedeemed(uint40 validatorIndex, address indexed recipient, uint256 withdrawalAmountWei);
+    event FullWithdrawalRedeemed(uint40 validatorIndex, uint64 timestamp, address indexed recipient, uint256 withdrawalAmountWei);
 
     /// @notice Emitted when a partial withdrawal claim is successfully redeemed
-    event PartialWithdrawalRedeemed(uint40 validatorIndex, address indexed recipient, uint64 partialWithdrawalAmountGwei);
+    event PartialWithdrawalRedeemed(uint40 validatorIndex, uint64 timestamp, address indexed recipient, uint64 partialWithdrawalAmountGwei);
```
ChaoticWalrus, line 112: why are the partial withdrawal and validator balance update event amounts of type uint64 / in Gwei, while the amount for the full withdrawal is in wei? seems like a recipe for messing something up in indexing

src/contracts/pods/EigenPod.sol
```diff
@@ -201,15 +217,13 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         onlyWhenNotPaused(PAUSED_EIGENPODS_VERIFY_CREDENTIALS)
         // check that the provided `oracleTimestamp` is after the `mostRecentWithdrawalTimestamp`
         proofIsForValidTimestamp(oracleTimestamp)
-        //ensure that caller has restaking enabled by calling "activateRestaking()"
+        // ensure that caller has previously enabled restaking by calling `activateRestaking()`
```
ChaoticWalrus, line 220: right now for a new EigenPod, won't a user have to:
1. deploy EigenPod
2. "enable restaking"
?

we should update the initializer for new EigenPods + the call made by the EigenPodManager, so that new pods automatically have "restaking enabled"

ChaoticWalrus:
Doing a bit of a review not, am pushing some minor changes / fixing nits.
I've hit a question:
it looks like we aren't checking the length of the `latestBlockHeaderProof` field in the `WithdrawalProofs` struct anywhere -- am I missing something, or is this a potential issue?

ChaoticWalrus:
Similar to my above question -- it seems we are not checking the length of any of the 3 proof fields of the `BalanceUpdateProofs` struct anywhere (namely, the `latestBlockHeaderProof`, the `validatorBalanceProof`, or the `validatorFieldsProof`), nor are we checking the length of the `latestBlockHeaderProof` or `validatorFieldsProof` of the `WithdrawalCredentialProofs` struct

ChaoticWalrus:
talked with G about my above 2 questions -- seems that these things are being checked, but that the param names just don't align. going to make a commit with some minor renaming for clarity.

src/contracts/core/StrategyManagerStorage.sol
Sidu28: fixed

src/contracts/libraries/BeaconChainProofs.sol
Sidu28: if Im reading this correct, each of the components in between the bitwise or's are wrapped in parenthesis?  Yeah that looks cleaner, added it

src/contracts/pods/EigenPod.sol
Sidu28: fixed

src/contracts/core/StrategyManagerStorage.sol
ChaoticWalrus: looks like this was addressed in commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/127/commits/feb53c860b20f3d3f164b4359f3ec59989347c9b

Sidu28:
> Doing a bit of a review now, am pushing some minor changes / fixing nits. I've hit a question: it looks like we aren't checking the length of the `latestBlockHeaderProof` field in the `WithdrawalProofs` struct anywhere -- am I missing something, or is this a potential issue?

Check this out: https://github.com/Layr-Labs/eigenlayer-contracts/blob/ce2792e65a54f95395ea2ce399612e0206dc2ea7/src/contracts/libraries/BeaconChainProofs.sol#L247

Sidu28:
> validatorBalanceProof

Ok this is the validatorBalanceProof length check: https://github.com/Layr-Labs/eigenlayer-contracts/blob/90ea6afb53487cde2736a53c1ea9cb7068f41b06/src/contracts/libraries/BeaconChainProofs.sol#L208


This is the validatorFieldsProof length check:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/90ea6afb53487cde2736a53c1ea9cb7068f41b06/src/contracts/libraries/BeaconChainProofs.sol#L184

src/contracts/pods/EigenPodManager.sol
```diff
@@ -29,14 +31,27 @@ import "./EigenPodPausingConstants.sol";
...
  * - withdrawing eth when withdrawals are initiated
  */
-contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenPodManager, EigenPodPausingConstants {
+contract EigenPodManager is 
+    Initializable, 
+    OwnableUpgradeable, 
+    Pausable, 
+    IEigenPodManager,
+    EigenPodPausingConstants,
+    ReentrancyGuardUpgradeable
```
ChaoticWalrus, line 40: I'm fairly certain that adding this into the inheritance graph of the EigenPodManager will destroy its storage pattern by shifting it downward 50 slots.
might need to do something like moving the storage into a separate contract and inheriting from `ReentrancyGuardUpgradeable` further down the tree?
please discuss with me if you are planning on doing this / unsure of the best approach here

src/contracts/pods/EigenPodManager.sol
```diff
@@ -134,68 +202,215 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+            ),
+            "EigenPodManager.slashShares: cannot slash the specified pod owner"
+        );
+        require(shareAmount > 0, "EigenPodManager.slashShares: shares must be greater than zero");
+        require(podOwnerShares[slashedPodOwner] >= shareAmount,
+            "EigenPodManager.slashShares: podOwnerShares[podOwner] must be greater than or equal to shares");
+
+        _removeShares(slashedPodOwner, shareAmount);
+        // TODO: @Sidu28 -- confirm that decrementing `withdrawableRestakedExecutionLayerGwei` is correct/intended here
+        _decrementWithdrawableRestakedExecutionLayerGwei(slashedPodOwner, shareAmount);
```
ChaoticWalrus, line 361: there's a TODO here in the code for @Sidu28

ChaoticWalrus: actually nevermind, I have a PR open to remove this function entirely, should probably just merge that first

src/contracts/pods/EigenPodManager.sol
```diff
@@ -207,6 +422,124 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+        address podOwner,
+        uint256 amountWei,
+        address withdrawer,
+        bool undelegateIfPossible
+    ) 
+        internal
+        returns (bytes32)
+    {
+        require(amountWei > 0, "EigenPodManager._queueWithdrawal: amount must be greater than zero");
+
```
ChaoticWalrus, line 435: TODO: make sure there's no edge case where we hit this and cause an unexpected revert

src/contracts/pods/EigenPodManager.sol
```diff
@@ -207,6 +422,124 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+        require(amountWei % GWEI_TO_WEI == 0,
+            "EigenPodManager._queueWithdrawal: cannot queue a withdrawal of Beacon Chain ETH for an non-whole amount of gwei");
```
ChaoticWalrus, line 437: do we also ensure that the user's shares only *increase* by a full Gwei amount? if not, seems like we probably should!

src/contracts/pods/EigenPodManager.sol
```diff
@@ -243,6 +576,120 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+
+    /**
+     * @notice Withdraws ETH from an EigenPod. The ETH must have first been withdrawn from the beacon chain.
+     * @param podOwner The owner of the pod whose balance must be withdrawn.
+     * @param recipient The recipient of the withdrawn ETH.
+     * @param amount The amount of ETH to withdraw.
+     * @dev Callable only by the StrategyManager contract.
+     */
+    function _withdrawRestakedBeaconChainETH(address podOwner, address recipient, uint256 amount)
+        internal onlyWhenNotPaused(PAUSED_WITHDRAW_RESTAKED_ETH)
```
ChaoticWalrus, line 646: I don't think having this modifier here makes sense.
Didn't we have a previous thread about this in another PR already?

ChaoticWalrus: IIRC the flows that arrive at calling this function already all check `onlyWhenNotPaused(PAUSED_WITHDRAWALS)`

src/contracts/pods/EigenPodPausingConstants.sol
```diff
@@ -18,4 +18,9 @@ abstract contract EigenPodPausingConstants {
     uint8 internal constant PAUSED_EIGENPODS_VERIFY_BALANCE_UPDATE = 3;
     /// @notice Index for flag that pauses the `verifyBeaconChainFullWithdrawal` function *of the EigenPods* when set. see EigenPod code for details.
     uint8 internal constant PAUSED_EIGENPODS_VERIFY_WITHDRAWAL = 4;
+
+    /// @notice Index for flag that pauses deposits to EigenPodManager when set.
+    uint8 internal constant PAUSED_DEPOSITS = 5;
+    /// @notice Index for flag that pauses withdrawals from EigenPodManager when set.
+    uint8 internal constant PAUSED_WITHDRAWALS = 6;
```
ChaoticWalrus, line 25: do we need a new flag here?
It seems like `PAUSED_DEPOSITS` is being used the same way  `PAUSED_EIGENPODS_VERIFY_CREDENTIALS` was designed, and likewise that `PAUSED_WITHDRAWALS` is being used the same way `PAUSED_WITHDRAW_RESTAKED_ETH` was designed. 

also looks like the comments on this file didn't get updated at all.

src/contracts/pods/EigenPod.sol
ChaoticWalrus: @gpsanant thoughts on the correct format here? I would think all uint256 / wei amounts makes the most sense?

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: I'm OK merging this PR without addressing this particular issue. I will give myself a TODO elsewhere.

src/contracts/pods/EigenPodManager.sol
Sidu28: We use the validator's effective balance to set the shares, which is always denominated in Gwei:

Initially:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/6525091148d1e10a0922c495b74d7c878c6c9d1c/src/contracts/pods/EigenPod.sol#L437

For balance updates, we convert it to Gwei:

https://github.com/Layr-Labs/eigenlayer-contracts/blob/6525091148d1e10a0922c495b74d7c878c6c9d1c/src/contracts/pods/EigenPod.sol#L304

So we're always accountfing for balances in Gwei (hwhich we convert to wei):
initially:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/6525091148d1e10a0922c495b74d7c878c6c9d1c/src/contracts/pods/EigenPod.sol#L445

balance update:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/6525091148d1e10a0922c495b74d7c878c6c9d1c/src/contracts/pods/EigenPod.sol#L320

src/contracts/pods/EigenPodManager.sol
Sidu28: Hmm ok, looking at it, slashShares(), slashQueuedWithdrawal(),  doesnt have a explicit pauser.  So what I'll do is add the onlyWhenNotPaused(PAUSED_WITHDRAWALS) to those functions?  I think that makes sense given that those functions act very similarly to normal withdrawals

src/contracts/pods/EigenPodManager.sol
Sidu28: From @ChaoticWalrus :
I think:
move the EigenPodManager storage (other than immutables) into a simple, standalone contract (EigenPodManagerStorage , don’t have it inherit from anything)
make the inheritance graph look like:
    Initializable, 
    OwnableUpgradeable, 
    Pausable, 
    IEigenPodManager,
    EigenPodPausingConstants,
    EigenPodManagerStorage,
    ReentrancyGuardUpgradeable

Sidu28: this works because all the storage is inherited before the ReentrancyGuardUpgradeable

src/contracts/pods/EigenPodPausingConstants.sol
Sidu28: Agreed, fixed!

src/contracts/pods/EigenPodManager.sol
Sidu28: Lol nvm, I see that it was  removed, fixed!

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: yeah, I mean makes sense. I'm OK w/ merging like this, may add a check that enforces this behavior more strictly but I'll think about it a touch more first.

Sidu28:
Ok made the storage change to EigenPodManager.  One question, moving the "IEigenPodManager" import from EigenPodManager to EigenPodManagerStorage shouldn't affect anything right? As far as I can rationalize it, it should be fine.

src/contracts/pods/EigenPodManager.sol
Sidu28: fixed!

ChaoticWalrus:
> moving the "IEigenPodManager" import from EigenPodManager to EigenPodManagerStorage shouldn't affect anything right?

yes, agreed. should be abstract anyway so that's fine.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -134,68 +137,156 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+     * @param amountWei The amount of ETH to 'deposit' (i.e. be credited to the podOwner).
      * @dev Callable only by the podOwner's EigenPod contract.
      */
-    function restakeBeaconChainETH(address podOwner, uint256 amount) external onlyEigenPod(podOwner) {
-        strategyManager.depositBeaconChainETH(podOwner, amount);
-        emit BeaconChainETHDeposited(podOwner, amount);
+    function restakeBeaconChainETH(address podOwner, uint256 amountWei) 
+        external 
+        onlyEigenPod(podOwner) 
+        onlyWhenNotPaused(PAUSED_EIGENPODS_VERIFY_CREDENTIALS)
```
ChaoticWalrus, line 146: isn't this behavior already enforced at `EigenPod.verifyWithdrawalCredentials`, which is the only path by which this function is called?

ChaoticWalrus (changes requested):
This seems really close.
Summarizing a few open threads:
looks like this is open https://github.com/Layr-Labs/eigenlayer-contracts/pull/127#discussion_r1332290356
I’d want to see some tracking for this, at least https://github.com/Layr-Labs/eigenlayer-contracts/pull/127#discussion_r1332291243
I’m OK leaving this as basically a note to myself https://github.com/Layr-Labs/eigenlayer-contracts/pull/127#discussion_r1333423866

There's other work happening in a separate branch, see https://github.com/Layr-Labs/eigenlayer-contracts/pull/180 -- I'm cool merging this PR first, just want to spell it out clearly.

src/contracts/pods/EigenPodManager.sol
Sidu28: yup, good point.  In general its probably ok to have redundant pauser modifier on external functions?  wdyt.

For this one, ill fux it since its perm'd to the pod only

src/contracts/pods/EigenPod.sol
gpsanant: uint64 timestamps are fine and maybe preferred because they work better with sql and are what's used on the beacon chain

gpsanant: either is fine, i have no opinion for amounts

Sidu28: Ok switched to Gwei for FullWithdrawalRedeemed

ChaoticWalrus (approved):

ChaoticWalrus:
> This seems really close. Summarizing a few open threads: looks like this is open [#127 (comment)](https://github.com/Layr-Labs/eigenlayer-contracts/pull/127#discussion_r1332290356) I’d want to see some tracking for this, at least [#127 (comment)](https://github.com/Layr-Labs/eigenlayer-contracts/pull/127#discussion_r1332291243) I’m OK leaving this as basically a note to myself [#127 (comment)](https://github.com/Layr-Labs/eigenlayer-contracts/pull/127#discussion_r1333423866)
> 
> There's other work happening in a separate branch, see #185 -- I'm cool merging this PR first, just want to spell it out clearly.

I'm approving this for merge now, after getting assurances offline that these open issues will be tracked and addressed promptly.

ChaoticWalrus:
I've also been assured that @gpsanant gave offline approval

src/contracts/pods/EigenPod.sol
Sidu28: Sorry I missed responding to this comment

I think the idea was that someone can deploy a pod, point their validator to it and then not be forced immediately into having to restake (which comes with proving, withdrawals both partial and full, etc.).,  When they're ready to turn it on, they can.

@gpsanant thoughts on this?

gpsanant: not worth it for complexity imo, let them decide when they create the pod. agree w @ChaoticWalrus.

Sidu28: Tracking this issue here @ChaoticWalrus : https://github.com/Layr-Labs/eignlayr-contracts/issues/965

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: This check (or at least a similar one) was actually the source of a bug, and has since been removed

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: check was added (along with a Prover rule) in a subsequent PR https://github.com/Layr-Labs/eigenlayer-contracts/pull/235

PR #129, opened 2023-08-18
Experimental reduce strategy manager bytecode size

ChaoticWalrus:
PR to reduce StrategyManager contract size.


bxmmm1:
@ChaoticWalrus 

Why don't you use Custom errors?
`In this comparison, one can see that custom errors decrease both deploy and runtime gas costs. `
Source: https://soliditylang.org/blog/2021/04/21/custom-errors/

ChaoticWalrus:
> @ChaoticWalrus
> 
> Why don't you use Custom errors? `In this comparison, one can see that custom errors decrease both deploy and runtime gas costs. ` Source: https://soliditylang.org/blog/2021/04/21/custom-errors/

Thanks for the suggestion. This PR was fairly thrown-together as a demo of some different potential measures. Personally, I view custom errors as having 2 main downsides:

1. I think the syntax is worse. I find it easier to be confident that things are correct when checking `require(something that should be true, revert message)` rather than `if (negation of something that should be true) {custom error}`. I realize this is subjective, just my 2 cents.
2. custom errors do not always surface as well. text errors have the added benefit of a clear and easily identifiable error message for any and all users

ChaoticWalrus:
this was solved indirectly, by splitting some functionality into the (more appropriate location of the) EigenPodManager.
closing this PR as not planned.

PR #131, opened 2023-08-24
Update DelegationManager Docs

stevennevins:


ChaoticWalrus (approved):
This looks great! Thank you!
Presumably we should duplicate the NatSpec changes in the `IDelegationManager.sol` interface as well. If you're happy with these changes then I'd do that + merge away.

PR #132, opened 2023-08-25
Add Build Size Check Step to CI

stevennevins:
@ChaoticWalrus PR'd to your reduce build size branch because adding this check to the CI will help us from running into this issue in the future

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,9 @@ jobs:
       - name: Install forge dependencies
         run: forge install
 
+      - name: Run build and check contract sizes
+        run: forge build --sizes
+
       - name: Run tests
         run: forge test -vvv
         env:


ChaoticWalrus:
LGTM, but will this actually fail if the contracts are too large? I thought it just output the table, like in this run https://github.com/Layr-Labs/eigenlayer-contracts/actions/runs/5979710165/job/16224359503?pr=132

stevennevins:
> LGTM, but will this actually fail if the contracts are too large? I thought it just output the table, like in this run https://github.com/Layr-Labs/eigenlayer-contracts/actions/runs/5979710165/job/16224359503?pr=132

It will fail the CI if any are over the 24kb limit. The behavior is a bit weird because it doesn’t explicitly throw an error

ChaoticWalrus:
> It will fail the CI if any are over the 24kb limit. The behavior is a bit weird because it doesn’t explicitly throw an error

Huh OK, agreed that's odd but if it works it works I guess. Will merge away.

PR #134, opened 2023-08-29
Fix existing Certora specs

ChaoticWalrus:
PR should make property/rule failures fail the CI, as well as fixing existing failures.


ChaoticWalrus:
CI is now passing, going to merge this. 
It will look like our CI is taking way, way longer than before, but this is how long these runs were always taking, it’s just tha the GH actions will now wait for the Certora Prover to run and hand back the results and then fail if a property fails in the Prover, rather than just passing as soon as the Prover starts running.

PR #135, opened 2023-08-29
EigenPod Delegation Changes

Sidu28:


src/contracts/pods/EigenPodManager.sol
```diff
@@ -46,15 +55,32 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
     /// @notice EigenLayer's Slasher contract
     ISlasher public immutable slasher;
 
     /// @notice Oracle contract that provides updates to the beacon chain's state
     IBeaconChainOracle public beaconChainOracle;
+
+    /// @notice Canonical beacon chain ETH strategy
+    IStrategy public constant beaconChainETHStrategy = IStrategy(0xbeaC0eeEeeeeEEeEeEEEEeeEEeEeeeEeeEEBEaC0);
+
+    IStrategy[] public beaconChainETHStrategyList;
```
ChaoticWalrus, line 70: ? what is this ?

ChaoticWalrus:
@Sidu28 OK, so giant question which currently remains unanswered in this implementation: what happens when a staker's beacon chain ETH shares are adjusted (especially _reduced_) after they have queued a withdrawal, e.g. as a result of a call to `EigenPod.verifyBalanceUpdate`?

I've tried to address this question myself in multiple docs already; happy to discuss whenever you want, but this can't just be left unaddressed; it is a glaring, massive design hole in this PR right now

src/contracts/core/DelegationManager.sol
```diff
@@ -181,12 +181,19 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
      */
-    function forceUndelegation(address staker) external returns (bytes32) {
+    function forceUndelegation(address staker) external returns (bytes32, bytes32) {
         address operator = delegatedTo[staker];
         require(staker != operator, "DelegationManager.forceUndelegation: operators cannot be force-undelegated");
         require(msg.sender == operator || msg.sender == _operatorDetails[operator].delegationApprover,
             "DelegationManager.forceUndelegation: caller must be operator or their delegationApprover");
-        return strategyManager.forceTotalWithdrawal(staker);
+        
+        bytes32[] memory queuedWithdrawals = new bytes32[](2);
```
ChaoticWalrus, line 190: seems like you meant for this to be the return values but aren't using it anymore? would recommend deleting

src/contracts/core/DelegationManager.sol
```diff
@@ -293,6 +300,24 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+            for (uint256 i = 0; i < numStrategies; i++) {
+                if(i == numStrategies - 1) {
+                    newStrategies[i] = beaconChainETHStrategy;
+                    newShares[i] = beaconChainETHShares;
+                } else {
+                    newStrategies[i] = strategies[i];
+                    newShares[i] = shares[i];
+                }
+            }
+        }
```
ChaoticWalrus, line 319: these memory operations seem quite inefficient -- couldn't we just have a single step outside of the other loop to add these shares?

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -75,5 +80,5 @@ abstract contract DelegationManagerStorage is IDelegationManager {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[44] private __gap;
+    uint256[43] private __gap;
```
ChaoticWalrus, line 83: immutable variables don't occupy storage slots.
I only see a single immutable variable getting added (none removed, no others added), so it seems like this change would ruin the storage gap and potentially corrupt storage in an upgrade!

src/contracts/core/StrategyManager.sol
```diff
@@ -612,6 +563,7 @@ contract StrategyManager is
         onlyStrategiesWhitelistedForDeposit(strategy)
         returns (uint256 shares)
     {
+        require(address(strategy) != address(eigenPodManager.beaconChainETHStrategy()), "StrategyManager._depositIntoStrategy: cannot deposit into the canonical beaconChainETHStrategy via the StrategyManager");
```
ChaoticWalrus, line 566: this is unnecessary, not sure why it was added

src/contracts/core/StrategyManager.sol
```diff
@@ -639,6 +591,7 @@ contract StrategyManager is
         internal
         returns (bool)
     {
+        require(address(strategy) != address(eigenPodManager.beaconChainETHStrategy()), "StrategyManager._depositIntoStrategy: cannot deposit into the canonical beaconChainETHStrategy via the StrategyManager");
```
ChaoticWalrus, line 594: this is extra unnecessary; I'm extra extra unsure why this was added

src/contracts/core/StrategyManager.sol
```diff
@@ -725,31 +678,9 @@ contract StrategyManager is
...
-                /**
-                * This decrements the withdrawableRestakedExecutionLayerGwei which is incremented only when a podOwner proves a full withdrawal.
-                * Remember that withdrawableRestakedExecutionLayerGwei tracks the currently withdrawable ETH from the EigenPod.  
-                * By doing this, we ensure that the number of shares in EigenLayer matches the amount of withdrawable ETH in 
-                * the pod plus any ETH still staked on the beacon chain via other validators pointed to the pod. As a result, a validator 
-                * must complete a full withdrawal from the execution layer prior to queuing a withdrawal of 'beacon chain ETH shares' 
-                * via EigenLayer, since otherwise withdrawableRestakedExecutionLayerGwei will be 0.
-                */         
-                eigenPodManager.decrementWithdrawableRestakedExecutionLayerGwei(msg.sender, shares[i]);   
+            require(address(strategies[i]) != address(eigenPodManager.beaconChainETHStrategy()), "StrategyManager._depositIntoStrategy: cannot deposit into the canonical beaconChainETHStrategy via the StrategyManager");
```
ChaoticWalrus, line 682: this is also unnecesssary

src/contracts/core/StrategyManagerStorage.sol
```diff
@@ -80,5 +80,5 @@ abstract contract StrategyManagerStorage is IStrategyManager {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[40] private __gap;
+    uint256[42] private __gap;
```
ChaoticWalrus, line 83: is this fixing a previous error in the storage gap size? not seeing it paired with any other storage changes?

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -168,4 +164,13 @@ interface IEigenPod {
     /// @notice called by the eigenPodManager to increment the withdrawableRestakedExecutionLayerGwei 
     /// in the pod, to reflect a completion of a queued withdrawal as shares
     function incrementWithdrawableRestakedExecutionLayerGwei(uint256 amountWei) external;
+
+    /// @notice called to deposit ETH into the pod
+    function receiveETH() external payable;
```
ChaoticWalrus, line 169: why is this named `receiveETH` instead of just `receive`? the `receive` function is treated specially, like a fallback!

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -13,6 +15,21 @@ import "./IPausable.sol";
...
+     * the data is resubmitted and the hash of the submitted data is computed by `calculateWithdrawalRoot` and checked against the
+     * stored hash in order to confirm the integrity of the submitted data.
+     */
+    struct BeaconChainQueuedWithdrawal {
+        uint256 shares;
+        address podOwner;
+        uint96 nonce;
+        uint32 withdrawalStartBlock;
+        address delegatedAddress;
+        bool alsoWithdraw;
```
ChaoticWalrus, line 30: should document the struct fields too

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -39,34 +56,58 @@ interface IEigenPodManager is IPausable {
...
      */
-    function withdrawRestakedBeaconChainETH(address podOwner, address recipient, uint256 amount) external;
+    function slashShares(address slashedPodOwner, address slashedFundsRecipient, uint256 shareAmount) external;
 
+
+    /**
+     * @notice Completes an existing queuedWithdrawal either by sending the ETH to the recipient or allowing the podOwner to re-delegate it
+     * @param queuedWithdrawal is the queued withdrawal to be completed
+     * @param middlewareTimesIndex is the index in the operator that the staker who triggered the withdrawal was delegated to's middleware times array
+     * @dev Callable only by the podOwner's EigenPod contract.
```
ChaoticWalrus, line 100: seems like potentially some very odd routing that the EigenPod is calling this rather than the user themselves?
I'll look into it more tho

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -86,4 +127,10 @@ interface IEigenPodManager is IPausable {
     function slasher() external view returns(ISlasher);
 
     function hasPod(address podOwner) external view returns (bool);
+
+    /// @notice returns shares of provided podOwner
+    function getBeaconChainETHShares(address podOwner) external returns (uint256);
+
+    /// @notice returns canonical beaconChainETH strategy
```
ChaoticWalrus, line 134: seems important to note that this is "virtual", i.e. this is not any actually existing contract

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -279,11 +282,15 @@ library BeaconChainProofs {
...
-                                                BLOCK_SUMMARY_ROOT_INDEX << (BLOCK_ROOTS_TREE_HEIGHT) | uint256(proofs.blockHeaderRootIndex);
-        }
+                                                uint256(proofs.historicalSummaryIndex) << (1 + (BLOCK_ROOTS_TREE_HEIGHT)) |
+                                                BLOCK_SUMMARY_ROOT_INDEX << (BLOCK_ROOTS_TREE_HEIGHT) | 
+                                                uint256(proofs.blockHeaderRootIndex);
 
-        {
+            require(Merkle.verifyInclusionSha256(proofs.historicalSummaryBlockRootProof, beaconStateRoot, proofs.blockHeaderRoot, historicalBlockHeaderIndex),
+                "BeaconChainProofs.verifyWithdrawalProofs: invalid historical summary proof");
+        } else {
```
ChaoticWalrus, line 293: what happens if I mark `proofs.proveHistoricalRoot` as 'false' but I supply a really old blockHeaderRoot?
where should the process fail?
do we have test coverage for this?

src/contracts/pods/EigenPod.sol
```diff
@@ -605,22 +600,23 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     }
 
         /// @notice payable fallback function that receives ether deposited to the eigenpods contract
-    function receive() external payable {
+    function receiveETH() external payable {
```
ChaoticWalrus, line 603: yeah, def want to revert this change, it should be called `receive` to work as a payable fallback function

src/contracts/pods/EigenPodManager.sol
```diff
@@ -51,17 +60,36 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
     /// @notice Oracle contract that provides updates to the beacon chain's state
     IBeaconChainOracle public beaconChainOracle;
-    
+
     /// @notice Pod owner to deployed EigenPod address
     mapping(address => IEigenPod) public ownerToPod;
 
+    /// @notice Mapping: podOwner => cumulative number of queued withdrawals of beaconchainETH they have ever initiated. only increments (doesn't decrement)
+    mapping(address => uint256) public numWithdrawalsQueued;
+
```
ChaoticWalrus, line 69: inserting this mapping here will mess up the storage layout and explode the system when we upgrade.
it needs to go after the existing storage

src/contracts/pods/EigenPodManager.sol
```diff
@@ -51,17 +60,36 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+    mapping(address => uint256) public podOwnerShares;
+
+
+    /// @notice Mapping: hash of withdrawal inputs, aka 'withdrawalRoot' => whether the withdrawal is pending
+    mapping(bytes32 => bool) public withdrawalRootPending;
+
+    /// @notice EigenLayer's DelegationManager contract
+    IDelegationManager public immutable delegationManager;
+
+    /// @notice Canonical beacon chain ETH strategy
```
ChaoticWalrus, line 87: again seems good to note this is "virtual" / not a real contract

src/contracts/pods/EigenPodManager.sol
```diff
@@ -134,68 +184,142 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+       _updateSharesToReflectBeaconChainETHBalance(podOwner, sharesDelta);
     }
 
     /**
-     * @notice Withdraws ETH from an EigenPod. The ETH must have first been withdrawn from the beacon chain.
-     * @param podOwner The owner of the pod whose balance must be withdrawn.
-     * @param recipient The recipient of the withdrawn ETH.
-     * @param amount The amount of ETH to withdraw.
-     * @dev Callable only by the StrategyManager contract.
+     * @notice Queues a withdrawal for beacon chain ETH shares on behalf of the owner of an EigenPod.
```
ChaoticWalrus, line 219: kinda weird description, isn't the podOwner the caller (i.e. `msg.sender`) here?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -134,68 +184,142 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+     * @param amountWei The amount of ETH to withdraw.
+     * @param undelegateIfPossible If true, the podOwner's shares will be undelegated if they are not currently delegated.
```
ChaoticWalrus, line 221: I'm worried by this description. it is either inaccurate or there is a bug in the system.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -134,68 +184,142 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+        withdrawalRootPending[withdrawalRoot] = false;
+
+        _withdrawRestakedBeaconChainETH(queuedWithdrawal.podOwner, slashedFundsRecipient, queuedWithdrawal.shares);
     }
 
     /**
-     * @notice This function is called to increment withdrawableRestakedExecutionLayerGwei when a validator's withdrawal is completed.
-     * @param amountWei is the amount of ETH in wei to increment withdrawableRestakedExecutionLayerGwei by
+     * @notice forces a withdrawal of the podOwner's beaconChainETHStrategy shares
+     * @param podOwner is the pod owner whose shares are to be removed
```
ChaoticWalrus, line 299: probably want a dev tag here which notes that only the DelegationManager can call this function

src/contracts/pods/EigenPodManager.sol
```diff
@@ -134,68 +184,142 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
      */
-    function incrementWithdrawableRestakedExecutionLayerGwei(address podOwner, uint256 amountWei) 
-        external onlyStrategyManager onlyWhenNotPaused(PAUSED_WITHDRAW_RESTAKED_ETH) 
+    function forceWithdrawal(address podOwner) external 
+        onlyDelegationManager 
+        onlyWhenNotPaused(PAUSED_WITHDRAWALS)
+        onlyNotFrozen(podOwner)
+        nonReentrant
```
ChaoticWalrus, line 305: I'm seeing nonReentrant modifiers on basically every function in this contract?
have you given any thought to which might actually need it?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -134,68 +184,142 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+        returns (bytes32)
     {
-        ownerToPod[podOwner].incrementWithdrawableRestakedExecutionLayerGwei(amountWei);
+        return _queueWithdrawal(podOwner, podOwnerShares[podOwner], true, true);
+    }
+
+    function getBeaconChainETHShares(address podOwner) external returns (uint256) {
+        return podOwnerShares[podOwner];
```
ChaoticWalrus, line 312: this should be marked 'view'.
It's also super unclear to me why you have a public mapping and a differently-named getter function for the public mapping.

src/contracts/core/DelegationManager.sol
Sidu28: yeah idk wtf I was thinking, fixed

src/contracts/core/DelegationManagerStorage.sol
Sidu28: fuck no this was because I added some storage and then deleted and forgot to update the gap uggh.  Good catch

src/contracts/core/StrategyManager.sol
Sidu28: yup

src/contracts/core/StrategyManagerStorage.sol
Sidu28: same thing, added storage, updated the gap and then deleted said storage.  big oof on my part, fixed

src/contracts/interfaces/IEigenPod.sol
Sidu28: ah forgot about the "receive" function, changed.

src/contracts/interfaces/IEigenPodManager.sol
Sidu28: oh I didn't touch the documentation really, thats all autocomplete

src/contracts/libraries/BeaconChainProofs.sol
Sidu28: doesnt exist, all the historical proof stuff is a wip

src/contracts/pods/EigenPodManager.sol
Sidu28: yeah whoops, i thought I placed all the new storage correctly, guess I missed this one

src/contracts/core/StrategyManager.sol
BOOFI-NearlyHeadlessNick: OK so this will revert anyway when the address with no code doesn't return any value, so going to remove it as it's unnecessary

ChaoticWalrus: OK so this function will revert anyway when the address with no code doesn't return any value, so I think I'm just going to delete it.

Sidu28: yeah fair enough, I thought it might be good to throw an informative error but im just overthinking it

src/contracts/pods/EigenPod.sol
ChaoticWalrus: @Sidu28 looks like you just deleted this instead? just trying to figure out what you're trying to do in this PR, frankly.

src/contracts/libraries/BeaconChainProofs.sol
ChaoticWalrus: I added a TODO + note in the code about this for now, at least.

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: resolved this with changes myself

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: resolved this with changes myself

ChaoticWalrus:
@gpsanant I've implemented a ton of these changes myself, so I would appreciate your review on this PR.
For a bit of context, I do expect these changes to get reviewed again prior to merging into master, at least.

src/contracts/core/DelegationManager.sol
```diff
@@ -37,9 +37,16 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         _;
     }
 
+    // @notice Simple permission for functions that are only callable by the StrategyManager contract OR by the EigenPodManagerContract
+    modifier onlyStrategyManagerOrEigenPodManager() {
```
gpsanant, line 41: Would prefer this being a mapping. I can see us wanting permissionless strategy managers that can push updates to the delegation manager for the strategies that they "own".

src/contracts/core/DelegationManager.sol
```diff
@@ -293,11 +310,24 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
 
-        // add strategy shares to delegated `operator`'s shares
+
+        //retrieve any beacon chain ETH shares the staker might have
+        uint256 beaconChainETHShares = eigenPodManager.podOwnerShares(staker);
+        IStrategy beaconChainETHStrategy = eigenPodManager.beaconChainETHStrategy();
+
         uint256 stratsLength = strategies.length;
+
+        if (beaconChainETHShares > 0){
```
gpsanant, line 320: why involve the loop with this? just
```soldity
if (beaconChainETHShares > 0){
    operatorShares[operator][beaconChainETHStrategy] += beaconChainETHShares;
}
```

src/contracts/core/DelegationManager.sol
```diff
@@ -214,7 +231,7 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
      */
     function decreaseDelegatedShares(address staker, IStrategy[] calldata strategies, uint256[] calldata shares)
         external
-        onlyStrategyManager
+        onlyStrategyManagerOrEigenPodManager
```
gpsanant, line 234: I would prefer adding a check if any of the strategies is beacon chain eth, then the sender must be eigenpodmanager

src/contracts/core/StrategyManager.sol
```diff
@@ -999,6 +889,11 @@ contract StrategyManager is
         );
     }
 
+    // @notice Returns 'true' if `staker` can undelegate and false otherwise
+    function stakerCanUndelegate(address staker) public view returns (bool) {
```
gpsanant, line 893: I would prefer addresses only being able to be pod owners or only being able to be liquid restakers. We could check on deposit. I feel like this makes the state of the system smaller and easier to reason about for reasonable tradeoff?

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -278,12 +281,18 @@ library BeaconChainProofs {
         require(proofs.timestampProof.length == 32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT),
             "BeaconChainProofs.verifyWithdrawalProofs: timestampProof has incorrect length");
 
+        // TODO: make historical proofs actually work. this is not compelete and does not function correctly right now
```
gpsanant, line 284: are we fine merging with this?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -134,68 +199,131 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
      *         balance of a validator is lower than how much stake they have committed to EigenLayer
      * @param podOwner is the pod owner whose balance is being updated.
-     * @param beaconChainETHStrategyIndex is the index of the beaconChainETHStrategy in case it must be removed,
      * @param sharesDelta is the change in podOwner's beaconChainETHStrategy shares
      * @dev Callable only by the podOwner's EigenPod contract.
      */
-    function recordBeaconChainETHBalanceUpdate(address podOwner, uint256 beaconChainETHStrategyIndex, int256 sharesDelta) external onlyEigenPod(podOwner) {
-        strategyManager.recordBeaconChainETHBalanceUpdate(podOwner, beaconChainETHStrategyIndex, sharesDelta);
+    function recordBeaconChainETHBalanceUpdate(address podOwner, int256 sharesDelta) external onlyEigenPod(podOwner) nonReentrant {
+       _updateSharesToReflectBeaconChainETHBalance(podOwner, sharesDelta);
```
gpsanant, line 224: can we name these `X` and `_X`?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -134,68 +199,131 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+            _enterUndelegationLimbo(podOwner);
+        }
     }
 
+    /**
+     * Sets the maximum number of pods that can be deployed
+     * @param newMaxPods The new maximum number of pods that can be deployed
+     * @dev Callable by the unpauser of this contract
+     */
+    function setMaxPods(uint256 newMaxPods) external onlyUnpauser {
```
gpsanant, line 324: can we put this near the top of the external functions?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -206,7 +334,169 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+        returns (bytes32)
+    {
+        require(amountWei > 0, "EigenPodManager._queueWithdrawal: amount must be greater than zero");
+
+        require(amountWei % GWEI_TO_WEI == 0,
+                    "EigenPodManager._queueWithdrawal: cannot queue a withdrawal of Beacon Chain ETH for an non-whole amount of gwei");
+
+        require(!isInUndelegationLimbo(podOwner),
+                    "EigenPodManager._queueWithdrawal: cannot queue a withdrawal when in undelegation limbo");
+
```
gpsanant, line 413: space

src/contracts/pods/EigenPodManager.sol
```diff
@@ -243,6 +533,119 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+            _addShares(podOwner, uint256(sharesDelta));
+        }      
+    }
+
+    /**
+     * @notice This function is called to decrement withdrawableRestakedExecutionLayerGwei when a validator queues a withdrawal.
+     * @param amountWei is the amount of ETH in wei to decrement withdrawableRestakedExecutionLayerGwei by
+     */
+    function _decrementWithdrawableRestakedExecutionLayerGwei(address podOwner, uint256 amountWei) 
+        internal onlyWhenNotPaused(PAUSED_WITHDRAW_RESTAKED_ETH)
```
gpsanant, line 590: why can't the pausing flag be `PAUSED_WITHDRAWALS`?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -206,7 +334,169 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+            "EigenPodManager._completeQueuedWithdrawal: withdrawal is not pending"
+        );
+
+        // verify that the withdrawal is completable
+        require(
+            slasher.canWithdraw(queuedWithdrawal.delegatedAddress, queuedWithdrawal.withdrawalStartBlock, middlewareTimesIndex),
+            "EigenPodManager._completeQueuedWithdrawal: shares pending withdrawal are still slashable"
+        );
+
+        /* TODO: decide definitively if minimum lag is enforced here or via use of the DelayedWithdrawalRouter
```
gpsanant, line 480: here

src/contracts/pods/EigenPodManager.sol
```diff
@@ -206,7 +334,169 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+     * @param withdrawFundsFromEigenLayer If marked as 'true', then the caller's beacon chain ETH shares will be withdrawn from EigenLayer, as they are when
+     * completing a withdrawal. If marked as 'false', then the caller's shares are simply returned to EigenLayer's delegation system.
+     */
+    function exitUndelegationLimbo(uint256 middlewareTimesIndex, bool withdrawFundsFromEigenLayer)
+        external
+        onlyWhenNotPaused(PAUSED_WITHDRAWALS)
+        onlyNotFrozen(msg.sender)
+        nonReentrant
+    {
+        require(isInUndelegationLimbo(msg.sender),
```
gpsanant, line 365: why not make this an internal function?

gpsanant:
@ChaoticWalrus @Sidu28 There are some preliminary changes I've suggested. I plan to read the EigenPod and BeaconChainProofs contract in detail over the next couple days.

src/contracts/pods/EigenPod.sol
```diff
@@ -86,8 +86,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     /// @notice This is a mapping that tracks a validator's information by their pubkey hash
     mapping(bytes32 => ValidatorInfo) internal _validatorPubkeyHashToInfo;
 
-
-    /// @notice This variable tracks any ETH deposited into this contract via the receive fallback
```
gpsanant, line 90: on line 76, `// STORAGE VARIABLE` should be moved up

src/contracts/pods/EigenPod.sol
```diff
@@ -437,7 +431,6 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         bytes calldata validatorFieldsProof,
```
gpsanant, line 431: in line 402, we are calling this the `latestBlockHeaderRoot`. seems like we should call this `blockHeaderRoot`

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -123,16 +125,13 @@ interface IEigenPod {
      *        Must be within `VERIFY_OVERCOMMITTED_WINDOW_BLOCKS` of the current block.
      * @param validatorIndex is the index of the validator being proven, refer to consensus specs 
      * @param proofs is the proof of the validator's balance and validatorFields in the balance tree and the balanceRoot to prove for
-     * @param beaconChainETHStrategyIndex is the index of the beaconChainETHStrategy for the pod owner for the callback to
```
gpsanant, line 126: in `verifyWithdrawalCredentials`, we do not set 2 fields of the validatorInfo struct?

src/contracts/pods/EigenPod.sol
```diff
@@ -552,7 +548,13 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         }
     }
 
-    function _processPartialWithdrawal(uint64 withdrawalHappenedSlot, uint64 partialWithdrawalAmountGwei, uint40 validatorIndex, bytes32 validatorPubkeyHash, address recipient) internal {
+    function _processPartialWithdrawal(
```
gpsanant, line 551: can we standardize the order of parameters `_processPartialWithdrawal` vs in `_processFullWithdrawal`?

src/contracts/libraries/BeaconChainProofs.sol
ChaoticWalrus: I'm OK merging into `newpodchanges` but I'm unclear on how this was supposed to be organized, really

src/contracts/core/DelegationManager.sol
ChaoticWalrus: fixed and made it work with the 'undelegation limbo' mechanism. good catch, not sure what sidu was thinking on this one https://github.com/Layr-Labs/eigenlayer-contracts/pull/135/commits/47d19c7b81ae980917026d7b3adaa95bd14cde39

src/contracts/libraries/BeaconChainProofs.sol
Sidu28: It was branched off of newpodchanges, which has some bug fixes + the new proofs.  If its alright with you, we merge this into newpodchanges, and then work off that branch.

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -86,4 +144,29 @@ interface IEigenPodManager is IPausable {
     function slasher() external view returns(ISlasher);
 
     function hasPod(address podOwner) external view returns (bool);
+
+    /// @notice returns shares of provided podOwner
+    function podOwnerShares(address podOwner) external returns (uint256);
```
gpsanant, line 149: line 138 has the wrong function name, should be `getBeaconChainBlockRootFromTimestamp(uint64 timestamp)`

src/contracts/core/DelegationManager.sol
ChaoticWalrus: > Would prefer this being a mapping. I can see us wanting permissionless strategy managers that can push updates to the delegation manager for the strategies that they "own".

This would break (rather than just modifying) the invariant we have w/r/t delegation, that someone can only undelegate if they have no delegated shares. for now, this is cheaper than a mapping since both addresses are stored immutably

src/contracts/core/StrategyManager.sol
ChaoticWalrus: > I would prefer addresses only being able to be pod owners or only being able to be liquid restakers. We could check on deposit. I feel like this makes the state of the system smaller and easier to reason about for reasonable tradeoff?

IMO this is a major design decision, not something that should be made off-hand in a GitHub comment.

We would need to, at minimum:

- look at the current state to see if anyone currently falls into both categories, and make a plan on how to handle any such users
- add checks that enforce this behavior

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: yes, fine by me, will make the change.

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: fixed locally, will push

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: this is already paused when `PAUSED_WITHDRAWALS` is set. I'm going to remove this modifier since it is redundant.

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: > here

I agree this feels more "correct" to me. please review this commit and comment yea or nay on it https://github.com/Layr-Labs/eigenlayer-contracts/pull/135/commits/7fa28e892f9e86190e297fe81205dbca25526066

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: I think this grouping makes more sense, with all the can-only-be-called-by-a-single-address functions grouped at the bottom of the external functions https://github.com/Layr-Labs/eigenlayer-contracts/pull/135/commits/4c6168d5cbc1d95c7accc14b283b88606f9d2723

src/contracts/core/DelegationManager.sol
ChaoticWalrus: > I would prefer adding a check if any of the strategies is beacon chain eth, then the sender must be eigenpodmanager

that kinda just feels more finnicky to me? I get the idea, feel like they arrive at the same place but this is just extra checking... let me know if this is something you feel strongly about, because my _slight_ preference is keeping things as-is here

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: didn't have a second usage of it 🤷

src/contracts/pods/EigenPod.sol
ChaoticWalrus: I believe I made the same comment elsewhere. Per @Sidu28 this is literally called the `latestBlockHeaderRoot` in the spec. I think at minimum we need to add some comments to clarify; if you have ideas where, please make a commit!

src/contracts/pods/EigenPod.sol
ChaoticWalrus: yes, will make a commit with this

src/contracts/interfaces/IEigenPodManager.sol
ChaoticWalrus: so this just should be timestamp rather than blocknumber? I recall us talking about this before, just want to confirm.
I thought @Sidu28 made this change already, maybe on a different branch or something?

src/contracts/pods/EigenPod.sol
ChaoticWalrus: https://github.com/Layr-Labs/eigenlayer-contracts/pull/135/commits/2c8b0c20b430307967a96fabf58c362fb9e400f0

ChaoticWalrus:
just brought in changes from master, particularly in order to fix failing CI

ChaoticWalrus:
I marked this as 'ready for review' despite the immense amount of missing test coverage. Since it is a merge into the `newpodchanges` branch, I recommend a looser standard to help keep us out of branch-management-hell; test coverage can be added later

src/contracts/pods/EigenPod.sol
```diff
@@ -437,7 +430,6 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         bytes calldata validatorFieldsProof,
         bytes32[] calldata validatorFields,
         bytes32[] calldata withdrawalFields,
-        uint256 beaconChainETHStrategyIndex,
         uint64 oracleTimestamp
```
gpsanant, line 433: can `_verifyWithdrawalCredentials` please emit a `ValidatorBalanceUpdated` event

src/contracts/pods/EigenPod.sol
```diff
@@ -110,11 +109,10 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     event RestakedBeaconChainETHWithdrawn(address indexed recipient, uint256 amount);
 
     /// @notice Emitted when podOwner enables restaking
-    event restakingActivated(address indexed podOwner);
+    event RestakingActivated(address indexed podOwner);
 
     /// @notice Emitted when ETH is received via the receive fallback
-    event nonBeaconChainETHReceived(uint256 amountReceived);
-    
+    event nonBeaconChainETHReceived(uint256 amountReceived);
```
gpsanant, line 115: capitalize

src/contracts/pods/EigenPodManager.sol
wadealexc: nR modifiers are fairly cheap - can we just use them by default and instead of justifying them, try to think of cases where we don't want them?

src/contracts/core/DelegationManager.sol
gpsanant: We can make the change when we need to, I guess

src/contracts/core/StrategyManager.sol
gpsanant: Agreed. Actually let's keep it this way. I like it less, but we cant stop people from restaking on beacon chain once they've already deposited LSTs.

src/contracts/pods/EigenPod.sol
gpsanant: This was where the proof was broken

src/contracts/core/DelegationManager.sol
```diff
@@ -186,18 +218,32 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+     * 
+     * @dev *If the staker is actively delegated*, then increases the `staker`'s delegated shares in `strategy` by `shares`. Otherwise does nothing.
      * @dev Callable only by the StrategyManager.
      */
     function increaseDelegatedShares(address staker, IStrategy strategy, uint256 shares)
         external
-        onlyStrategyManager
+        onlyStrategyManagerOrEigenPodManager
     {
+        require(msg.sender == address(strategyManager) || msg.sender == address(eigenPodManager),
```
gpsanant, line 243: redundant with modifier

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -13,6 +15,40 @@ import "./IPausable.sol";
...
+
+    /**
+     * @notice Struct used to track a pod owner's "undelegation limbo" status and associated variables.
+     * @dev Undelegation limbo is a mode which a staker can enter into, in which they remove their virtual "beacon chain ETH shares" from EigenLayer's delegation
+     * system but do not necessarily withdraw the associated ETH from EigenLayer itself. This mode allows users who have restaked native ETH a route via
+     * which they can undelegate from an operator without needing to exit any of their validators from the Consensus Layer.
+     */
+    struct UndelegationLimboStatus {
+        // @notice Whether or not the pod owner is in the "undelegation limbo" mode.
+        bool undelegationLimboActive;
```
gpsanant, line 45: the `undelegationLimbo` in the names of these are redundant imo

src/contracts/pods/EigenPod.sol
ChaoticWalrus: > can `_verifyWithdrawalCredentials` please emit a `ValidatorBalanceUpdated` event

It doesn't seem to be as simple as I thought to add this, since the `ValidatorBalanceUpdated` event has a 'slotNumber' field, but we are not checking the slot number as part of verifying withdrawal credentials.
Unsure if there is a bug here, or what the best approach is with the event. If there's no bug in the proof-checking, then I'd ask if we can remove the 'slotNumber' field from the event.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -243,6 +541,120 @@ contract EigenPodManager is Initializable, OwnableUpgradeable, Pausable, IEigenP
...
+        delegationManager.undelegate(staker);
+    }
+
+    /**
+     * @notice Internal function to enter `podOwner` into undelegation limbo
+     * @dev Does nothing if the `podOwner` has no delegated shares (i.e. they are already in undelegation limbo or have no shares)
+     * OR if they are not actively delegated to any operator.
+     */
+    function _enterUndelegationLimbo(address podOwner) internal {
+        if (!stakerHasNoDelegatedShares(podOwner) && delegationManager.isDelegated(podOwner)) {
```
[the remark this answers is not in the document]
ChaoticWalrus: not using a require statement here means the `forceIntoUndelegationLimbo` function will not revert in weird edge cases (where it might, with the require). Open to suggestions, but I specifically made this change on purpose, in this commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/135/commits/f08e02560d726d7abea29685da8671150784e3a8

src/contracts/pods/EigenPod.sol
gpsanant: we need to change these to timestamps everwhere

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -13,6 +15,40 @@ import "./IPausable.sol";
  */
 
 interface IEigenPodManager is IPausable {
+    /**
+     * Struct type used to specify an existing queued withdrawal. Rather than storing the entire struct, only a hash is stored.
+     * In functions that operate on existing queued withdrawals -- e.g. `startQueuedWithdrawalWaitingPeriod` or `completeQueuedWithdrawal`,
+     * the data is resubmitted and the hash of the submitted data is computed by `calculateWithdrawalRoot` and checked against the
+     * stored hash in order to confirm the integrity of the submitted data.
+     */
+    struct BeaconChainQueuedWithdrawal {
```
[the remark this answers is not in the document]
ChaoticWalrus: Agreed, will add.

src/contracts/interfaces/IEigenPodManager.sol
ChaoticWalrus: agreed, holdover from before I combined into a struct, will revise names

src/contracts/interfaces/IEigenPodManager.sol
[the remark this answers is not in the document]
ChaoticWalrus: https://github.com/Layr-Labs/eigenlayer-contracts/pull/135/commits/b16368ba48b11de89bc5291ff0b644a9d1fe1019

src/contracts/interfaces/IEigenPodManager.sol
ChaoticWalrus: https://github.com/Layr-Labs/eigenlayer-contracts/pull/135/commits/b16368ba48b11de89bc5291ff0b644a9d1fe1019

src/contracts/interfaces/IEigenPodManager.sol
ChaoticWalrus: looks like he just forgot to update _some_ of the usage. adding a commit to fix this, will push it shortly.

src/contracts/interfaces/IEigenPod.sol
ChaoticWalrus: I see 'validatorIndex', 'restakedBalanceGwei', and 'status' getting updated, but _not_ 'mostRecentBalanceUpdateSlot' -- this is just one field, but kind of awkward / doesn't seem desired.
@Sidu28 can you confirm if bug?

src/contracts/pods/EigenPod.sol
ChaoticWalrus: > This was where the proof was broken

please expand on this idea / I guess file an issue for @Sidu28 for this

gpsanant (approved):
I'm approving the PR in its current state, given we've agreed that a PR addressing the following will be opened as soon as possible:
- Outstanding PR comments in EigenPods
- Full withdrawal proof needs to be fixed
- Switch to use timestamps instead of slots in storage and events
- Fix order of functions in EigenPods

Nice work!

Issue #139, opened 2023-09-05
Task: move middleware contracts into a separate repo


samlaf:
## Description

Would be best to have the middleware contracts in a separate repo. They are currently in the [avs-unstable](https://github.com/Layr-Labs/eigenlayer-contracts/tree/avs-unstable/src/contracts/middleware) branch. I had to ask Jeff today to merge master into avs-unstable since it had fallen behind by 300+ commits and had a few merge conflicts.

The dependencies with eigensdk and incredible-squaring currently look like:
![IMG_5C022598C8DE-1 2](https://github.com/Layr-Labs/eigenlayer-contracts/assets/9342524/e669962e-fe88-4677-a7cd-a7f9fc017f8e)

the deploy script for eigenlayer-core contracts is in master (eigensdk uses the contract addresses stored in the output config file in that branch), and incredible-squaring deploys the middleware contracts (registry contracts are stored in the avs-unstable and avs-unstable-require-opt-into-slashing branches). It would be best if incredible-squaring was pointing to another repo with the contracts, instead of a branch, the middleware contracts' dependency on the core contracts would be more obvious (that new repo would branch to a specific commit of eigenlayer-contracts).

### Action Items
- [ ] Move middleware contracts into another repo (maybe named eigenlayer-avs-contracts?)
- [ ] Move blspubkey compendium to core contracts directory (it should now be shared by all avs teams)
- [ ] Make sure to also move tests and deployment scripts

ChaoticWalrus:
Agreed this is something we should do.
Currently development is spread across several branches which would make this fairly hellish to accomplish; we anticipate being able to get this done after closing out the major development branches.
This should align well with plans to improve our "git hygiene" with better branch management, and splitting the repos may also help with differentiating the docs for AVS developers vs. the docs for developers who want to integrate with EigenLayer "core".

PR #141, opened 2023-09-06
update m2_deploy_from_scratch script to be able to deploy from different chains

samlaf:
Updated the script. Merging to master, then will cherrypick to `experimental-reduce-strategy-manager-bytecode-size` branch since that's the only branch that can currently be deployed (strategymanager contract is too big on master).

I've used this script to deploy eigenlayer core contracts for incredible-squaring, and plan to keep doing this for updates until we find a better strategy (aka master is deployable probably)

EDIT: one more thing that I haven't been able to do. Currently the blspubkey compendium that is being deployed as part of this script is outdated. I tried copying the one from `avs-unstable` but that caused a bunch of downstream issues. Perhaps we should just delete the middleware folder from master since its going to be in a separate repo after #139? Also blspubkeycompendium should be moved to core dir I believe.


ChaoticWalrus (approved):
changes look fine to me. will sync offline on the issues you've raised

ChaoticWalrus:
adding a note: I'm seeing a couple CI failures but _these are flaky tests_ -- the failures have nothing to do with the changes you've submitted, so shouldn't be blocking for merging.

PR #142, opened 2023-09-06
Propagate changes

gpsanant:
- added ejector
- added weightOfOperatorForQuorumView
- allow for operators that are already registered to register for new quorums


src/contracts/middleware/VoteWeigherBase.sol
```diff
@@ -59,7 +59,7 @@ contract VoteWeigherBase is VoteWeigherBaseStorage {
      * @notice This function computes the total weight of the @param operator in the quorum @param quorumNumber.
      * @dev reverts in the case that `quorumNumber` is greater than or equal to `quorumCount`
```
stevennevins, line 60: This comment is no longer accurate

What's the reason for essentially duplicating the function without a modifier?

gpsanant: oops, my mistake. good catch! b9071182

stevennevins:
@gpsanant I also wanted to ask why the diff is so large? is this also merging updates from avs-unstable that aren't in mutliquorums yet?

gpsanant:
> @gpsanant I also wanted to ask why the diff is so large? is this also merging updates from avs-unstable that aren't in mutliquorums yet?

yep i believe it's also merging master into mutliquorums

src/contracts/interfaces/IBLSRegistryCoordinatorWithIndices.sol
```diff
@@ -27,12 +27,13 @@ interface IBLSRegistryCoordinatorWithIndices is ISignatureUtils, IRegistryCoordi
...
 
     /**
-     * @notice Data structure for the parameters needed to kick an operator from a quorum, used during registration churn.
+     * @notice Data structure for the parameters needed to kick an operator from a quorum with number `quorumNumber`, used during registration churn.
      * Specifically the `operator` is the address of the operator to kick, `pubkey` is the BLS public key of the operator,
      * `operatorIdsToSwap` is the list of operatorIds to swap with the operator being kicked in the indexRegistry,
      * and `globalOperatorListIndex` is the index of the operator in the global operator list in the indexRegistry.
      */
     struct OperatorKickParam {
+        uint8 quorumNumber;
```
ChaoticWalrus, line 34: why is this a singlet but `operatorIdsToSwap` is an array?

ChaoticWalrus: ahh, I see it was eliminated in a later commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/142/commits/af36906f1f085006984375572a5b711f804a7331

src/contracts/interfaces/IBLSRegistryCoordinatorWithIndices.sol
```diff
@@ -42,6 +42,8 @@ interface IBLSRegistryCoordinatorWithIndices is ISignatureUtils, IRegistryCoordi
 
     event ChurnApproverUpdated(address churnApprover);
 
+    event EjectorUpdated(address ejector);
```
ChaoticWalrus, line 45: AFAIK it's more standard to emit both past address and the new one, but I think this is OK

src/contracts/interfaces/IBLSRegistryCoordinatorWithIndices.sol
```diff
@@ -50,4 +52,18 @@ interface IBLSRegistryCoordinatorWithIndices is ISignatureUtils, IRegistryCoordi
     function blsPubkeyRegistry() external view returns (IBLSPubkeyRegistry);
     /// @notice the Index Registry contract that will keep track of operators' indexes
     function indexRegistry() external view returns (IIndexRegistry);
+
+    /**
+     * @notice Ejects the provided operator from the provided quorums from the AVS
+     * @param operator is the operator to eject
+     * @param quorumNumbers are the quorum numbers to eject the operator from
+     * @param pubkey is the BLS public key of the operator
+     * @param operatorIdsToSwap is the list of the operator ids tho swap the index of the operator with in each
```
ChaoticWalrus, line 61: incomplete sentence(?) -- feels like this just trails off without completing the thought
also typo: `tho` => `to`

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
```diff
@@ -665,6 +682,54 @@ contract BLSRegistryCoordinatorWithIndicesUnit is MockAVSDeployer {
         registryCoordinator.registerOperatorWithCoordinator(quorumNumbers, operatorToRegisterPubKey, defaultSocket, operatorKickParams, signatureWithSaltAndExpiry);
     }
 
+    function testEjectOperatorFromCoordinator_Valid() public {
```
ChaoticWalrus, line 685: might be nice to have a version of this that only ejects the operator from some quorums?
also would really appreciate some comments on these tests!!

ChaoticWalrus:
@gpsanant based on your comment about [d1fa577](https://github.com/Layr-Labs/eigenlayer-contracts/pull/142/commits/d1fa5779fb18355790d6e8527709f856c698bc1d) on being the important parts to review:
In general this LGTM, have dropped a few nits as comments, _however_, looks like you also merged in commits from the `experimental-reduce-strategy-manager-bytecode-size` branch (https://github.com/Layr-Labs/eigenlayer-contracts/tree/experimental-reduce-strategy-manager-bytecode-size) -- was this on purpose? I didn't think we were planning on merging those changes.
The aforementioned merge was made in this commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/142/commits/0fd74e96bb6d12d507265843484efd666b27af5c -- I would recommend reverting this commit.

src/contracts/interfaces/IBLSRegistryCoordinatorWithIndices.sol
gpsanant: c5501a83

src/contracts/interfaces/IBLSRegistryCoordinatorWithIndices.sol
gpsanant: c1740bc8 this work with ya?

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
gpsanant: c4f9e92a done :)

gpsanant:
@ChaoticWalrus pls rereview
One new large change: *"allow for operators that are already registered to register for new quorums"*

Also, I would want to keep the strategy manager change merged in since we are actively deploying these contracts for EigenDA. Since the SM changes are orthagonal, when the SM is smaller, i'll just paste it in. does that work?

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -344,8 +388,15 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
     }
     
     function _setChurnApprover(address newChurnApprover) internal {
+        address prevChurnApprover = churnApprover;
         churnApprover = newChurnApprover;
-        emit ChurnApproverUpdated(newChurnApprover);
+        emit ChurnApproverUpdated(prevChurnApprover, newChurnApprover);
```
ChaoticWalrus, line 393: typically you just emit the event first rather than storing things in memory.
so it would look like:
```
emit ChurnApproverUpdated(churnApprover, newChurnApprover);
churnApprover = newChurnApprover;
```

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
```diff
@@ -80,10 +82,25 @@ contract BLSRegistryCoordinatorWithIndicesUnit is MockAVSDeployer {
...
+        registryCoordinator.setEjector(newEjector);
+    }
+
+    function testSetEjector_Valid() public {
+        address newEjector = address(uint160(uint256(keccak256("newEjector"))));
+        cheats.prank(serviceManagerOwner);
+        cheats.expectEmit(true, true, true, true, address(registryCoordinator));
+        emit EjectorUpdated(ejector, newEjector);
+        registryCoordinator.setEjector(newEjector);
+    }
```
ChaoticWalrus, line 102: you should check that the ejector gets updated correctly after this call

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
```diff
@@ -665,6 +749,104 @@ contract BLSRegistryCoordinatorWithIndicesUnit is MockAVSDeployer {
...
+            keccak256(abi.encode(IRegistryCoordinator.Operator({
+                operatorId: defaultOperatorId,
+                status: IRegistryCoordinator.OperatorStatus.REGISTERED
+            })))
+        );
+        // make sure the operator is not in any quorums
+        assertEq(
+            registryCoordinator.getCurrentQuorumBitmapByOperatorId(defaultOperatorId), 
+            BitmapUtils.orderedBytesArrayToBitmap(quorumNumbers) & ~BitmapUtils.orderedBytesArrayToBitmap(quorumNumbersToEject) // quorumsRegisteredFor & ~quorumsEjectedFrom
+        );
```
ChaoticWalrus, line 830: wait so it's possible for an operator to be registered but not in any quorums now (as a result of ejection)
shouldn't this be checking that the operator is *not* registered?

feels like something is off here :/

ChaoticWalrus (changes requested):
few nits, one question on tests, looks pretty good from a fairly quick review

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
gpsanant: no, this only works since `quorumNumbersToEject` is a strict subset of `quorumNumbers`. if they were the same, the operator would be deregistered.

src/contracts/middleware/BLSRegistryCoordinatorWithIndices.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/142/commits/7eef90501cca5c18d0bd82e76d092013992e4f31 events after ftw, but ye u right

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/142/commits/7eef90501cca5c18d0bd82e76d092013992e4f31

src/test/unit/BLSRegistryCoordinatorWithIndicesUnit.t.sol
ChaoticWalrus: OK gotcha, makes sense

ChaoticWalrus (approved):

PR #150, opened 2023-09-08
ServiceManagerBase implementation + minor change to IServiceManager interface

samlaf:
We add a ServiceManagerBase abstract class that implements all of the "trivial" functions that just proxy to the slasher, and only leave for implementation:
- latestServeUntilBlock
- freezeOperator
- taskNumber

This way, AVS teams can derive their service manager from this base implementation without having to write a lot of redundant proxy contract calls.

This ServiceManagerBase depends on our registry contracts (registryCoordinator and stakeRegistry) so any team that uses their own registry will still need to implement their service manager from scratch. We could have an even more barebone ServiceManagerBase contract, but not sure how useful it would be.

There's one more change included in this PR, which is that I removed the `latestServeUntil` parameter from all the recordStakeUpdate calls in the IServiceManager interface, since all call sites were doing something like
```
uint32 latestServeUntilBlock = serviceManager.latestServeUntilBlock();
serviceManager.recordLastStakeUpdateAndRevokeSlashingAbility(operator, latestServeUntilBlock);
```
and I figured instead of doing these multiple separate calls to serviceManager, might as well instead of the serviceManager call its own `latestServeUntilBlock()` implementation. I left `latestServeUntilBlock` in the interface to make it explicit that the servicemanager should implement this public function, but I think it only makes sense to be called internally, so wondering whether to remove it entirely from the interface.

### DIFF
--- a/src/contracts/middleware/ServiceManagerBase.sol
+++ b/src/contracts/middleware/ServiceManagerBase.sol
@@ -1,0 +1,139 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.9;
+
+import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
+import "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
+import "./BLSSignatureChecker.sol";
+import "../permissions/Pausable.sol";
+
+import "../interfaces/IDelegationManager.sol";
+import "../interfaces/IServiceManager.sol";
+import "../interfaces/IStrategyManager.sol";
+
+/**
+ * @title Base implementation of `IServiceManager` interface, designed to be inherited from by more complex ServiceManagers.
+ * @author Layr Labs, Inc.
+ * @notice This contract is used for:
+ * - proxying calls to the Slasher contract
+ * - implementing the two most important functionalities of a ServiceManager:
+ *  - freezing operators as the result of various "challenges"
+ *  - defining the latestServeUntilBlock which is used by the Slasher to determine whether a withdrawal can be completed
+ */
+abstract contract ServiceManagerBase is
+    IServiceManager,
+    Initializable,
+    OwnableUpgradeable,
+    BLSSignatureChecker,
+    Pausable
+{
+    /// @notice Called in the event of challenge resolution, in order to forward a call to the Slasher, which 'freezes' the `operator`.
+    /// @dev This function should contain slashing logic, to make sure operators are not needlessly being slashed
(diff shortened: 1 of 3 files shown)


src/contracts/middleware/ServiceManagerBase.sol
```diff
@@ -0,0 +1,145 @@
...
+        uint32 updateBlock,
+        uint256 prevElement
+    ) external onlyRegistryCoordinatorOrStakeRegistry {
+        slasher.recordStakeUpdate(
+            operator,
+            updateBlock,
+            latestServeUntilBlock(),
+            prevElement
+        );
+    }
```
ChaoticWalrus, line 119: we might want to mark these functions as 'virtual' (so they can be overridden) and/or add extensible "hooks" to them, but this seems like a fine start

ChaoticWalrus:
Seems fine to me, in general.
Looks like there's a couple merge conflicts which need to be sorted out.
Requesting that @gpsanant take a quick look and resolve these merge conflicts.

gpsanant:
I don't think this change makes sense for AVS devs. I see the service manager as a proxy for the AVS to talk to the slasher and other EigenLayer contracts. What if we want more inputs into the `latestServeUntilBlock` calculation?

samlaf:
> I don't think this change makes sense for AVS devs. I see the service manager as a proxy for the AVS to talk to the slasher and other EigenLayer contracts. What if we want more inputs into the `latestServeUntilBlock` calculation?

It's a Base contract, so they don't *have* to use it. It's just meant as an example for teams to use. We're using it for Incredible-Squaring for eg. We can generalize it or add different ServiceManager base/template contracts like this as we expand and get a broader understanding of the types of AVS that are building on AVS.

src/contracts/middleware/ServiceManagerBase.sol
samlaf: Made them virtual.

samlaf:
@ChaoticWalrus @gpsanant reverted the IServiceManager changes to keep the same interface. Just removed the word "permissioned" in the comments since the interface doesn't enforce that permission, but can add it back if you guys think its better to keep it.

Also changed the merging branch from multiquorums to avs-unstable.

ChaoticWalrus:
> @ChaoticWalrus @gpsanant reverted the IServiceManager changes to keep the same interface. Just removed the word "permissioned" in the comments since the interface doesn't enforce that permission, but can add it back if you guys think its better to keep it.
> 
> Also changed the merging branch from multiquorums to avs-unstable.

My bad, I missed this update -- looks good to me, merge if you want.

PR #151, opened 2023-09-12
merge master into avs-unstable

samlaf:
I need the updated test deploy_from_scratch script to deploy a new image for incredible-squaring.

This merges my changes, but also 2 other commits. Is that OK @ChaoticWalrus ?


samlaf:
Merging with permission from Jeffrey.

PR #164, opened 2023-09-14
BLSSignatureChecker Updates

0x0aa0:
[Implement: BLSSignatureChecker changes #949](https://github.com/Layr-Labs/eignlayr-contracts/issues/949?notification_referrer_id=NT_kwDOA1tA77M3NjgxODk2NDk1OjU2MzEyMDQ3&notifications_query=reason%3Aassign)

- check sorting of nonsigners
- sort nonsigners in tests
- update scalar_mul_tiny for use

### DIFF
--- a/src/contracts/libraries/BN254.sol
+++ b/src/contracts/libraries/BN254.sol
@@ -139,12 +139,20 @@ library BN254 {
      */ 
     function scalar_mul_tiny(BN254.G1Point memory p, uint16 s) internal view returns (BN254.G1Point memory) {
         require(s < 2**9, "scalar-too-large");
+
+        uint256 n = 1;
+        uint256 m = 1;
+        while(s > m){
+            m <<= 1;
+            ++n;
+        }
+
         // the accumulated product to return
         BN254.G1Point memory acc = BN254.G1Point(0, 0);
         // the 2^n*p to add to the accumulated product in each iteration
         BN254.G1Point memory p2n = p;
         // loop through each bit of s
-        for (uint8 i = 0; i < 9; i++) {
+        for (uint8 i = 0; i < n; i++) {
             // if the bit is 1, add the 2^n*p to the accumulated product
             if (s >> i & 1 == 1) {
                 acc = plus(acc, p2n);
--- a/src/contracts/libraries/BN254.sol
+++ b/src/contracts/libraries/BN254.sol
@@ -140,36 +140,34 @@ library BN254 {
     function scalar_mul_tiny(BN254.G1Point memory p, uint16 s) internal view returns (BN254.G1Point memory) {
         require(s < 2**9, "scalar-too-large");
 
+        // if s is 1 return p
+        if(s == 1) {
+            return p;
+        }
+
         // the accumulated product to return
         BN254.G1Point memory acc = BN254.G1Point(0, 0);
+        // the 2^n*p to add to the accumulated product in each iteration
+        BN254.G1Point memory p2n = p;
         // value of most signifigant bit
         uint16 m = 1;
+        // index of most signifigant bit
+        uint8 i = 0;
 
-        //check if s is greater than 1
-        if(s > m){
-            // the 2^n*p to add to the accumulated product in each iteration
-            BN254.G1Point memory p2n = p;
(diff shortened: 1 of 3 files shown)


src/contracts/libraries/BN254.sol
```diff
@@ -139,12 +139,20 @@ library BN254 {
      */ 
     function scalar_mul_tiny(BN254.G1Point memory p, uint16 s) internal view returns (BN254.G1Point memory) {
         require(s < 2**9, "scalar-too-large");
+
+        uint256 n = 1;
+        uint256 m = 1;
+        while(s > m){
```
gpsanant, line 145: this is close to what i was thinking, but couldn't we fit this inside the main loop? or would that make the gas larger for some reason?

also could we have a special case for `s == 1`? to avoid the doubling at the end?

src/contracts/middleware/BLSSignatureChecker.sol
```diff
@@ -116,9 +116,11 @@ contract BLSSignatureChecker {
 
                 for (uint i = 0; i < nonSignerStakesAndSignature.nonSignerPubkeys.length; i++) {
                     nonSignerPubkeyHashes[i] = nonSignerStakesAndSignature.nonSignerPubkeys[i].hashG1Point();
-                    // if (i != 0) {
-                    //     require(uint256(nonSignerPubkeyHashes[i]) > uint256(nonSignerPubkeyHashes[i - 1]), "BLSSignatureChecker.checkSignatures: nonSignerPubkeys not sorted");
-                    // }
+                    
+                    if (i != 0) {
```
gpsanant, line 120: add comment about check for lack of duplicates

gpsanant:
Could you also import this branch on EigenDA and get some nonsigner gas costs to tell what the benefit we are getting is?

In addition, could we get the gas costs of scalar mul tiny for different values of `s`? i think it may make sense to switch between `scalar_mul_tiny` and `scalar_mul` depending on what `s`. Likely an optimization that doesn't need to be done before testnet though. Maybe create a ticket for this last bit in the backlog?

src/contracts/libraries/BN254.sol
```diff
@@ -139,19 +139,40 @@ library BN254 {
...
-        // loop through each bit of s
-        for (uint8 i = 0; i < 9; i++) {
-            // if the bit is 1, add the 2^n*p to the accumulated product
-            if (s >> i & 1 == 1) {
-                acc = plus(acc, p2n);
+        // value of most signifigant bit
+        uint16 m = 1;
+
+        //check if s is greater than 1
+        if(s > m){
```
gpsanant, line 149: this function returns the wrong thing if s == 0, probably, just add a conditional at the top if s == 1

src/contracts/libraries/BN254.sol
```diff
@@ -139,19 +139,40 @@ library BN254 {
...
+            // the 2^n*p to add to the accumulated product in each iteration
+            BN254.G1Point memory p2n = p;
+            // index of most signifigant bit
+            uint8 i = 0;
+
+            //loop until we reach the most signifigant bit
+            while(s > m){
+                unchecked {
+                    // if the  current bit is 1, add the 2^n*p to the accumulated product
+                    if (s >> i & 1 == 1) {
```
gpsanant, line 159: `s & m` should work here?

0x0aa0 (commented):
# 1 Quorum
### scalar_mul
- 0 nonsign = 162804
- 1 nonsign = 181308
- 2 nonsign =  197517
###  scalar_mul_tiny
- 0 nonsign = 162804
- 1 nonsign = 175020
- 2 nonsign = 184940
# 100 Quorum
### scalar_mul
- 0 nonsign = 1095123
- 1 nonsign = 1486747
- 2 nonsign =  1878193
### scalar_mul_tiny
- 0 nonsign = 1095123
- 1 nonsign = 1488391
- 2 nonsign = 1881601

gpsanant:
The above might imply that any `scalar_mul_tiny` where `s` has more than 3 1s in its binary repr is more expensive than `scalar_mul`?

We should verify this by fuzzing `scalar_mul_tiny`.

PR #167, opened 2023-09-14
add events

gpsanant:
adds events on share delegation increases and decreases


src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -87,10 +87,16 @@ interface IDelegationManager {
      */
     event OperatorMetadataURIUpdated(address indexed operator, string metadataURI);
 
-    // @notice Emitted when @param staker delegates to @param operator.
+    /// @notice Emitted whenever an operator's shares are increased for a given strategy
+    event OperatorSharesIncreased(address indexed operator, address staker, IStrategy strategy, uint256 shares);
+
+    /// @notice Emitted whenever an operator's shares are decreased for a given list of strategies
+    event OperatorSharesDecreased(address indexed operator, address staker, IStrategy[] strategies, uint256[] shares);
```
bowenli86, line 94: why increase is by one strategy, but decrease is by a list of strategy?

gpsanant: gas

src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -87,10 +87,16 @@ interface IDelegationManager {
      */
     event OperatorMetadataURIUpdated(address indexed operator, string metadataURI);
 
-    // @notice Emitted when @param staker delegates to @param operator.
+    /// @notice Emitted whenever an operator's shares are increased for a given strategy
+    event OperatorSharesIncreased(address indexed operator, address staker, IStrategy strategy, uint256 shares);
```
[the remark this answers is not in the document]
gpsanant: good callout. Fixed: https://github.com/Layr-Labs/eigenlayer-contracts/pull/167/commits/282e2410dc95b04eb8d9e01094f9c34cf27a1c1a

src/contracts/interfaces/IDelegationManager.sol
bowenli86: IMHO this is potentially a bad idea from data perspective, better keep APIs uniformed - either all single strategy, or all list of strategy

ChaoticWalrus: we could always emit events in a loop when decrementing. I'm OK with that.

ChaoticWalrus (approved):
LGTM

PR #171, opened 2023-09-16
add shares getter in IStrategy.sol

daweth:
Added the shares() view function to the strategy interface, only the userUnderlying() method exists currently.

### DIFF
--- a/src/contracts/interfaces/IStrategy.sol
+++ b/src/contracts/interfaces/IStrategy.sol
@@ -54,6 +54,12 @@ interface IStrategy {
      */
     function userUnderlying(address user) external returns (uint256);
 
+    /**
+     * @notice convenience function for fetching the current total shares of `user` in this strategy, by
+     * querying the `strategyManager` contract
+     */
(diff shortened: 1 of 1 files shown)


ChaoticWalrus:
Seems like a good suggestion. I'm generally hesitant to add functions to interfaces, but I think it's reasonable + the existing Strategy contracts already support this.
Any opinions here @gpsanant ?

gpsanant:
> Any opinions here @gpsanant ?

Fine by me! What's wrong with the CI? @ChaoticWalrus

ChaoticWalrus:
> Fine by me! What's wrong with the CI? @ChaoticWalrus

OK, will approve + merge.

CI doesn't get access to GitHub secrets even when manually approved to run -- I think even for foundry tests, it's missing some necessary environment variable(s)? Definitely something we should sort out!
To be fair, this is "safer" in that it prevents leaking secrets.

ChaoticWalrus (approved):
LGTM, thanks for the submission.

PR #172, opened 2023-09-18
Cleanup, Add Check, and Write InitialOwner to Json

stevennevins:


script/M2_Deploy.s.sol
```diff
@@ -34,12 +38,13 @@ contract Deployer_M2 is ExistingDeploymentParser {
...
-        address oracleInitialOwner = executorMultisig;
-        uint256 initialThreshold = stdJson.readUint(config_data, ".oracleInitialization.threshold");
+        oracleInitialOwner = executorMultisig;
+        initialThreshold = stdJson.readUint(config_data, ".oracleInitialization.threshold");
         bytes memory oracleSignerListRaw = stdJson.parseRaw(config_data, ".oracleInitialization.signers");
-        address[] memory initialOracleSigners = abi.decode(oracleSignerListRaw, (address[]));
+        initialOracleSigners = abi.decode(oracleSignerListRaw, (address[]));
 
         require(initialThreshold <= initialOracleSigners.length, "invalid initialThreshold");
+        require(initialThreshold >= 1, "invalid initialThreshold");
```
gpsanant, line 47: We don't need the beacon chain oracle file actually, this is deployed by Succinct.

We do need all of the contracts in this script: https://github.com/Layr-Labs/eigenlayer-contracts/blob/m2-upgrade-script/script/upgrade/GoerliM2Upgrade.s.sol

Could you add them?

script/output/M2_deployment_data.json
```diff
@@ -1,9 +1,11 @@
 {
   "addresses": {
-    "beaconChainOracle": "0x90193C961A926261B756D1E5bb255e67ff9498A1"
```
gpsanant, line 3: .gitignore

script/M2_Deploy.s.sol
stevennevins: 🫡 added

stevennevins:
@gpsanant added the beacon chain oracle from succinct

script/output/M2_deployment_data.json
```diff
@@ -1,9 +1,20 @@
...
```
gpsanant, line 3: still not ignored

stevennevins: copied the .gitignore line for M1 and realized it doesn't work lol

updated

should I fix the one for the M1 and remove that as well?

gpsanant: ah, yes i guess. not sure if we're keeping them here for clarity? doubt it

gpsanant (approved):
lgtm

PR #173, opened 2023-09-18
Assorted EigenPod Fixes/Cleanup

Sidu28:
This PR is to add historical summary proofs to our withdrawal verification to prove older withdrawals.

It also addresses some comments in the "newpodchanges" PR.

- update validatorIndex in the validatorInfo struct
- I think that mostRecentBalanceUpdateSlot doesn't need to be set when you verify your withdrawal credentials.  0 here seems fine to me, but I will let the reviewers see if they agree.
- Switch from slot to timestamp in storage and events
- Syntax change in call to verifyProof functions to spedify individual inputs

### DIFF
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -282,6 +282,8 @@ library BeaconChainProofs {
             "BeaconChainProofs.verifyWithdrawalProofs: timestampProof has incorrect length");
 
         if(proofs.proveHistoricalRoot){
+            require(proofs.historicalSummaryBlockRootProof.length == 32 * (BEACON_STATE_FIELD_TREE_HEIGHT + (HISTORICAL_SUMMARIES_TREE_HEIGHT + 1) + 1 + (BLOCK_ROOTS_TREE_HEIGHT)),
+            "BeaconChainProofs.verifyWithdrawalProofs: historicalSummaryBlockRootProof has incorrect length");
             /**
             * Note: Here, the "1" in "1 + (BLOCK_ROOTS_TREE_HEIGHT)" signifies that extra step of choosing the "block_root_summary" within the individual 
             * "historical_summary". Everywhere else it signifies merkelize_with_mixin, where the length of an array is hashed with the root of the array,
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -283,11 +283,10 @@ library BeaconChainProofs {
 
         if(proofs.proveHistoricalRoot){
             uint256 historicalBlockHeaderIndex = HISTORICAL_SUMMARIES_INDEX << ((HISTORICAL_SUMMARIES_TREE_HEIGHT + 1) + 1 + (BLOCK_ROOTS_TREE_HEIGHT)) | 
+                                                uint256(proofs.historicalSummaryIndex) << 1 + (BLOCK_ROOTS_TREE_HEIGHT) |
                                                 BLOCK_SUMMARY_ROOT_INDEX << (BLOCK_ROOTS_TREE_HEIGHT) | uint256(proofs.blockHeaderRootIndex);
             require(Merkle.verifyInclusionSha256(proofs.historicalSummaryBlockRootProof, beaconStateRoot, proofs.blockHeaderRoot, historicalBlockHeaderIndex), 
                 "BeaconChainProofs.verifyWithdrawalProofs: Invalid historicalsummary merkle proof");
-            
-
         } else {
             /**
             * Computes the block_header_index relative to the beaconStateRoot.  It concatenates the indexes of all the
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -282,6 +282,11 @@ library BeaconChainProofs {
             "BeaconChainProofs.verifyWithdrawalProofs: timestampProof has incorrect length");
 
         if(proofs.proveHistoricalRoot){
+            /**
+            * Note: Here, the "1" in "1 + (BLOCK_ROOTS_TREE_HEIGHT)" signifies that extra step of choosing the "block_root_summary" within the individual 
+            * "historical_summary". Everywhere else it signifies merkelize_with_mixin, where the length of an array is hashed with the root of the array,
+            * but not here.
+            */
             uint256 historicalBlockHeaderIndex = HISTORICAL_SUMMARIES_INDEX << ((HISTORICAL_SUMMARIES_TREE_HEIGHT + 1) + 1 + (BLOCK_ROOTS_TREE_HEIGHT)) | 
                                                 uint256(proofs.historicalSummaryIndex) << 1 + (BLOCK_ROOTS_TREE_HEIGHT) |
                                                 BLOCK_SUMMARY_ROOT_INDEX << (BLOCK_ROOTS_TREE_HEIGHT) | uint256(proofs.blockHeaderRootIndex);
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -136,9 +136,7 @@ library BeaconChainProofs {
         bytes latestBlockHeaderProof;
         bytes validatorBalanceProof;
         bytes validatorFieldsProof;
-        bytes slotProof;
         bytes32 balanceRoot;
-        bytes32 slotRoot;
     }
 
     // @notice This struct contains the merkle proofs and leaves needed to verify a validator's withdrawal credential
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -281,7 +281,14 @@ library BeaconChainProofs {
         require(proofs.timestampProof.length == 32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT),
             "BeaconChainProofs.verifyWithdrawalProofs: timestampProof has incorrect length");
 
-        {
+        if(proofs.proveHistoricalRoot){
+            uint256 historicalBlockHeaderIndex = HISTORICAL_SUMMARIES_INDEX << ((HISTORICAL_SUMMARIES_TREE_HEIGHT + 1) + 1 + (BLOCK_ROOTS_TREE_HEIGHT)) | 
+                                                BLOCK_SUMMARY_ROOT_INDEX << (BLOCK_ROOTS_TREE_HEIGHT) | uint256(proofs.blockHeaderRootIndex);
+            require(Merkle.verifyInclusionSha256(proofs.historicalSummaryBlockRootProof, beaconStateRoot, proofs.blockHeaderRoot, historicalBlockHeaderIndex), 
+                "BeaconChainProofs.verifyWithdrawalProofs: Invalid historicalsummary merkle proof");
+            
+
+        } else {
             /**
             * Computes the block_header_index relative to the beaconStateRoot.  It concatenates the indexes of all the
             * intermediate root indexes from the bottom of the sub trees (the block header container) to the top of the tree
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -190,7 +190,7 @@ library BeaconChainProofs {
         bytes32 validatorRoot = Merkle.merkleizeSha256(validatorFields);
 
         // verify the proof of the validatorRoot against the beaconStateRoot
-        require(Merkle.verifyInclusionSha256(proof, beaconStateRoot, validatorRoot, index), "BeaconChainProofs.verifyValidatorFields: Invalid merkle proof");
+        require(Merkle.verifyInclusionSha256({proof: proof, root: beaconStateRoot, leaf: validatorRoot, index: index}), "BeaconChainProofs.verifyValidatorFields: Invalid merkle proof");
     }
 
     /**
@@ -215,7 +215,7 @@ library BeaconChainProofs {
         uint256 balanceIndex = uint256(validatorIndex/4);
         balanceIndex = (BALANCE_INDEX << (BALANCE_TREE_HEIGHT + 1)) | balanceIndex;
 
-        require(Merkle.verifyInclusionSha256(proof, beaconStateRoot, balanceRoot, balanceIndex), "BeaconChainProofs.verifyValidatorBalance: Invalid merkle proof");
+        require(Merkle.verifyInclusionSha256({proof: proof, root: beaconStateRoot, leaf: balanceRoot, index: balanceIndex}), "BeaconChainProofs.verifyValidatorBalance: Invalid merkle proof");
     }
 
     /**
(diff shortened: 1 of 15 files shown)


src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -281,7 +281,14 @@ library BeaconChainProofs {
         require(proofs.timestampProof.length == 32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT),
             "BeaconChainProofs.verifyWithdrawalProofs: timestampProof has incorrect length");
 
-        {
+        if(proofs.proveHistoricalRoot){
```
gpsanant, line 284: my opinion is that we only allow proofs against hisorical summaries. not worth maintaining 2 things.

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -281,7 +281,14 @@ library BeaconChainProofs {
...
+            uint256 historicalBlockHeaderIndex = HISTORICAL_SUMMARIES_INDEX << ((HISTORICAL_SUMMARIES_TREE_HEIGHT + 1) + 1 + (BLOCK_ROOTS_TREE_HEIGHT)) |
```
gpsanant, line 285: parentheses here are awkward, can you regroup to `((HISTORICAL_SUMMARIES_TREE_HEIGHT + 1) + (1 + BLOCK_ROOTS_TREE_HEIGHT))`

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -281,7 +281,14 @@ library BeaconChainProofs {
         require(proofs.timestampProof.length == 32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT),
             "BeaconChainProofs.verifyWithdrawalProofs: timestampProof has incorrect length");
 
-        {
+        if(proofs.proveHistoricalRoot){
+            uint256 historicalBlockHeaderIndex = HISTORICAL_SUMMARIES_INDEX << ((HISTORICAL_SUMMARIES_TREE_HEIGHT + 1) + 1 + (BLOCK_ROOTS_TREE_HEIGHT)) | 
+                                                BLOCK_SUMMARY_ROOT_INDEX << (BLOCK_ROOTS_TREE_HEIGHT) | uint256(proofs.blockHeaderRootIndex);
```
gpsanant, line 286: talked offline, this needs to be fixed

src/contracts/libraries/BeaconChainProofs.sol
Sidu28: I disagree - ok so now they have to wait up to 27hours until they can withdraw? Its not like we're "maintaining" two things.  We just generate an additional proof when you're proving an older root.

src/contracts/libraries/BeaconChainProofs.sol
Sidu28: ok should be fixed

src/contracts/libraries/BeaconChainProofs.sol
gpsanant: It's 2 different proof generation codes, no?

Sidu28: whats a proof generation code?

src/contracts/libraries/BeaconChainProofs.sol
Sidu28: no I think it should be 1 + (BLOCK_ROOTS_TREE_HEIGHT) not (1 + (BLOCK_ROOTS_TREE_HEIGHT))

because there the "1" signifies that extra step of choosing the "block_root_summary" within the individual "historical_summary". Everywhere else it signifies merkelize_with_mixin, where the length of an array is hashed with the root of the array.  I've added a comment to clear this up

gpsanant: ah, sounds good. something that would be helpful is literally commending on every single thing being summed what it's supposed to be. maybe we can do this after things are merged into master

src/contracts/pods/EigenPod.sol
```diff
@@ -44,6 +44,12 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     /// @notice Maximum "staleness" of a Beacon Chain state root against which `verifyBalanceUpdate` or `verifyWithdrawalCredental` may be proven.
     uint256 internal constant VERIFY_BALANCE_UPDATE_WINDOW_SECONDS = 4.5 hours;
 
+    //this is the genesis time of the beacon state, to help us calculate conversions between slot and timestamp
+    uint64 internal constant GENESIS_TIME = 1616508000;
```
gpsanant, line 48: this needs to be a contruction parameter?

gpsanant:
> * I think that mostRecentBalanceUpdateSlot doesn't need to be set when you verify your withdrawal credentials.  0 here seems fine to me, but I will let the reviewers see if they agree.

I think it's just nice to have the invariant there because we know it works if it's set. Saves us all some brain power :)

src/contracts/libraries/BeaconChainProofs.sol
gpsanant: agreed offline to keep 1 proof

src/contracts/pods/EigenPod.sol
```diff
@@ -540,7 +549,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         // now that the validator has been proven to be withdrawn, we can set their restaked balance to 0
         _validatorPubkeyHashToInfo[validatorPubkeyHash].restakedBalanceGwei = 0;
 
-        provenWithdrawal[validatorPubkeyHash][withdrawalHappenedSlot] = true;
+        provenWithdrawal[validatorPubkeyHash][_computeTimestampAtSlot(withdrawalHappenedSlot)] = true;
 
         emit FullWithdrawalRedeemed(validatorIndex, recipient, withdrawalAmountGwei * GWEI_TO_WEI);
```
gpsanant, line 554: add timestamp to this event

src/contracts/pods/EigenPod.sol
```diff
@@ -558,7 +567,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         uint64 partialWithdrawalAmountGwei
     ) internal {
 
-        provenWithdrawal[validatorPubkeyHash][withdrawalHappenedSlot] = true;
+        provenWithdrawal[validatorPubkeyHash][_computeTimestampAtSlot(withdrawalHappenedSlot)] = true;
         emit PartialWithdrawalRedeemed(validatorIndex, recipient, partialWithdrawalAmountGwei);
```
gpsanant, line 571: add timestamp to this event

Sidu28:
> > * I think that mostRecentBalanceUpdateSlot doesn't need to be set when you verify your withdrawal credentials.  0 here seems fine to me, but I will let the reviewers see if they agree.
> 
> I think it's just nice to have the invariant there because we know it works if it's set. Saves us all some brain power :)

Ok but then I have to somehow prove which slot it was in which doesn't make sense?  Maybe I could set it to validator. activation_epoch but Im not sure.  I could also explicityl set it to 0 or the genesis slot or something.

src/contracts/pods/EigenPod.sol
Sidu28: yeah fair, fixed

src/contracts/pods/EigenPod.sol
Sidu28: added

gpsanant:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/5a169dac7728bf98463e6e88c0c61b9efaaf1cc5/src/contracts/pods/EigenPod.sol#L264

https://github.com/Layr-Labs/eigenlayer-contracts/blob/5a169dac7728bf98463e6e88c0c61b9efaaf1cc5/src/contracts/pods/EigenPod.sol#L284

Could you take out the `'`? Github hates it

gpsanant:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/5a169dac7728bf98463e6e88c0c61b9efaaf1cc5/src/contracts/pods/EigenPod.sol#L271
pls also fix the name of this function for the love of god im going crazy

src/contracts/pods/EigenPod.sol
Sidu28: fixed

gpsanant:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/5a169dac7728bf98463e6e88c0c61b9efaaf1cc5/src/contracts/pods/EigenPod.sol#L267
Can we bring this line to where it's used?

gpsanant:
Could we include the explicit param name syntax for all of these?
https://github.com/Layr-Labs/eigenlayer-contracts/blob/5a169dac7728bf98463e6e88c0c61b9efaaf1cc5/src/contracts/pods/EigenPod.sol#L317

gpsanant:
Could we order the functions as following? Maybe do this in a separate PR after we merge into master?
```
// external callable by anyone
// external callable by EPM
// external callable by pod owner
// internal state changing
// internal pure
```

gpsanant:
remove: https://github.com/Layr-Labs/eigenlayer-contracts/blob/1426edbf414918fdc15682a45c58884555b262bc/src/contracts/pods/EigenPod.sol#L418

gpsanant:
> > > * I think that mostRecentBalanceUpdateSlot doesn't need to be set when you verify your withdrawal credentials.  0 here seems fine to me, but I will let the reviewers see if they agree.
> > 
> > 
> > I think it's just nice to have the invariant there because we know it works if it's set. Saves us all some brain power :)
> 
> Ok but then I have to somehow prove which slot it was in which doesn't make sense? Maybe I could set it to validator. activation_epoch but Im not sure. I could also explicityl set it to 0 or the genesis slot or something.

should just be oracle timestamp everywhere except withdrawal proofs, no?

gpsanant:
Could we move timestamp to be the 2nd param after index in all of these?

gpsanant:
no need to pass down slot, just pass timestamp:

https://github.com/Layr-Labs/eigenlayer-contracts/blob/10fa596adcfe7c42d8d5889c313fdf6b9001957c/src/contracts/pods/EigenPod.sol#L496

gpsanant:
Set `provenWithdrawal` here and take it out of child functions
https://github.com/Layr-Labs/eigenlayer-contracts/blob/10fa596adcfe7c42d8d5889c313fdf6b9001957c/src/contracts/pods/EigenPod.sol#L469

gpsanant:
Don't pass `status` here, just load below
https://github.com/Layr-Labs/eigenlayer-contracts/blob/10fa596adcfe7c42d8d5889c313fdf6b9001957c/src/contracts/pods/EigenPod.sol#L496

Sidu28:
> > > > * I think that mostRecentBalanceUpdateSlot doesn't need to be set when you verify your withdrawal credentials.  0 here seems fine to me, but I will let the reviewers see if they agree.
> > > 
> > > 
> > > I think it's just nice to have the invariant there because we know it works if it's set. Saves us all some brain power :)
> > 
> > 
> > Ok but then I have to somehow prove which slot it was in which doesn't make sense? Maybe I could set it to validator. activation_epoch but Im not sure. I could also explicityl set it to 0 or the genesis slot or something.
> 
> should just be oracle timestamp everywhere except withdrawal proofs, no?

I just set it to slotRoot converted to timestamp in the withdrawal and balance update proofs

gpsanant:
Can we emit a balance update event here:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/10fa596adcfe7c42d8d5889c313fdf6b9001957c/src/contracts/pods/EigenPod.sol#L425

src/contracts/pods/EigenPod.sol
```diff
@@ -68,6 +66,9 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     */
     uint64 public immutable RESTAKED_BALANCE_OFFSET_GWEI;
 
+    //this is the genesis time of the beacon state, to help us calculate conversions between slot and timestamp
```
gpsanant, line 69: @ notice

src/contracts/pods/EigenPod.sol
```diff
@@ -493,32 +497,32 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
             */
             // reference: uint64 withdrawableEpoch = Endian.fromLittleEndianUint64(validatorFields[BeaconChainProofs.VALIDATOR_WITHDRAWABLE_EPOCH_INDEX]);
             if (Endian.fromLittleEndianUint64(validatorFields[BeaconChainProofs.VALIDATOR_WITHDRAWABLE_EPOCH_INDEX]) <= slot/BeaconChainProofs.SLOTS_PER_EPOCH) {
-                _processFullWithdrawal(validatorIndex, validatorPubkeyHash, slot, podOwner, withdrawalAmountGwei, _validatorPubkeyHashToInfo[validatorPubkeyHash].status);
+                _processFullWithdrawal(validatorIndex, validatorPubkeyHash, withdrawalHappenedTimestamp, podOwner, withdrawalAmountGwei, validatorInfo);
             } else {
-                _processPartialWithdrawal(validatorIndex, validatorPubkeyHash, slot, podOwner, withdrawalAmountGwei);
+                _processPartialWithdrawal(validatorIndex, slot, podOwner, withdrawalAmountGwei);
```
gpsanant, line 502: pass timestamp here

src/contracts/pods/EigenPod.sol
```diff
@@ -452,39 +452,43 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-        require(_validatorPubkeyHashToInfo[validatorPubkeyHash].status != VALIDATOR_STATUS.INACTIVE,
+        ValidatorInfo memory validatorInfo = _validatorPubkeyHashToInfo[validatorPubkeyHash];
+        require(validatorInfo.status != VALIDATOR_STATUS.INACTIVE,
             "EigenPod._verifyAndProcessWithdrawal: Validator never proven to have withdrawal credentials pointed to this contract");
 
-        uint64 timestamp = _computeTimestampAtSlot(Endian.fromLittleEndianUint64(withdrawalProofs.slotRoot));
-        require(!provenWithdrawal[validatorPubkeyHash][timestamp],
-            "EigenPod._verifyAndProcessWithdrawal: withdrawal has already been proven for this slot");
-
+        {
```
gpsanant, line 467: why need scoping?

src/contracts/pods/EigenPod.sol
Sidu28: fixed

src/contracts/pods/EigenPod.sol
Sidu28: whoops, fixed

src/contracts/pods/EigenPod.sol
Sidu28: vestigal, fixed it now

gpsanant:
Can we take slot root out entirely here and just use oracleTimestamp?

https://github.com/Layr-Labs/eigenlayer-contracts/blob/4569bb479fbfbb852eab49d30edbea1e5e870e75/src/contracts/pods/EigenPod.sol#L266

src/contracts/pods/EigenPod.sol
```diff
@@ -20,6 +20,8 @@ import "../interfaces/IPausable.sol";
 
 import "./EigenPodPausingConstants.sol";
 
+import "forge-std/Test.sol";
```
gpsanant, line 23: noooooooo

src/contracts/pods/EigenPod.sol
```diff
@@ -288,12 +291,6 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
             proof: proofs.validatorBalanceProof,
             validatorIndex: validatorIndex
         });
-        //verify provided slot is valid against beaconStateRoot
-        BeaconChainProofs.verifySlotRoot({
-            beaconStateRoot: proofs.beaconStateRoot,
-            slotRoot: proofs.slotRoot,
```
gpsanant: should remove this from the struct?

gpsanant (approved):
LGTM, great work. we so close

src/contracts/pods/EigenPod.sol
Sidu28: done

PR #176, opened 2023-09-19
Simplify Undelegation Interactions

ChaoticWalrus:
This PR seeks to (as the title suggests) simplify the undelegation interactions in the EigenLayer core contracts.
Specifically, the primary goal is to move checks and functions from the StrategyManager and EigenPodManager into the DelegationManager itself.
This should clarify and "clean" the design, help ensure that state is being stored and properties enforced in more "correct" locations, and reduce the amount of back-and-forth calls between system contracts.

I am initially marking this PR as a 'draft'. I plan to review it myself / potentially add a bit more to the PR before marking it as 'ready for review'.

Edit (as of commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/dfccfb9a011cb44bee033b3fc9249e2c2a90599f):
I am marking this PR as ready-for-review now. CI currently has a couple failures in Prover rules. From a cursory investigation, these appear to be failures that are indicative of the rules needing minor refinements, rather than indicative of the changes causing "true failures" in important contract properties. I intend to push commits that fix these failures. I still consider the PR ready-for-review since I do not anticipate getting much review on these minor refinements to specs; I will update and ping reviewers if there end up needing to be substantial changes to anything beyond these spec files.


src/contracts/core/DelegationManager.sol
```diff
@@ -368,6 +381,15 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         emit StakerDelegated(staker, operator);
     }
 
+    function _undelegate(address staker) internal onlyNotFrozen(staker) {
+        address operator = delegatedTo[staker];
```
gpsanant, line 385: load inside if

src/contracts/pods/EigenPodManager.sol
```diff
@@ -266,17 +266,18 @@ contract EigenPodManager is
 
     /**
      * @notice Called by a staker who owns an EigenPod to enter the "undelegation limbo" mode.
+     * @param undelegateIfPossible If this param is marked as 'true', then this function will also inform the DelegationManager of the caller's desire to undelegate
      * @dev Undelegation limbo is a mode which a staker can enter into, in which they remove their virtual "beacon chain ETH shares" from EigenLayer's delegation
      * system but do not necessarily withdraw the associated ETH from EigenLayer itself. This mode allows users who have restaked native ETH a route via
      * which they can undelegate from an operator without needing to exit any of their validators from the Consensus Layer.
      */
-    function enterUndelegationLimbo()
+    function enterUndelegationLimbo(bool undelegateIfPossible)
```
gpsanant, line 274: you only go into undelegation limbo if you want to undelegate? :)

gpsanant:
LGTM mostly except a couple comments. Does the following case now exist?
- Remove all shares and forget to set `undelegateIfPossible`
- In order to undelegate, need to 
  - For SM, deposit and then withdraw with `undelegateIfPossible = true` 
  - For EPM, go into undelegation limbo and wait for delay to pass

ChaoticWalrus:
> LGTM mostly except a couple comments. Does the following case now exist?
> 
>     * Remove all shares and forget to set `undelegateIfPossible`
> 
>     * In order to undelegate, need to
>       
>       * For SM, deposit and then withdraw with `undelegateIfPossible = true`
>       * For EPM, go into undelegation limbo and wait for delay to pass

Cool, will address your comments immediately. For the case you're suggesting, the staker _should_ be able to just undelegate by calling the `DelegationManager.undelegate` function, which is no longer permissioned (and no longer takes an address input) -- see these lines here https://github.com/Layr-Labs/eigenlayer-contracts/blob/a559b4c8dd0f0cfbba5275b491ff1c1dac90cf1e/src/contracts/core/DelegationManager.sol#L202-L206 -- please take a look and me know if I've missed something!

gpsanant:
Ah yes, forgot this is now public. Your solution is correct.

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: hmm yeah good point, this probably just shouldn't have an input param, will make that change

ChaoticWalrus: removed the input in this commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/3c4c5e6bed5f94e0ae329046b041373aa250a67f

src/contracts/core/DelegationManager.sol
ChaoticWalrus: ty, done https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/282372bf2f13d8fa396c39b2fc1bbd0fb534fd50

ChaoticWalrus:
@gpsanant one notable commit since your review https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/058e65044a824eec0b4d014a9081780f3c686794 -- please take a quick look and confirm my reasoning (in the commit message)? This is in-line with the undelegation behavior we had prior to this PR.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -675,12 +666,7 @@ contract EigenPodManager is
...
             IStrategy[] memory strategies = new IStrategy[](1);
             strategies[0] = beaconChainETHStrategy;
-            delegationManager.decreaseDelegatedShares(podOwner, strategies, shareAmounts);
-
-            // undelegate the pod owner, if possible
-            if (stakerCanUndelegate(podOwner)) {
-                _undelegate(podOwner);
-            }
+            delegationManager.decreaseDelegatedShares(podOwner, strategies, shareAmounts, true/*undelegateIfPossible*/);
         }
```
gpsanant, line 670: inline comment is mistake?

gpsanant (approved):
LGTM except last comment?

signing off for next few hours

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: are you talking about this comment?
```
        // if the podOwner has delegated shares, record a decrease in their delegated shares
```
I think this is correct? Maybe a bit unclear though; would appreciate any suggestions you have for replacement.

src/contracts/core/DelegationManager.sol
```diff
@@ -223,8 +217,12 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         eigenPodManager.forceIntoUndelegationLimbo(staker);
         
         // force a withdrawal of all of the staker's shares from the StrategyManager
-        return (strategyManager.forceTotalWithdrawal(staker));
+        bytes32 queuedWithdrawal = strategyManager.forceTotalWithdrawal(staker);
+
+        // actually undelegate the staker
+        _undelegate(staker);
```
wadealexc, line 223: This method actually gets really funky if you think about it. `epm.forceIntoUndelegationLimbo` needs to query the staker's operator, and then also calls back into `DelegationManager.decreaseDelegatedShares(..., undelegateIfPossible: true)`

And this happens before we try to call the strategymanager, or even undelegate. I imagine there are similar weird loopy patterns there.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -675,12 +666,7 @@ contract EigenPodManager is
             shareAmounts[0] = podOwnerShares[podOwner];
             IStrategy[] memory strategies = new IStrategy[](1);
             strategies[0] = beaconChainETHStrategy;
-            delegationManager.decreaseDelegatedShares(podOwner, strategies, shareAmounts);
-
-            // undelegate the pod owner, if possible
-            if (stakerCanUndelegate(podOwner)) {
-                _undelegate(podOwner);
-            }
+            delegationManager.decreaseDelegatedShares(podOwner, strategies, shareAmounts, true/*undelegateIfPossible*/);
```
wadealexc, line 669: Weird looping behavior here. I think this can be simplified and maybe we should pass "false" in here.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -642,23 +643,13 @@ contract EigenPodManager is
...
-    }
-
     /**
      * @notice Internal function to enter `podOwner` into undelegation limbo
      * @dev Does nothing if the `podOwner` has no delegated shares (i.e. they are already in undelegation limbo or have no shares)
      * OR if they are not actively delegated to any operator.
      */
     function _enterUndelegationLimbo(address podOwner) internal {
-        if (!stakerHasNoDelegatedShares(podOwner) && delegationManager.isDelegated(podOwner)) {
+        if (!podOwnerHasNoDelegatedShares(podOwner) && delegationManager.isDelegated(podOwner)) {
```
wadealexc, line 652: Not a fan of the additional dependencies here - feels like implementing delegationmgr logic in epmgr

wadealexc (changes requested):
Still a lot of weird interdependencies. I think it can be cleaned up further, but maybe this is a good step for M2 and we can make it really nice for the next release.

I have a few changes requested in the comments.

I do have some tentative changes on a local branch, but then I found a withdrawal delay bug and got sidetracked.

src/contracts/core/DelegationManager.sol
```diff
@@ -480,6 +494,23 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+
+    }
+
+    /**
+     * @notice Returns 'true' if `staker` has "active" shares in EigenLayer (i.e. the staker has shares which are currently in the StrategyManager
+     * or in the EigenPodManager + not in "undelegation limbo"), and returns 'false' otherwise.
+     */
+    function stakerHasActiveShares(address staker) public view returns (bool) {
+        return ((strategyManager.stakerStrategyListLength(staker) != 0) || !eigenPodManager.podOwnerHasNoDelegatedShares(staker));
+    }
```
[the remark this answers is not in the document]
ChaoticWalrus: agreed it's pretty convoluted. I prefer your second suggestion and will implement it.

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: do you think it would make more sense for entering this "undelegation limbo" mode to always originate from a call to the DelegationManager? then this flow would probably be more logical / checks could happen more "in place"?

wadealexc: yes, that's what i've spent some time implementing locally! i got sidetracked by that bug in withdrawal queues, but my initial findings were that things were much more logically separated.

ChaoticWalrus: > yes, that's what i've spent some time implementing locally! i got sidetracked by that bug in withdrawal queues, but my initial findings were that things were much more logically separated.

makes sense! if you're implementing it yourself, I can avoid replicating your work? feel free to push when ready and I can review? Or let me know if you prefer I implement it myself and we can compare.

src/contracts/core/DelegationManager.sol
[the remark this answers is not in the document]
ChaoticWalrus: implemented in this commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/50d27e997d68c15791c1f3d8ce16dda763152201

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: agreed, especially with the commit that combines the two undelegation functions (https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/c5b9f55f650cecb3f75beb0d2114a322f73f4b3c) this has some weird reentrant behavior. simple fix would probably be passing "false" here

src/contracts/core/DelegationManager.sol
```diff
@@ -480,6 +494,23 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         return approverDigestHash;
     }
 
+    /** 
+     * @notice Returns 'true' if the staker can undelegate or  if the staker is already undelegated, and 'false' otherwise
+     * @dev A staker can only undelegate if they have no "active" shares in EigenLayer and are not themselves an operator
+     */
+    function stakerCanUndelegate(address staker) public view returns (bool) {
```
stevennevins, line 499: Outside the scope of this PR, the below naming is more straightforward.
```
function canUndelegate(address staker) public view returns (bool)
```
The notice could also be shortened to:
```
@notice Checks if `staker` can undelegate.
```

src/contracts/core/DelegationManager.sol
```diff
@@ -254,11 +252,13 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
      * @param shares An array of the number of shares to decrease for a operator and strategy.
+     * @param undelegateIfPossible If marked 'true', then this contract will check if the `staker` can undelegate, and undelegate them if possible.
+     * If the check fails, then the `staker` will simply remain delegated.
      * 
      * @dev *If the staker is actively delegated*, then decreases the `staker`'s delegated shares in each entry of `strategies` by its respective `shares[i]`. Otherwise does nothing.
-     * @dev Callable only by the StrategyManager.
+     * @dev Callable only by the StrategyManager or EigenPodManager.
      */
-    function decreaseDelegatedShares(address staker, IStrategy[] calldata strategies, uint256[] calldata shares)
+    function decreaseDelegatedShares(address staker, IStrategy[] calldata strategies, uint256[] calldata shares, bool undelegateIfPossible)
```
stevennevins, line 259: Is there a reason we're not auto-undelegating a user if their shares are being decreased by an amount that zeros out their entire share balance in all the strategies instead of undelegateIfPossible?

src/contracts/core/DelegationManager.sol
```diff
@@ -480,6 +494,23 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        return (!stakerHasActiveShares(staker) && !isOperator(staker));
+
+    }
+
+    /**
+     * @notice Returns 'true' if `staker` has "active" shares in EigenLayer (i.e. the staker has shares which are currently in the StrategyManager
+     * or in the EigenPodManager + not in "undelegation limbo"), and returns 'false' otherwise.
+     */
+    function stakerHasActiveShares(address staker) public view returns (bool) {
+        return ((strategyManager.stakerStrategyListLength(staker) != 0) || !eigenPodManager.podOwnerHasNoDelegatedShares(staker));
```
stevennevins: Double negation here:
```
!eigenPodManager.podOwnerHasNoDelegatedShares(staker))
```
Is more complicated to understand than:
```
eigenPodManager.podOwnerHasDelegatedShares(staker));
```
Similar clarity comment for the stakerStategyListLength check.  It seems like, ideally, this comparison should be
```
return strategyManager.stakerHasDelegatedShares(staker) || eigenPodManager.podOwnerHasDelegatedShares(staker);
```

src/contracts/core/DelegationManager.sol
ChaoticWalrus: > Adding undelegateIfPossible feels like it adds a bit of complexity
yes, sorry for lack of clarity, this PR to this branch further simplifies / deletes this https://github.com/Layr-Labs/eigenlayer-contracts/pull/183

src/contracts/core/DelegationManager.sol
ChaoticWalrus: yes, sorry for lack of clarity, this PR to this branch further simplifies this https://github.com/Layr-Labs/eigenlayer-contracts/pull/183

src/contracts/core/DelegationManager.sol
ChaoticWalrus: I like this suggestion, will plan to implement

ChaoticWalrus: addressed in minor commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/1124853d65e922c18369cc94a84dd806ffea7ac1

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: addressed as part of PR https://github.com/Layr-Labs/eigenlayer-contracts/pull/183 -- or see the merge commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/75e1c96437486846dce1b391fe59c71a88598113

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: addressed as part of PR https://github.com/Layr-Labs/eigenlayer-contracts/pull/183 -- or see the merge commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/75e1c96437486846dce1b391fe59c71a88598113

src/contracts/core/DelegationManager.sol
ChaoticWalrus: addressed as part of PR https://github.com/Layr-Labs/eigenlayer-contracts/pull/183 -- or see the merge commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/75e1c96437486846dce1b391fe59c71a88598113

src/contracts/pods/EigenPodManager.sol
```diff
@@ -434,7 +441,7 @@ contract EigenPodManager is
             "EigenPodManager._queueWithdrawal: cannot queue a withdrawal when in undelegation limbo");
 
 
-        _removeShares(podOwner, amountWei);  
+        _removeShares(podOwner, amountWei, undelegateIfPossible);
```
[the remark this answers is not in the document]
ChaoticWalrus: addressed as part of PR https://github.com/Layr-Labs/eigenlayer-contracts/pull/183 -- or see the merge commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/75e1c96437486846dce1b391fe59c71a88598113

ChaoticWalrus:
@stevennevins the sub-PR to this https://github.com/Layr-Labs/eigenlayer-contracts/pull/183 has been merged in this commit: https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/75e1c96437486846dce1b391fe59c71a88598113, please give a review of this PR when you can, and thanks for bearing with me!

src/contracts/core/DelegationManager.sol
```diff
@@ -483,6 +501,17 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         return approverDigestHash;
     }
 
+    /** 
+     * @notice Returns 'true' if the `staker` can immediately undelegate without queuing a new withdrawal OR if the staker is already undelegated,
+     * and 'false' otherwise
+     * @dev A staker can only undelegate if they have no "active" shares in EigenLayer and are not themselves an operator
+     */
+    function canUndelegate(address staker) public view returns (bool) {
+        return (!isOperator(staker) &&
```
stevennevins, line 510: This should also have a condition for `isDelegated(staker)` otherwise, this will return true when they're already undelegated

src/contracts/core/DelegationManager.sol
```diff
@@ -185,46 +185,59 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        require(!isOperator(staker), "DelegationManager.undelegate: operators cannot be undelegated");
+        require(staker != address(0), "DelegationManager.undelegate: cannot undelegate zero address");
+        require(
+            msg.sender == staker ||
+            msg.sender == operator ||
+            msg.sender == _operatorDetails[operator].delegationApprover,
+            "DelegationManager.undelegate: caller cannot undelegate staker"
+        );
+        
+        if (!canUndelegate(staker)) {
```
stevennevins, line 209: Is this just for EigenPod owner (stakers)?

stevennevins (approved):

src/contracts/core/DelegationManager.sol
ChaoticWalrus: yeah, good point. it has a comment which notes the current behavior, but I agree your suggested change aligns with the stricter behavior added to the `undelegate` function (reverting rather than being a no-op if the staker is already undelegated)

ChaoticWalrus: or this function could be renamed to something like `hasActivelyDelegatedShares` ?
I think this would make much more sense with the current behavior, actually.
any thoughts on this @stevennevins ?

src/contracts/core/DelegationManager.sol
ChaoticWalrus: This is a result of combining the `undelegate` and `forceUndelegate` functions. See my comment above -- I think renaming the function to `hasActivelyDelegatedShares` is probably the right change to be made here; I can also add a comment to the effect of
```
// remove any shares from the delegation system that the staker currently has delegated, if necessary
```

src/contracts/core/DelegationManager.sol
stevennevins: > or this function could be renamed to something like `hasActivelyDelegatedShares` ?
> I think this would make much more sense with the current behavior, actually.
> any thoughts on this @stevennevins ?

Yeah, I like that better too

ChaoticWalrus: This ended up being a touch more complicated than I was thinking -- can you do a bit of a spot-check on the commit? https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/be4a017e0c0b3535318a25f759601bcf07bc22a0

ChaoticWalrus:
@wadealexc I think you've had a good look at this already, but wasn't sure if you wanted a final pass -- PR is currently marked as "change requested" from you.

gpsanant:
This patches an oversight in our logging: https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/37abdcd543b7b1918c312ac51ff26d936ff451ed

We need to emit `OperatorSharesIncreased` during delegation

src/contracts/core/DelegationManager.sol
ChaoticWalrus: I added one more subsequent commit that just deletes the function 😄 -- see https://github.com/Layr-Labs/eigenlayer-contracts/pull/176/commits/96cbe826c83cf1c7ec531a4f8b01b4ecfa7e7ab1
I think the logic is both clearer and cleaner now

wadealexc (approved):
LGTM :)

Thanks for tweaking the wording again, J!

PR #177, opened 2023-09-20
M2 upgrade script

gpsanant:
The deploy script for M2

TODO: Keep compatible with EP and EPM updates


script/upgrade/GoerliM2Upgrade.s.sol
```diff
@@ -0,0 +1,121 @@
...
+
+// # To load the variables in the .env file
+// source .env
+
+// # To deploy and verify our contract
+// forge script script/upgrade/GoerliM2Deployment.s.sol:GoerliM2Deployment --rpc-url $RPC_URL  --private-key $PRIVATE_KEY --broadcast -vvvv
+contract GoerliM2Deployment is Script, Test {
+    Vm cheats = Vm(HEVM_ADDRESS);
+
+    string public deploymentOutputPath = string(bytes("script/output/M1_deployment_goerli_2023_3_23.json"));
```
ChaoticWalrus, line 39: update this file?

script/upgrade/GoerliM2Upgrade.s.sol
```diff
@@ -0,0 +1,121 @@
...
+            slasher,
+            delegation
+        );
+        eigenPodImplementation = new EigenPod(
+            ethPOS,
+            delayedWithdrawalRouter,
+            eigenPodManager,
+            31 gwei,
+            0.5 gwei
+        );
```
ChaoticWalrus, line 91: might want the more explicit syntax here, to extra-ensure that the last two args can't be in the wrong order

src/test/EigenLayerDeployer.t.sol
```diff
@@ -101,7 +99,7 @@ contract EigenLayerDeployer is Operators {
     address podAddress;
     address delayedWithdrawalRouterAddress;
     address eigenPodBeaconAddress;
-    address beaconChainOracleAddress;
+    address beaconChainOracleAddress = 0x40B10ddD29a2cfF33DBC420AE5bbDa0649049f2c;
```
ChaoticWalrus, line 102: seems odd to be hard-coding this address in the tests? I guess we just aren't using it so it's OK?

ChaoticWalrus (approved):
Seems fine, dropped some nits.
We will need to execute the upgrade separately; there's still a lot to do on this operationally.

src/test/EigenLayerDeployer.t.sol
stevennevins: What do you think is best here? We removed the Oracle contract that wasn't being used in the tests, but we still needed to pass it as a parameter to some contracts. It could have been an uninitialized value, but that similarly felt weird.

ChaoticWalrus: ah, good point. I think it's OK as-is since we aren't using it.
Going forward we should probably make some mock for this and use it in tests as needed.

stevennevins: That’s a good point. I will make a mock so it’s clearer

script/upgrade/GoerliM2Upgrade.s.sol
stevennevins: Addressed [d9a548c](https://github.com/Layr-Labs/eigenlayer-contracts/pull/177/commits/d9a548c6fab99936f3c71a0aac68bbdee21bcce7)

script/upgrade/GoerliM2Upgrade.s.sol
stevennevins: This path is to get the addresses from the M1 deployment

The output of this script is getting written to: `script/output/M2_deployment_data.json`

src/test/EigenLayerDeployer.t.sol
stevennevins: addressed in [935a629](https://github.com/Layr-Labs/eigenlayer-contracts/pull/177/commits/935a629e2814ce8ab8c4f8897839861116358e54)

PR #178, opened 2023-09-20
HistoricalSummary Proofs Working

Sidu28:
This PR is specifically to add the historical summary proof to the existing withdrawal proofs (full and partial withdrawals) such that the tests work.

THis PR also makes the change that mandates that all withdrawal proofs include the historical summary proof (previously it was optional)


ChaoticWalrus:
> THis PR also makes the change that mandates that all withdrawal proofs include the historical summary proof (previously it was optional)

Can you give a brief summary of why you landed on requiring historical summary proofs for all withdrawal proofs? Trying to understand why/if this was the correct decision, myself.

Sidu28:
> > THis PR also makes the change that mandates that all withdrawal proofs include the historical summary proof (previously it was optional)
> 
> Can you give a brief summary of why you landed on requiring historical summary proofs for all withdrawal proofs? Trying to understand why/if this was the correct decision, myself.

Yup!  Sorry this was a discussion that was had on slack, and I didn't record it on here.  This was my original post:

Ok, this post is regarding a point of disagreement between my and , we want to solicit opinions of the other folks here.  Once again, if something is unclear I apologize and will try to address it
Historical summary proofs allow us to prove withdrawals from more than 27 hours after they are included in the beacon chain.  In my initial implementation, I made the submission of a historical summary proof optional in case a large pod owner wanted to prove withdrawals within the 27 hour period to save gas on computing an additonal proof.
The historical summary proof, by my estimation, will cost very close to a normal withhdrawal credential proof (which costs 33k gas).  The normal withdrawal cred proof is about 46 hashes long.  The historical summary proof is about 24 hashes longer so safe to say they will cost about 15k gas.
I think someone shouldn;t have to bear the burden of waiting for the next historical summary to get added to the state (which can take up to 27 hours) and then have to pay for an extra proof.  They should have the option to save 15k gas per withdrawal proof.
On the flip side, I don't think this is much overhead for us.  Its a matter of generating the historical summary proof or not.  So we need logic on the backend that knows whether or not to generate that additional proof


Both @gpsanant and @wadealexc expressed that they feel less complexity should be the priority here and that we should just make it one pathway for proving withdrawals that involves every user also submitting a historical summary proof even if their withdrawal doesn't need it.

gpsanant (approved):
LGTM

PR #179, opened 2023-09-20
Update inline commenting for RegistryCoordinator

8sunyuan:
Removed comments mentioning that the `BLSRegistryCoordinatorWithIndices ` contract is the `StakeRegistry` itself via inheritance.
Updated for clarity.


gpsanant (approved):
thanks for this

PR #181, opened 2023-09-20
Remove Slashing Functionality (and related tests)

ChaoticWalrus:
Description copied from ticket:
We do not plan to have slashing as part of M2. Leaving this in production code confuses auditors & security researchers, requires undue maintenance, and just generally "muddies the waters". Slashing functionality has remained in our codebase despite these plans, partly for historical reasons and partly to ensure that we are not "painting ourselves into a corner" to too great a degree. Removing this functionality means clear acceptance of a decision that has already been made explicitly, as well as being made implicitly many times over -- slashing will not be "unpaused" without an upgrade. Given that the functionality is very likely to change without upgrading and changing it, by removing the functionality from our main codebase, we are simplifying what we actually plan on shipping, with no decrease in "real" / usable functions. If we wish to reintroduce the same functionality, we can simply revert the changes that remove it, or similar. Development of slashing features can still of course also be done at any time, on a branch.


wadealexc:
Missing calls to the slasher and slasher-related modifiers:
1. `DelegationManager._delegate` checks `!slasher.isFrozen`
2. `StrategyManager` modifiers: `onlyNotFrozen` and `onlyFrozen`
3. `StrategyManager._completeQueuedWithdrawal` checks `slasher.canWithdraw`
4. `EigenPodManager` modifiers: `onlyNotFrozen` and `onlyFrozen`
5. `EigenPodManager.exitUndelegationLimbo` and `_completeQueuedWithdrawal` both check `slasher.canWithdraw`

Do we want to remove these? I feel like we should

ChaoticWalrus:
> Do we want to remove these? I feel like we should

Personally I'm more hesitant to remove these.
We're already using them, and I think the design (or at least interface) is decently well-established, especially for the calls to `Slasher.canWithdraw`. I realize that as planned for M2, `canWithdraw` will always return 'true' and `isFrozen` will always return 'false' though, so I can see how these calls are also unnecessary for M2.

ChaoticWalrus:
Had some offline discussion on this with Alex + G.
Agreed on it being OK to leave these existing calls to `Slasher.canWithdraw` and `Slasher.isFrozen` in place, with auditors / similar being instructed to basically just treat these calls as no-ops.

PR #183, opened 2023-09-20
DelegationManager now initiates all undelegation-type actions

wadealexc:
1. Removes top-level undelegation functions from `EigenPodManager`, and any `undelegateIfPossible` parameters in both `EigenPodManager` and `StrategyManager`
2. `DelegationManager` initiates all undelegation operations and decreases operator shares in the same function, without the need for callbacks
3. Removes `EigenPodManager.enterUndelegationLimbo` method in favor of users calling `DelegationManager.undelegate`


src/contracts/core/StrategyManager.sol
```diff
@@ -290,16 +290,17 @@ contract StrategyManager is
         uint256[] calldata strategyIndexes,
         IStrategy[] calldata strategies,
         uint256[] calldata shares,
-        address withdrawer,
-        bool undelegateIfPossible
+        address withdrawer
```
ChaoticWalrus, line 293: I realize we agreed on this change, but any concerns about changing this function signature?
I guess we can still support the same function signature with a second method that just ignores the bool parameter, if we decide it's necessary

src/contracts/pods/EigenPodManager.sol
```diff
@@ -654,26 +640,22 @@ contract EigenPodManager is
...
-            address delegatedAddress = delegationManager.delegatedTo(podOwner);
-
+    function _enterUndelegationLimbo(address podOwner, address delegatedTo) internal returns (uint) {
+        if (podOwnerShares[podOwner] != 0 && !isInUndelegationLimbo(podOwner)) {
             // store the undelegation limbo details
             _podOwnerUndelegationLimboStatus[podOwner].active = true;
             _podOwnerUndelegationLimboStatus[podOwner].startBlock = uint32(block.number);
-            _podOwnerUndelegationLimboStatus[podOwner].delegatedAddress = delegatedAddress;
+            // TODO: is this even necessary to keep in state? i think it's literally only used to emit an event later
+            _podOwnerUndelegationLimboStatus[podOwner].delegatedAddress = delegatedTo;
```
ChaoticWalrus, line 653: This is necessary because we treat withdrawals as completeable or not (basically, "can they be slashed") based on the address that the staker was delegated to _at the time the withdrawal was created_. "Undelegation Limbo" is subject to the same delays/conditions as a withdrawal.

wadealexc: that makes sense. i definitely did the EPM first and when looking at the SM, started wondering why it doesn't have a more concrete concept of "undelegation limbo." Or, why the EPM's undelegation limbo didn't look more like the SM's queued withdrawal concept.

As I said, I think there's further room for simplifying this - but we can pin it for post M2!

ChaoticWalrus:
@gpsanant could you take a glance at this when you get a chance tomorrow? Wanting to merge this into my other PR which simplifies the undelegation interactions & then merge the changes

src/contracts/core/StrategyManager.sol
wadealexc: From my perspective, it's not a big deal - but I guess frontend or someone else might have opinions? idk who might be relying on it

ChaoticWalrus: Yeah, this could be a frontend issue. I'll can make a quick commit that should give us the ability to call with consistent function signature, and explain in a comment

ChaoticWalrus: actually, working through this, it might be a pain to maintain both versions of the function. @gpsanant any estimate on the pain the frontend team would have dealing with a change in function signature here?

ChaoticWalrus: discussed offline. This is definitely easier to coordinate one time on the frontend than keep maintaining two parallel functions, so I'm just going to close this thread out.

src/contracts/core/DelegationManager.sol
```diff
@@ -206,20 +207,33 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
             // force the staker into "undelegation limbo" in the EigenPodManager if necessary
-            eigenPodManager.forceIntoUndelegationLimbo(staker);
+            uint256 podShares = eigenPodManager.forceIntoUndelegationLimbo(staker, operator);
+
+            IStrategy[] memory strategies;
+            uint256[] memory strategyShares;
 
             // force a withdrawal of all of the staker's shares from the StrategyManager
-            queuedWithdrawal = strategyManager.forceTotalWithdrawal(staker);
+            (strategies, strategyShares, queuedWithdrawal)
```
gpsanant, line 216: weird syntax here, can we declare these in the call to SM?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -651,29 +634,26 @@ contract EigenPodManager is
...
+     *
+     * This method assumes the podOwner is delegated, as it's being called by the
+     * DelegationManager (and supplied with `delegatedTo`)
      */
-    function _enterUndelegationLimbo(address podOwner) internal {
-        if (!podOwnerHasNoDelegatedShares(podOwner) && delegationManager.isDelegated(podOwner)) {
-            // look up the address that the pod owner is currrently delegated to in EigenLayer
-            address delegatedAddress = delegationManager.delegatedTo(podOwner);
-
+    function _enterUndelegationLimbo(address podOwner, address delegatedTo) internal returns (uint) {
```
gpsanant, line 644: uint256

src/contracts/pods/EigenPodManager.sol
```diff
@@ -651,29 +634,26 @@ contract EigenPodManager is
...
             // emit event
             emit UndelegationLimboEntered(podOwner);
 
-            // undelegate all shares
-            uint256[] memory shareAmounts = new uint256[](1);
-            shareAmounts[0] = podOwnerShares[podOwner];
-            IStrategy[] memory strategies = new IStrategy[](1);
-            strategies[0] = beaconChainETHStrategy;
-            delegationManager.decreaseDelegatedShares(podOwner, strategies, shareAmounts, true/*undelegateIfPossible*/);
+            return podOwnerShares[podOwner];
```
gpsanant, line 654: would prefer to simplify the conditional here to `if (!isInUndelegationLimbo(podOwner))` and `return podOwnerShares[podOwner];` outside of the conditional

src/contracts/core/DelegationManager.sol
gpsanant: also, this should be called `withdrawalRoot` and an `@returns` should be added

src/contracts/core/DelegationManager.sol
```diff
@@ -206,20 +207,33 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
-            if (msg.sender != staker) {
-                emit StakerForceUndelegated(staker, operator);
+                unchecked {
+                    ++i;
+                }
             }
         }
 
         // actually undelegate the staker
         _undelegate(staker);
```
gpsanant, line 230: Looks like we are only calling `_undelegate` here. no need to make it internal then

src/contracts/pods/EigenPodManager.sol
wadealexc: out of curiosity, why do you like the long form? i tend to use `uint` to keep things concise, since everyone knows it's 256.

i'm fine with being verbose if that's the style we're using. unified style is important :)

src/contracts/pods/EigenPodManager.sol
wadealexc: hmm, that would be a meaningful change though - because now users that have never touched eigenpods would get state changes via this method.

wadealexc: i.e. someone trying to undelegate that purely holds LSTs in stmgr would not be able to skip the conditional in your version

wadealexc: what i think we want to capture with the conditional is a clear semantic distinction between addresses that have opted into restaking, and addresses that have not.

maybe we leave this as is and flag this as a future change?

src/contracts/core/DelegationManager.sol
ChaoticWalrus: > also, this should be called `withdrawalRoot` and an `@returns` should be added

there is a `@return` tag, but I'll correct the name.
Solidity doesn't appear to like assigning to a mix of newly-declared params and already-declared params in a single call, which is why I have that uglier syntax. 
I tried something like 
```
            (IStrategy[] memory strategies, uint256[] memory strategyShares, withdrawalRoot)
                = strategyManager.forceTotalWithdrawal(staker);
```
but the compiler throws an error:
```
Compiler run failed:
Error (2314): Expected identifier but got ')'
   --> src/contracts/core/DelegationManager.sol:216:92:
    |
216 |             (IStrategy[] memory strategies, uint256[] memory strategyShares, withdrawalRoot)
```

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: not sure I'm fully understanding -- we agreed we want the property that someone can't enter undelegation limbo if they have zero shares, right? can you write the suggestion you have?

src/contracts/core/DelegationManager.sol
ChaoticWalrus: good suggestion, IMO. addressed in commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/183/commits/bf1de8a06817eb639c1c5a265b1aedfe0c9e1a33

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: fixed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/183/commits/bf1de8a06817eb639c1c5a265b1aedfe0c9e1a33

src/contracts/core/DelegationManager.sol
ChaoticWalrus: naming changed in commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/183/commits/bf1de8a06817eb639c1c5a265b1aedfe0c9e1a33
let me know if you can sort out making the cleaner syntax work correctly?

src/contracts/core/DelegationManager.sol
```diff
@@ -226,15 +226,18 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
             }
         }
 
-        // actually undelegate the staker
-        _undelegate(staker);
+        // actually undelegate the staker -- only make storage changes + emit an event if the staker is actively delegated, otherwise do nothing
+        if (isDelegated(staker)) {
+            emit StakerUndelegated(staker, operator);
```
wadealexc, line 231: Do we need `if (isDelegated(staker))`? Surely that's enforced above, and we would want this to fail if false.

Hmm, looking above, I think there may not be a hard requirement? Let's add `require(operator != 0)` or `require(isDelegated(staker))`

ChaoticWalrus: yeah, there's documentation on this being a no-op in this case, but I agree given the other changes that this require statement makes sense to add now.

ChaoticWalrus: addressed this in a commit. https://github.com/Layr-Labs/eigenlayer-contracts/pull/183/commits/be217b232c368f700dc4cb7cbacf5a805cdd450d

gpsanant (approved):
Good change, approved from my end

src/contracts/pods/EigenPodManager.sol
gpsanant: Yep, I was ok with that, but don't feel strongly

PR #184, opened 2023-09-20
Add pausing to EigenDA/Registry contracts

8sunyuan:
Added pausing to the following functions. Although we may not want to have pausing ability for deregistering operators and let them disassociate from unethical services.

BLSRegistryCoordinatorWithIndices
- `registerOperatorWithCoordinator`
- `deregisterOperatorWithCoordinator`

Update:
Reverted on adding pausing to BLSPublicKeyCompendium


gpsanant:
i think the plan is ideally to make the BLSPublicKeyCompendium non upgradable soon (post audit), so we i think should remove pausing from it.

gpsanant (approved):
LGTM

PR #185, opened 2023-09-21
Further changes to newpodchanges

Sidu28:
There are several small tasks we need to accomplish once the "newpodchanges" branch is merged into master:

- [x] We need to order the functions in EigenPods according to this:
- - external callable by anyone
- - external callable by pod owner
- - external callable by EPM
- - internal state changing
- - internal pure
- [x] More explicit comments explaining every one of the constructed beacon chain proof indicies. These are particularly hard to understand why they are the way they are so comments are helpful here.
- [x] Route ETH withdrawn via the withdrawNonBeaconChainETHBalanceWei() function through Delayed Withdrawal Router.
- [x] Change events tfrom validatorIndex to validatorPKHash
- [x] Change naming of timestamp field in events to be more specific

The ticket covering this is https://github.com/Layr-Labs/eignlayr-contracts/issues/962

### DIFF
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -357,7 +357,7 @@ library BeaconChainProofs {
      *  hh.Reset()
      */
     function hashValidatorBLSPubkey(bytes memory validatorPubkey) internal pure returns (bytes32 pubkeyHash) {
-        require(validatorPubkey.length == 48, "Input should be 32 bytes in length");
+        require(validatorPubkey.length == 48, "Input should be 48 bytes in length");
         bytes memory padding = new bytes(16);
         bytes memory result = new bytes(64);
 
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -213,6 +213,10 @@ library BeaconChainProofs {
         * Therefore, the index of the balance of a validator is validatorIndex/4
         */
         uint256 balanceIndex = uint256(validatorIndex/4);
+        /**
+        * Note: Merkleization of the balance root tree uses MerkleizeWithMixin, i.e., the length of the array is hashed with the root of 
+        * the array.  Thus we shift the BALANCE_INDEX over by BALANCE_TREE_HEIGHT + 1 and not just BALANCE_TREE_HEIGHT.
+        */
         balanceIndex = (BALANCE_INDEX << (BALANCE_TREE_HEIGHT + 1)) | balanceIndex;
 
         require(Merkle.verifyInclusionSha256({proof: validatorBalanceProof, root: beaconStateRoot, leaf: balanceRoot, index: balanceIndex}),
@@ -327,6 +331,10 @@ library BeaconChainProofs {
             * intermediate root indexes from the bottom of the sub trees (the withdrawal container) to the top, the blockHeaderRoot.
             * Then we calculate merkleize the withdrawalFields container to calculate the the withdrawalRoot.
             * Finally we verify the withdrawalRoot against the executionPayloadRoot.
+            *
+            *
+            * Note: Merkleization of the withdrawals root tree uses MerkleizeWithMixin, i.e., the length of the array is hashed with the root of 
+            * the array.  Thus we shift the WITHDRAWALS_INDEX over by WITHDRAWALS_TREE_HEIGHT + 1 and not just WITHDRAWALS_TREE_HEIGHT.
             */
             uint256 withdrawalIndex = WITHDRAWALS_INDEX << (WITHDRAWALS_TREE_HEIGHT + 1) | uint256(withdrawalProofs.withdrawalIndex);
             bytes32 withdrawalRoot = Merkle.merkleizeSha256(withdrawalFields);
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -349,4 +349,26 @@ library BeaconChainProofs {
         }
     }
 
+    /**
+     * @notice This function replicates the ssz hashing of a validator's pubkey, outlined below:
+     *  hh := ssz.NewHasher()
+     *  hh.PutBytes(validatorPubkey[:])
+     *  validatorPubkeyHash := hh.Hash()
+     *  hh.Reset()
+     */
+    function hashValidatorBLSPubkey(bytes memory validatorPubkey) internal pure returns (bytes32 pubkeyHash) {
+        require(validatorPubkey.length == 48, "Input should be 32 bytes in length");
+        bytes memory padding = new bytes(16);
+        bytes memory result = new bytes(64);
+
+        for (uint i = 0; i < validatorPubkey.length; i++) {
+            result[i] = validatorPubkey[i];
+        }
+        for (uint i = 0; i < padding.length; i++) {
+            result[i + validatorPubkey.length] = padding[i];
+        }
+        
+        return sha256(abi.encodePacked(result));
+    }
+
 }
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -361,10 +361,10 @@ library BeaconChainProofs {
         bytes memory padding = new bytes(16);
         bytes memory result = new bytes(64);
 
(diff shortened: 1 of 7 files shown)


src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -213,6 +213,10 @@ library BeaconChainProofs {
         * Therefore, the index of the balance of a validator is validatorIndex/4
         */
         uint256 balanceIndex = uint256(validatorIndex/4);
+        /**
+        * Note: Merkleization of the balance root tree uses MerkleizeWithMixin, i.e., the length of the array is hashed with the root of 
+        * the array.  Thus we shift the BALANCE_INDEX over by BALANCE_TREE_HEIGHT + 1 and not just BALANCE_TREE_HEIGHT.
+        */
```
ChaoticWalrus, line 219: I'm trying to actually fully understand this.
Looks like you're referring to this code https://github.com/ferranbt/fastssz/blob/v0.1.3/hasher.go#L269-L288 
But the spec itself is described here https://github.com/ethereum/consensus-specs/blob/28e3c807e29f0125f6871fbc06ec3465597b8175/ssz/simple-serialize.md?plain=1#L236
I can reference your proof-generation code here https://github.com/Layr-Labs/apg/blob/master/pkg/eigenlayer/proofs.go#L119-L160
which refers to the `beaconstate_ssz.go` file (seems like here https://github.com/attestantio/go-eth2-client/blob/v0.18.3/spec/deneb/beaconstate_ssz.go#L763-L776 ?)
So is the crux of the idea here that the apparent size of the container holding balances ends up being increased by 1 field because of this "mixin" step?

src/contracts/pods/EigenPod.sol
```diff
@@ -99,17 +99,17 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     event EigenPodStaked(bytes pubkey);
 
     /// @notice Emitted when an ETH validator's withdrawal credentials are successfully verified to be pointed to this eigenPod
-    event ValidatorRestaked(uint40 validatorIndex);
+    event ValidatorRestaked(uint40 validatorIndex, bytes32 validatorPubkeyHash, bytes validatorPubkey);
```
ChaoticWalrus, line 102: to my knowledge, events with dynamic length arguments like `bytes validatorPubkey` will _actually_ output the keccak256 hash of the argument. might be fine for indexing, but if `validatorPubkeyHash` is already the keccak256 hash (and not a different kind of hash) then this is might be redundant info to be emitting?
will dig in more, this is a bit a note to myself.

src/contracts/pods/EigenPod.sol
```diff
@@ -706,6 +718,28 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
 
+    /**
+     * @notice This function replicates the ssz hashing of a validator's pubkey, outlined below:
+     *  hh := ssz.NewHasher()
+     *  hh.PutBytes(validatorPubkey[:])
+     *  validatorPubkeyHash := hh.Hash()
+     *  hh.Reset()
+     */
+    function _hashPubkey(bytes memory validatorPubkey) internal pure returns (bytes32 pubkeyHash) {
+        require(validatorPubkey.length == 48, "Input should be 32 bytes in length");
```
ChaoticWalrus, line 729: incorrect revert message, '32' in place of '48'.
might make more sense for this function to live in the BeaconChainProofs library, especially given it's 'pure' nature? IDK though, am not personally strongly opinionated on where this lives

src/contracts/pods/EigenPod.sol
```diff
@@ -706,6 +718,28 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        bytes memory padding = new bytes(16);
+        bytes memory result = new bytes(64);
+
+        for (uint i = 0; i < validatorPubkey.length; i++) {
+            result[i] = validatorPubkey[i];
+        }
+        for (uint i = 0; i < padding.length; i++) {
+            result[i + validatorPubkey.length] = padding[i];
+        }
```
ChaoticWalrus, line 738: nit: `uint` => `uint256`

ChaoticWalrus: bigger issue here though -- is there a reason you aren't just using bytes.concat or abi.encodePacked to combine these bytes objects? looping byte-by-byte seems quite painful/inefficient

ChaoticWalrus: documentation on `bytes.concat` with the version of solidity we're using: https://docs.soliditylang.org/en/v0.8.12/types.html#the-functions-bytes-concat-and-string-concat

ChaoticWalrus (changes requested):
Dropped a few comments, please respond/resolve as you can.
The EigenPod changes are a little hard to review given that they're coupled with reordering the functions, but it is what it is.

Looks like this PR is leaving one open thread from the previous PR. Note: I think this probably makes sense to accomplish in another separate PR (see my above comment about the changes to EigenPods being a little hard to parse in this PR), but I want to link it again to make sure we're tracking it https://github.com/Layr-Labs/eigenlayer-contracts/pull/127#discussion_r1332291243 -- @Sidu28 if you could file an issue for this I'd appreciate it.

src/contracts/pods/EigenPod.sol
Sidu28: Oh interesting, didn't know that.  Its really important for the indexing that they have both the hash and the preimage.

src/contracts/pods/EigenPod.sol
Sidu28: I agree, much more appropriate for BeaconChainProofs lib, moved

src/contracts/pods/EigenPod.sol
Sidu28: Fixed

src/contracts/pods/EigenPod.sol
ChaoticWalrus: gotcha. I think because there's some extra padding happening here before hashing, I can see how having both in the event makes sense.

src/contracts/libraries/BeaconChainProofs.sol
Sidu28: yes, that's right.  Lets take a simple example:

Lets say I have array X with 8 elements.  When I merkleize it, it will have log 8 = 3 levels.  MerkleizeWithMixin takes root(X) and then hashes it again:  final_root = hash(len(X), root(X)).  So this adds an additional level to the tree.

Sidu28: The "mixin" refers to mixing in the length of te array into the merkelization

Sidu28:
> Dropped a few comments, please respond/resolve as you can. The EigenPod changes are a little hard to review given that they're coupled with reordering the functions, but it is what it is.
> 
> Looks like this PR is leaving one open thread from the previous PR. Note: I think this probably makes sense to accomplish in another separate PR (see my above comment about the changes to EigenPods being a little hard to parse in this PR), but I want to link it again to make sure we're tracking it [#127 (comment)](https://github.com/Layr-Labs/eigenlayer-contracts/pull/127#discussion_r1332291243) -- @Sidu28 if you could file an issue for this I'd appreciate it.

Responded to the comment in #127 : https://github.com/Layr-Labs/eigenlayer-contracts/pull/127#discussion_r1334842699

src/contracts/pods/EigenPod.sol
Sidu28: yup, exactly

ChaoticWalrus: cool, resolving thread, thanks for going on this journey with me.

src/contracts/libraries/BeaconChainProofs.sol
ChaoticWalrus: Gotcha, that makes sense. I think we can continue to improve documentation -- including clarifying this -- but I understand it's difficult to explain the complexity / nuance of these things.

ChaoticWalrus (approved):

ChaoticWalrus:
Approving as of commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/185/commits/1e0816863ec02aec7f597396ce457d3089dd5e4e
LGTM, overall.

PR #189, opened 2023-09-23
Update eigenpod proof structs

gpsanant:
- update naming of certain fields
- remove uneeded fields


src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -110,38 +110,37 @@ library BeaconChainProofs {
...
         bytes executionPayloadProof;
         bytes timestampProof;
         bytes historicalSummaryBlockRootProof;
-        uint64 blockHeaderRootIndex;
+        uint64 blockRootIndex;
         uint64 historicalSummaryIndex;
         uint64 withdrawalIndex;
-        bytes32 blockHeaderRoot;
+        bytes32 blockRoot;
         bytes32 blockBodyRoot;
```
Sidu28, line 125: idk man its technically the blockHeader root.  idk if we should perpetuate the misnomer

Sidu28: So blockHeaderRoot is one level above the blockBodyRoot, which is why I liked the distinction.  But that being said,
blockHeader and block are basically the same:

`class BeaconBlockHeader(Container):
    slot: Slot
    proposer_index: ValidatorIndex
    parent_root: Root
    state_root: Root
    body_root: Root`
    
    `class BeaconBlock(Container):
    slot: Slot
    proposer_index: ValidatorIndex
    parent_root: Root
    state_root: Root
    body: BeaconBlockBody`
    
    
    So blockHeader is like an intermediate step to hashing block.  So I actually agree blockRoot is honestly the more clear option.

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -227,125 +226,120 @@ library BeaconChainProofs {
...
      * @param beaconStateRoot is the beacon chain state root to be proven against.
-     * @param latestBlockHeaderProof is the provided merkle proof
-     * @param latestBlockHeaderRoot is hashtree root of the latest block header in the beacon state
+     * @param stateRootProof is the provided merkle proof
+     * @param latestBlockRoot is hashtree root of the latest block header in the beacon state
      */
-    function verifyStateRootAgainstLatestBlockHeaderRoot(
+    function verifyStateRootAgainstLatestBlockRoot(
         bytes32 beaconStateRoot,
-        bytes32 latestBlockHeaderRoot,
```
Sidu28, line 235: maybe we should switch latestBlockRoot and beaconStateRoot in the param ordering?  I think everywhere else it is (root, leaf, proof) as input order

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -227,125 +226,120 @@ library BeaconChainProofs {
...
 
     /**
      * @notice This function verifies the slot and the withdrawal fields for a given withdrawal
-     * @param beaconStateRoot is the beacon chain state root to be proven against.
-     * @param withdrawalProofs is the provided set of merkle proofs
+     * @param withdrawalProof is the provided set of merkle proofs
      * @param withdrawalFields is the serialized withdrawal container to be proven
      */
-    function verifyWithdrawalProofs(
-        bytes32 beaconStateRoot,
```
Sidu28, line 252: I always like passing the root separately because it feels like its "separate" from the leaf/proofs

gpsanant: gas doesn't care about your feelings

src/contracts/libraries/BeaconChainProofs.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/189/commits/26c02279a2e18f8b6d4f98f7cc53db029628d056 cood callout

src/contracts/libraries/BeaconChainProofs.sol
Sidu28: so the optimization here is adding it to the struct?  Why does that save gas?

src/contracts/libraries/BeaconChainProofs.sol
ChaoticWalrus: great, then we have agreement on this. I just pushed a commit eliminating the unused 'blockBodyRoot' field as well: https://github.com/Layr-Labs/eigenlayer-contracts/pull/189/commits/e131d56f58ee455f8093a29f3e2a524ceb1e7d78

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -125,31 +125,31 @@ interface IEigenPod {
...
      * @param validatorFields are the fields of the "Validator Container", refer to consensus specs
      * @dev For more details on the Beacon Chain spec, see: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#validator
      */
     function verifyBalanceUpdate(
+        uint64 oracleBlockNumber,
         uint40 validatorIndex,
-        BeaconChainProofs.BalanceUpdateProofs calldata proofs,
-        bytes32[] calldata validatorFields,
-        uint64 oracleBlockNumber
+        BeaconChainProofs.BalanceUpdateProof calldata balanceUpdateProofs,
```
ChaoticWalrus, line 135: looks like the struct was renamed to singular 'BalanceUpdateProof' but the param was left as plural?

ChaoticWalrus (approved):
LGTM, added a commit + identified one nit to fix prior to merging.
Nice simplification and good renaming to help clarify things.

PR #190, opened 2023-09-24
emit by pubkeyhash

gpsanant:
this accounts for https://ethereum-magicians.org/t/eip-6914-reuse-withdrawn-validator-indices/15253 by connecting indices to pubkey hashes


gpsanant:
Converting to draft, because I don't actually think we need this because the same public key could redeposit.

We will need to upgrade eigenpods to allow proving restaked after a certain time has passed since https://github.com/Layr-Labs/eigenlayer-contracts/blob/9bc321e7e10a5ab0e73f5201d095104f168bdfe1/src/contracts/pods/EigenPod.sol#L657

PR #191, opened 2023-09-24
get rid of pubkeys

gpsanant:
don't do validator pubkey checks


ChaoticWalrus (approved):
Discussed offline with @gpsanant and I am in agreement with this change.
These event params were added to try to refine offchain indexing to conform with an EIP that would allow reusing validatorIndices, but since it would also allow reusing pubkeys, just indexing by pubkey instead of index is not strictly sufficient -- we'll solve this problem later.
The correct thing to do here is to ultimately update off-chain proof/indexing logic.

PR #192, opened 2023-09-25
Removing some vestigal code in Proofs

Sidu28:
We switched to requiring a historical summary proof which means we no longer need the blockHeaderProof (the proof for withdrawals within 27 hours of the oracle root).  This PR removes this from the proof lib and adds new test proof files without that proof


gpsanant (approved):
lgtm

PR #193, opened 2023-09-25
add flag for pausing undelegations and an associated unit test

ChaoticWalrus:
also add a comment to the existing unit test for pausing new delegations


wadealexc (approved):
LGTM. Thanks for adding tests :)

PR #195, opened 2023-09-25
Use default RPC and chain id during tests

8sunyuan:
Description:
Tests fail if `RPC_MAINNET` and `CHAIN_ID` are not set prior to running tests. 
Updated setUp() functions to use try catch blocks if env vars are not defined.
- `CHAIN_ID` defaults to local 31337
- `RPC_MAINNET` defaults to LlamaNodes RPC url


ChaoticWalrus (approved):
LGTM, thanks for taking care of this!

PR #196, opened 2023-09-25
Combine flows in EP

Sidu28:
This PR removes the "activate restaking" pathway and intializes pods as "restaked" upon deployment.  

It also removes the "withdrawbeforerestaking" method and adds that logic to the verifyWithdrawalCredentials method that withdraws any remaining balance to the DWR before proving the withdrawal credentials.


ChaoticWalrus:
> This PR removes the "activate restaking" pathway

You can't just remove this. We need to make this work for existing pod owners as well.

Sidu28:
> > This PR removes the "activate restaking" pathway
> 
> You can't just remove this. We need to make this work for existing pod owners as well.

Now that hasRestaked = true is in the initialze() function, when we upgrade the implementation, everyone will be "restaked" right?

ChaoticWalrus:
> Now that hasRestaked = true is in the initialze() function, when we upgrade the implementation, everyone will be "restaked" right?

No, not at all. The updated initialize function will only be called on _new_ EigenPods; the old version has already been called on existing ones!

Sidu28:
> > Now that hasRestaked = true is in the initialze() function, when we upgrade the implementation, everyone will be "restaked" right?
> 
> No, not at all. The updated initialize function will only be called on _new_ EigenPods; the old version has already been called on existing ones!

Ah I see, didn't know that.  Opened a new PR [here](https://github.com/Layr-Labs/eigenlayer-contracts/pull/199)

PR #197, opened 2023-09-25
combined all DWR calls into one for withdrawal

Sidu28:
Combines calls to DWR and EPM into one call across multiple withdrawals.  Before, we were making calls for every withdrawal that was proven


src/contracts/pods/EigenPod.sol
```diff
@@ -317,8 +317,18 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
             (withdrawalProofs.length == withdrawalFields.length), "EigenPod.verifyAndProcessWithdrawals: inputs must be same length"
         );
 
+        uint256 totalWithdrawalAmount;
```
gpsanant, line 320: totalAmountToSend

src/contracts/pods/EigenPod.sol
```diff
@@ -546,8 +557,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
          * If the validator status is inactive, then withdrawal credentials were never verified for the validator,
          * and thus we cannot know that the validator is related to this EigenPod at all!
          */
-        ValidatorInfo memory validatorInfo = _validatorPubkeyHashToInfo[validatorPubkeyHash];
```
gpsanant, line 549: why are we loading twice? stack too deep?

src/contracts/pods/EigenPod.sol
```diff
@@ -605,8 +614,8 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         address recipient,
         uint64 withdrawalAmountGwei,
         ValidatorInfo memory validatorInfo
-    ) internal {
-        uint256 amountToSend;
+    ) internal returns(WithdrawalInfo memory) {
```
gpsanant, line 617: imo the struct is bulky and not needed.

returning uint256,int256 is better

Sidu28: stack too deep ma boi

src/contracts/pods/EigenPod.sol
Sidu28: yup

src/contracts/pods/EigenPod.sol
Sidu28: fixed

src/contracts/pods/EigenPod.sol
```diff
@@ -317,8 +317,18 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        uint256 totalAmountToSend;
+        int256 totalSharesDelta;
         for (uint256 i = 0; i < withdrawalFields.length; i++) {
-            _verifyAndProcessWithdrawal(oracleTimestamp, withdrawalProofs[i], validatorFieldsProofs[i], validatorFields[i], withdrawalFields[i]);
+            WithdrawalInfo memory withdrawalInfo = _verifyAndProcessWithdrawal(oracleTimestamp, withdrawalProofs[i], validatorFieldsProofs[i], validatorFields[i], withdrawalFields[i]);
+            totalAmountToSend += withdrawalInfo.amountToSend;
+            totalSharesDelta += withdrawalInfo.sharesDelta;
+        }
+        // send ETH to the `recipient` via the DelayedWithdrawalRouter, if applicable
+        _sendETH_AsDelayedWithdrawal(podOwner, totalAmountToSend);
```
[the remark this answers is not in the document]
Sidu28: fixed

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -50,6 +50,13 @@ interface IEigenPod {
         VALIDATOR_STATUS status;
     }
 
+    struct WithdrawalInfo {
+        //amount to send to a podOwner from a proven withdrawal
+        uint256 amountToSend;
+        //difference in shares to be recorded in the eigenPodManager
+        int256 sharesDelta;
+    }
```
ChaoticWalrus, line 58: is there something else we can name this, perhaps more specific or descriptive?
even just adding a top-level comment to the struct definition (perhaps just capturing its usage) might help.
also going to note a nit about spacing; should have a space between the slashes and start of comment

src/contracts/pods/EigenPod.sol
```diff
@@ -317,8 +317,20 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+            WithdrawalInfo memory withdrawalInfo = _verifyAndProcessWithdrawal(oracleTimestamp, withdrawalProofs[i], validatorFieldsProofs[i], validatorFields[i], withdrawalFields[i]);
+            totalAmountToSend += withdrawalInfo.amountToSend;
+            totalSharesDelta += withdrawalInfo.sharesDelta;
+        }
+        // send ETH to the `recipient` via the DelayedWithdrawalRouter, if applicable
+        if (totalAmountToSend != 0) {
+            _sendETH_AsDelayedWithdrawal(podOwner, totalAmountToSend);
+        }
+        //update podOwner's shares in the strategy manager
+        if (totalSharesDelta != 0){
```
ChaoticWalrus, line 332: nit: spacing
also noticing that we are inputting the `oracleTimestamp` into `_verifyAndProcessWithdrawal` repeatedly -- can we not move that logic out of the loop here too?

src/contracts/interfaces/IEigenPod.sol
ChaoticWalrus: suggestion:
```
    /** 
     * @notice struct used to store amounts related to proven withdrawals in memory. Used to help
     * manage stack depth and optimize the number of external calls, when batching withdrawal operations.
     */
    struct ProvenWithdrawal {
        // amount to send to a podOwner from a proven withdrawal
        uint256 amountToSend;
        // difference in shares to be recorded in the eigenPodManager, as a result of the withdrawal
        int256 sharesDelta;
    }
```

ChaoticWalrus (changes requested):
This PR is a really good idea. Added a few comments that I'd like to see addressed.

src/contracts/interfaces/IEigenPod.sol
Sidu28: fixed

src/contracts/pods/EigenPod.sol
```diff
@@ -583,17 +595,16 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
 
             /**
             * if the validator's withdrawable epoch is less than or equal to the slot's epoch, then the validator has fully withdrawn because
             * a full withdrawal is only processable after the withdrawable epoch has passed.
             */
             // reference: uint64 withdrawableEpoch = Endian.fromLittleEndianUint64(validatorFields[BeaconChainProofs.VALIDATOR_WITHDRAWABLE_EPOCH_INDEX]);
-            if (Endian.fromLittleEndianUint64(validatorFields[BeaconChainProofs.VALIDATOR_WITHDRAWABLE_EPOCH_INDEX]) <= slot/BeaconChainProofs.SLOTS_PER_EPOCH) {
-                _processFullWithdrawal(validatorIndex, validatorPubkeyHash, withdrawalHappenedTimestamp, podOwner, withdrawalAmountGwei, validatorInfo);
+            if (Endian.fromLittleEndianUint64(validatorFields[BeaconChainProofs.VALIDATOR_WITHDRAWABLE_EPOCH_INDEX]) <= (Endian.fromLittleEndianUint64(withdrawalProof.slotRoot))/BeaconChainProofs.SLOTS_PER_EPOCH) {
+                return _processFullWithdrawal(validatorIndex, validatorPubkeyHash, withdrawalHappenedTimestamp, podOwner, withdrawalAmountGwei, _validatorPubkeyHashToInfo[validatorPubkeyHash]);
```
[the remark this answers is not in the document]
Sidu28: fixed, let me know if the break up follows our general style guideline

src/contracts/pods/EigenPod.sol
Sidu28: good call, fixed

src/contracts/pods/EigenPod.sol
```diff
@@ -537,6 +550,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
          * *prior* to the proof provided in the `verifyWithdrawalCredentials` function.
          */
         proofIsForValidTimestamp(Endian.fromLittleEndianUint64(withdrawalProof.timestampRoot))
+        returns (VerifiedWithdrawal memory)
     {
         uint64 withdrawalHappenedTimestamp = _computeTimestampAtSlot(Endian.fromLittleEndianUint64(withdrawalProof.slotRoot));
```
ChaoticWalrus, line 555: does this timestamp check still make sense here / is it even correct?
seems quite odd that we essentially have 3 different timestamps here -- the `oracleTimestamp`, `withdrawalProof.timestampRoot`, and converting the `withdrawalProof.slotRoot`  to a timestamp !

Sidu28: G added timestamp root as a field in the proof but this line didn't get updated - thanks for noting it, fixing it now

ChaoticWalrus: OK excellent, really glad to see the simplification and adoption of more consistency here.
Should we move the timestamp check (`proofIsForValidTimestamp(Endian.fromLittleEndianUint64(withdrawalProof.timestampRoot))`) outside of the loop, so we aren't checking it each time?

ChaoticWalrus: nevermind, seems like we have to do this in the loop, am re-resolving this thread.

ChaoticWalrus (approved):

src/contracts/pods/EigenPod.sol
Sidu28: Discussed on slack, this is the withdrawal timestamp not the oracle timestamp, it needs to be checked for each proof

PR #198, opened 2023-09-26
comply with telepathy interfaces

gpsanant:


script/misc/SubmitCredProof.s.sol
```diff
@@ -0,0 +1,58 @@
...
+// source .env
+
+// # To deploy and verify our contract
+// forge script script/misc/DeployStrategy.s.sol:DeployStrategy --rpc-url $RPC_URL  --private-key $PRIVATE_KEY --broadcast -vvvv
+
+contract SumitCredProof is Script, Test {
+    Vm cheats = Vm(HEVM_ADDRESS);
+
+    // EigenLayer Contracts
+    EigenPod public eigenPod = EigenPod(payable(0x77Da2A808Cd18Fe8Be57D06EcDBC13be92CD3879));
```
Sidu28, line 19: shouldn't we load this from an env file

Sidu28 (changes requested):
LGTM for the most part, feel like the script should be reading from some input files not hardcoding this values?  So we can use it to verify other shit

gpsanant:
script meant to be in .gitignore myb

Sidu28 (approved):

PR #199, opened 2023-09-26
Combine activate restaking and deploy flows

Sidu28:
New pods deployed post M2 will not have an activate restaking() option.  They will have to prove their WC to withdraw

### DIFF
--- a/src/contracts/pods/EigenPod.sol
+++ b/src/contracts/pods/EigenPod.sol
@@ -188,6 +188,12 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     function initialize(address _podOwner) external initializer {
         require(_podOwner != address(0), "EigenPod.initialize: podOwner cannot be zero address");
         podOwner = _podOwner;
+        /**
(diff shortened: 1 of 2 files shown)


src/test/EigenPod.t.sol
```diff
@@ -262,6 +262,9 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
     function testWithdrawBeforeRestaking() public {
```
gpsanant, line 262: we should test that new pods have hasRestaked set to true

Sidu28: added

gpsanant (approved):
lgtm

src/contracts/pods/EigenPod.sol
```diff
@@ -188,6 +188,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     function initialize(address _podOwner) external initializer {
         require(_podOwner != address(0), "EigenPod.initialize: podOwner cannot be zero address");
         podOwner = _podOwner;
+        hasRestaked = true;
```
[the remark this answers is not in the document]
Sidu28: added

PR #200, opened 2023-09-26
Draft: Add Prettier formatting and slither-action to workflows

8sunyuan:
Description: 
- Adding linting/formatting rules to the repo. Using [slither-action](https://github.com/marketplace/actions/slither-action) to our CI for more visibility on static analysis.
- Added .prettierrc config and updated .solhint.json
- Styling should mostly align with https://docs.soliditylang.org/en/latest/style-guide.html 

Perhaps the slither-action workflow can only be used for merging into `master` as it is quite slow and we don't need to check too often.
If formatting rules agreed on, I will format src/contracts files before merging into master with 
`npx prettier --write --plugin=prettier-plugin-solidity 'src/contracts/**/*.sol'`


ChaoticWalrus (changes requested):
Looks like a good PR. Did a bit of a spot-check on the formatting, which seems like an improvement to me -- there's a couple things which aren't necessarily the best, but I feel like that'll always happen! 
Looks like CI is breaking due to the Prover patch file not having the new format -- I'll sort that out quickly if I can.

ChaoticWalrus (approved):

ChaoticWalrus:
Just marked as approved; the small commit I pushed looks like it fixed the CI error.

PR #204, opened 2023-09-28
DelegationFaucet for M2 Testnet

8sunyuan:
Description:
- For M2, we are creating a quorum with a dummy ERC20 token. Operators can call `DelegationFaucet.mintDepositAndDelegate` which uses a `DelegationFaucetStaker` contract (deploys per-operator if it doesn't already exist) that gets minted the dummy ERC20 token and delegates to the operator.
- Copied most of the logic over from `Whitelister.sol`

Todo:
- Events
- Further testing

### DIFF
--- a/script/whitelist/delegationFaucet/DelegationFaucet.sol
+++ b/script/whitelist/delegationFaucet/DelegationFaucet.sol
@@ -1,0 +1,173 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
+
+import "src/contracts/interfaces/IStrategyManager.sol";
+import "src/contracts/interfaces/IStrategy.sol";
+import "src/contracts/interfaces/IDelegationManager.sol";
+import "src/contracts/interfaces/IWhitelister.sol";
+import "src/contracts/interfaces/IDelegationFaucet.sol";
+import "script/whitelist/delegationFaucet/DelegationFaucetStaker.sol";
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/Create2.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+
+import "../ERC20PresetMinterPauser.sol";
+/**
+ * @title DelegationFaucet for M2
+ * @author Layr Labs, Inc.
+ * @notice Faucet contract setup with a dummy ERC20 token and strategy for M2 testnet release.
+ * Each operator address gets a staker contract assigned to them deterministically.
+ * This contract assumes minting role of the ERC20 token and that the ERC20 token has a whitelisted strategy.
+ */
+contract DelegationFaucet is IDelegationFaucet, Ownable {
+    IStrategyManager immutable strategyManager;
+    ERC20PresetMinterPauser immutable stakeToken;
+    IStrategy immutable stakeStrategy;
+    IDelegationManager delegation;
+
+    uint256 public constant DEFAULT_AMOUNT = 100e18;
+
+    //TODO: Deploy ERC20PresetMinterPauser and a corresponding StrategyBase for it
+    //TODO: Transfer ownership of Whitelister to multisig after deployment
+    //TODO: Give mint/admin/pauser permssions of whitelistToken to Whitelister and multisig after deployment
+    //TODO: Give up mint/admin/pauser permssions of whitelistToken for deployer
+    constructor(
+        IStrategyManager _strategyManager,
+        IDelegationManager _delegation,
+        ERC20PresetMinterPauser _token,
+        IStrategy _strategy
+    ) {
+        strategyManager = _strategyManager;
+        delegation = _delegation;
+        stakeToken = _token;
+        stakeStrategy = _strategy;
+    }
+
+    /**
+     * Deploys a Staker contract if not already deployed for operator. Staker gets minted _depositAmount or
+     * DEFAULT_AMOUNT if _depositAmount is 0. Then Staker contract deposits into the strategy and delegates to operator.
+     * @param _operator The operator to delegate to
+     * @param _depositAmount The amount of stakeToken to mint to the staker and deposit into the strategy
+     */
+    function mintDepositAndDelegate(address _operator, uint256 _depositAmount) public onlyOwner {
+        address staker = getStaker(_operator);
+        // Set deposit amount
+        if (_depositAmount == 0) {
+            _depositAmount = DEFAULT_AMOUNT;
+        }
+        stakeToken.mint(staker, _depositAmount);
+
+        // Deploy staker address if not already deployed, staker constructor will approve the StrategyManager to spend the stakeToken
+        if (!Address.isContract(staker)) {
+            Create2.deploy(
+                0,
+                bytes32(uint256(uint160(_operator))),
+                abi.encodePacked(type(DelegationFaucetStaker).creationCode, abi.encode(strategyManager, stakeToken))
+            );
+        }
+
+        // deposit into stakeToken strategy
+        depositIntoStrategy(staker, stakeStrategy, stakeToken, _depositAmount);
+
+        // delegate to operator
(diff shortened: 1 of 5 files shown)


script/whitelist/delegationFaucet/DelegationFaucet.sol
```diff
@@ -0,0 +1,189 @@
...
+
+        // deposit into stakeToken strategy
+        uint256 shares = abi.decode(depositIntoStrategy(staker, stakeStrategy, stakeToken, _depositAmount), (uint256));
+        // increaseDelegatedShares if staker contract already delegated, o/w delegateTo operator
+        if (delegation.isDelegated(staker)) {
+            delegation.increaseDelegatedShares(staker, stakeStrategy, shares);
+        } else {
+            IDelegationManager.SignatureWithExpiry memory signatureWithExpiry;
+            delegateTo(_operator, signatureWithExpiry, bytes32(0));
+        }
```
ChaoticWalrus, line 78: the 'if' part of this should revert, since the caller won't have permission to call `DelegationManager.increaseDelegatedShares` (this function is only called by the StrategyManager or EigenPodManager contracts).
the 'else' part looks correct.
I would amend this to be:
```
        if (!delegation.isDelegated(staker)) {
            IDelegationManager.SignatureWithExpiry memory signatureWithExpiry;
            delegateTo(_operator, signatureWithExpiry, bytes32(0));
        }
```

I'll note here as well that this implementation won't work if the `_operator` has set a `delegationApprover` in the DelegationManager -- it might make sense to add a `signatureWithExpiry` input to this function, instead of always using an empty struct here

script/whitelist/delegationFaucet/DelegationFaucet.sol
```diff
@@ -0,0 +1,189 @@
...
+        uint256 shares
+    ) public onlyOwner returns (bytes memory) {
+        bytes memory data = abi.encodeWithSelector(
+            IDelegationManager.increaseDelegatedShares.selector,
+            strategy,
+            shares
+        );
+
+        return Staker(staker).callAddress(address(delegation), data);
+    }
```
ChaoticWalrus, line 119: as noted above, this should always revert. if the intention here is to add shares to a staker, we'd need to mint more tokens and have the staker deposit them.

script/whitelist/delegationFaucet/DeployDelegationFaucet.sol
```diff
@@ -0,0 +1,84 @@
...
+                    eigenLayerProxyAdminAddress,
+                    abi.encodeWithSelector(StrategyBase.initialize.selector, stakeToken, eigenLayerPauserRegAddress)
+                )
+            )
+        );
+
+        // Needs to be strategyManager.strategyWhitelister() to add STK strategy
+        IStrategy[] memory _strategy = new IStrategy[](1);
+        _strategy[0] = stakeTokenStrat;
+        strategyManager.addStrategiesToDepositWhitelist(_strategy);
```
ChaoticWalrus, line 68: good point, presumably needs to be accomplished separately / after deployment

src/test/DelegationFaucet.t.sol
```diff
@@ -0,0 +1,219 @@
...
+        // Mint token to Staker, deposit minted amount into strategy, and delegate to operator
+        assertTrue(
+            !Address.isContract(stakerContract),
+            "test_mintDepositAndDelegate_DeployStakerContract: staker contract shouldn't be deployed"
+        );
+        delegationFaucet.mintDepositAndDelegate(operator, _depositAmount);
+        assertTrue(
+            Address.isContract(stakerContract),
+            "test_mintDepositAndDelegate_DeployStakerContract: staker contract not deployed"
+        );
```
ChaoticWalrus, line 71: seems like here we can also check:
delegation status (staker now delegated to operator)
change in delegated amount (increase to operator, in strategy shares)
perhaps increase in deposited amount (shares that the staker has in strategy, seems like overkill though)

ChaoticWalrus: ah, looks like some of these checks are in the `test_mintDepositAndDelegate_TokenBalances` test below?
personally I think if we're doing the exact same operations, it makes sense to combine these tests and just do all the checks in a single one.

src/test/DelegationFaucet.t.sol
```diff
@@ -0,0 +1,219 @@
...
+            "test_mintDepositAndDelegate_StrategyAndOperatorShares: operator shares not updated correctly"
+        );
+    }
+
+    /**
+     * @notice Invariant test the before/after values for strategy shares and operator shares from multiple runs
+     */
+    /// forge-config: default.invariant.runs = 5
+    /// forge-config: default.invariant.depth = 20
+    function invariant_test_mintDepositAndDelegate_StrategyAndOperatorShares() public {
```
ChaoticWalrus, line 166: just so I understand correctly, what exactly is this being labeled an invariant test doing?
seems like its set to do 5 runs of 20 calls, but like, what's being done in each run? I would think you define something like a setup to be done once and a bunch of conditions to check after each call, but this doesn't quite look like that -- does this run all of the below, do a random function call, then another run through all of the below + another random function call, etc.?

ChaoticWalrus (changes requested):
Seems like a good start.
I pointed out a few things I'd like to see fixed / questions I have.
In general it could be nice to have a bit of coverage for some of the other functions (even just minimally testing the 'happy path'), but I don't want to waste time, so... I think this is probably fine. The most important pieces are probably:
a) the combined `mintDepositAndDelegate` function, which you have some coverage for already
b) making sure all the functions are marked onlyOwner (LGTM)
c) having the generalized `callAddress` functions, in case we need to force the system into doing something it's not explicitly designed for (already LGTM as well)

script/whitelist/delegationFaucet/DelegationFaucet.sol
8sunyuan: Ahh I see, looked over the contract and repo docs again. Calling `depositIntoStrategy()` which is done on L76 will already addShares to the staker and increase delegated shares if applicable. 
Will add the signature params as well.

script/whitelist/delegationFaucet/DelegationFaucet.sol
8sunyuan: This function is not needed and won't work anyways, adding more shares to a staker will just be called by the same function `mintDepositAndDelegate()`

src/test/DelegationFaucet.t.sol
8sunyuan: You can set targetContracts that will be called over course of a invariant test run. It defaults to all contracts defined in setUp. In each of those 5 runs, it will call 20 functions randomly with fuzzed inputs and then check your invariant function and make sure the assertions are still true.

ChaoticWalrus: Gotcha, makes sense.
So it's really just checking that every time `mintDepositAndDelegate` is called, the share amounts all change by the right amount and the staker contract is properly delegated, right? And the random calls here are kind of seeing if you can get into some state which messes up that behavior?
LGTM.

script/whitelist/delegationFaucet/DelegationFaucet.sol
8sunyuan: Added the fix to L80
```
        // delegateTo operator if not delegated
        if (!delegation.isDelegated(staker)) {
            delegateTo(_operator, _approverSignatureAndExpiry, _approverSalt);
        }

script/whitelist/delegationFaucet/DeployDelegationFaucet.sol
8sunyuan: Sounds good, commented this out in the script for now

src/test/DelegationFaucet.t.sol
8sunyuan: Updated with `test_mintDepositAndDelegate_CheckBalancesAndDeploys` test now

ChaoticWalrus (approved):

ChaoticWalrus:
LGTM now, thanks for this.

PR #205, opened 2023-09-28
Move state root against block root proof outside loop

Sidu28:
Move state root against block root proof outside validator and withdrawal proof loop

Created a new struct for stateRootProof

### DIFF
--- a/src/contracts/interfaces/IEigenPod.sol
+++ b/src/contracts/interfaces/IEigenPod.sol
@@ -174,6 +174,7 @@ interface IEigenPod {
     function verifyBalanceUpdate(
         uint64 oracleTimestamp,
         uint40 validatorIndex,
+        BeaconChainProofs.StateRootProof calldata stateRootProof,
         BeaconChainProofs.BalanceUpdateProof calldata balanceUpdateProof,
         bytes32[] calldata validatorFields
     ) external;
--- a/src/contracts/interfaces/IEigenPod.sol
+++ b/src/contracts/interfaces/IEigenPod.sol
@@ -153,8 +153,7 @@ interface IEigenPod {
      */
     function verifyWithdrawalCredentials(
         uint64 oracleTimestamp,
-        bytes32 beaconStateRoot,
-        bytes calldata stateRootProof,
+        BeaconChainProofs.StateRootProof calldata stateRootProof,
         uint40[] calldata validatorIndices,
         bytes[] calldata withdrawalCredentialProofs,
         bytes32[][] calldata validatorFields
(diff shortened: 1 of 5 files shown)


wadealexc (commented):
a

wadealexc (commented):
Left some ideas to make things read a bit nicer/clearer. Didn't go through the semantic changes as I'm still getting comfortable with this part of the code.

Won't make anything blocking though!

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -187,6 +189,8 @@ interface IEigenPod {
      */
     function verifyAndProcessWithdrawals(
         uint64 oracleTimestamp,
+        bytes32 beaconStateRoot,
+        bytes calldata stateRootProof,
```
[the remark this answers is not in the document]
Sidu28: actually I like that, will add

src/contracts/pods/EigenPod.sol
```diff
@@ -335,8 +342,9 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
      */
     function verifyWithdrawalCredentials(
         uint64 oracleTimestamp,
+        BeaconChainProofs.StateRootProof calldata stateRootProof,
         uint40[] calldata validatorIndices,
-        BeaconChainProofs.WithdrawalCredentialProof[] calldata withdrawalCredentialProofs,
+        bytes[] calldata withdrawalCredentialProofs,
```
ChaoticWalrus, line 347: seems like you should be deleting the `WithdrawalCredentialProof` struct type? AFAICT we aren't using it any more.

Sidu28: should be deleted, check the diffs in BeaconCHainProofs

ChaoticWalrus: also seems like the input might be better named `validatorFieldsProofs` instead of `withdrawalCredentialProofs`? since that is the semantic meaning -- this would make it clearer that these proofs are for the `validatorFields` input.

ChaoticWalrus (approved):
Splitting these proofs up feels like an excellent change; it feels like the data structures have clearer semantics now.
AFAICT the slot proof is still being checked everywhere it should be, _but I'd really like to see test coverage to be sure of this_ (could be in another PR, just expecting this to be tracked/noted somewhere).
I added comments about cleanup of the deprecated(?) WithdrawalCredentialProof and clarifying variable naming; get these done and I'm OK with merging.

src/contracts/pods/EigenPod.sol
ChaoticWalrus: > should be deleted, check the diffs in BeaconCHainProofs.

ah mb it's because it was "renamed" in the diffs.
makes sense, would still like the parameter name change.

Sidu28:
> Splitting these proofs up feels like an excellent change; it feels like the data structures have clearer semantics now. AFAICT the slot proof is still being checked everywhere it should be, _but I'd really like to see test coverage to be sure of this_ (could be in another PR, just expecting this to be tracked/noted somewhere). I added comments about cleanup of the deprecated(?) WithdrawalCredentialProof and clarifying variable naming; get these done and I'm OK with merging.

Will add to the test cases sheet

PR #207, opened 2023-09-29
Remove validator status check in processFullWithdrawal

Sidu28:
We removed the check to allow already withdrawn validators to prove full withdrawals for post-exit deposits on the beacon chain.  However this check was not removed in the _processFullWithdrawal function

PR adds two regression tests to ensure this bug does not resurface


ChaoticWalrus (approved):
LGTM, merging now.
I confirmed that the key regression test (`testWithdrawAfterFullWithdrawal`) fails with the bug present and passes with the modifications -- looks like a proper regression test, nice job 👍

PR #208, opened 2023-09-29
Move "undelegation limbo" concept from EigenPodManager to DelegationM…

ChaoticWalrus:
…anager

Conceptually, this makes more sense to live in the DelegationManager. Moving the functionality allows for better segmentation of system components, and removes a weird asymmetry where the EigenPodManager and StrategyManager were treated differently. This PR is mostly moving some logic from EigenPodManager to DelegationManager, modifying interfaces + behavior of the StrategyManager a bit.

## Summary of Changes:
- undelegation limbo is mututally exclusive with being delegated (a staker can _either_ be delegated or in undelegation limbo, never both at the same time)
- undelegation limbo is entered _exclusively_ via calls to `DelegationManager.undelegate`
- removal of `StrategyManager.forceTotalWithdrawal`, `EigenPodManager.forceIntoUndelegationLimbo`, and `EigenPodManager.exitUndelegationLimbo`, since these are no longer needed
- EigenPodManager and StrategyManager always communicate changes in staker shares to the DelegationManager, which handles undelegation limbo logic internally
- a staker cannot queue a new withdrawal when in undelegation limbo (potentially pending other changes)
- stakers can no longer withdraw funds directly when exiting undelegation limbo (this must be done in a separate call)

### Test Coverage (likely in a following PR):

- [x] tests that specifically trigger "undelegation limbo" and check for event emission / correct modification of delegated shares
- [x] tests that ensure a staker cannot queue a new withdrawal when in undelegation limbo (enforced in StrategyManager + EigenPodManager)
- [x] verify that an operator cannot enter undelegation limbo
- [x] test that a staker cannot exit undelegation limbo without already being in it!
- [x] test that a staker cannot re-delegate while in undelegation limbo
- [x] test that a staker can successfully call `DelegationManager.exitUndelegationLimbo`, verifying that the correct event is emitted and that state updates appropriately (specifically, that the staker's undelegation limbo status is reset to empty/null/default)
- [x] verify that a staker cannot be simultaneously in undelegation limbo and delegated to an operator
- [x] ~~tests to ensure that operator shares do not increase (or decrease) if the share amounts of a staker who is delegated to the operator change *and* the staker is in undelegation limbo~~ not needed -- see commit comment here https://github.com/Layr-Labs/eigenlayer-contracts/pull/208/commits/aaf90480dcdd80e665501d4b7442f2bdab56a69f
- [x] ~~verify that `DelegationManager.exitUndelegationLimbo` enforces the same delays as a queued withdrawal~~ edit: this was removed in commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/208/commits/6b53ba9c1d31cec10bd93fc15ee460bbd049d3b9 following discussion


ChaoticWalrus:
Marking this as ready-for-review now, after having 
See the top-level comment -- test coverage is _definitely_ missing here still.

src/contracts/core/DelegationManager.sol
```diff
@@ -201,19 +209,20 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
      * @dev Reverts if the `staker` is already undelegated.
+     * @dev Note that if a staker enters undelegation limbo as a result of this function, then they will be unable to redelegate until after
+     * successfully calling `exitUndelegationLimbo`.
      */
     function undelegate(
         address staker
-    ) external onlyWhenNotPaused(PAUSED_UNDELEGATION) returns (bytes32 withdrawalRoot) {
+    ) external onlyWhenNotPaused(PAUSED_UNDELEGATION) {
         require(isDelegated(staker), "DelegationManager.undelegate: staker must be delegated to undelegate");
+        require(!isInUndelegationLimbo(staker), "DelegationManager.undelegate: staker is already in undelegation limbo");
```
ChaoticWalrus, line 225: this check that the staker is not in undelegation limbo is technically redundant with the one above (that the staker *is* delegated), since that check passing already ensures this one will pass.
a new property of undelegation limbo is that a staker cannot simutaneously be both (a) in delegation limbo and (b) delegated to an operator. This is enforced with a check in the `_delegate` function.

TODO for myself: write a simple prover invariant showing that `isInUndelegationLimbo(staker) => !isDelegated(staker)`

src/contracts/core/DelegationManager.sol
```diff
@@ -224,34 +233,40 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
-                staker: staker,
-                strategy: beaconChainETHStrategy,
-                shares: podShares
-            });
-        }
-        // force-queue a withdrawal of all of the staker's shares from the StrategyManager, if necessary
-        if (strategyManager.stakerStrategyListLength(staker) != 0) {
-            IStrategy[] memory strategies;
-            uint256[] memory strategyShares;
-            (strategies, strategyShares, withdrawalRoot) = strategyManager.forceTotalWithdrawal(staker);
```
gpsanant, line 243: where did `forceTotalWithdrawal` go?

ChaoticWalrus: With this change, we no longer have a need for `StrategyManager.forceTotalWithdrawal` or `EigenPodManager.forceIntoUndelegationLimbo` -- delegation limbo is more "contained" just in the DelegationManager contract itself.
The flip side of this is that if a staker is in delegation limbo, we currently:

- prevent them from queuing a new withdrawal (in either the StrategyManager or DelegationManager)
- prevent them from re-delegating

I think however that we could allow queuing a withdrawal when in delegation limbo? Would just need to store the proper operator address for the queued withdrawal.
We could also then allow exiting delegation limbo early if a user subsequently removes all their shares via withdrawals.
I feel like these features could improve UX, but it's kind of weird because if implemented naively it might "restart the clock" on the withdrawal delay for the funds, so if we decide to add this we need to be careful about implementation.

src/contracts/core/DelegationManager.sol
ChaoticWalrus: resolved in https://github.com/Layr-Labs/eigenlayer-contracts/pull/208/commits/7d681c2bacf58f224f9ad818e5964063f3d9f48f
leaving this thread open because I think the note about the new property/invariant is important

src/contracts/core/DelegationManager.sol
```diff
@@ -264,8 +278,72 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+                    staker: msg.sender,
+                    strategy: strategies[i],
+                    shares: strategyShares[i]
+                });
+
+                unchecked {
+                    ++i;
+                }
+            }
+        }
```
ChaoticWalrus, line 346: I've realized that this logic is actually entirely unnecessary -- since we are enforcing that a staker cannot re-delegate while in undelegation limbo, `operator` here should always be the zero address, meaning no shares should ever get delegated here!

ChaoticWalrus:
Edited the top-level comment on this PR to capture some missing test coverage, and checked off the 2 boxes covered by the new Prover rules. I will try to add some more test coverage to this PR before it's merged

ChaoticWalrus:
Question for reviewers: should undelegation limbo still enforce the minimum withdrawal delay?
This felt necessary when stakers could exit undelegation limbo and withdraw funds from EigenLayer protocol in the same call. 
Now that a separate call / going through the withdrawal queue is necessary, users can end up in a situation where they essentially suffer 2x the minimum delay.
Since the minimum delay exists basically to try to ensure that funds cannot leave EigenLayer without providing some chance for automated/manual review and the ability to catch bugs, leaving this in seems unnecessary.
I am of the opinion that removing this delay likely makes sense, but would appreciate other input.

src/contracts/core/DelegationManager.sol
ChaoticWalrus: removed the logic in this commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/208/commits/f6164740c93e6b8360d536c17fd91f05936e5f62

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -66,6 +66,9 @@ abstract contract DelegationManagerStorage is IDelegationManager {
      */
     mapping(address => mapping(bytes32 => bool)) public delegationApproverSaltIsSpent;
 
+    // @notice Mapping: staker => UndelegationLimboStatus struct. Mapping is internal so we can have a getter that returns a memory struct.
+    mapping(address => IDelegationManager.UndelegationLimboStatus) internal _stakerUndelegationLimboStatus;
```
gpsanant, line 70: undelegation limbo might top pubkey compendium in terms of obnoxious naming

gpsanant:
Ok, to be clear, this sort of exposes LSD restaking to the same sort of "if im in limbo, new deposits cant be delegated" behaviour?

gpsanant:
The withdrawa

> Question for reviewers: should undelegation limbo still enforce the minimum withdrawal delay? This felt necessary when stakers could exit undelegation limbo and withdraw funds from EigenLayer protocol in the same call. Now that a separate call / going through the withdrawal queue is necessary, users can end up in a situation where they essentially suffer 2x the minimum delay. Since the minimum delay exists basically to try to ensure that funds cannot leave EigenLayer without providing some chance for automated/manual review and the ability to catch bugs, leaving this in seems unnecessary. I am of the opinion that removing this delay likely makes sense, but would appreciate other input.

ok great, i was looking for this comment!

I agree, on exiting undelegation limbo, require the "min withdrawal delay blocks" has passed.

ChaoticWalrus:
> Ok, to be clear, this sort of exposes LSD restaking to the same sort of "if im in limbo, new deposits cant be delegated" behaviour?

Yes. This gives consistent behavior to both LSTs and natively-restaked ETH. A staker who is in limbo does not have new shares delegated / cannot delegate new deposits prior to exiting limbo.

ChaoticWalrus:
> I agree, on exiting undelegation limbo, require the "min withdrawal delay blocks" has passed.

I'm suggesting we remove this check, actually! Tried to explain my reasoning above.

gpsanant:
> I'm suggesting we remove this check, actually! Tried to explain my reasoning above.

ok, so this would mean `timeToWithdraw = withdrawalDelayBlocks + unbondingPeriod`?

or?

src/contracts/core/StrategyManager.sol
```diff
@@ -515,6 +480,10 @@ contract StrategyManager is
     ) internal returns (bytes32) {
         require(strategies.length == shares.length, "StrategyManager.queueWithdrawal: input length mismatch");
         require(withdrawer != address(0), "StrategyManager.queueWithdrawal: cannot withdraw to zero address");
+        require(
```
gpsanant, line 483: can we remove the `receiveAsTokens` argument now? avoids complexity that will never be used now

gpsanant: refering to completeWithdrawal ofc

gpsanant:
sort of related thought: It will help for us to keep a level of "docs" that essentially go through the calls needed to delegate/undelegate/withdaw fully/withdraw partially. Would help the frontend team.

gpsanant:
Tests and changes overall look very good. Remaining detail to figure out wrt delays. would also want @wadealexc's approval before merge.

src/contracts/core/DelegationManager.sol
gpsanant: got it, thanks :)

src/contracts/core/DelegationManagerStorage.sol
stevennevins: yeah can we rename to Queue or something similar

stevennevins:
> Question for reviewers: should undelegation limbo still enforce the minimum withdrawal delay? This felt necessary when stakers could exit undelegation limbo and withdraw funds from EigenLayer protocol in the same call. Now that a separate call / going through the withdrawal queue is necessary, users can end up in a situation where they essentially suffer 2x the minimum delay. Since the minimum delay exists basically to try to ensure that funds cannot leave EigenLayer without providing some chance for automated/manual review and the ability to catch bugs, leaving this in seems unnecessary. I am of the opinion that removing this delay likely makes sense, but would appreciate other input.

I’m for aligning to a single delay. Seems possible with these changes

ChaoticWalrus:
> > I'm suggesting we remove this check, actually! Tried to explain my reasoning above.
> 
> ok, so this would mean `timeToWithdraw = withdrawalDelayBlocks + unbondingPeriod`?
> 
> or?

the delay for undelegation limbo would be the "unbondingPeriod" and the delay for a subsequent withdrawal (after the staker has exited undelegation limbo) would be "withdrawalDelayBlocks".

src/contracts/core/StrategyManager.sol
ChaoticWalrus: I think we still want something like this, to deal with the edge case of a reverting strategy -- allowing users to "cancel" a withdrawal still seems good, which is essentially what marking `receiveAsTokens = false` is.

gpsanant:
> the delay for undelegation limbo would be the "unbondingPeriod" and the delay for a subsequent withdrawal (after the staker has exited undelegation limbo) would be "withdrawalDelayBlocks".

this seems... long. is it worth adding custom logic to the SM/EPM to allow full withdrawal on exit if the delegation limbo lasted longer than the withdrawal delay?

src/contracts/core/DelegationManagerStorage.sol
ChaoticWalrus: so this is a state where a staker is essentially "waiting to finalize / fully enact" their undelegation from an operator. I think I suggested "purgatory" previously but don't like the associations.
'UndelegationQueue' could potentially work, although we've also had confusion around the word 'queue' with people perceiving these things to have some kind of ordering to them.

ChaoticWalrus:
> this seems... long. is it worth adding custom logic to the SM/EPM to allow full withdrawal on exit if the delegation limbo lasted longer than the withdrawal delay?

IMO not worth it / it's acceptable for now, at least.
In practice for M2, users will be able to immediately exit undelegation limbo with the change I'm suggesting. I'm OK revisiting if we should add other associated functionality as we prepare for the next release.

docs/core/DelegationManager.md
```diff
@@ -43,7 +43,7 @@ Registers the caller as an Operator in EigenLayer. The new Operator provides the
 * Sets `OperatorDetails` for the Operator in question
 * Delegates the Operator to itself
 * If the Operator has deposited into the `EigenPodManager` and is not in undelegation limbo, the `DelegationManager` adds these shares to the Operator's shares for the beacon chain ETH strategy.
-* For each of the three strategies in the `StrategyManager`, if the Operator holds shares in that strategy they are added to the Operator's shares under the corresponding strategy.
+* For each of the three strategies in the `StrategyManager`, if the staker is not in undelegation limbo and holds shares in that strategy, then they are added to the Operator's shares under the corresponding strategy.
```
wadealexc, line 46: I've been capitalizing "Staker" and "Operator" because they're "vocab" i defined in `RolesAndActors.md`

src/contracts/core/DelegationManager.sol
```diff
@@ -264,8 +278,38 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+     * @notice Called by a staker to exit the "undelegation limbo" mode.
+     * @param middlewareTimesIndex Passed on as an input to the `slasher.canWithdraw` function, to ensure that the caller can exit undelegation limbo.
+     * This is because undelegation limbo is subject to some of the same restrictions as completing a withdrawal, to ensure that a staker cannot use undelegation
+     * limbo to avoid potentially being subject to slashing.
+     * @dev Note that this checks against the frozen status of the operator *that the caller was delegated to when they entered undelegation limbo*, to provide
+     * the same kind of slashing eligibility as withdrawals via the StrategyManager or DelegationManager.
+     */
+    function exitUndelegationLimbo(
+        uint256 middlewareTimesIndex
+    ) external onlyNotFrozen(_stakerUndelegationLimboStatus[msg.sender].delegatedAddress) {
```
wadealexc, line 293: Is `onlyNotFrozen(msg.sender)` relevant here? That's what the current `EPM.exitUndelegationLimbo` has.

I think I'm becoming less of a fan of keeping in this slasher stuff - it makes me question its use during refactors, and the answer is inevitably something like "we don't know yet." But I worry that by leaving it in, we're going to become blind to its presence and not think about it hard enough when we eventually do add it.

src/contracts/core/DelegationManager.sol
```diff
@@ -264,8 +278,38 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+                middlewareTimesIndex
+            ),
+            "DelegationManager.exitUndelegationLimbo: shares in limbo are still slashable"
+        );
+
+        // a staker cannot be simultaneously in delegation limbo and delegated to an operator, so we do not need to return any shares to the delegation system here
+
+        // delete the pod owner's undelegation limbo details and emit an event
+        delete _stakerUndelegationLimboStatus[msg.sender];
+        emit UndelegationLimboExited(msg.sender);
```
wadealexc, line 312: I'm not sure I follow the logic behind not enforcing a delay here. Seems like this allows you to very quickly swap out your Operator if you think they're about to get slashed or something.

src/contracts/core/DelegationManagerStorage.sol
gpsanant: i meant this as a complement @stevennevins

src/contracts/core/DelegationManager.sol
ChaoticWalrus: specifically it is the address that the Staker *was* delegated to being checked against here.
That being said, this appears to be redundant with the call to `Slasher.canWithdraw` below; I will remove it.
We'll need to revisit these calls regardless. I can understand your stance, and my own position feels increasingly harder to justify. Perhaps the best explanation I can offer is "this seems compliant with the (preliminary) interface of the Slasher, and ensures that the DelegationManager has the sufficient info to communicate here / perform the same check as the StrategyManager + EigenPodManager.

ChaoticWalrus: removed the modifier in a commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/208/commits/a3963f2c4f446269477e759e7c2332450917759a

docs/core/DelegationManager.md
ChaoticWalrus: I gave this a shot in a commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/208/commits/e77dd42906faa166eae58c8583448809752602a1
Feels like an uphill battle, but I'll try to be consistent with this going forwards

src/contracts/core/DelegationManager.sol
ChaoticWalrus: discussed this some offline. the intention is that
- the minimum withdrawal delay is an extra safety measure on funds leaving EigenLayer, not designed to have anything to do with ensuring slashing
- the call to `slasher.canWithdraw` should enforce slashing eligibility, when slashing is actually enabled

src/contracts/core/DelegationManager.sol
wadealexc: I'm happy for now knowing that it is in fact redundant with the call to `Slasher.canWithdraw`, and can be removed.

Is the current `onlyNotFrozen(msg.sender)` check completely unneeded in the EPM then?

ChaoticWalrus: > Is the current onlyNotFrozen(msg.sender) check completely unneeded in the EPM then?

on completing a queued withdrawal? yes, seems redundant as well. likewise with the StrategyManager.

I guess the conundrum here is that the `Slasher.canWithdraw` function doesn't actually check frozen status of the operator -- so perhaps for now it actually makes sense to put this modifier back in place 😞

ChaoticWalrus: I actually think leaving the modifier in is the sanest thing to do right now, so am reverting my previous commit which removed it.
For M2 testnet we are treating these as no-ops, so it shouldn't really matter either way, but I think what we should have here is consistency for now.
Going forward I think we should absorb this modifier's behavior into the `Slasher.canWithdraw` function (and further refine the behavior of the function), but I would prefer just leaving that off the table for this PR.
Change reverted here https://github.com/Layr-Labs/eigenlayer-contracts/pull/208/commits/c967c7441bcea58d9d836d4325a29deee59762af -- if we want to remove this then we should also do it in the `StrategyManager._completeQueuedWithdrawal` and `EigenPodManager._completeQueuedWithdrawal` functions as well, as part of a combined change.

ChaoticWalrus:
Closing this as not planned.
We've opted instead to move withdrawals to the DelegationManager and use a "deficit" accounting system to remove the need for the "undelegation limbo" mechanic.
See this branch / associated PR(s) for referencce: https://github.com/Layr-Labs/eigenlayer-contracts/blob/Combined-Withdrawal-Flow-No-Limbo

PR #210, opened 2023-09-30
Add stake updates to eigen layer core

gpsanant:
Adds stake registry to the delegation manager and forces stake updates.

Notably, the EigenLayer core is dependent on the non reversion of the StakeRegistry for liveness. This is fine because the governance of the StakeRegistry will be the same as EigenLayer core.


ChaoticWalrus:
Seems like this is also incorporating many of the formatting updates, which is making it very hard for me to review.
@gpsanant can you merge these formatting updates into the `multiquorums` branch separately please?

src/contracts/core/DelegationManager.sol
```diff
@@ -197,41 +222,55 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+                });
 
                 unchecked {
                     ++i;
                 }
             }
         }
 
+        // push the operator's new stake to the StakeRegistry
+        _pushOperatorStakeUpdate(operator);
```
ChaoticWalrus, line 272: looks like this will already happen above, in the calls to `_decreaseOperatorShares` (if they happen), but seems OK to have this here for extra coverage 🤷 This is a testnet-only change / not intended to be gas optimized.

ChaoticWalrus: nevermind, it's only in the external version of the function. we need this here; resolving the thread.

ChaoticWalrus:
AFAICT these changes look good. I'd like to see the new changes in isolation though!

ChaoticWalrus (approved):
LGTM. Only curious about the decision to only allow setting the `stakeRegistry` once, but I think it's OK.

ChaoticWalrus:
Seeing a flaky test failure in forge tests in the CI
idk what the root cause is, seems like potentially infinite looping? https://github.com/Layr-Labs/eigenlayer-contracts/actions/runs/6385829402/job/17331395276?pr=210#step:6:426
failure does not appear at all related to the changes in this PR, so shouldn't be blocking for merging.

gpsanant:
> LGTM. Only curious about the decision to only allow setting the `stakeRegistry` once, but I think it's OK.

we should only need to set it once. kept the requirement as a safety measure mainly against mistakes as it's an easy check. no good reason essentially :)

PR #221, opened 2023-10-04
WIP: Reintroduce deficit concept to support combined withdrawal flow no limbo

ChaoticWalrus:


ChaoticWalrus:
The notion here is that we:
- don't allow queuing a withdrawal for more 'beacon chain ETH shares' than the user currently has
- if we would otherwise decrease a user's beacon chain ETH shares (e.g. as a result of a call to `EigenPod.verifyBalanceUpdate`), then if the amount to decrease exceeds the user's current share balance, their balance is reduced to zero and they incur a "deficit"
- when increasing shares we prioritize decrementing the deficit first (instead of actually increasing shares)
- when completing any withdrawals, we also prioritize using part/all of the withdrawal to reduce the deficit first

ChaoticWalrus:
Closing in favor of https://github.com/Layr-Labs/eigenlayer-contracts/pull/222

PR #217, opened 2023-10-03
WIP - undelegation refactor proposal

wadealexc:
First pass complete so yall can take a look. Need to fix compiler errors and remove some now-uneeded methods.

PR #222, opened 2023-10-05
Combined withdrawal flow no limbo

ChaoticWalrus:
Combines changes in https://github.com/Layr-Labs/eigenlayer-contracts/pull/217 and https://github.com/Layr-Labs/eigenlayer-contracts/pull/221

## Key features:
- The DelegationManager is now the single system entry-point for queuing or completing a withdrawal
- No more "undelegation limbo" -- we're back to undelegating requiring/creating a withdrawal for all of the user's shares (but it makes more sense now with the withdrawal flows combined)
- Return to 'deficit' accounting in EigenPods -- this allows the removal of the "undelegation limbo" concept, putting restaked LSTs and natively-restaked ETH on "even footing"
- Migration process for existing queued withdrawals in the StrategyManager to be transformed into Withdrawals in the DelegationManager (so that we can remove the logic for queuing + completing withdrawals from the StrategyManager)

This _should_ unify flows, make the call stack between system contracts clearer/more sensible, and reduce complexity + the size of the attack surface of the core EigenLayer contracts.

PR initially marked as a draft, since it breaks **tons** of tests and still needs a bit of polish on functionality.


src/contracts/pods/EigenPodManagerStorage.sol
```diff
@@ -57,14 +57,23 @@ abstract contract EigenPodManagerStorage is IEigenPodManager {
...
     /// @notice Mapping: podOwner => cumulative number of queued withdrawals of beaconchainETH they have ever initiated. only increments (doesn't decrement)
     mapping(address => uint256) public numWithdrawalsQueued;
 
     /// @notice Mapping: hash of withdrawal inputs, aka 'withdrawalRoot' => whether the withdrawal is pending
     mapping(bytes32 => bool) public withdrawalRootPending;
 
     // @notice Mapping: pod owner => UndelegationLimboStatus struct. Mapping is internal so we can have a getter that returns a memory struct.
-    mapping(address => IEigenPodManager.UndelegationLimboStatus) internal _podOwnerUndelegationLimboStatus;
+    // mapping(address => IEigenPodManager.UndelegationLimboStatus) internal _podOwnerUndelegationLimboStatus;
+    uint private __deprecated__UndelegationLimboStatus;
```
gpsanant, line 76: can remove entirely?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -141,114 +146,41 @@ contract EigenPodManager is
...
-            _decrementWithdrawableRestakedExecutionLayerGwei(msg.sender, currentPodOwnerShares);
-            // withdraw through the ETH from the EigenPod
-            _withdrawRestakedBeaconChainETH(msg.sender, msg.sender, currentPodOwnerShares);
-            // or else return the "shares" to the delegation system
-        } else {
-            delegationManager.increaseDelegatedShares(msg.sender, beaconChainETHStrategy, podOwnerShares[msg.sender]);
-        }
+    // @notice Increases the `podOwner`'s shares by `shares`, paying off deficit if possible
+    // @dev Returns the number of shares added to `podOwnerShares[podOwner]`
+    function awardShares(
```
gpsanant, line 160: addShares

src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -75,6 +75,16 @@ interface IDelegationManager {
         uint256 expiry;
     }
 
+    struct Withdrawal {
```
gpsanant, line 78: reminder to comment

src/contracts/pods/EigenPodManagerStorage.sol
ChaoticWalrus: ahh good point, will remove

src/contracts/core/DelegationManager.sol
```diff
@@ -224,50 +229,247 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+            withdrawal.startBlock + withdrawalDelayBlocks <= block.number, 
+            "DelegationManager.completeQueuedAction: withdrawalDelayBlocks period has not yet passed"
+        );
+
+        require(
+            msg.sender == withdrawal.withdrawer, 
+            "DelegationManager.completeQueuedAction: only withdrawer can complete action"
+        );
+
+        if (receiveAsTokens) {
```
gpsanant, line 314: would be good to not that `recieveAsTokens` is at the level of the withdrawal

src/contracts/pods/EigenPodManagerStorage.sol
```diff
@@ -71,10 +71,6 @@ abstract contract EigenPodManagerStorage is IEigenPodManager {
     /// @notice Mapping: hash of withdrawal inputs, aka 'withdrawalRoot' => whether the withdrawal is pending
     mapping(bytes32 => bool) public withdrawalRootPending;
 
-    // @notice Mapping: pod owner => UndelegationLimboStatus struct. Mapping is internal so we can have a getter that returns a memory struct.
-    // mapping(address => IEigenPodManager.UndelegationLimboStatus) internal _podOwnerUndelegationLimboStatus;
-    uint private __deprecated__UndelegationLimboStatus;
```
wadealexc: Double-checking, are we removing this because it doesn't exist in any of our current deployed code?

ChaoticWalrus: yes. it doesn't have any dedicated storage slot, so it doesn't make sense to reserve the slot

src/contracts/core/DelegationManager.sol
ChaoticWalrus: I added some documentation, please improve more if you think it needs it

src/contracts/core/StrategyManagerStorage.sol
```diff
@@ -63,9 +63,8 @@ abstract contract StrategyManagerStorage is IStrategyManager {
 
     /*
      * Reserved space previously used by the deprecated mapping(address => uint256) beaconChainETHSharesToDecrementOnWithdrawal.
-     * This mapping tracked beaconChainETH "debt" in case updates were made to shares retroactively.  However, this design was
-     * replaced by a simpler design that prevents withdrawals from EigenLayer before withdrawals from the beacon chain, which
-     * makes this tracking unnecessary.
+     * This mapping tracked beaconChainETH "deficit" in cases where updates were made to shares retroactively.  However, this construction was
```
gpsanant, line 66: take out `withdrawalDelayBlocks`?

src/contracts/core/StrategyManager.sol
```diff
@@ -186,129 +183,38 @@ contract StrategyManager is
...
-        uint256[] calldata strategyIndexes,
-        IStrategy[] calldata strategies,
-        uint256[] calldata shares,
-        address withdrawer
-    ) external onlyWhenNotPaused(PAUSED_WITHDRAWALS) onlyNotFrozen(msg.sender) nonReentrant returns (bytes32) {
-        bytes32 queuedWithdrawal = _queueWithdrawal(msg.sender, strategyIndexes, strategies, shares, withdrawer);
-        delegation.decreaseDelegatedShares(msg.sender, strategies, shares);
-        return queuedWithdrawal;
+    function addShares(
+        address grantee,
```
gpsanant, line 195: staker?

src/contracts/core/StrategyManager.sol
```diff
@@ -186,129 +183,38 @@ contract StrategyManager is
...
-        bool receiveAsTokens
-    )
-        external
-        onlyWhenNotPaused(PAUSED_WITHDRAWALS)
-        // check that the address that the staker *was delegated to* – at the time that they queued the withdrawal – is not frozen
-        nonReentrant
-    {
-        _completeQueuedWithdrawal(queuedWithdrawal, tokens, middlewareTimesIndex, receiveAsTokens);
+    function withdrawSharesAsTokens(
+        address destination,
```
gpsanant, line 203: recipient?

src/contracts/core/StrategyManager.sol
```diff
@@ -186,129 +183,38 @@ contract StrategyManager is
...
+        IStrategy strategy,
+        uint256 shares,
+        IERC20 token
```
gpsanant, line 206: token, shares is a nicer order imo

src/contracts/core/StrategyManager.sol
```diff
@@ -186,129 +183,38 @@ contract StrategyManager is
...
-            strategies[i] = stakerStrategyList[staker][index];
-            shares[i] = stakerStrategyShares[staker][strategies[i]];
-            strategyIndexes[i] = index;
-            unchecked {
-                ++i;
-            }
-        }
-        bytes32 queuedWithdrawal = _queueWithdrawal(staker, strategyIndexes, strategies, shares, staker);
-        return (strategies, shares, queuedWithdrawal);
+    function removeShares(
```
gpsanant, line 186: natspec comments needed in this file in general?

src/contracts/core/StrategyManager.sol
```diff
@@ -713,8 +456,14 @@ contract StrategyManager is
         }
     }
 
+    // @notice Internal function for calculating the current domain separator of this contract
+    function _calculateDomainSeparator() internal view returns (bytes32) {
```
gpsanant, line 460: we should eventually move this to OZ EIP712. eventually.

src/contracts/core/StrategyManager.sol
```diff
@@ -98,20 +98,17 @@ contract StrategyManager is
      * @param initialOwner Ownership of this contract is transferred to this address.
      * @param initialStrategyWhitelister The initial value of `strategyWhitelister` to set.
```
gpsanant, line 99: some unused constants in this file i believe

src/contracts/pods/EigenPodManager.sol
```diff
@@ -142,113 +147,41 @@ contract EigenPodManager is
...
-        onlyNotFrozen(msg.sender)
-        nonReentrant
-        returns (bytes32)
-    {
-        return _queueWithdrawal(msg.sender, amountWei, withdrawer);
+    function removeShares(
+        address podOwner, 
+        uint256 shares
+    ) external onlyDelegationManager {
+        require(shares <= podOwnerShares[podOwner], "EigenPodManager.removeShares: shares amount too high");
```
gpsanant, line 157: we need to get rid of this check now?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -407,94 +236,73 @@ contract EigenPodManager is
...
-            strategies[0] = beaconChainETHStrategy;
-            delegationManager.decreaseDelegatedShares(podOwner, strategies, shareAmounts);
+        // skip dealing with deficit if there isn't any need for it
+        if (shareAmount <= currentPodOwnerShares) {
+            podOwnerShares[podOwner] = currentPodOwnerShares - shareAmount;
+            return shareAmount;
+        // otherwise, add to the deficit as necessary
+        } else {
+            podOwnerShares[podOwner] = 0;
+            uint256 newDeficitAmount = (shareAmount - currentPodOwnerShares);
```
gpsanant, line 280: parentheses only if inlining below?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -142,113 +147,41 @@ contract EigenPodManager is
...
-        onlyDelegationManager
-        onlyWhenNotPaused(PAUSED_WITHDRAW_RESTAKED_ETH)
-        onlyNotFrozen(podOwner)
-        nonReentrant
-        returns (uint256 sharesRemovedFromDelegation)
-    {
-        // put the `podOwner` into undelegation limbo, and return the amount of shares which were entered into undelegation limbo
-        return _enterUndelegationLimbo(podOwner, delegatedTo);
+    /// @notice Used by the DelegationManager to complete a withdrawal, sending tokens to some destination address
+    // TODO the 2 calls here can probably be combined?
```
gpsanant, line 176: nice, we should do that!

src/contracts/pods/EigenPodManager.sol
```diff
@@ -407,94 +236,73 @@ contract EigenPodManager is
...
-        // if the podOwner has delegated shares, record a decrease in their delegated shares
-        if (!isInUndelegationLimbo(podOwner)) {
-            uint256[] memory shareAmounts = new uint256[](1);
-            shareAmounts[0] = shareAmount;
-            IStrategy[] memory strategies = new IStrategy[](1);
-            strategies[0] = beaconChainETHStrategy;
-            delegationManager.decreaseDelegatedShares(podOwner, strategies, shareAmounts);
+        // skip dealing with deficit if there isn't any need for it
+        if (shareAmount <= currentPodOwnerShares) {
+            podOwnerShares[podOwner] = currentPodOwnerShares - shareAmount;
```
gpsanant, line 275: seems like the following invariant is true: `(deficit == 0 && podOwnerShares >= 0) || (deficit <= 0 && podOwnerShares == 0)`.

What if we just made `podOwnerShares` an `int256` and got rid of `deficit` entirely?

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -66,7 +71,14 @@ abstract contract DelegationManagerStorage is IDelegationManager {
      */
     mapping(address => mapping(bytes32 => bool)) public delegationApproverSaltIsSpent;
 
-    IEigenPodManager public immutable eigenPodManager;
+    /// @notice Mapping: hash of withdrawal inputs, aka 'withdrawalRoot' => whether the withdrawal is pending
+    mapping(bytes32 => bool) public pendingWithdrawals;
+
+    /// @notice Mapping: staker => cumulative number of queued withdrawals they have ever initiated.
+    /// @dev This only increments (doesn't decrement), and is used to help ensure that otherwise identical withdrawals have unique hashes.
+    mapping(address => uint96) public numWithdrawalsQueued;
```
gpsanant, line 79: can we add cumulative to the name here? confusing with whether this is the number of withdrawals queued *at the time of calling*

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -66,7 +71,14 @@ abstract contract DelegationManagerStorage is IDelegationManager {
...
+
+    uint256 public withdrawalDelayBlocks;
```
gpsanant, line 81: comment

src/contracts/core/DelegationManager.sol
```diff
@@ -513,6 +732,57 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+            shares[0] = podShares;
+        } else {
+            // Has shares in both
+            
+            // 1. Allocate return arrays
+            strategies = new IStrategy[](strategyManagerStrats.length + 1);
+            shares = new uint256[](strategies.length);
+            
+            // 2. Place StrategyManager strats/shares in return arrays
+            for (uint256 i = 0; i < strategyManagerStrats.length; ) {
```
gpsanant, line 766: nice to leave a note that this is inefficiency is only encountered by people who natively restake and liquid restake on the same address. we should discourage that in the docs. use different addresses.

src/contracts/core/DelegationManager.sol
```diff
@@ -224,48 +227,170 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+                });
+            }
+
+            unchecked { ++i; }
+        }
+
+        emit WithdrawalCompleted(withdrawalRoot);
+    }
+
+    // @notice Migrates an existing queued withdrawal from the StrategyManager contract to this contract.
```
gpsanant, line 366: `/// @notice ... this will be live until the next updrade`

src/contracts/core/DelegationManager.sol
```diff
@@ -224,48 +227,170 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+    function migrateQueuedWithdrawal(IStrategyManager.DeprecatedStruct_QueuedWithdrawal memory strategyManagerWithdrawalToMigrate) external {
+        // check for existence and delete the old storage
+        bytes32 oldWithdrawalRoot = strategyManager.calculateWithdrawalRoot(strategyManagerWithdrawalToMigrate);
+        strategyManager.migrateQueuedWithdrawal(oldWithdrawalRoot);
```
gpsanant, line 370: might as well just pass the whole struct here if we're gonna calculate the withdrawal root by passing it the line above. the other option is to calculate in an internal function in the DM.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -142,113 +147,41 @@ contract EigenPodManager is
...
-        // put the `podOwner` into undelegation limbo, and return the amount of shares which were entered into undelegation limbo
-        return _enterUndelegationLimbo(podOwner, delegatedTo);
+    /// @notice Used by the DelegationManager to complete a withdrawal, sending tokens to some destination address
+    // TODO the 2 calls here can probably be combined?
+    function withdrawSharesAsTokens(
+        address podOwner, 
+        address destination, 
+        uint256 shares
+    ) external onlyDelegationManager {
+        ownerToPod[podOwner].decrementWithdrawableRestakedExecutionLayerGwei(shares);
```
gpsanant, line 182: pay off deficit here?

gpsanant:
This PR looks great. Mostly nits that don't need to be addressed. A few logic questions and recommendations that we should discuss.

src/contracts/core/StrategyManager.sol
8sunyuan, line 1: Should remove top level contract Natspec functionalities
 * - enabling withdrawal of assets from specified strategy(s)
 * - recording deposit of ETH into settlement layer

src/contracts/core/StrategyManagerStorage.sol
8sunyuan: `numWithdrawalsQueued` unused as well

src/contracts/core/StrategyManager.sol
```diff
@@ -186,129 +183,38 @@ contract StrategyManager is
...
-            }
-        }
-        bytes32 queuedWithdrawal = _queueWithdrawal(staker, strategyIndexes, strategies, shares, staker);
-        return (strategies, shares, queuedWithdrawal);
+    function removeShares(
+        address staker,
+        IStrategy strategy,
+        uint256 shares
+    ) external onlyDelegationManager {
+        _removeShares(staker, 0, strategy, shares);
```
8sunyuan, line 191: Seems like we decided on removing the `strategyIndexes` param. We should remove the index logic from `_removeShares()` and `_removeStrategyFromStakerStrategyList()` as well

src/contracts/core/StrategyManager.sol
stevennevins: maybe staker?

src/contracts/core/StrategyManager.sol
ChaoticWalrus: we use 'staker, Strategy, shares' in multiple other functions, so I feel like this ordering is at least pretty consistent.

src/contracts/core/StrategyManager.sol
wadealexc: I think in this case staker doesn't make as much sense. I think of a staker as someone that has shares, but here the destination/recipient may not have any shares at all.

I like either `destination` or `recipient`

src/contracts/core/StrategyManager.sol
wadealexc: agreed

src/contracts/core/StrategyManager.sol
wadealexc: it's consistent with the way the EPmgr is called, and makes it clear that `tokens` is this extra thing the Smgr needs. I like this order.

src/contracts/core/StrategyManager.sol
```diff
@@ -186,129 +183,38 @@ contract StrategyManager is
...
-        // check that the address that the staker *was delegated to* – at the time that they queued the withdrawal – is not frozen
-        nonReentrant
-    {
-        _completeQueuedWithdrawal(queuedWithdrawal, tokens, middlewareTimesIndex, receiveAsTokens);
+    function withdrawSharesAsTokens(
+        address destination,
+        IStrategy strategy,
+        uint256 shares,
+        IERC20 token
+    ) external onlyDelegationManager {
```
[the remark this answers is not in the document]
wadealexc: Personally I hate these unused modifiers, but even if we do feel the need to add some of them back, I think they belong in the DelegationManager. These methods are only called from there, so if we want these modifiers let's at least place them in a single location

src/contracts/core/DelegationManager.sol
wadealexc: ... surely this isn't too bad of a gas cost? It hasn't been intuitive to me that people should use different addresses - why do we want that?

Most people not tryna keep up with addresses like engineers.

src/contracts/core/StrategyManager.sol
ChaoticWalrus: I like the unity/consistency of language, will make the change

wadealexc:
One big question I have - do we want a `completeQueuedWithdrawals`-type plural method? Did we originally have that because it was useful for people with multiple partial withdrawals?

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: no, it's enforcing that a staker can't create a share deficit as the result of a withdrawal

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: yes, this is true, and the root issue of the previous deficit system was that it _wasn't_ always true.
I've tried to think through if using a signed integer would be much cleaner. thinking it'll make sense to do a quick PR to settle the debate / show everyone what it would look like.

src/contracts/core/DelegationManager.sol
ChaoticWalrus: I think it's fine. I will say, by my estimation, worst case you end up with:
66*32 = 2112 memory expansion -- not sure on cost, depends on other memory already used
(66 * some small number) memory operations -- probably on the order of hundreds of gas

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: ahhh good catch

src/contracts/core/DelegationManager.sol
```diff
@@ -412,46 +531,146 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+            // update shares amount depending upon the returned value
+            // the return value will be lower than the input value in the case where the staker has an existing share deficit
+            shares = eigenPodManager.addShares({
+                podOwner: staker,
+                shares: shares
+            });
+            // since these shares are given back to the pod owner, they must be delegated to the *pod owner*'s operator, which could be different from the withdrawer's
+            operator = delegatedTo[staker];
+            // we also want to emit the event within `_increaseOperatorShares` with the correct fields. Since the staker receives the shares, they should be in the event.
+            withdrawer = staker;
```
wadealexc, line 658: Good bug fix here. I think we can simplify this just a touch - currently this method is passed the operator from `completeQueuedWithdrawal`.

However, `completeQueuedWithdrawal` doesn't need the operator for anything -- it's only used in `_addAndDelegateShares`.

Maybe we just don't pass it in, and calculate it in both branches within this function? Probably clearer logic overall, and reduces a storage read for certain cases.

src/contracts/core/StrategyManagerStorage.sol
```diff
@@ -32,43 +38,43 @@ abstract contract StrategyManagerStorage is IStrategyManager {
...
-    mapping(address => uint256) public numWithdrawalsQueued;
+
+    /*
+     * Reserved space previously used by the deprecated mapping(address => uint256) cumulativeWithdrawalsQueued.
+     * This mapping tracked the cumulative number of queued withdrawals initiated by a staker.
+     * Withdrawals are now initiated in the DlegationManager, so the mapping has moved to that contract.
+     */
+    // slither-disable-next-line incorrect-shift-in-assembly
+    uint256[1] internal _deprecatedStorage_cumulativeWithdrawalsQueued;
+
```
wadealexc, line 67: IMO deprecated storage values should just retain old naming, even if it's bad. People that actually care about what it used to point to will have an easier time cross-referencing

src/contracts/core/DelegationManager.sol
```diff
@@ -224,48 +227,171 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+            withdrawer: strategyManagerWithdrawalToMigrate.withdrawerAndNonce.withdrawer,
+            nonce: nonce,
+            startBlock: strategyManagerWithdrawalToMigrate.withdrawalStartBlock,
+            strategies: strategyManagerWithdrawalToMigrate.strategies,
+            shares: strategyManagerWithdrawalToMigrate.shares
+        });
+
+        // create the new storage
+        bytes32 newRoot = calculateWithdrawalRoot(migratedWithdrawal);
+        pendingWithdrawals[newRoot] = true;
```
wadealexc, line 390: Just for sanity's sake, we should require that the new calculated root is "not" currently pending. Just in case.

src/contracts/core/StrategyManagerStorage.sol
ChaoticWalrus: I'm fine with that, can change back after responding to other comments

src/contracts/core/StrategyManagerStorage.sol
ChaoticWalrus: I tried changing these, but Alex pointed this out as well https://github.com/Layr-Labs/eigenlayer-contracts/pull/222#discussion_r1348947449
Take a look at the new comments. I think reverting the names but marking these as 'internal' and noting the deprecation in the comments is probably the best move here.

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: yes, done https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/c587b10684a8f37e68ae1787e3007bb64560dd80

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: ehhh I'm fine leaving it

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: fixed here https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/b4bfe165a67ef10ab51b7285e3349db04bc3ab98

src/contracts/core/StrategyManager.sol
ChaoticWalrus: done here https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/d3707d463b3c93611b3de0ace320e6ed573b4d02

src/contracts/core/DelegationManager.sol
ChaoticWalrus: > might as well just pass the whole struct here if we're gonna calculate the withdrawal root by passing it the line above.

That's exactly what I was doing before the request to only have the old root in the event.

>  the other option is to calculate in an internal function in the DM.

Hadn't considered this. Is having a public method on the StrategyManager still useful at all? Otherwise I do think this is the right approach.

src/contracts/core/DelegationManagerStorage.sol
ChaoticWalrus: done https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/430c04321010b5ebdd487b569e1bac560b0818e2

src/contracts/core/DelegationManagerStorage.sol
ChaoticWalrus: good suggestion, done https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/3ee27af5e0379de1ca0c486a923e4a6c9687da8f

src/contracts/core/StrategyManager.sol
ChaoticWalrus: deleted 'em

src/contracts/core/StrategyManager.sol
ChaoticWalrus: I've added a bunch. please note specific places and/or just add more comments yourself

src/contracts/core/StrategyManager.sol
ChaoticWalrus: I updated this to 'recipient'. we use this language elsewhere, should improve consistency.

src/contracts/core/StrategyManager.sol
ChaoticWalrus: I updated this to 'staker'. it's consistent with our other usage.

src/contracts/core/DelegationManager.sol
ChaoticWalrus: comment added here https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/281990899b647118a6257af3d5fc033a24029343

src/contracts/core/StrategyManager.sol
ChaoticWalrus: removed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/153a805da021fb7b9553e6699890b647e55a20af

src/contracts/core/DelegationManager.sol
```diff
@@ -224,48 +227,171 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+
+            unchecked { ++i; }
+        }
+
+        emit WithdrawalCompleted(withdrawalRoot);
+    }
+
+    /// @notice Migrates an existing queued withdrawal from the StrategyManager contract to this contract.
+    /// @dev This function is expected to be removed in the next upgrade, after all queued withdrawals have been migrated.
+    function migrateQueuedWithdrawal(IStrategyManager.DeprecatedStruct_QueuedWithdrawal memory strategyManagerWithdrawalToMigrate) external {
```
ypatil12, line 368: Can this function accept an array of withdrawals to migrate? That way we don't have to go through a multicall contract for a batched withdrawal migration

src/contracts/core/DelegationManager.sol
ypatil12: +1 for having the withdrawal be calculated in the DM for gas savings on migration. Doesn't appear that a public method is needed given other contracts don't have a dependency on this function

wadealexc: Downside is additional bloat in the Dmgr contract.

Here's a solution that gives both, I think:
- Pass the struct to `Smgr.migrateQueuedWithdrawal`
- It returns the calculated root to the Dmgr

src/contracts/core/DelegationManager.sol
ChaoticWalrus: yes, confirming here that I am cool with you making this change, should just modify it so the previous withdrawal not existing is a no-op rather than causing reversion (otherwise there is a potential race condition)

src/contracts/core/DelegationManager.sol
ypatil12: Updated here: https://github.com/Layr-Labs/eigenlayer-contracts/pull/228

ChaoticWalrus: LGTM. can close this thread out once the PR is merged.

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: PR is here https://github.com/Layr-Labs/eigenlayer-contracts/pull/227
I recommend this change; I think it helps clarify how things are supposed to work.

src/contracts/core/StrategyManagerStorage.sol
ChaoticWalrus: switched back to old naming here. I think the comments are sufficient about deprecation, but please flag anything that's a bit lacking on clarity (or just make a quick commit to clarify the deprecation).

src/contracts/core/StrategyManagerStorage.sol
ChaoticWalrus: switched back to old naming here https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/f1ebd25d35edbf213ed3d79d75bf8ef12a3a6fbd. I think the comments are sufficient about deprecation, but please flag anything that's a bit lacking on clarity (or just make a quick commit to clarify the deprecation).

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: PR merged.

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: addressed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/227 (merged)

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: I think this is clearer in now that https://github.com/Layr-Labs/eigenlayer-contracts/pull/227 is merged. Please resolve if it's well-documented / clear now; otherwise would really appreciate if you can add to the documentation here!

ChaoticWalrus:
> > One big question I have - do we want a `completeQueuedWithdrawals`-type plural method? Did we originally have that because it was useful for people with multiple partial withdrawals?
> 
> Great call out. Yes, we need this for the UI.

Added this in commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/3afac4628270077ba2979aa7cdae7a2adeef227e

src/contracts/core/DelegationManager.sol
ChaoticWalrus: sure, it's basically an assertion, since it really, really should never happen.
added here https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/726aceb6e9eb320ece7057b432948e0c91f6f471

src/contracts/core/DelegationManager.sol
ChaoticWalrus: Are you opposed to just doing this? https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/0df17301a7f834adb2bbc52e14849adcbdaa5b89
I think it's clearer this way, especially after merging https://github.com/Layr-Labs/eigenlayer-contracts/pull/227 and clarifying the concept of "(change in) delegateable shares"

src/contracts/core/StrategyManager.sol
```diff
@@ -412,260 +315,82 @@ contract StrategyManager is
...
     function _removeShares(
-        address depositor,
-        uint256 strategyIndex,
+        address staker,
         IStrategy strategy,
         uint256 shareAmount
     ) internal returns (bool) {
         // sanity checks on inputs
-        require(depositor != address(0), "StrategyManager._removeShares: depositor cannot be zero address");
+        require(staker != address(0), "StrategyManager._removeShares: staker cannot be zero address");
```
8sunyuan, line 342: Can we move this `staker != address(0)` check back up to the function `DelegationManager._removeSharesAndQueueWithdrawal()` or elsewhere in DM? This function `_removeShares()` is only called in a loop so we are doing redundant checks.

ChaoticWalrus: yes! I think that's a great suggestion, please go ahead and make it if you'd like 👍

src/contracts/core/DelegationManager.sol
```diff
@@ -513,6 +769,57 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+    function getDelegatableShares(address staker) public view returns (IStrategy[] memory, uint256[] memory) {
+        // Get currently active shares and strategies for `staker`
+        int256 podShares = eigenPodManager.podOwnerShares(staker);
+        (IStrategy[] memory strategyManagerStrats, uint256[] memory strategyManagerShares) 
+            = strategyManager.getDeposits(staker);
+
+        // Has shares in StrategyManager, but not in EigenPodManager
+        if (podShares <= 0) {
+            return (strategyManagerStrats, strategyManagerShares);
+        }
```
wadealexc, line 784: Maybe clarify comment - this branch might also mean the staker has no shares anywhere. If so, this is returning empty arrays.

src/contracts/core/StrategyManagerStorage.sol
wadealexc: Ah, I had meant keep the `_deprecated` prefix, just also keep the suffix as the exact prior name. The change here was that `numWithdrawalsQueued` -> `_deprecated_cumulativeWithdrawalsQueued`, and I was just advocating for `_deprecated_numWithdrawalsQueued`.

That said... I have no problem with dropping the `_deprecated` prefix and just leaving a comment. So, the changes are fine by me!

src/contracts/core/DelegationManager.sol
```diff
@@ -224,48 +228,141 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        bool[] calldata receiveAsTokens
+    ) external onlyWhenNotPaused(PAUSED_EXIT_WITHDRAWAL_QUEUE) nonReentrant {
+        for (uint256 i = 0; i < withdrawals.length; ++i) {
+            _completeQueuedWithdrawal(withdrawals[i], tokens[i], middlewareTimesIndexes[i], receiveAsTokens[i]);
+        }
+    }
+
+    /// @notice Migrates an array of queued withdrawals from the StrategyManager contract to this contract.
+    /// @dev This function is expected to be removed in the next upgrade, after all queued withdrawals have been migrated.
+    function migrateQueuedWithdrawals(IStrategyManager.DeprecatedStruct_QueuedWithdrawal[] memory strategyManagerWithdrawalsToMigrate) external {
```
wadealexc, line 329: The naming here is so long. 

Although this is happily endemic in our codebase, can we just call this long variable `toMigrate`? or `withdrawalToMigrate`? It's word soup otherwise.

wadealexc (approved):
Left some comments, but nothing blocking. LGTM otherwise.

src/contracts/core/DelegationManager.sol
ChaoticWalrus: yeah, I'm cool with `withdrawalToMigrate`. I'll make this and the other clarity change you recommended.

src/contracts/core/DelegationManager.sol
```diff
@@ -412,46 +503,217 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+                            staker: staker,
+                            strategy: withdrawal.strategies[i],
+                            shares: increaseInDelegateableShares
+                        });
+                    }
+                } else {
+                    strategyManager.addShares(msg.sender, withdrawal.strategies[i], withdrawal.shares[i]);
+                    // Similar to `isDelegated` logic
+                    if (currentOperator != address(0)) {
+                        _increaseOperatorShares({
```
wadealexc, line 606: Nit: could we read and assign this variable just above, like in the EPmgr branch? Currently we're grabbing `currentOperator` from way earlier in the method, and it's only used here.

Would also save a storage read for the EPmgr branch!

script/whitelist/delegationFaucet/DelegationFaucet.sol
```diff
@@ -134,34 +134,34 @@ contract DelegationFaucet is IDelegationFaucet, Ownable {
...
-        bytes memory data = abi.encodeWithSelector(
-            IStrategyManager.queueWithdrawal.selector,
-            strategyIndexes,
-            strategies,
-            shares,
-            withdrawer
-        );
-        return Staker(staker).callAddress(address(strategyManager), data);
+        // bytes memory data = abi.encodeWithSelector(
+        //     IStrategyManager.queueWithdrawal.selector,
```
stevennevins, line 138: should we uncomment this out before merging? seems like switching to IDelegationMananger and adjusting the params

ChaoticWalrus: good point, I'll adjust this

src/contracts/core/DelegationManager.sol
ChaoticWalrus: addressed here https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/08a3c0642b959ba1c56e401fb0048572acaa9469

src/contracts/core/DelegationManager.sol
ChaoticWalrus: addressed here https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/08a3c0642b959ba1c56e401fb0048572acaa9469

script/whitelist/delegationFaucet/DelegationFaucet.sol
ChaoticWalrus: fixed in this commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/56160b7c022ac7c6d7dbc0b4170add6f39b00620

src/contracts/core/DelegationManager.sol
ChaoticWalrus: I added this commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/f0ecb9272b8b87c99c7154566a6feef02c9b76de
We still end up SLOAD'ing in the EPM branch, but I think it's fine.
The code is a little awkward with the two ~identical loop definitions, but it saves an SLOAD in some cases + helps make the semantic meaning of `receiveAsTokens` clearer (IMO, at least).
Appreciate the comment.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -138,117 +121,98 @@ contract EigenPodManager is
...
+    }
+
+    /// @notice Used by the DelegationManager to complete a withdrawal, sending tokens to some destination address
+    /// @dev Prioritizes decreasing the podOwner's share deficit, if they have one
+    function withdrawSharesAsTokens(
+        address podOwner, 
+        address destination, 
+        uint256 shares
+    ) external onlyDelegationManager {
+        require(int256(shares) >= 0, "EigenPodManager.withdrawSharesAsTokens: shares cannot be negative");
```
wadealexc, line 197: For consistency with `addShares` and `removeShares`, let's make the same "zero checks" for `podOwner` and `destination` here

src/contracts/core/DelegationManager.sol
ChaoticWalrus: yup yup. We could revert the commit, but I prefer the improvement in semantics with the change, especially w/r/t `receiveAsTokens`. Planning to leave this as-is.

stevennevins (approved):

src/contracts/pods/EigenPodManager.sol
ChaoticWalrus: good suggestion, added in https://github.com/Layr-Labs/eigenlayer-contracts/pull/222/commits/e9782d99ba964787727d4fdf7d83980630f493bc

ChaoticWalrus:
Merging this now despite it breaking CI (automated tests failing). Expecting to push fixes to tests soon.

PR #224, opened 2023-10-05
Feat: StrategyManager Unit testing

8sunyuan:
Additional Unit Tests:
- Fixed commented out tests
- improved fuzzing
- added additional coverage

Update:
- Fixing broken tests from withdrawal changes to StrategyManager and DelegationManager
- comprehensive withdrawals tests including both EPM and SM withdrawals should be written later


ChaoticWalrus (commented):
LGTM, seems like improvements all around.
This will definitely have conflicts with https://github.com/Layr-Labs/eigenlayer-contracts/pull/222
I feel like it might actually make more sense to try merging from this branch into that one (or vice versa?) rather than merging either PR into master first.
Let me know your thoughts; I'm sure we can sort things out.

8sunyuan:
> LGTM, seems like improvements all around. This will definitely have conflicts with #222 I feel like it might actually make more sense to try merging from this branch into that one (or vice versa?) rather than merging either PR into master first. Let me know your thoughts; I'm sure we can sort things out.

Yeah I agree, I'll merge #222 into this branch, resolve conflicts then rebase.

.github/workflows/certora-prover.yml
```diff
@@ -38,10 +38,10 @@ jobs:
           python-version: '3.10'
           cache: 'pip'
       - name: Install java
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v2
         with:
-          java-version: '11'
-          java-package: 'jre'
+          distribution: temurin
```
stevennevins, line 43: What's the reason for the prover updates?

stevennevins:
Great work on handling all the upstream changes here. Can you add a couple of tests for addShares and removeShares where you prank the MockDelegationManager to ensure the updates happen accurately? We should also hit the case if the strategy isn't in the stakerStrategyList since that should be possible since we pass in the strategy list in queueWithdrawal. It looks like the expected behavior is we ignore it in the StrategyManager, but I want to confirm (potentially we might want to revert instead)

.github/workflows/certora-prover.yml
8sunyuan: Actually not entirely sure where these changes came from. Originally thought it was from pulling from the `Combined-Withdrawal-Flow` branch. @ChaoticWalrus Should I revert these changes?

ChaoticWalrus: > What's the reason for the prover updates?

This was an attempt to fix some CI failures, which unfortunately I'm still seeing somewhat intermittently -- they're hard to track down because everything works locally.
IDK they're showing up as diffs in this PR (seems like a GH error?) because these changes are already [merged into master](https://github.com/Layr-Labs/eigenlayer-contracts/blob/master/.github/workflows/certora-prover.yml#L40-L44)

> Should I revert these changes?

No, see above, IDK why this is even showing up in diffs

stevennevins: sg just seemed off

stevennevins (approved):

ChaoticWalrus:
Merging this PR for now, as it seems in a good place / well-reviewed.

> Great work on handling all the upstream changes here. Can you add a couple of tests for addShares and removeShares where you prank the MockDelegationManager to ensure the updates happen accurately?

 I agree these tests would be good to add, which can be done in a following PR 😄

PR #226, opened 2023-10-06
Hexens EIG-14 Addressal

Sidu28:
Addressing Hexen's EIG14 Bug.

### DIFF
--- a/src/test/unit/EigenPodUnit.t.sol
+++ b/src/test/unit/EigenPodUnit.t.sol
@@ -231,4 +231,41 @@ contract EigenPodUnitTests is EigenPodTests {
         cheats.stopPrank();
     }
 
+    function testPodReceiveFallBack(uint256 amountETH) external {
+        cheats.assume(amountETH > 0);
+        setJSON("./src/test/test-data/withdrawal_credential_proof_302913.json");
+        _testDeployAndVerifyNewEigenPod(podOwner, signature, depositDataRoot);
+        IEigenPod pod = eigenPodManager.getPod(podOwner);
+        cheats.deal(address(this), amountETH);
+
+        Address.sendValue(payable(address(pod)), amountETH);
+    }
+
+    /**
+    * This is a regression test for a bug (EIG-14) found by Hexens.  Lets say podOwner sends 32 ETH to the EigenPod, 
+    * the nonBeaconChainETHBalanceWei increases by 32 ETH. podOwner calls withdrawBeforeRestaking, which 
+    * will simply send the entire ETH balance (32 ETH) to the owner. The owner activates restaking, 
+    * creates a validator and verifies the withdrawal credentials, receiving 32 ETH in shares.  
+    * They can exit the validator, the pod gets the 32ETH and they can call withdrawNonBeaconChainETHBalanceWei
+    * And simply withdraw the 32ETH because nonBeaconChainETHBalanceWei is 32ETH.
+     */
+    function testValidatorBalanceCannotBeRemovedFromPodViaNonBeaconChainETHBalanceWei() external {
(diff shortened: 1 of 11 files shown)


src/test/unit/EigenPodUnit.t.sol
```diff
@@ -231,4 +231,41 @@ contract EigenPodUnitTests is EigenPodTests {
...
 
+    function testPodReceiveFallBack(uint256 amountETH) external {
+        cheats.assume(amountETH > 0);
+        setJSON("./src/test/test-data/withdrawal_credential_proof_302913.json");
+        _testDeployAndVerifyNewEigenPod(podOwner, signature, depositDataRoot);
+        IEigenPod pod = eigenPodManager.getPod(podOwner);
+        cheats.deal(address(this), amountETH);
+
+        Address.sendValue(payable(address(pod)), amountETH);
+    }
```
wadealexc, line 242: Should this check something after sending value? Doesn't seem like this tests anything in particular

wadealexc: Should have a check for `nonBeaconChainETHBalanceWei` I guess?

wadealexc: Or are we just checking that it doesnt revert

src/test/unit/EigenPodUnit.t.sol
```diff
@@ -231,4 +231,41 @@ contract EigenPodUnitTests is EigenPodTests {
...
+    * This is a regression test for a bug (EIG-14) found by Hexens.  Lets say podOwner sends 32 ETH to the EigenPod, 
+    * the nonBeaconChainETHBalanceWei increases by 32 ETH. podOwner calls withdrawBeforeRestaking, which 
+    * will simply send the entire ETH balance (32 ETH) to the owner. The owner activates restaking, 
+    * creates a validator and verifies the withdrawal credentials, receiving 32 ETH in shares.  
+    * They can exit the validator, the pod gets the 32ETH and they can call withdrawNonBeaconChainETHBalanceWei
+    * And simply withdraw the 32ETH because nonBeaconChainETHBalanceWei is 32ETH.
+     */
+    function testValidatorBalanceCannotBeRemovedFromPodViaNonBeaconChainETHBalanceWei() external {
+        cheats.startPrank(podOwner);
+        IEigenPod newPod = eigenPodManager.getPod(podOwner);
```
wadealexc, line 254: I think this tests the very bare minimum oh EIG-14, but I assume setting up a proper test that follows their 7 steps exactly is much harder, because you'd need to mock out verifying withdrawal creds?

src/test/unit/EigenPodUnit.t.sol
ChaoticWalrus: Agreed this is fairly vacuous as-is. it should 100% check that `nonBeaconChainETHBalanceWei` increases by the sent amount. Could also verify that the event is emitted

ChaoticWalrus (approved):
LGTM. Agreed with Alex's evaluation.

gpsanant (approved):
lgtm

src/test/unit/EigenPodUnit.t.sol
Sidu28: Ok i think I somehow didn't complete this test rippp

src/test/unit/EigenPodUnit.t.sol
Sidu28: yeah, I felt like the core of the issue is that nonBeaconChainETHBalanceWei was never being decremented so checking that that was the case ensured that its covered.  Im not sure how much value mocking out the entire flow would add

PR #227, opened 2023-10-06
combine podOwnerShares and podOwnerShareDeficit into a single, signed…

ChaoticWalrus:
… integer mapping

The key concept is that negative/“deficit” shares are not delegateable — if we correctly track changes in “delegateable shares” then the system should be consistent.


src/contracts/pods/EigenPodManager.sol
```diff
@@ -238,74 +269,25 @@ contract EigenPodManager is
...
-                    decreaseInPositiveShares = uint256(currentPodOwnerShares);
-                } else {
-                    decreaseInPositiveShares = uint256(sharesDelta);
-                }
-                delegationManager.decreaseDelegatedShares({
-                    staker: podOwner,
-                    strategy: beaconChainETHStrategy,
-                    shares: decreaseInPositiveShares
-                });
+    function _calculateChangeInDelegatableShares(int256 sharesBefore, int256 sharesAfter) internal pure returns (int256) {
```
gpsanant, line 275: really nice abstraction

gpsanant (approved):
LGTM, great change

PR #228, opened 2023-10-06
Migrate multiple queued withdrawals

ypatil12:
This PR updates the migrateQueuedWithdrawal function within the DM to migrate multiple queued withdrawals to easily batch multiple migrations. To handle updating multiple withdrawals without reverting, the `migrateQueuedWithdrawal` in the SM does not revert if the withdrawal has already been provided. We also remove all withdrawalRoot calculations from the DM. 

Migration tests will be added in subsequent PR.

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -365,33 +365,41 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
 
     /// @notice Migrates an existing queued withdrawal from the StrategyManager contract to this contract.
     /// @dev This function is expected to be removed in the next upgrade, after all queued withdrawals have been migrated.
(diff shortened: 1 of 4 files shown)


src/contracts/core/DelegationManager.sol
```diff
@@ -365,33 +365,41 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
 
     /// @notice Migrates an existing queued withdrawal from the StrategyManager contract to this contract.
```
ChaoticWalrus, line 366: tiny nit: this should be updated to reflect the fact that this is now accepting an array, rather than singlet input.

ChaoticWalrus (approved):
LGTM. Noted one tiny nit; would prefer if you can fix it before merging.

src/contracts/core/DelegationManager.sol
ypatil12: Fixed

PR #229, opened 2023-10-07
rename NonSignerStakesAndSignature fields to be more consistent

samlaf:
Renamed the fields of NonSignerStakesAndSignature to be more consistent and added a huge comment describing how it's meant to be used above, since it was definitely the trickiest part to understand when writing an avs.

Talked to @gpsanant about this offline. This doesn't need to be merged now (since this would trickle down to a lot of needed offchain code changes), but leaving here as a suggestion, and we can merge this whenever. I personally think it would make this struct easier to understand, but I'm open to changing some of these.

Currently I followed the convention:
```
    Each field array can either be over signers, nonsigners, or quorums.
    We use the first word of the variable to represent what the array is over
      eg: []nonSignersPubkeysG1 -> array of nonSigners
      eg: []quorumApkG1 -> array of quorums
```
But we could also potentially use a `per...` suffix convention
```
nonSignersPubkeysG1 -> pubkeysG1PerNonSigners
quorumApkG1 -> apkG1PerQuorum
```

### Tests
Some unit tests are failing, but I don't think it's related to my changes.

### DIFF
--- a/src/contracts/middleware/BLSOperatorStateRetriever.sol
+++ b/src/contracts/middleware/BLSOperatorStateRetriever.sol
@@ -19,130 +19,196 @@ contract BLSOperatorStateRetriever {
     }
 
     struct CheckSignaturesIndices {
-        uint32[] nonSignerQuorumBitmapIndices;
+        uint32[] nonSignersQuorumBitmapIndices;
         uint32[] quorumApkIndices;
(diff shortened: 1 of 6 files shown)


ChaoticWalrus:
Closing this out as stale; these contracts no longer live in this repo.

PR #230, opened 2023-10-09
added check for validator balance update against withdrawable epoch

Sidu28:
From @gpsanant 👍 
Possible bug in eigenpods: https://github.com/Layr-Labs/eigenlayer-contracts/blob/05e754dac084dda744626884af9ed44d9ca5c7ba/src/contracts/pods/EigenPod.sol#L208
A validator withdraws fully but doesn't prove their withdrawal, a balance update can be proven to bring their balance to 0. You should ideally only be able to prove the full withdrawal once the validator is withdrawn.

This PR addresses this issue by adding a check that only allows balance updates from before withdrawable epoch is set: https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#is_fully_withdrawable_validator


Sidu28:
@gpsanant note that this changes the EP constructor

script/upgrade/GoerliUpgrade1.s.sol
```diff
@@ -73,7 +73,8 @@ contract GoerliUpgrade1 is Script, Test {
                 delayedWithdrawalRouter,
                 eigenPodManager,
                 32e9,
-                75e7
+                75e7,
+                1616508000
             )
```
ChaoticWalrus, line 78: seems like a good opportunity for named parameter syntax (having a few integer args in a row, could be easy to confuse these)

src/test/EigenPod.t.sol
```diff
@@ -554,14 +555,17 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
     /// @notice verifies that multiple full withdrawals for a single validator fail
     function testDoubleFullWithdrawal() public returns(IEigenPod newPod) {
         newPod = testFullWithdrawalFlow();
+        uint64 withdrawalAmountGwei = Endian.fromLittleEndianUint64(withdrawalFields[BeaconChainProofs.WITHDRAWAL_VALIDATOR_AMOUNT_INDEX]);
+        emit log_named_uint("HELLOWOWOWOWO", newPod.nonBeaconChainETHBalanceWei());
+        uint64 leftOverBalanceWEI = uint64(withdrawalAmountGwei - newPod.MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR()) * uint64(GWEI_TO_WEI);
+        emit log("hello");
```
ChaoticWalrus, line 561: looks like some leftover logs that you probably meant to delete?

ChaoticWalrus (approved):
Seems technically correct, would appreciate just a little more polish prior to merging -- pointed out a couple nits I'd like to see addressed.

src/contracts/pods/EigenPod.sol
```diff
@@ -761,6 +775,13 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         return (int256(newAmountWei) - int256(currentAmountWei));
     }
 
+    // reference: https://github.com/ethereum/consensus-specs/blob/ce240ca795e257fc83059c4adfd591328c7a7f21/specs/bellatrix/beacon-chain.md#compute_timestamp_at_slot
+    function _computeSlotAtTimestamp(uint64 timestamp) internal view returns (uint64) {
```
gpsanant, line 779: fine with this, but does it make snese to put this in the beacon chain lib and rename to BeaconChainUtils.sol?

cc @ChaoticWalrus

src/contracts/pods/EigenPod.sol
```diff
@@ -205,6 +211,14 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
             "EigenPod.verifyBalanceUpdate: specified timestamp is too far in past"
         );
 
+        /**
+        * checks that a balance update can only be made before the validator is withdrawable.  If this is not checked
+        * anyone can prove the withdrawn validator's balance as 0 before the validator is able to prove their full withdrawal
+        */
+        require(Endian.fromLittleEndianUint64(validatorFields[BeaconChainProofs.VALIDATOR_WITHDRAWABLE_EPOCH_INDEX]) >
```
gpsanant, line 218: this should also check that the balance == 0. it could be that the validator is withdrawable but not withdrawn yet. pls update revert message accordingly

gpsanant (changes requested):
comment above

src/contracts/pods/EigenPod.sol
ChaoticWalrus: hmm, it would need to take the 'genesis timestamp' as input then, right?
I'm OK with either approach honestly.

src/contracts/pods/EigenPod.sol
```diff
@@ -205,6 +211,14 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+                (_computeSlotAtTimestamp(oracleTimestamp)) / BeaconChainProofs.SLOTS_PER_EPOCH, "EigenPod.verifyBalanceUpdate: balance update is being proven after a validator is withdrawable");
+        
+
```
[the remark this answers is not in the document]
Sidu28: Added!

script/upgrade/GoerliUpgrade1.s.sol
Sidu28: Is this script still in commission?

src/test/EigenPod.t.sol
Sidu28: cleared

src/contracts/pods/EigenPod.sol
Sidu28: I kinda like having the proofs be their own thing, we already have a few such functions in the EP

wadealexc (approved):
LGTM pending G and J feedback

src/contracts/pods/EigenPod.sol
Sidu28: fixed, good catch

src/contracts/pods/EigenPod.sol
```diff
@@ -205,6 +211,26 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        * uint64 oracleEpoch = _computeSlotAtTimestamp(oracleTimestamp)) / BeaconChainProofs.SLOTS_PER_EPOCH;
+        * require(validatorWithdrawableEpoch > oracleEpoch)
+        * checks that a balance update can only be made before the validator is withdrawable.  If this is not checked
+        * anyone can prove the withdrawn validator's balance as 0 before the validator is able to prove their full withdrawal
+        */
+        if (
+            Endian.fromLittleEndianUint64(validatorFields[BeaconChainProofs.VALIDATOR_WITHDRAWABLE_EPOCH_INDEX]) <=
+            (_computeSlotAtTimestamp(oracleTimestamp)) / BeaconChainProofs.SLOTS_PER_EPOCH
+        ) {
+            require(newRestakedBalanceGwei > 0, "EigenPod.verifyBalanceUpdate: validator is withdrawable but has not withdrawn");
```
gpsanant, line 231: balanceRoot

Sidu28: wut?

gpsanant (approved):
lgtm

PR #232, opened 2023-10-10
Update m2 deploy script

gpsanant:
- Updates our M2 deploy script


ChaoticWalrus (approved):
LGTM. Seems a little odd that the output is essentially overwriting existing files though; might want it to do something else instead?

gpsanant:
> LGTM. Seems a little odd that the output is essentially overwriting existing files though; might want it to do something else instead?

`deploymentOutputPath` is a bad name, we are not writing to it, i will call it `m1DeploymentOutputPath` instead

PR #234, opened 2023-10-11
Removed references to the slasher (left state variables alone)

wadealexc:
Reasoning - these are just simple method calls. We can always add them back in a future upgrade, even without a migration.

I'm removing them now because people (auditors, bounty hunters, ...) find them confusing and because the code has changed so much that it's no longer clear that the slasher will work the way it was intended to when these modifiers were added.


ChaoticWalrus (approved):
LGTM. IIRC there’s a couple tests that verify things happen (specific functions revert) when users are frozen. probably want to remove those too so they don’t just fail indefinitely

PR #238, opened 2023-10-12
Simplify EigenPod logic

wadealexc:
Changes:
* Got rid of `GWEI_TO_WEI` conversions everywhere in favor of just using Gwei and only converting to wei when e.g. sending to EPmgr or sending ETH
* Added a bunch of helper methods to BeaconChainProofs.sol that make grabbing values from validatorFields, withdrawalFields, etc much easier to read
* Reordered some checks and rewrote some comments for mental and spiritual clarity
* Rewrote `_processFullWithdrawal` to do the same thing but with less code


docs/core/EigenPodManager.md
```diff
@@ -266,6 +266,7 @@ For the validator whose balance should be updated, the caller must supply:
 
 *Requirements*:
 * Pause status MUST NOT be set: `PAUSED_EIGENPODS_VERIFY_BALANCE_UPDATE`
+* If the validator is withdrawable (withdrawable epoch is set), balance being proven MUST NOT be zero
```
ChaoticWalrus, line 269: the balance is only zero in the case where the validator has become exited (i.e. the "withdrawal has completed"), right?
Two thoughts here:
1) want to be sure there isn't some weird edge case where a validator gets slashed down to actual zero and pushing an update + proving a withdrawal are _both forbidden at the same time_
2) I feel like the intention of this requirement should be clarified here, because otherwise it kinda just seems like a weird quirk / raises the question of "why"

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -414,4 +406,100 @@ library BeaconChainProofs {
...
+
+    /**
+     * Getters for validator fields:
+     * 0: pubkey
+     * 1: withdrawal credentials
+     * 2: effective balance
+     * 3: slashed?
+     * 4: activation elligibility epoch
+     * 5: activation epoch
+     * 6: exit epoch
```
ChaoticWalrus, line 449: maybe I'm missing something, but not seeing these 4 getter functions

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -414,4 +406,100 @@ library BeaconChainProofs {
...
+        return 
+            Endian.fromLittleEndianUint64(validatorFields[VALIDATOR_WITHDRAWABLE_EPOCH_INDEX]);
+    }
+
+    /**
+     * Getters for withdrawal fields:
+     * 0: withdrawal index
+     * 1: validator index
+     * 2: execution address
+     * 3: withdrawal amount
```
ChaoticWalrus, line 487: likewise, not seeing getters for `withdrawal index` or  `execution address`

src/contracts/libraries/BeaconChainProofs.sol
wadealexc: Yeah, we don't have getters for all of these, because we don't need to use all of these fields. Should I update the comment to mention that?

wadealexc: I really just wanted a short and dirty reference to the indices being used in the getters we do have

docs/core/EigenPodManager.md
wadealexc: Good notes. I can fix the latter one now. 

The former, yes - good thing to raise to auditors, and hopefully some work improving our testing can make us more assured that this can't happen.

src/contracts/pods/EigenPod.sol
```diff
@@ -199,102 +188,93 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        // 2. Balance updates should be more recent than the most recent update
         require(
             validatorInfo.mostRecentBalanceUpdateTimestamp < oracleTimestamp,
             "EigenPod.verifyBalanceUpdate: Validators balance has already been updated for this timestamp"
         );
 
-        {
-            // verify ETH validator proof
-            bytes32 latestBlockRoot = eigenPodManager.getBlockRootAtTimestamp(oracleTimestamp);
+        // 3. Balance updates should not be "stale" (older than 4.5 hrs)
```
Sidu28, line 211: grouping this checks together is lit, can we make this comment a bit more clear lik "balance updates must be made within VERIFY_BALANCE_UPDATE_WINDOW_SECONDS of the oracleTimeStamp" or smthg

docs/core/EigenPodManager.md
wadealexc: Question on this -- is it possible for a validator's balance to reach 0 without having the "withdrawable epoch" set?

Because if so, the check this requirement references prevents `verifyBalanceUpdate` from functioning, maybe.

Sidu28 (approved):
Overall LGTM, seems like one EP test is breaking?

testProveSingleWithdrawalCredential

wadealexc:
> Overall LGTM, seems like one EP test is breaking?
> 
> testProveSingleWithdrawalCredential

Yes... it's also failing on master so I didn't think about it!

src/contracts/libraries/BeaconChainProofs.sol
Sidu28: probably not neessary to add getters we dont use?

src/contracts/pods/EigenPod.sol
wadealexc: The comment you linked specifically references the value backed by that constant, and the code itself uses the constant -- do you think this needs additional clarity?

Sidu28:
> > Overall LGTM, seems like one EP test is breaking?
> > testProveSingleWithdrawalCredential
> 
> Yes... it's also failing on master so I didn't think about it!

Ah ok, I will take a look at it separately

src/contracts/pods/EigenPod.sol
wadealexc: I could update the comment to mention the oracle timestamp - would that be better?

src/contracts/pods/EigenPod.sol
```diff
@@ -576,151 +554,150 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        /**
+         * Finally, the validator is fully withdrawn. Update their status and place in state:
+         */
 
-        // now that the validator has been proven to be withdrawn, we can set their restaked balance to 0
         validatorInfo.restakedBalanceGwei = 0;
         validatorInfo.status = VALIDATOR_STATUS.WITHDRAWN;
-        validatorInfo.mostRecentBalanceUpdateTimestamp = withdrawalHappenedTimestamp;
-
+        validatorInfo.mostRecentBalanceUpdateTimestamp = withdrawalTimestamp;
```
ChaoticWalrus, line 675: going to point out that elsewhere we set this field equal to the `oracleTimestamp`, whereas here we are setting it to when the withdrawal actually occurred
AFAICT there's no actual issue present here, but wanted to note the inconsistency in case there is one

ChaoticWalrus (approved):
Excellent PR! Added a few comments in places where I think a little more could be done, but this is a definite improvement as-is.

src/contracts/pods/EigenPod.sol
Sidu28: yup, thats because its the last balance update made becuase its a withdrawal.  With the other proofs, the oracleTimestamp is the timstamp of the update, whereas here its timestamp of the withdrawal which may be far in the past

wadealexc: that's a great thing to mention. I was thinking that this line might not be quite right.

Imagine a scenario where you manage to hit `_processFullWithdrawal` with multiple withdrawals, but you purposefully prove them out of order. Then the validator's `mostRecentBalanceUpdateTimestamp` isn't really accurate.

WDYT?

src/contracts/pods/EigenPod.sol
ChaoticWalrus: IMO it's OK, but I do think referencing the constant rather than the value tends to make more sense, just to ensure that the comment remains accurate, e.g. if the definition of the constant is ever changed.

src/contracts/libraries/BeaconChainProofs.sol
ChaoticWalrus: > probably not neessary to add getters we dont use?

yeah, I agree with that, just noting the seeming inconsistency with the comment here

src/contracts/libraries/BeaconChainProofs.sol
ChaoticWalrus: OK yeah, fair enough. I think it's fine.

src/contracts/libraries/BeaconChainProofs.sol
ChaoticWalrus: I think it's fine as-is, overall.

docs/core/EigenPodManager.md
ChaoticWalrus: > Question on this -- is it possible for a validator's balance to reach 0 without having the "withdrawable epoch" set?
> 
> Because if so, the check this requirement references prevents `verifyBalanceUpdate` from functioning, maybe.

I'm not personally sure. will page @gpsanant and @Sidu28 on this

src/contracts/pods/EigenPod.sol
wadealexc: aight -- that's a good point. pushed a fix

src/contracts/pods/EigenPod.sol
gpsanant: great point. i think this is not critical, since the logic around `mostRecentBalanceUpdateTimestamp` only matters when the validator is not active but still we should probably only allow it to increase

ChaoticWalrus: > WDYT?

Despite this appearing to have no security implications, I would prefer to not leave known inconsistencies in variable usage in.
My recommendation is passing the `oracleTimestamp` into this function and using it here. **However**, in the edge case of proving multiple withdrawals for the same validator at the same time, I think this would lead to a revert still -- would there be an issue with making these checks against the oracleTimestamp  (`validatorInfo.mostRecentBalanceUpdateTimestamp < oracleTimestamp`) _non-strict_ instead of strict? Because then we'd also solve this pesky edge case.

PR #243, opened 2023-10-13
Remove tests that were failing due to deleted slasher checks

wadealexc:
Removes tests that are now failing because we got rid of "onlyNotFrozen" et al.

Also fixes broken EigenPod test. The test was broken because it was using the wrong value to calculate restaked balance after verifying withdrawal credentials.

I fixed it by just removing the line entirely because, as it turns out, the helper function used to deploy the pod and prove credentials also has a test for the exact same thing, except it uses the correct value.

### DIFF
--- a/src/test/EigenPod.t.sol
+++ b/src/test/EigenPod.t.sol
@@ -701,17 +701,11 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
     // //                     validator status should be marked as ACTIVE
 
     function testProveSingleWithdrawalCredential() public {
-        // get beaconChainETH shares
-        int256 beaconChainETHBefore = eigenPodManager.podOwnerShares(podOwner);
-
         // ./solidityProofGen "ValidatorFieldsProof" 302913 true "data/withdrawal_proof_goerli/goerli_block_header_6399998.json"  "data/withdrawal_proof_goerli/goerli_slot_6399998.json" "withdrawal_credential_proof_302913.json"         setJSON("./src/test/test-data/withdrawal_credential_proof_302913.json");
          setJSON("./src/test/test-data/withdrawal_credential_proof_302913.json");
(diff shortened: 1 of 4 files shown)


src/test/EigenPod.t.sol
```diff
@@ -701,17 +701,11 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
...
-        int256 beaconChainETHBefore = eigenPodManager.podOwnerShares(podOwner);
-
         // ./solidityProofGen "ValidatorFieldsProof" 302913 true "data/withdrawal_proof_goerli/goerli_block_header_6399998.json"  "data/withdrawal_proof_goerli/goerli_slot_6399998.json" "withdrawal_credential_proof_302913.json"         setJSON("./src/test/test-data/withdrawal_credential_proof_302913.json");
          setJSON("./src/test/test-data/withdrawal_credential_proof_302913.json");
         IEigenPod pod = _testDeployAndVerifyNewEigenPod(podOwner, signature, depositDataRoot);
         bytes32 validatorPubkeyHash = getValidatorPubkeyHash();
 
-
-        int256 beaconChainETHAfter = eigenPodManager.podOwnerShares(pod.podOwner());
-        assertTrue(beaconChainETHAfter - beaconChainETHBefore == int256(_calculateRestakedBalanceGwei(pod.MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR())*GWEI_TO_WEI), "pod balance not updated correcty");
```
ChaoticWalrus, line 714: this test doesn't seem like it should be deleted ?

ChaoticWalrus (approved):
PR seems good, just noted the one test that I don't think should be deprecated -- if it's failing, I don't think it's because of the Slasher-integration changes

src/test/EigenPod.t.sol
ChaoticWalrus: oh, I see, it gets checked inside of `_verifyWithdrawalCredentials`, inside of `_testDeployAndVerifyNewEigenPod`

PR #246, opened 2023-10-13
Added slasher deprecation notice to README

samlaf:
Slasher contract is under heavy changes and redesign, so we should add an impossible-to-miss notice on our README.

This shouldn't however prevent AVS teams from starting to write their service managers and integrating with eigenlayer contracts. I have thus moved the AVS-Guide to `docs/experimental` instead of `docs/outdated`, and added a warning notice in there as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 <a name="introduction"/></a>
+
 # EigenLayer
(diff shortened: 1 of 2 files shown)


ChaoticWalrus (approved):
Works for me, thanks.

Issue #240, opened 2023-10-12
EigenPod: Edge case in `mostRecentBalanceUpdateTimestamp` when proving multiple withdrawals out of order


wadealexc:
From discussion here: https://github.com/Layr-Labs/eigenlayer-contracts/pull/238#discussion_r1357421778

The gist is that `mostRecentBalanceUpdateTimestamp` can be updated inconsistently depending on the order in which you prove withdrawals.

Sidu28:
Ok, I don't think I agree with the suggestion made to set mostRecentBalanceUpdateTimestamp  to oracleTimestamp.  Apart from not being as intuitive, proving withdrawals is permissionless, and setting it as such would brick proving any older withdrawals.  I think the correct solution here is to  set it to withdrawal timestamp, and if the withdrawals are out of order, just check if withdrawal timestamp > validatorInfo.mostRecentBalanceUpdateTimestamp, and only update it then

ChaoticWalrus:
> Ok, I don't think I agree with the suggestion made to set mostRecentBalanceUpdateTimestamp to oracleTimestamp. Apart from not being as intuitive, proving withdrawals is permissionless, and setting it as such would brick proving any older withdrawals. I think the correct solution here is to set it to withdrawal timestamp, and if the withdrawals are out of order, just check if withdrawal timestamp > validatorInfo.mostRecentBalanceUpdateTimestamp, and only update it then

Doesn't this mean that the current logic would brick withdrawals when proving them out of order? Am I missing something here? Would it just make more sense to set the timestamp to zero or something in these cases?

wadealexc:
Closing this because the consensus is that no update to the balance update timestamp is needed when processing full withdrawals.

PR #247, opened 2023-10-14
ep testing

Sidu28:
More EP testing, reconfigure the CI to work efficiently and not take 10 minutes lol.  It now takes 2


ChaoticWalrus:
@Sidu28 Cool idea with the parallelization of running the tests, but they just aren't running at all right now!
See e.g. https://github.com/Layr-Labs/eigenlayer-contracts/actions/runs/6525043953/job/17717316889?pr=247#step:5:120
I also think it being split into 77 different runs might be a tad much -- can we split this any more sanely, perhaps into like 5-10 suites?

src/contracts/pods/EigenPod.sol
```diff
@@ -672,7 +673,14 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
         validatorInfo.status = VALIDATOR_STATUS.WITHDRAWN;
-        validatorInfo.mostRecentBalanceUpdateTimestamp = withdrawalTimestamp;
+        /**
+        * if multiple full withdrawals are proven for the same validator, we ensure the latest timestamp is recorded
+        * if the withdrawals are proven out of order. A validator may have to make multiple full withdrawals if 
+        * they redeposit into their exited validator.
+        */
+        if (validatorInfo.mostRecentBalanceUpdateTimestamp < withdrawalTimestamp){
+            validatorInfo.mostRecentBalanceUpdateTimestamp = withdrawalTimestamp;   
+        }
```
ChaoticWalrus, line 683: so this change is specifically in reference to https://github.com/Layr-Labs/eigenlayer-contracts/issues/240 -- right? 
I commented some on the issue in Slack -- seems like we need to agree on the correct behavior before merging this change.

src/test/EigenPod.t.sol
```diff
@@ -1488,8 +1487,8 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
         address recipient,
         uint64 withdrawalAmountGwei,
         ValidatorInfo memory validatorInfo
-    ) public {
-        _processFullWithdrawal(
+    ) public returns(IEigenPod.VerifiedWithdrawal memory) {
+        return _processFullWithdrawal(
```
ChaoticWalrus, line 1491: why this change? seems sensible + presumably better for integrations, but noting that this changes the interface.

ChaoticWalrus: nvm, test file lol

src/test/EigenPod.t.sol
```diff
@@ -1498,4 +1497,18 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
...
+        address recipient,
+        uint64 withdrawalAmountGwei
+    ) public returns(IEigenPod.VerifiedWithdrawal memory) {
+        return _processPartialWithdrawal(
+            validatorIndex,
+            withdrawalHappenedTimestamp,
+            recipient,
+            withdrawalAmountGwei
+        );
+    }
```
ChaoticWalrus, line 1513: I think this 'wrapper' contract (that you've named `EPInternalFunctions`) should probably be broken out into a separate file

src/test/unit/EigenPodUnit.t.sol
```diff
@@ -275,4 +275,143 @@ contract EigenPodUnitTests is EigenPodTests {
...
+        cheats.deal(address(this), amount);
+        // simulate a withdrawal processed on the beacon chain, pod balance goes to 32 ETH
+        Address.sendValue(payable(address(newPod)), amount);
+        require(newPod.nonBeaconChainETHBalanceWei() == amount, "nonBeaconChainETHBalanceWei should be 32 ETH");
+
+        cheats.startPrank(podOwner);
+        newPod.withdrawBeforeRestaking();
+        cheats.stopPrank();
+
+        require(_getLatestDelayedWithdrawalAmount(podOwner) == amount, "Payment amount should be stake amount");
```
ChaoticWalrus, line 317: can we also check that whatever the name is for ETH sent to the EigenPod via the fallback is now zero?

src/test/unit/EigenPodUnit.t.sol
```diff
@@ -275,4 +275,143 @@ contract EigenPodUnitTests is EigenPodTests {
...
+        });
+        podInternalFunctionTester.processFullWithdrawal(0, pubkeyHash, timestamp += 5 , podOwner, withdrawalAmount, validatorInfo);
+        IEigenPod.ValidatorInfo memory info = podInternalFunctionTester.validatorPubkeyHashToInfo(pubkeyHash);
+
+        podInternalFunctionTester.processFullWithdrawal(0, pubkeyHash, lesserTimestamp , podOwner, withdrawalAmount, podInternalFunctionTester.validatorPubkeyHashToInfo(pubkeyHash));
+
+        IEigenPod.ValidatorInfo memory info2 = podInternalFunctionTester.validatorPubkeyHashToInfo(pubkeyHash);
+
+        require(info2.mostRecentBalanceUpdateTimestamp == timestamp, "mostRecentBalanceUpdateTimestamp should not have been updated");
+    }
```
ChaoticWalrus, line 338: this should have a comment or something at least explaining what the intended behavior is here.
The naming makes it sound like maybe it should revert to me (which it shouldn't)

src/test/unit/EigenPodUnit.t.sol
```diff
@@ -275,4 +275,143 @@ contract EigenPodUnitTests is EigenPodTests {
...
+            status: IEigenPod.VALIDATOR_STATUS.ACTIVE
+        });
+        IEigenPod.VerifiedWithdrawal memory vw = podInternalFunctionTester.processFullWithdrawal(0, pubkeyHash, 0, podOwner, withdrawalAmount, validatorInfo);
+
+        if(withdrawalAmount > podInternalFunctionTester.MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR()){
+            require(vw.amountToSendGwei == withdrawalAmount - podInternalFunctionTester.MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR(), "newAmount should be MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR");
+        }
+        else{
+            require(vw.amountToSendGwei == 0, "newAmount should be withdrawalAmount");
+        }
```
ChaoticWalrus, line 355: shouldn't you also check that the 2 amounts add up to the total?

Sidu28:
After discussion with @ChaoticWalrus , we decided to simply not update mostRecentBalanceupdateTimestamp at all in processFullWithdrawal.  It was unnecessary in the first place.  Relevant issue: #240

src/contracts/pods/EigenPod.sol
Sidu28: Resolved!

src/test/EigenPod.t.sol
Sidu28: haha yeah just to test the VerifiedWithdrawal being correct

src/test/EigenPod.t.sol
Sidu28: Fair, done

src/test/unit/EigenPodUnit.t.sol
Sidu28: added

src/test/unit/EigenPodUnit.t.sol
Sidu28: this test is no longer relevant

src/test/unit/EigenPodUnit.t.sol
Sidu28: discussed offline, resolved

PR #248, opened 2023-10-15
ZK partial withdrawal changes

Sidu28:
https://docs.google.com/document/d/17YRCKfg_5u_uT_o5ljEA2nop5qYP8SAw8bFvYOGKcyg/edit?usp=sharing


src/contracts/pods/EigenPod.sol
```diff
@@ -280,6 +289,41 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+            address(this),
+            oracleTimestamp,
+            requestNonce,
+            this.handleCallback.selector
+        );
+        /**
+        * Zero out the running total of partial withdrawals claimed.  Once another merkle proof for partial withdrawals is proven
+        * this running total wil once again start accumulating.
+        */
+        sumOfPartialWithdrawalsClaimedGwei = 0;
```
ChaoticWalrus, line 311: This doesn't seem right at all. Shouldn't we be checking the reported amount in the callback function and decreasing this (+ the amount of ETH forwarded on) as appropriate?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -57,12 +58,14 @@ contract EigenPodManager is
     function initialize(
         uint256 _maxPods,
         IBeaconChainOracle _beaconChainOracle,
+        IFunctionGateway _functionGateway,
```
ChaoticWalrus, line 61: do you want to maybe just make a setter function for this instead of including in the initializer?
I guess it depends what you find to be easier in tests, but if we won't be reinitializing anywhere, then either option works

src/contracts/pods/EigenPod.sol
```diff
@@ -280,6 +289,41 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+
+    /// @notice The callback function for the ZK proof fulfiller.
+    function handleCallback(bytes memory output, bytes memory context) external onlyFunctionGateway {
+        uint256 partialWithdrawalSumWei;
+        assembly {
+            partialWithdrawalSumWei := mload(add(output, 0x20))
+        }
+        uint256 requestNonce = abi.decode(context, (uint256));
+        emit PartialWithdrawalProven(requestNonce, partialWithdrawalSumWei);
+    }
```
ChaoticWalrus, line 325: so after here is when we would calculate the amount of ETH to send + actually forward it on

PR #250, opened 2023-10-16
update specs to include "nondet" summary of the calls to `ServiceMana…

ChaoticWalrus:
…ger.updateStakes`

This basically means treating the function as 'view' -- see documentation here https://docs.certora.com/en/latest/docs/cvl/methods.html#summaries Seems like a reasonable assumption to add.

### DIFF
--- a/certora/specs/core/DelegationManager.spec
+++ b/certora/specs/core/DelegationManager.spec
@@ -6,6 +6,9 @@ methods {
     function decreaseDelegatedShares(address,address,uint256) external;
 	function increaseDelegatedShares(address,address,uint256) external;
 
+    // external calls from DelegationManager to ServiceManager
+    function _.updateStakes(address[]) external => NONDET;
+
 	// external calls to Slasher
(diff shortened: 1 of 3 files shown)


ChaoticWalrus:
Passing run of the StrategyManager: https://prover.certora.com/output/83341/471cb2a31873431f81731c31c6541eaf?anonymousKey=69b08be166b147f843457c284402eca9c6182154

ChaoticWalrus:
I'm unsure why the BLSRegistryCoordinatorWithIndices spec is failing -- it is hitting 'havoc' behavior when it seems that the behavior _should_ be dispatched.
E.g. failure https://github.com/Layr-Labs/eigenlayer-contracts/actions/runs/6536819518/job/17749363704#step:9:41 and the run itself https://prover.certora.com/output/83341/c7e73cda434a43f7807c9a4893141750?anonymousKey=6c0e0c779caa20fc925e82d0a9dde4b28aff28d3

PR #252, opened 2023-10-16
Remove useless stake updates and fix issue where operator was set inc…

wadealexc:
…orrectly for withdrawal processing


ChaoticWalrus (approved):
LGTM, good catch on the bug

PR #253, opened 2023-10-16
add test which verifies that delegated shares are added to the correc…

ChaoticWalrus:
…t operator when completing a withdrawal as shares

also fix the EigenPodManagerMock to return an appropriate value for the 'increase in delegateable shares'


wadealexc (approved):
LGTM! thanks for the test.

PR #255, opened 2023-10-17
Validator Balance update Refactor

Sidu28:
This PR refactors the validator balance updater to process multiple balance updates.  We allow this for verifying withdrawal credentials and withdrawals, it makes sense to do it here as well

### DIFF
--- a/src/contracts/pods/EigenPod.sol
+++ b/src/contracts/pods/EigenPod.sol
@@ -503,19 +503,16 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         bytes32 validatorPubkeyHash = validatorFields.getPubkeyHash();
         ValidatorInfo memory validatorInfo = _validatorPubkeyHashToInfo[validatorPubkeyHash];
 
-
-        // Verify balance update timing:
-
-        // 1. Balance updates should only be performed on "ACTIVE" validators
-        require(
-            validatorInfo.status == VALIDATOR_STATUS.ACTIVE, 
-            "EigenPod.verifyBalanceUpdate: Validator not active"
-        );
-
-        // 2. Balance updates should be more recent than the most recent update
+        // 1. Balance updates should be more recent than the most recent update
         require(
             validatorInfo.mostRecentBalanceUpdateTimestamp < oracleTimestamp,
             "EigenPod.verifyBalanceUpdate: Validators balance has already been updated for this timestamp"
+        );
+
+        // 2. Balance updates should only be performed on "ACTIVE" validators
+        require(
+            validatorInfo.status == VALIDATOR_STATUS.ACTIVE, 
+            "EigenPod.verifyBalanceUpdate: Validator not active"
         );
 
         // 4. Balance updates should only be made before a validator is fully withdrawn. 
--- a/src/contracts/pods/EigenPod.sol
+++ b/src/contracts/pods/EigenPod.sol
@@ -212,8 +212,9 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
             stateRootProof: stateRootProof.proof
         });
 
+        int256 sharesDeltaGwei;
         for (uint256 i = 0; i < validatorIndices.length; i++) {
-            _verifyBalanceUpdate(
+            sharesDeltaGwei += _verifyBalanceUpdate(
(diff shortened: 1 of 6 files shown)


Sidu28:
@gpsanant this changes interfaces a bit, shouldn't be too bad

src/contracts/pods/EigenPod.sol
```diff
@@ -172,41 +172,31 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         emit NonBeaconChainETHReceived(msg.value);
     }
 
+//     BeaconState(epochToTimestamp(validator.withdrawableEpoch))
+//     .next_withdrawal_validator_index 
+// <= validator.Index 
+// < BeaconState(oracleTimestamp).next_withdrawal_validator_index
+
```
ChaoticWalrus, line 179: why is this comment being added here? seems totally out of place

src/contracts/pods/EigenPod.sol
```diff
@@ -538,6 +491,78 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        bytes32 beaconStateRoot,
+        BeaconChainProofs.BalanceUpdateProof calldata balanceUpdateProof,
+        bytes32[] calldata validatorFields
+    ) internal {
+        
+        uint64 validatorBalance = balanceUpdateProof.balanceRoot.getBalanceAtIndex(validatorIndex);
+        bytes32 validatorPubkeyHash = validatorFields.getPubkeyHash();
+        ValidatorInfo memory validatorInfo = _validatorPubkeyHashToInfo[validatorPubkeyHash];
+
+        // 1. Balance updates should be more recent than the most recent update
```
ChaoticWalrus, line 506: nit: the numbering of these checks is weird now, since (3) is now before (1)

src/contracts/pods/EigenPod.sol
```diff
@@ -538,6 +491,78 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        // If our new and old balances differ, calculate the delta and send to the EigenPodManager
+        if (newRestakedBalanceGwei != currentRestakedBalanceGwei) {
+            emit ValidatorBalanceUpdated(validatorIndex, oracleTimestamp, newRestakedBalanceGwei);
+
+            int256 sharesDeltaGwei = _calculateSharesDelta({
+                newAmountGwei: newRestakedBalanceGwei,
+                previousAmountGwei: currentRestakedBalanceGwei
+            });
+            eigenPodManager.recordBeaconChainETHBalanceUpdate(podOwner, sharesDeltaGwei * int256(GWEI_TO_WEI));
+        }
```
ChaoticWalrus, line 563: shouldn't you group these together and just make a single call at the end, outside of the loop, as we now do with withdrawals?

src/contracts/pods/EigenPod.sol
Sidu28: whoops that was supposed to be deleted, mbad

src/contracts/pods/EigenPod.sol
ChaoticWalrus: to elaborate, seems like the calls to `eigenPodManager.recordBeaconChainETHBalanceUpdate` could be bundled together inside the main body of `verifyBalanceUpdates` -- the internal function `_verifyBalanceUpdate` could return the sharesDeltaGwei for this.

ChaoticWalrus (changes requested):
Seems like a good change. Should also update docs, and we should align w/ FE etc on merging this.

src/contracts/pods/EigenPod.sol
Sidu28: fixed

src/contracts/pods/EigenPod.sol
Sidu28: done

ChaoticWalrus (approved):

ChaoticWalrus:
Seems good now. @Sidu28 would be good to see test coverage for all these functions that accept arrays (even just a simple test w/ 2 inputs, preferably with the validators having different amounts so we can be sure that we don't have an error of e.g. reading always from one entry of the array) -- capturing these as TODOs for the moment is OK by me though.

PR #256, opened 2023-10-17
add event

gpsanant:
we need to emit this for the backend to work

### DIFF
--- a/src/contracts/pods/EigenPod.sol
+++ b/src/contracts/pods/EigenPod.sol
@@ -164,6 +164,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
          * initialized with hasRestaked set to true.
          */
         hasRestaked = true;
+        emit RestakingActivated(podOwner);
     }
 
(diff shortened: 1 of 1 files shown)


src/contracts/pods/EigenPod.sol
```diff
@@ -164,6 +164,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
          * initialized with hasRestaked set to true.
          */
         hasRestaked = true;
+        emit RestakingActivated(podOwner);
```
ChaoticWalrus, line 167: IIRC we set `hasRestaked = true` in other places as well!

ChaoticWalrus: are those just already covered then?

gpsanant: ye we set it at one other place, and we are only emitting there before this change :)

ChaoticWalrus:
LGTM

ChaoticWalrus (approved):
LGTM'd

PR #259, opened 2023-10-17
Added ZK storage to EP contracts for forward compatibility

Sidu28:


wadealexc:
Is this ready for review now, or still WIP?

src/contracts/pods/EigenPod.sol
```diff
@@ -93,6 +93,18 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     /// @notice This variable tracks any ETH deposited into this contract via the `receive` fallback function
     uint256 public nonBeaconChainETHBalanceWei;
 
+     /// @notice This variable tracks the total amoutn of partial withdrawals claimed via merkle proofs prior to a switch to ZK proofs for claiming partial withdrawals
+    uint64 public totalPartialWithdrawalAmountClaimedGwei;
```
wadealexc, line 97: This name is word soup. Can we do something shorter?

How about `sumPartialWithdrawalsClaimedGwei`

If we can get something shorter that's dope too

wadealexc:
Why are we adding unused variables?

Sidu28:
> Why are we adding unused variables?

to help with forward compatibility, ie, storage already exists for when we upgrade implementation to ZK proofs

ChaoticWalrus:
> > Why are we adding unused variables?
> 
> to help with forward compatibility, ie, storage already exists for when we upgrade implementation to ZK proofs

We should not be adding purely unused storage variables. This PR should cover logic + storage changes that are _necessary_ for forwards compatibility, and exclude adding totally unused storage variables that you think you might be added later.

src/contracts/pods/EigenPod.sol
```diff
@@ -93,6 +93,9 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     /// @notice This variable tracks any ETH deposited into this contract via the `receive` fallback function
     uint256 public nonBeaconChainETHBalanceWei;
 
+     /// @notice This variable tracks the total amoutn of partial withdrawals claimed via merkle proofs prior to a switch to ZK proofs for claiming partial withdrawals
```
ChaoticWalrus, line 96: typo: `amoutn` => `amount`

ChaoticWalrus (approved):
LGTM. noted one minor typo in a comment.

Sidu28:
> > > Why are we adding unused variables?
> > 
> > 
> > to help with forward compatibility, ie, storage already exists for when we upgrade implementation to ZK proofs
> 
> We should not be adding purely unused storage variables. This PR should cover logic + storage changes that are _necessary_ for forwards compatibility, and exclude adding totally unused storage variables that you think you might be added later.

removed the extraneous stuff, take a look now

src/contracts/pods/EigenPod.sol
Sidu28: fixed

PR #262, opened 2023-10-18
added multiple queuedwithdrawals function

Sidu28:
Adding ability to prove multiple queuedWithdrawals in a single function call

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -264,6 +264,34 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         });
     }
 
+    function queueWithdrawals(
+         IStrategy[][] calldata strategies,
+        uint256[][] calldata shares,
+        address[] calldata withdrawers
+    ) external onlyWhenNotPaused(PAUSED_ENTER_WITHDRAWAL_QUEUE) returns (bytes32[] memory) {
+        require(strategies.length == shares.length && shares.length == withdrawers.length, "DelegationManager.queueWithdrawals: input length mismatch");
+        bytes32[] memory withdrawalRoots = new bytes32[](strategies.length);
+
+        for (uint256 i = 0; i < strategies.length; i++) {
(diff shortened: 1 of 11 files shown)


src/contracts/core/DelegationManager.sol
```diff
@@ -264,6 +264,34 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         });
     }
 
+    function queueWithdrawals(
+         IStrategy[][] calldata strategies,
```
gpsanant, line 268: i think we should create a `QueuedWithdrawaParam` struct and the param here should be `QueuedWithdrawaParam[]`

src/contracts/core/DelegationManager.sol
```diff
@@ -264,6 +264,34 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        uint256[][] calldata shares,
+        address[] calldata withdrawers
+    ) external onlyWhenNotPaused(PAUSED_ENTER_WITHDRAWAL_QUEUE) returns (bytes32[] memory) {
+        require(strategies.length == shares.length && shares.length == withdrawers.length, "DelegationManager.queueWithdrawals: input length mismatch");
+        bytes32[] memory withdrawalRoots = new bytes32[](strategies.length);
```
gpsanant, line 273: also, please delete the other queueWithdrawal function

ChaoticWalrus (commented):
Seems reasonable, would really like to see some test coverage added.

Should we leave a singleton method? I guess it's fine just having the array-ified version though.

Have you already coordinated with others on this interface change?

ChaoticWalrus (approved):
I am approving but I'd say wait for @gpsanant to confirm prior to merging

Sidu28:
> Seems reasonable, would really like to see some test coverage added.
> 
> Should we leave a singleton method? I guess it's fine just having the array-ified version though.
> 
> Have you already coordinated with others on this interface change?

Gautham asked for this to replace the singleton method

UI is ok with the interfact change since apparently no balance updates for M2

PR #263, opened 2023-10-19
Storage Upgrade Checker

ypatil12:
This PR has scripts to validate the storage slots when upgrading contracts. To validate the upgrade of a contract on-chain to a local contract, run:

`bash script/upgrade/validateUpgrade.sh -n <network> -c <contract> -a <implementation address> -k <optional param to keep storage layout csvs>`. 

This script uses cast and forge inspect to get the layouts of the local and on-chain contract. Your `ETHERSCAN_API_KEY` must be set as an environment variable. The storage layouts are saved into csv files and passed into the typescript helper `validateStorage.ts`, which takes paths to two csv layouts and validates the storage slots. 

Storage slot validation is NOT comprehensive, and errs on the side of caution. We recommend using this script as a tool along with manual storage slot verification. The validation is opinionated on storage for each contract consuming 50 slots and gaps being sized accordingly.

The script _does not_ validate [legal type changes](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/65cf285bd36af24570186ca6409341540c67238a/packages/core/contracts/test/RetypeFromContract.sol) (ie. from bool to uint8) and errors out if the types of slots have updated, including having different struct names. A manual check will need to be done to validate this conversion. In addition, the script does not support [non-contiguous gaps](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/65cf285bd36af24570186ca6409341540c67238a/packages/core/contracts/test/Storage.sol#L784)

This is version 1 of the script, and further improvements will be added as we have more complex storage layouts to check against.


ypatil12:
Outputs for 5 upgradeable contracts. Will make PRs for the incorrect gaps

DelegationManager:
image

StrategyManager:
image

EigenPodManager:
image

EigenPod:
image

Slasher: No warnings/errors

ChaoticWalrus (approved):
PR looks 🔥 🔥 🔥 , thank you immensely for this.

I think adding some instructions / context like what's already in your PR to the documentation (not sure where is best, tbh) would be great, if you can.

I also think an example for the syntax would be helpful, as I wasn't totally sure on the format. `bash script/upgrade/validateUpgrade.sh -n goerli -c strategyManager  -a 0x3FFa9daE46d15f15c925d4694c19c49e624b3dE6` worked well for me.

Would love to see this further automated (maybe something for the CI eventually) in the future, but this is great as-is.

Please do make a follow-up PR addressing the issues you've identified (sounds like just gap sizes? although we should make sure the semantic changes are all good as well).

ypatil12:
@ChaoticWalrus Added a README in the script/upgrades folder

PR #264, opened 2023-10-19
Add `bool isPartialWithdrawal` to DelayedWithdrawalRouter struct

JasonVranek:
This PR adds a bool `isPartialWithdrawal` to the `DelayedWithdrawal` struct and modifies the necessary EigenPod logic and tests.  The goal is to specify whether a `DelayedWithdrawal` pertains to partial or full withdrawal ETH which is helpful for proof of reserves for protocols that leverage native restaking.


wadealexc:
As of the upcoming upgrade, full withdrawals no longer go through the DWR. Closing this as outdated :)

PR #260, opened 2023-10-18
Remove Middleware from Core

stevennevins:

PR #267, opened 2023-10-19
DeDupe EigenPod and Withdrawals Tests

ypatil12:
- EigenPod: Combined unit test file into integration test file, as unit tests were not using mocks anyways
- Withdrawals: Removed inheritance from DelegationTest. Ported over helper functions to withdrawals

No more duped tests!


src/test/Withdrawals.t.sol
```diff
@@ -17,12 +18,85 @@ contract WithdrawalTests is DelegationTests {
         uint96 nonce;
     }
 
+    address public registryCoordinator = address(uint160(uint256(keccak256("registryCoordinator"))));
+    ServiceManagerMock public serviceManager;
+    StakeRegistryHarness public stakeRegistry;
+    StakeRegistryHarness public stakeRegistryImplementation;
```
ChaoticWalrus, line 24: Flagging the use of this as a potential merge conflict with the branch that removes all middleware contracts -- see https://github.com/Layr-Labs/eigenlayer-contracts/pull/260 or more specifically https://github.com/Layr-Labs/eigenlayer-contracts/blob/209e6c85d760b7913bfa286e59ab0189714219f9/src/test/EigenLayerTestHelper.t.sol#L7-L9
@stevennevins tagging you here for visibility
My thought is that this is probably OK to merge rn, we'll just have to iron out the issue afterwards.

ChaoticWalrus (approved):
Good changes, don't know how I hadn't realized that the EigenPod "unit" tests inherited from the other file.
Just had the one note about merge conflicts, but I think this should still be fine to merge at the moment.

PR #270, opened 2023-10-20
Update Gaps for EigenPodManager and StrategyManager

ypatil12:
- Gap on StrategyManager decreased from size 41 to 40. Any contracts inheriting from SM will begin storage at the 201st slot
- 2nd to last gap on EigenPodManager increased from 40 to 43. The `status` variable now correctly begins at slot 201

This PR addresses the gap warnings from the storage layout script in #263


ChaoticWalrus (approved):
I manually double-checked these and they both look correct. Thanks for fixing these!

PR #275, opened 2023-10-24
Repo Cleanup: Separating test files into separate contracts

8sunyuan:
Setup
- Create a separate directory dedicated for testing of a single contract. Separate specific external functions into separate contracts, mimicking describe blocks.
- Create parent contract with setUp and internal helper functions defined for all other function specific test contracts to inherit from. For ex. StrategyManagerUnit.t.sol for StrategyManagerUnit directory
- For testing StrategyManangerUnit tests in the corresponding directory
`forge test --match-contract StrategyManangerUnitTests`

Motivation
- Test files were getting really longer and this makes it much easier to navigate testing for a contract, especially if you just want to add/check testing for a specific function.
- Better organization as the parent testing contract contains the helpers, events, constants, and other info that are needed across all tests.
- If the file doesn't exist for a specific function or is mostly empty, we can easily recognize lack of coverage.
- Supposedly compiles faster according to foundry docs

Resources/Examples
https://book.getfoundry.sh/tutorials/best-practices?highlight=best#general-test-guidance
https://github.com/sablier-labs/v2-core/tree/main/test/unit/concrete/nft-descriptor


ypatil12 (commented):
I'm not sure if we get much benefit in terms of organization from splitting out all of the StrategyManager unit tests into separate files. There is a lot of moving around from one file to another to really understand what is going on. IMO, it might be better if we combine the files back into one and organize the tests by functionality with headers and clear/consistent naming on each: [example](https://github.com/Layr-Labs/eigenlayer-contracts/blob/feat/EigenPodManager-test-refactor/src/test/unit/EigenPodManagerUnit.t.sol).

stevennevins:
I'm a fan of splitting them up by Function, but not specifically in a new file. It can reduce compile/run time for tests by having smaller test contracts. I am not super opposed to creating a new file for each, though

FooTest is ContractUnderTest {}

BarTest is ContractUnderTest {}

Exist in ContractUnderTest.t.sol

PR #287, opened 2023-10-25
EP quick bug fix

Sidu28:
Bug fix related to a line that was added that required those proving WC to have greater than 0 balance - this bricks validators that have exited before proving WC: https://github.com/Layr-Labs/eigenlayer-contracts/commit/7f9eb4789dbfea73442472dae8bacc20c1f9c920


ChaoticWalrus (approved):
LGTM
I confirmed the regression test properly fails without the fix + passes with the fix. We agreed offline on the removal of the check being OK since overflow is completely unrealistic here (which is what the check was designed to catch).

PR #289, opened 2023-10-25
require cannot q withdrawal for 0 strategies and update docs

gpsanant:
Make it so that withdrawals for 0 strats cannot occur

Special case:
- return bytes32(0) if undelegating with no delegatable shares

### DIFF
--- a/docs/core/DelegationManager.md
+++ b/docs/core/DelegationManager.md
@@ -241,7 +241,7 @@ The withdrawal can be completed by the `withdrawer` after `withdrawalDelayBlocks
 *Requirements*:
 * Pause status MUST NOT be set: `PAUSED_ENTER_WITHDRAWAL_QUEUE`
 * `strategies.length` MUST equal `shares.length`
-* `strategies.length` MUST be not equal to 0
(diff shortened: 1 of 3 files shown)


docs/core/DelegationManager.md
```diff
@@ -240,6 +241,7 @@ The withdrawal can be completed by the `withdrawer` after `withdrawalDelayBlocks
 *Requirements*:
 * Pause status MUST NOT be set: `PAUSED_ENTER_WITHDRAWAL_QUEUE`
 * `strategies.length` MUST equal `shares.length`
+* `strategies.length` MUST be not equal to 0
```
wadealexc, line 244: `strategies.length` MUST NOT be equal to 0

wadealexc:
LGTM pending a tweak to docs. Thanks for updating them :)

wadealexc (changes requested):
LGTM pending tweak to docs. Thanks :)

ChaoticWalrus (approved):
LGTM. Agreed with Alex's nit though.

PR #290, opened 2023-10-25
EigenPodManager Unit Test Refactor

ypatil12:
This PR refactors unit tests for the EigenPodManager per the [new smart contract testing guide](https://docs.google.com/document/d/107Wn4cVjSOTng1T72TWhhzswj1Afw4Dz7fUXQWqzLT0/edit?usp=sharing)

Working doc with invariants to monitor and integration test todos [here](https://docs.google.com/document/d/1GHfkPHR6asAjSqz2u8ZCgXdLbs6HX8HEzotLT3HIDE4/edit?usp=sharing).


src/test/tree/EigenPodManager.tree
```diff
@@ -0,0 +1,95 @@
...
+│       └── it should set the max pods
+├── when updateBeaconChainOracle is called
+│   ├── given the user is not the owner
+│   │   └── it should revert
+│   └── given the user is the owner
+│       └── it should set the beacon chain oracle
+├── when addShares is called
+│   ├── given that the caller is not the delegationManager
+│   │   └── it should revert
+│   ├── given that the podOwne address is 0
```
ChaoticWalrus, line 31: tiny typo `podOwne` => `podOwner`

src/test/tree/EigenPodManager.tree
```diff
@@ -0,0 +1,95 @@
...
+│   │   └── it should revert
+│   ├── given that the shares amount is negative
+│   │   └── it should revert
+│   ├── given that the shares is not a whole gwei amount
+│   │   └── it should revert
+│   ├── given that too many shares are withdrawn
+│   │   └── it should revert
```
ChaoticWalrus, line 38: this seems out of place? it's under `when addShares is called` right now

src/test/tree/EigenPodManager.tree
```diff
@@ -0,0 +1,95 @@
...
+│       └── it should update the podOwnerShares
+├── when removeShares is called
+│   ├── given that the caller is not the delegationManager
+│   │   └── it should revert
+│   ├── given that the shares amount is negative
+│   │   └── it should revert
+│   ├── given that the shares is not a whole gwei amount
+│   │   └── it should revert
+│   ├── given that too many shares are removed
+│   │   └── it should revert
```
ChaoticWalrus, line 49: I'm unsure what the correct amount of detail is for these tree diagrams, but think that "too many shares are removed" is quite vague -- specifically, we revert if the amount removed would cause the podOwner's shares to become negative.
Let me know if you think being this specific in the tree diagram is overkill! I'm not personally sure what's better, since more concise is more readable.

src/test/tree/EigenPodManager.tree
```diff
@@ -0,0 +1,95 @@
...
+│   └── given that all of the above conditions are satisfied
+│       └── it should update the podOwnerShares
+├── when shares are withdrawn as tokens
```
ChaoticWalrus, line 52: should this be `when withdrawSharesAsTokens is called` ?

src/test/tree/EigenPodManager.tree
```diff
@@ -0,0 +1,95 @@
...
+│   │   └── it should revert
+│   ├── given that the destination is address 0
+│   │   └── it should revert
+│   ├── given that the shares amount is negative
+│   │   └── it should revert
+│   ├── given that the shares is not a whole gwei amount
+│   │   └── it should revert
+│   ├── given that the current podOwner shares are negative
+│   │   ├── given that the shares to withdraw are greater than the negative shares of the podOwner
+│   │   │   └── it should set the podOwnerShares to 0 and decrement shares to withdraw by the share defecit
```
ChaoticWalrus, line 63: I find this wording hard to parse. maybe
`given that the shares to withdraw are larger in magnitude than the shares of the podOwner` instead?
should we also explain that any remainder moves on to the 'withdraw restaked ETH from the eigenPod' step?

also minor typo: `defecit` => `deficit`

src/test/tree/EigenPodManager.tree
```diff
@@ -0,0 +1,95 @@
...
+│   ├── given that sharesBefore is negative or 0
+│   │   ├── given that sharesAfter is negative or zero
+│   │   │   └── the change in delegateable shares should be 0
+│   │   └── given that sharesAfter is positive
+│   │       └── the change in delegateable shares should be positive
+│   └── given that sharesBefore is positive
+│       ├── given that sharesAfter is negative or zero
+│       │   └── the change in delegateable shares is negative sharesBefore
+│       └── given that sharesAfter is positive
+│           └── the change in delegateable shares is the difference between sharesAfter and sharesBefore
```
ChaoticWalrus, line 78: feels like a description of the paths in the internal `_calculateChangeInDelegatableShares` function -- should we perhaps just use wrappers that expose these internal functions for unit testing (e.g. `contract EigenPodManagerWrapper is EigenPodManager ...`), and include these paths in the tree structure diagrams in general?

src/test/tree/EigenPodManager.tree
```diff
@@ -0,0 +1,95 @@
...
+└── when recordBeaconChainETHBalanceUpdate is called
+    ├── given that the podOwner is not the caller
+    │   └── it should revert
```
ChaoticWalrus, line 81: the modifier might be somewhat confusing, but the caller here has to be the podOwner's EigenPod, not the podOwner themself!

ypatil12: Yup, typo

src/test/unit/EigenPodManagerUnit.t.sol
```diff
@@ -117,220 +85,438 @@ contract EigenPodManagerUnitTests is Test, EigenPodPausingConstants {
...
-        cheats.startPrank(address(improperCaller));
-        eigenPodManager.recordBeaconChainETHBalanceUpdate(staker, int256(0));
-        cheats.stopPrank();
+    function _initializePodWithShares(address podOwner, int256 shares) internal {
+        _deployAndReturnEigenPodForStaker(podOwner);
+        // Signature of `podOwnerShares(address)
+        bytes4 signature = 0x60f4062b;
+        bytes32 sharesToSet = bytes32(uint256(shares));
+        stdstore.target(address(eigenPodManager)).sig(signature).with_key(podOwner).checked_write(sharesToSet);
+    }
```
ChaoticWalrus, line 110: is there a reason for using stdstore here instead of calling the `recordBeaconChainETHBalanceUpdate`?

seems potentially convenient for some setup but also like it could result in situations in which we are simulating otherwise inaccessible states (e.g. if `shares` is a non-whole Gwei amount, and we are otherwise successfully enforcing that user shares are always a whole-Gwei amount)

not trying to bash this, it's cool, honestly just trying to understand + offer my perspective

src/test/tree/EigenPodManager.tree
ypatil12: We should strive to be as concise as possible to get the gist of the execution. Updated this to be: `given that removing shares results in the pod owner having negative shares`

src/test/unit/EigenPodManagerUnit.t.sol
```diff
@@ -117,220 +85,438 @@ contract EigenPodManagerUnitTests is Test, EigenPodPausingConstants {
...
-
-    //     // actually create the queued withdrawal, and check for event emission
-    //     cheats.startPrank(staker);
+    function _checkPodDeployed(address staker, uint256 numPodsBefore) internal {
+        // Get expected pod address
+        IEigenPod expectedPod = eigenPodManager.getPod(staker);
+
+        assertEq(address(eigenPodManager.ownerToPod(staker)), address(expectedPod));
+        assertEq(eigenPodManager.numPods(), numPodsBefore + 1);
+    }
```
ChaoticWalrus, line 131: should this be called inside of `_deployAndReturnEigenPodForStaker`? or would that be overkill / less clear than calling this in other places?

src/test/unit/EigenPodManagerUnit.t.sol
```diff
@@ -117,220 +85,438 @@ contract EigenPodManagerUnitTests is Test, EigenPodPausingConstants {
...
-    //     uint256 nonceBefore = eigenPodManager.cumulativeWithdrawalsQueued(staker);
-    //     int256 sharesBefore = eigenPodManager.podOwnerShares(staker);
-
-    //     // actually create the queued withdrawal, and check for event emission
-    //     cheats.startPrank(staker);
+    function _checkPodDeployed(address staker, uint256 numPodsBefore) internal {
+        // Get expected pod address
+        IEigenPod expectedPod = eigenPodManager.getPod(staker);
+
+        assertEq(address(eigenPodManager.ownerToPod(staker)), address(expectedPod));
```
ChaoticWalrus, line 129: seems vacuous -- since `eigenPodManager.getPod` reads from `eigenPodManager.ownerToPod`, this will always pass. seems like we'd need to store the `expectedPod` _prior_ to deploying an EigenPod, if we wanted to check what I think this is designed to check

src/test/tree/EigenPodManager.tree
ypatil12: Good suggestion, was planning on testing in integration but this is a better way to do it

src/test/unit/EigenPodManagerUnit.t.sol
ypatil12: Likely overkill, since we there is no need to assert the pod was properly deployed every time in every test. Tests should only test what they are designed to test and ideally nothing more/less

src/test/unit/EigenPodManagerUnit.t.sol
```diff
@@ -117,220 +85,438 @@ contract EigenPodManagerUnitTests is Test, EigenPodPausingConstants {
...
+        bytes32 value = bytes32(eigenPodManager.maxPods());
+        cheats.store(address(eigenPodManager), slot, value);
+
+        //Create pod & expect revert based on maxPods size
+        if (eigenPodManager.maxPods() == type(uint256).max) {
+            // Arithmetic over/underflow not working with foundry
+            cheats.expectRevert();
+        } else {
+            cheats.expectRevert("EigenPodManager._deployPod: pod limit reached");
+        }
```
ChaoticWalrus, line 241: since we _are_ setting maxPods to the max uint256 in the setUp, this won't ever actually hit the `else` block / check that this reversion is happening.
I think it probably makes sense to reduce `maxPods` first, or otherwise make sure we are checking both paths here?

src/test/unit/EigenPodManagerUnit.t.sol
```diff
@@ -117,220 +85,438 @@ contract EigenPodManagerUnitTests is Test, EigenPodPausingConstants {
...
+    
+    function testFuzz_removeShares_revert_sharesNotWholeGwei(uint256 shares) public {
+        cheats.assume(int256(shares) >= 0);
+        cheats.assume(shares % GWEI_TO_WEI != 0);
+        cheats.prank(address(delegationManagerMock));
+        cheats.expectRevert("EigenPodManager.removeShares: shares must be a whole Gwei amount");
+        eigenPodManager.removeShares(defaultStaker, shares);
+    }
+
+    // Fuzzer rejects too many inputs, unit testsing
```
ChaoticWalrus, line 353: you can start with a small uint size fuzzed input and then multiply it by GWEI_TO_WEI ?
like `uint96 fuzzedInput ... uint256 sharesToRemove = fuzzedInput * GWEI_TO_WEI`

src/test/unit/EigenPodManagerUnit.t.sol
ypatil12: Yup, good catch - updated order

src/test/unit/EigenPodManagerUnit.t.sol
```diff
@@ -117,220 +85,438 @@ contract EigenPodManagerUnitTests is Test, EigenPodPausingConstants {
...
+        // Deploy Pod And initialize with negative shares
+        _initializePodWithShares(defaultStaker, sharesBeginning);
+
+        // Withdraw shares
+        cheats.prank(address(delegationManagerMock));
+        eigenPodManager.withdrawSharesAsTokens(defaultStaker, defaultStaker, sharesToWithdraw);
+
+        // Check storage update
+        assertEq(eigenPodManager.podOwnerShares(defaultStaker), int256(0));
+    }
```
ChaoticWalrus, line 445: it would be cool if after this (and similarly in the `test_withdrawSharesAsTokens_withdrawPositive` test) we could check that the contract is correctly calling `EigenPod.withdrawRestakedBeaconChainETH` ... might be a more general problem to sort out a solution to, definitely not going to block this PR on this

src/test/unit/EigenPodManagerUnit.t.sol
ypatil12: My philosophy here is to create a pre-test state without having to depend on other functions.

You do make a good point on enforcing that `shares` is a non-whole Gwen amount, which we should do in this function.

ChaoticWalrus (changes requested):
Looks very good, a really appreciating these changes.
Dropped a bunch of nits / a few questions on specific lines.

src/test/tree/EigenPodManager.tree
ChaoticWalrus: cool, works for me, ty ty

src/test/unit/EigenPodManagerUnit.t.sol
ChaoticWalrus: hmm, fair enough.
I am a bit on-the-fence about whether or not we should do more redundant checks which potentially slow down tests but might eventually catch something. I'm fine with this in particular as it is.

src/test/unit/EigenPodManagerUnit.t.sol
ypatil12: Although now that I think about it, if we ever change `recordBeaconChainEThBalanceUpdate` logic, then the tests should be updated to reflect that. Will update this to not use std store - thanks for the suggestion.

ChaoticWalrus: cool, makes sense + works for me

src/test/unit/EigenPodManagerUnit.t.sol
ypatil12: Fixed

src/test/unit/EigenPodManagerUnit.t.sol
ypatil12: Planning to include that in the integration test

src/test/unit/EigenPodManagerUnit.t.sol
ypatil12: Hmm, good point. I think for more complex properties we should err on the side of over-test. For simple ones like this perhaps it is fine...

src/test/unit/EigenPodManagerUnit.t.sol
ChaoticWalrus: > Planning to include that in the integration test

makes sense. it would kind of be stretching what a unit test should cover.

src/test/tree/EigenPodManager.tree
8sunyuan, line 1: Can we rename this to EigenPodManagerUnit.tree? So we can distinguish between our unit and integration tests

src/test/utils/EigenLayerUnitTestSetup.sol
8sunyuan, line 1: Perhaps we should set this up similarly to the EigenLayerDeployer and EigenLayerTestHelper files where we deploy and initialize all of our mocked contracts. I'm not sure how useful some of the helper functions will be since the contracts will all be mocks but would be much cleaner in each of our `setUp()` definitions

ypatil12: As we notice that multiple of our unit test contracts initialize the same set of mock contracts we should add them to the EigenLayerUnitTestHelper file. For now, will keep it empty and add on as needed

src/test/EigenPod.t.sol
```diff
@@ -1885,4 +1885,192 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
     ) public view {
         BeaconChainProofs.verifyWithdrawal(beaconStateRoot, withdrawalFields, proofs);
     }
- }
\ No newline at end of file
+ }
+
+
+//TODO: Integration Tests from old EPM unit tests:
+// TODO: salvage / re-implement a check for reentrancy guard on functions, as possible
```
Sidu28, line 1892: General question, when we're commenting out these tests in several places, whats the general goal for the re-implementation?

ypatil12: Rewrite the tests for clarity per the EigenLayer testing guide

The tests above will be reimplemented once we begin writing integration tests

ypatil12: Okay, after discussion in standup going to move this to a combined EigenPodManager-Pod unit test file...

8sunyuan:
Could you also move `EigenPodHarness.sol` to the `test/harnesses` folder?

src/test/unit/EigenPodManagerUnit.t.sol
```diff
@@ -1,98 +1,67 @@
...
-    event BeaconChainETHWithdrawalCompleted(address indexed podOwner, uint256 shares, uint96 nonce, address delegatedAddress, address withdrawer, bytes32 withdrawalRoot);
-
-    // @notice Emitted when `podOwner` enters the "undelegation limbo" mode
-    event UndelegationLimboEntered(address indexed podOwner);
-
-    // @notice Emitted when `podOwner` exits the "undelegation limbo" mode
-    event UndelegationLimboExited(address indexed podOwner);
+    // Constants
+    uint256 public constant GWEI_TO_WEI = 1e9;
+    uint256 public constant REQUIRED_BALANCE_WEI = 31 ether;
```
ChaoticWalrus, line 42: pretty sure we can delete this?

ChaoticWalrus (approved):
LGTM.
Guess we'll see where the combined Pod-Manager test file goes. Sometimes things take a bit of work to shake out; I'm OK with this as it is right now.

src/test/unit/EigenPodManagerUnit.t.sol
ChaoticWalrus: comment was meant to be just about `REQUIRED_BALANCE_WEI`

PR #293, opened 2023-10-26
StrategyManager Unit Test Refactor

8sunyuan:
Refactoring tests for `StrategyManagerUnit.t.sol` following our [testing guide ](https://docs.google.com/document/d/107Wn4cVjSOTng1T72TWhhzswj1Afw4Dz7fUXQWqzLT0/edit?usp=sharing)

Changes include:
- Added tree file for code path coverage
- Added additional test coverage
- Deploying contracts through `EigenLayerUnitTestSetup.sol` and `EigenLayerUnitTestBase.sol`
- moving external functions into separate contracts within the test file
- renaming function tests to be more explicit between being fuzz vs unit
- assert statements as opposed to require
- absolute import paths


src/test/utils/EigenLayerUnitTestBase.sol
```diff
@@ -0,0 +1,39 @@
...
+        cheats.assume(!addressIsExcludedFromFuzzedInputs[fuzzedAddress]);
+        _;
+    }
+
+    function setUp() public virtual {
+        address[] memory pausers = new address[](1);
+        pausers[0] = pauser;
+        pauserRegistry = new PauserRegistry(pausers, unpauser);
+        eigenLayerProxyAdmin = new ProxyAdmin();
+        reenterer = new Reenterer();
```
ypatil12, line 33: Don't think reenter needs to be declared here since we use it in a minority of unit tests - only SM, DM, and EPM have a reentrancy guard. IMO this contract should only be declared in the sub-test contract that is testing reentrancy

src/test/utils/EigenLayerUnitTestSetup.sol
```diff
@@ -32,34 +24,15 @@ abstract contract EigenLayerUnitTestSetup is Test, Utils {
...
     uint256 public constant eigenTotalSupply = 1e50;
 
-    // Helper Functions/Modifiers
-    modifier filterFuzzedAddressInputs(address fuzzedAddress) {
-        cheats.assume(!addressIsExcludedFromFuzzedInputs[fuzzedAddress]);
-        _;
-    }
-
-    function setUp() public virtual {
+    function setUp() public virtual override {
```
ypatil12, line 30: This looks good - although I don't think the strategies need to be deployed here as those are just specific to unit tests for strategies and not other files

ypatil12 (commented):
Comment on Test Setup Files

src/test/utils/EigenLayerUnitTestBase.sol
8sunyuan: Removed and fixed

src/test/utils/EigenLayerUnitTestSetup.sol
8sunyuan: Removed as well !

src/test/utils/EigenLayerUnitTestSetup.sol
```diff
@@ -0,0 +1,46 @@
...
+
+abstract contract EigenLayerUnitTestSetup is EigenLayerUnitTestBase, Utils {
+    // Declare Mocks
+    StrategyManagerMock strategyManagerMock;
+    DelegationManagerMock public delegationManagerMock;
+    SlasherMock public slasherMock;
+    EigenPodManagerMock public eigenPodManagerMock;
+    StakeRegistryMock stakeRegistryMock;
+    // strategies
+    IERC20 public weth;
```
ypatil12, line 22: These strategies can be removed, as can the imports. Also it looks like the contract doesn't need to inherit from Utils anymore

src/test/utils/EigenLayerUnitTestSetup.sol
```diff
@@ -0,0 +1,46 @@
...
+    IERC20 public eigenToken;
+    StrategyBase public wethStrat;
+    StrategyBase public eigenStrat;
+    StrategyBase public baseStrategyImplementation;
+    uint256 wethInitialSupply = 10e50;
+    uint256 public constant eigenTotalSupply = 1e50;
+
+    function setUp() public virtual override {
```
ypatil12, line 30: Should the pauser registry also be initialized here?

src/test/utils/EigenLayerUnitTestSetup.sol
8sunyuan: Yup check my commit here https://github.com/Layr-Labs/eigenlayer-contracts/pull/293/commits/3a00d9367128a0bede8b648924f279b7523404f8
I removed the Utils and made the StrategyManagerUnit.t.sol inherit it as its the only contract using the `deployNewStrategy()` function

src/test/utils/EigenLayerUnitTestSetup.sol
8sunyuan: `EigenLayerUnitTestSetup` is inheriting `EigenLayerUnitTestBase` which is setting up the pauser registry

src/test/utils/EigenLayerUnitTestSetup.sol
8sunyuan: Actually just moved that function into `StrategyManagerUnit.t.sol` and deleted `utils/Utils.sol` altogether. Its not being used anywhere else

8sunyuan:
Not sure if why there are changes for `tree/EigenPodManager.tree` as it is up to date with master

ChaoticWalrus:
> Not sure if why there are changes for `tree/EigenPodManager.tree` as it is up to date with master

Hmm yes that's very odd, am assuming some GH error

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,89 @@
+StrategyManagerUnit.t.sol
+├── initialize
+│   └── when initialize called again
+│       └── it should revert
+├── depositIntoStrategy()
+│   ├── when deposit successfully
+│   │   └── shares should increase and stakerStrategyListLength should increase (deposit new strategy)
+│   ├── when deposit successfully twice
+│   │   └── shares should increase and stakerStrategyListLength should increase (deposit new strategy)
```
ChaoticWalrus, line 9: I think this should be combined into two sub-branches of 'when deposit successfully'
they are 'when the caller does not already have shares in the strategy' and 'when the caller already has shares in the strategy'
technically there is also an edge case of 'when the caller does not already have shares in the strategy' -- it is 'when the caller's strategy list is full (max length)'

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,89 @@
...
+│   ├── when safeTransferFrom() function reverts
+│   │   └── it should revert
+│   ├── when IERC20 token does not exist
+│   │   └── it should revert
+│   ├── when strategy.deposit() function reverts
+│   │   └── it should revert
+│   ├── when strategy does not exist
+│   │   └── it should revert
+│   ├── when strategy is not whitelisted
+│   │   └── it should revert
```
ChaoticWalrus, line 23: technically the whitelisting feature can be deactivated entirely, so this should probably be 'when the whitelist is active and the strategy is not whitelisted' ?

src/test/tree/StrategyManangerUnit.tree
ChaoticWalrus: ahh, the latter part is noted further down. I think that works ?

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,89 @@
...
+│   ├── when deposit with signature successfully
+│   │   └── staker nonce should increase and shares should increase
+│   ├── when deposit with signature successfully and replay same deposit signature
+│   │   └── it should revert the second time
+│   ├── when deposit with signature with EIP1271 compliant contract signature
+│   │   └── it should deposit successfully with shares and nonce increase
+│   ├── when deposit with signature with EIP1271 compliant contract but with bad signature
+│   │   └── it should revert
+│   ├── when deposit with signature with non-compliant EIP1271 contract
+│   │   └── it should revert
```
ChaoticWalrus, line 39: this feels a little like its describing tests as opposed to conditions?
there should probably be a branching point for 'given the staker is a contract' vs. 'given the staker is an EOA'

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,89 @@
...
+│   ├── when not called by DelegationManager
+│   │   └── it should revert
+│   ├── when share amount is too high, higher than deposited amount
+│   │   └── it should revert
+│   ├── when deposit single strategy and remove all shares
+│   │   └── it should remove all shares and decrease stakerStrategyListLength (set to 0)
+│   ├── when deposited 3 strategies and randomly select one to remove all shares
+│   │   └── it should remove all shares and decrease stakerStrategyListLength (set to 2)
+│   └── when deposited 3 strategies and removeShares() from all 3 strategies with fuzzed inputs
+│       └── for each strategy, decrease shares properly but if sharesAfter is 0, then should decrement stakerStrategyListLength
```
ChaoticWalrus, line 58: this is definitely describing tests rather than logical branches. the important branch is something like 'given the share amount is equal to the deposited amount'

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,89 @@
...
+├── addShares()
+│   ├── when not called by DelegationManager
+│   │   └── it should revert
+│   ├── when staker is 0 address
+│   │   └── it should revert
+│   ├── when adding shares back with 0 existing shares
+│   │   └── it should increase shares and append to stakerStrategyListLength
+│   └── when adding shares back with > 0 existing shares
+│       └── it should increase shares without increasing stakerStrategyListLength
```
ChaoticWalrus, line 67: remove the word 'back' here? feels unnecessary; from a technical perspective, the shares could technically be 'transferred' in which case the staker never had them before.

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,89 @@
...
+├── withdrawSharesAsTokens()
+│   ├── when not called by DelegationManager
+│   │   └── it should revert
+│   ├── when shares amount is too high, higher than deposited amount
+│   │   └── it should revert
+│   └── when deposit single strategy and withdraw shares <= deposited amount
+│       └── it should withdraw shares with token.balanceOf() increasing
```
ChaoticWalrus, line 74: the specific behavior on the level of this contract is calling `strategy.withdraw` ... not sure if it's too technical for this tree diagram; I think this wording works, it's just also approaching describing test behavior rather than 'spec' behavior

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,89 @@
...
+├── setStrategyWhitelister()
+│   ├── when called by owner address
+│   │   └── it should update strategyWhitelister address
+│   └── when not called by owner but by fuzzed address
+│       └── it should revert
+├── addStrategiesToDepositWhitelist()
+│   ├── when fuzz test number of strategies to add <= 16
+│   │   └── it should add all new strategies to deposit whitelist up to 16
```
ChaoticWalrus, line 82: should be something like
│   ├── when called by strategyWhitelister address
│   │   └── it should add all new strategies to the deposit whitelist
this wording is too specific to the (existing?) testing

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,89 @@
...
+│   └── when called by non-strategyWhitelister address
+│       └── it should revert
+└── removeStrategiesFromDepositWhitelist()
+    ├── when called by non-strategyWhitelister address
+    │   └── it should revert
+    └── when fuzz test numStrats to add <= 16 and numStrats to remove <= 16
+        └── it should first add all new strategies to deposit whitelist up to 16, then remove all strategies from deposit whitelist up to numStrats to add
```
ChaoticWalrus, line 89: see above comment, about `addStrategiesToDepositWhitelist`.
I think this should also maybe have a branch or note about what's supposed to happen when strategies are already in the whitelist?

src/test/tree/StrategyManangerUnit.tree
ChaoticWalrus: I think this should also maybe have a branch or note about what's supposed to happen when strategies are already in the whitelist?

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -163,20 +121,249 @@ contract StrategyManagerUnitTests is Test, Utils {
...
+                "stakerStrategyListLengthAfter != stakerStrategyListLengthBefore + 1"
+            );
+            require(
+                strategyManager.stakerStrategyList(staker, stakerStrategyListLengthAfter - 1) == strategy,
+                "strategyManager.stakerStrategyList(staker, stakerStrategyListLengthAfter - 1) != strategy"
+            );
+        }
+    }
+
+    function _setUpQueuedWithdrawalStructSingleStrat_MultipleStrategies(
```
ChaoticWalrus, line 217: I think this should be named something like `_setUpQueuedWithdrawalStruct_MultipleStrategies`, at least? feels like this name was from a copy-paste and has just ended up carried over

ChaoticWalrus: also might be an opportunity to delete this if we aren't using it?

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -240,7 +427,7 @@ contract StrategyManagerUnitTests is Test, Utils {
         strategyManager.depositIntoStrategy(dummyStrat, dummyToken, amount);
     }
 
-    function testDepositIntoStrategyRevertsWhenReentering() public {
+    function test_Revert_WhenReentering() public {
```
ChaoticWalrus, line 430: technically wouldn't we want an equivalent test for `depositIntoStrategyBySignature` as well?

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -532,187 +897,241 @@ contract StrategyManagerUnitTests is Test, Utils {
...
+                require(
+                    sharesAfter[i] == sharesBefore[i] - sharesAmounts[i],
+                    "sharesAfter != sharesBefore - sharesAmounts"
+                );
+            }
+        }
+        require(
+            stakerStrategyListLengthBefore - numPoppedStrategies == strategyManager.stakerStrategyListLength(staker),
+            "stakerStrategyListLengthBefore - numPoppedStrategies != strategyManager.stakerStrategyListLength(staker)"
+        );
```
ChaoticWalrus, line 1037: if we're bothering to check `_isDepositedStrategy` before, we should probably check this afterwords as well? i.e. verify that the correct strategy(ies) are removed, rather than just that the correct _number of strategies_ is removed

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -532,187 +897,241 @@ contract StrategyManagerUnitTests is Test, Utils {
...
-        cheats.expectEmit(true, true, true, true, address(strategyManager));
-        emit StrategyAddedToDepositWhitelist(dummyStrat);
-        strategyManager.addStrategiesToDepositWhitelist(_strategy);
-        cheats.stopPrank();
+contract StrategyManagerUnitTests_addShares is StrategyManagerUnitTests {
+    function test_Revert_DelegationManagerModifier() external {
+        DelegationManagerMock invalidDelegationManager = new DelegationManagerMock();
+        cheats.expectRevert(bytes("StrategyManager.onlyDelegationManager: not the DelegationManager"));
+        invalidDelegationManager.addShares(strategyManager, address(this), dummyStrat, 1);
+    }
```
ChaoticWalrus, line 1046: this test seems OK but it's structure is a bit unusual, in particular since it relies on the details of the 'DelegationManagerMock' contract. seems OK, mostly just pointing this out as it stood out to be as "different"

ChaoticWalrus (changes requested):
Changes seem good in general.
I think there's plenty more room here to continue modularizing + de-duping some operations, but I'm not about to block this PR based on thoughts that "more could be done".
Requesting changes primarily to the tree diagram; everything else seems pretty good although I think I've dropped a few nits here-and-there.

src/test/tree/StrategyManangerUnit.tree
8sunyuan: Active as in when deposits are not paused right? From the onlyWhenNotPaused(PAUSED_DEPOSITS) modifier. 
Then it could be 'when deposits are not paused and strategy is not whitelisted'

src/test/tree/StrategyManangerUnit.tree
8sunyuan: Right, the withdrawer address from a queuedWithdrawal could be a different address

src/test/tree/StrategyManangerUnit.tree
8sunyuan: You're right, I will rewrite the tree file according to all possible code paths instead and make appropriate changes. 
Will also include the more integration-like scenarios such as increaseDelegatedShares() involving the DelegationManager

wadealexc:
Looking clean :)

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -271,34 +458,212 @@ contract StrategyManagerUnitTests is Test, Utils {
...
+        strategyManager.addStrategiesToDepositWhitelist(_strategy);
+        cheats.stopPrank();
+
+        address staker = address(this);
+        IERC20 token = dummyToken;
+        uint256 amount = 1e18;
+        IStrategy strategy = dummyStrat;
+
+        cheats.startPrank(staker);
+        cheats.expectRevert();
```
stevennevins, line 530: Make explicit the revert reason

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -271,34 +458,212 @@ contract StrategyManagerUnitTests is Test, Utils {
...
+    function test_Revert_WhenStrategyDepositFunctionReverts() external {
+        // replace 'dummyStrat' with one that always reverts
+        dummyStrat = StrategyBase(address(new Reverter()));
+
+        // whitelist the strategy for deposit
+        cheats.startPrank(strategyManager.owner());
+        IStrategy[] memory _strategy = new IStrategy[](1);
+        _strategy[0] = dummyStrat;
+        cheats.expectEmit(true, true, true, true, address(strategyManager));
+        emit StrategyAddedToDepositWhitelist(dummyStrat);
```
stevennevins, line 520: It seems unnecessary to check the event

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -271,34 +458,212 @@ contract StrategyManagerUnitTests is Test, Utils {
...
+        strategyManager.addStrategiesToDepositWhitelist(_strategy);
+        cheats.stopPrank();
+
+        address staker = address(this);
+        IERC20 token = dummyToken;
+        uint256 amount = 1e18;
+        IStrategy strategy = dummyStrat;
+
+        cheats.startPrank(staker);
+        cheats.expectRevert();
```
stevennevins, line 554: explicit revert reason

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -271,34 +458,212 @@ contract StrategyManagerUnitTests is Test, Utils {
...
+
+    function test_Revert_WhenStrategyDoesNotExist() external {
+        // replace 'dummyStrat' with one that does not exist
+        dummyStrat = StrategyBase(address(5678));
+
+        // whitelist the strategy for deposit
+        cheats.startPrank(strategyManager.owner());
+        IStrategy[] memory _strategy = new IStrategy[](1);
+        _strategy[0] = dummyStrat;
+        cheats.expectEmit(true, true, true, true, address(strategyManager));
```
stevennevins, line 543: It seems unnecessary to check the event

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -271,34 +458,212 @@ contract StrategyManagerUnitTests is Test, Utils {
...
+        IERC20 token = dummyToken;
+        uint256 amount = 1e18;
+        IStrategy strategy = dummyStrat;
+
+        // uint256 MAX_STAKER_STRATEGY_LIST_LENGTH = strategyManager.MAX_STAKER_STRATEGY_LIST_LENGTH();
+        uint256 MAX_STAKER_STRATEGY_LIST_LENGTH = 32;
+
+        // loop that deploys a new strategy and deposits into it
+        for (uint256 i = 0; i < MAX_STAKER_STRATEGY_LIST_LENGTH; ++i) {
+            cheats.startPrank(staker);
```
stevennevins, line 612: In a lot of places, you can use cheats.prank(caller) since only the proceeding call requires to be pranked

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -532,187 +897,241 @@ contract StrategyManagerUnitTests is Test, Utils {
...
-            IStrategy[] memory _strategy = new IStrategy[](1);
-            _strategy[0] = dummyStrat;
-            cheats.expectEmit(true, true, true, true, address(strategyManager));
-            emit StrategyAddedToDepositWhitelist(dummyStrat);
-            strategyManager.addStrategiesToDepositWhitelist(_strategy);
-            cheats.stopPrank();
-        }
+        uint256 stakerStrategyListLengthBefore = strategyManager.stakerStrategyListLength(staker);
+        uint256 sharesBefore = strategyManager.stakerStrategyShares(staker, strategy);
+        require(sharesBefore == sharesAmount, "Staker has not deposited amount into strategy");
```
stevennevins, line 939: assertEq() instead of require.  The trace is nicer and outputs similar to below:
```
left : shows value
right : shows value
optional reason string
```

stevennevins: in the case the equality is violated*

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -1,69 +1,35 @@
...
+    IERC20 public dummyToken;
     StrategyBase public dummyStrat;
     StrategyBase public dummyStrat2;
     StrategyBase public dummyStrat3;
 
-    IStrategy public beaconChainETHStrategy;
-
-    IERC20 public dummyToken;
-
     Reenterer public reenterer;
```
ypatil12, line 28: Nit: to save deploy time, you could declare this and deploy in the test contracts that only test reentrancy instead of deploying it for every test contract

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -146,9 +104,9 @@ contract StrategyManagerUnitTests is Test, Utils {
...
         dummyToken = new ERC20Mock();
-        dummyStrat = deployNewStrategy(dummyToken, strategyManager, pauserRegistry, dummyAdmin);
-        dummyStrat2 = deployNewStrategy(dummyToken, strategyManager, pauserRegistry, dummyAdmin);
-        dummyStrat3 = deployNewStrategy(dummyToken, strategyManager, pauserRegistry, dummyAdmin);
+        dummyStrat = _deployNewStrategy(dummyToken, strategyManager, pauserRegistry, dummyAdmin);
+        dummyStrat2 = _deployNewStrategy(dummyToken, strategyManager, pauserRegistry, dummyAdmin);
+        dummyStrat3 = _deployNewStrategy(dummyToken, strategyManager, pauserRegistry, dummyAdmin);
 
         // whitelist the strategy for deposit
         cheats.startPrank(strategyManager.owner());
```
ypatil12, line 112: Nit: pluralize `_strategy` array. Also, we can use just cheats.prank since there's only one call that needs to be pranked

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -163,20 +121,249 @@ contract StrategyManagerUnitTests is Test, Utils {
...
-        addressIsExcludedFromFuzzedInputs[address(0)] = true;
-        addressIsExcludedFromFuzzedInputs[address(proxyAdmin)] = true;
-        addressIsExcludedFromFuzzedInputs[address(eigenPodManagerMock)] = true;
+        return strategyArray;
     }
+}
 
-    function testCannotReinitialize() public {
+contract StrategyManagerUnitTests_initialize is StrategyManagerUnitTests {
+    function test_CannotReinitialize() public {
```
ypatil12, line 359: Nit: test for verifying that storage was properly initialized, as too the pause, owner, and whitelister

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -271,34 +458,212 @@ contract StrategyManagerUnitTests is Test, Utils {
...
-        cheats.assume(expiry > block.timestamp);
+    function test_Revert_WhenTokenSafeTransferFromReverts() external {
+        // replace 'dummyStrat' with one that uses a reverting token
+        dummyToken = IERC20(address(new Reverter()));
+        dummyStrat = _deployNewStrategy(dummyToken, strategyManager, pauserRegistry, dummyAdmin);
 
-        address staker = cheats.addr(privateKey);
-        // not expecting a revert, so input an empty string
+        // whitelist the strategy for deposit
+        cheats.startPrank(strategyManager.owner());
```
ypatil12, line 467: nit: can just use cheats.prank(address) if there's one call to be pranked. Won't need cheats.stopPrank() if so

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -271,34 +458,212 @@ contract StrategyManagerUnitTests is Test, Utils {
...
+        strategyManager.addStrategiesToDepositWhitelist(_strategy);
+        cheats.stopPrank();
+
+        address staker = address(this);
+        IERC20 token = dummyToken;
+        uint256 amount = 1e18;
+        IStrategy strategy = dummyStrat;
+
+        cheats.startPrank(staker);
+        cheats.expectRevert();
```
ypatil12, line 481: Can we add the revert reasons to all tests

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -271,34 +458,212 @@ contract StrategyManagerUnitTests is Test, Utils {
...
+            cheats.stopPrank();
+        }
+
+        require(
+            strategyManager.stakerStrategyListLength(staker) == MAX_STAKER_STRATEGY_LIST_LENGTH,
+            "strategyManager.stakerStrategyListLength(staker) != MAX_STAKER_STRATEGY_LIST_LENGTH"
+        );
+
+        cheats.startPrank(staker);
+        cheats.expectRevert(bytes("StrategyManager._addShares: deposit would exceed MAX_STAKER_STRATEGY_LIST_LENGTH"));
```
ypatil12, line 635: We've done this throughout the repo, but you don't need to case the revert reason string to bytes. I'm fine either way, but just flagging

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -532,187 +897,241 @@ contract StrategyManagerUnitTests is Test, Utils {
...
 
-    function test_depositIntoStrategyRevertsWhenStrategyDoesNotExist() external {
-        // replace 'dummyStrat' with one that does not exist
-        dummyStrat = StrategyBase(address(5678));
+    function testFuzz_AppendsStakerStrategyList(address staker, uint256 amount) external {
+        cheats.assume(staker != address(0) && amount != 0);
+        uint256 stakerStrategyListLengthBefore = strategyManager.stakerStrategyListLength(staker);
+        uint256 sharesBefore = strategyManager.stakerStrategyShares(staker, dummyStrat);
+        require(sharesBefore == 0, "Staker has already deposited into this strategy");
+        require(!_isDepositedStrategy(staker, dummyStrat), "strategy shouldn't be deposited");
```
ypatil12, line 1064: Nit: possible to not use contractions so GitHub highlighting syntax doesn't break?

ypatil12 (commented):
Looks good, added a couple of nit/styling comments

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,89 @@
...
+│   ├── when not called by DelegationManager
+│   │   └── it should revert
+│   ├── when shares amount is too high, higher than deposited amount
+│   │   └── it should revert
+│   └── when deposit single strategy and withdraw shares <= deposited amount
+│       └── it should withdraw shares with token.balanceOf() increasing
+├── setStrategyWhitelister()
+│   ├── when called by owner address
+│   │   └── it should update strategyWhitelister address
+│   └── when not called by owner but by fuzzed address
```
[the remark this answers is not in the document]
8sunyuan: Done

src/test/tree/StrategyManangerUnit.tree
8sunyuan: Just simplified it to should withdraw tokens from strategy with token balanceOf() update. This path requires more integration implementation

src/test/tree/StrategyManangerUnit.tree
8sunyuan: Refactored this to describe logical branches now

src/test/tree/StrategyManangerUnit.tree
8sunyuan: Yup, added now that the event should not be emitted when strategies already whitelisted. Similarly for the removal of non-whitelisted strategies.

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: Within the for loop after each `removeShares()` call its checking that `_isDepositedStrategy` is false when the strategy is removed. But I moved this into a separate loop after calling `removeShares()` in all strategies https://github.com/Layr-Labs/eigenlayer-contracts/pull/293/commits/f011f90397319e8de8784535dc15b08fe943d165

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: Removed this and unused internal functions https://github.com/Layr-Labs/eigenlayer-contracts/pull/293/commits/69b0897bff77b3ad2b7df26162cc1bdef4f222ae

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: Changed requires into asserts https://github.com/Layr-Labs/eigenlayer-contracts/pull/293

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: I believe Reenterer is currently setup that way, just deployed per unit test as needed. Its not being deployed every `setUp()` call before each test

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: Noticed this too throughout tests and assumed its for being explicit. Can remove this though for reducing line count and being cleaner

8sunyuan: updated now in https://github.com/Layr-Labs/eigenlayer-contracts/pull/293/commits/d7c322b5a0bbd86e10c45377f3c89c0633c0d89e

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: added https://github.com/Layr-Labs/eigenlayer-contracts/pull/293/commits/df853c09d6e39c168b5efaf5ecd4dcea94dbd751

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: For reverts that give the error `Reason: Call reverted as expected, but without data` how do you write the `expectRevert()`? This is the case for the two test cases I didn't update

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: Added changes here but left in certain areas. See my commit message https://github.com/Layr-Labs/eigenlayer-contracts/commit/d7c322b5a0bbd86e10c45377f3c89c0633c0d89e

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: Removed https://github.com/Layr-Labs/eigenlayer-contracts/pull/293/commits/9355fb4a61033a3217420aed9539fd62229173e7

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: Removed https://github.com/Layr-Labs/eigenlayer-contracts/pull/293/commits/9355fb4a61033a3217420aed9539fd62229173e7

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,89 @@
...
+│   │   └── it should deposit successfully with shares and nonce increase
+│   ├── when deposit with signature with EIP1271 compliant contract but with bad signature
+│   │   └── it should revert
+│   ├── when deposit with signature with non-compliant EIP1271 contract
+│   │   └── it should revert
+│   ├── when deposits paused
+│   │   └── it should revert
+│   ├── when re-entering
+│   │   └── it should revert
+│   ├── when signature expired
```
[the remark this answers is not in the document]
8sunyuan: Personally prefer for #1 to simply have the external function name for better readability and makes more logical sense since we are trying to cover all the possible code paths from an external user. Will change #2 and #3 to be more consistent

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: pluralized

src/test/tree/StrategyManangerUnit.tree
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/293/commits/3b6a00fa9b6b51a6cf1705b5e0e2f6e0c4fd6581

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: I'm for it, removed now https://github.com/Layr-Labs/eigenlayer-contracts/pull/293/commits/45b1fcb80c90e9b979487411b3a4caac93373d50

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: Makes sense, updated tree file with changes to reflect the code paths as opposed to the testing coverage.
Will go back now to add missing tests and coverage

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: added https://github.com/Layr-Labs/eigenlayer-contracts/pull/293/commits/f69019cbc2231fe2635341ebcc15f9992ea305c2

8sunyuan:
Initially intended PR to just have some minor refactoring changes but a lot of people have given solid input on
improvements and missing coverage so I'll make sure this PR is comprehensive of changes to StrategyManagerUnit testing. Will add additional testing coverage for code paths not covered in the tree file as well.

src/test/tree/StrategyManangerUnit.tree
ChaoticWalrus: ah I remembered wrong, I thought there was literally a 'whitelist enabled/disabled' kind of switch -- guess we decided an upgrade would be the appropriate place to remove the whitelisting functionality, if it ever happens

src/test/unit/StrategyManagerUnit.t.sol
ChaoticWalrus: agreed, sometimes you have a revert without data and this is just the required syntax

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,107 @@
...
+│   │   └── it should revert
+│   ├── given that the strategy is not whitelisted
+│   │   └── it should revert
+│   ├── given the token safeTransferFrom() reverts
+│   │   └── it should revert
+│   └── given that token safeTransferFrom() succeeds
+│       ├── given the staker has existing shares in strategy (not first deposit)
+│       │   └── it should increase shares, nonce. while stakerStrategyListLength is unchanged
+│       ├── given the staker has no existing shares in strategy (first deposit)
+│       │   └── stakerStrategyListLength increases by 1 and shares, nonce increase
```
ChaoticWalrus, line 18: seems like there's errant notes about nonce increasing in calls to `depositIntoStrategy`  (it should only increase when calling `depositIntoStrategyWithSignature`, right?)

src/test/tree/StrategyManangerUnit.tree
```diff
@@ -0,0 +1,107 @@
...
+│       ├── given the staker has delegated to a operator ***
+│       │   └── it should deposit successfully with shares and nonce increase, including delegated shares
```
ChaoticWalrus, line 20: probably want to add the same note as in the EigenPodManager tree
`(*** denotes that integration tests are needed to validate the path)`

src/test/tree/StrategyManangerUnit.tree
8sunyuan: Good callout will remove

ChaoticWalrus (approved):

ChaoticWalrus:
> Initially intended PR to just have some minor refactoring changes but a lot of people have given solid input on improvements and missing coverage so I'll make sure this PR is comprehensive of changes to StrategyManagerUnit testing. Will add additional testing coverage for code paths not covered in the tree file as well. More integration based testing will follow after this

Personally am feeling OK with this PR as-is + would probably prefer a "follow-up" PR to add more coverage, but I can live with things either way (i.e. just adding coverage to this PR is also OK with me).

PR #294, opened 2023-10-26
Use stake registry stub in DelegationManager

ChaoticWalrus:
Enables removing the middleware interfaces from this repo.
This required a lot of test cleanup -- as far as I can tell, it doesn't diminish the tests whatsoever, just gets rid of a lot of garbage. Please do review though.


src/test/Withdrawals.t.sol
```diff
@@ -289,14 +208,12 @@ contract WithdrawalTests is EigenLayerTestHelper {
         cheats.warp(uint32(block.timestamp) + 2 days);
         cheats.roll(uint32(block.number) + 2);
 
-        uint256 prevElement = uint256(uint160(address(generalServiceManager2)));
-        generalReg1.propagateStakeUpdate(operator, uint32(block.number), prevElement);
+        // uint256 prevElement = uint256(uint160(address(generalServiceManager2)));
```
stevennevins, line 211: Can remove these comments and probably also the warp/rolls

stevennevins (approved):

src/test/Withdrawals.t.sol
ChaoticWalrus: agreed, but am feeling OK leaving this for later / whenever we rework this test. I think there's a lot more in this file that can get deleted.

PR #295, opened 2023-10-26
Removing understimation in Pods

Sidu28:
Removing understimation (hysterisis) in Pods


src/contracts/pods/EigenPod.sol
```diff
@@ -545,7 +545,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         // Done with proofs! Now update the validator's balance and send to the EigenPodManager if needed
 
         uint64 currentRestakedBalanceGwei = validatorInfo.restakedBalanceGwei;
-        uint64 newRestakedBalanceGwei = _calculateRestakedBalanceGwei(validatorBalance);
+        uint64 newRestakedBalanceGwei = validatorBalance;
```
gpsanant, line 548: should this now be capped to 32 ETH to not take yield into account?

src/contracts/pods/EigenPod.sol
```diff
@@ -483,7 +483,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         validatorInfo.status = VALIDATOR_STATUS.ACTIVE;
```
gpsanant, line 483: remove _RESTAKED_BALANCE_OFFSET_GWEI

docs/core/EigenPod.md
```diff
@@ -30,12 +30,6 @@ Since the execution layer cannot yet synchronously read arbitrary state from the
 
 This will either be implemented by Succinct Labs or eventually by Ethereum natively in EIP-4788 ([read here](https://eips.ethereum.org/EIPS/eip-4788)). The interface will have requests be for a block number and responses be the block roots corresponding to the provided block numbers.
 
-### Hysteresis
-
-We want to understimate validator balances on EigenLayer to be tolerant to small slashing events that occur on the beacon chain.
-
-We underestimate validator stake on EigenLayer through the following equation: $\text{eigenlayerBalance} = \text{min}(32, \text{floor}(x-C))$ where $C$ is some offset which we can assume is equal to 0.75. Since a validator's effective balance on the beacon chain can at most 0.25 ETH more than its actual balance on the beacon chain, we subtract 0.75 and floor it for simplicity.
-
```
ChaoticWalrus, line 38: Looks like the EPM doc also has a section about `_calculateRestakedBalanceGwei` that should be edited

ChaoticWalrus: and a mention of the function within the section about `EigenPod.verifyWithdrawalCredentials`

src/contracts/pods/EigenPod.sol
```diff
@@ -483,7 +475,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
         validatorInfo.validatorIndex = validatorIndex;
         validatorInfo.mostRecentBalanceUpdateTimestamp = oracleTimestamp;
-        validatorInfo.restakedBalanceGwei = _calculateRestakedBalanceGwei(validatorEffectiveBalanceGwei);
+        validatorInfo.restakedBalanceGwei = validatorEffectiveBalanceGwei;
```
ChaoticWalrus, line 478: if we're checking against `MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR` inside of the `_verifyBalanceUpdate` function, it feels like we should here as well.
Otherwise in the event of a hard-fork we will have some weird, inconsistent behavior

ChaoticWalrus (changes requested):
Only have a couple smaller comments, otherwise looks good.

src/contracts/pods/EigenPod.sol
Sidu28: my thought is that the effective balance is capped at 32ETH, and if that changes, lets say to 64ETH, that wouldn't be an issue since we'd make MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR = 64 ETH as well.

ChaoticWalrus: What is the sequencing you're imagining here?
It feels like we'd end up with inconsistent behavior unless we either (a) coordinated an upgrade _exactly at_ the hard fork or (b) coordinated an upgrade prior to the hard fork, which checked against block/slot number to know when to change the behavior.

If we are anticipating needing to upgrade for a hard fork, I'd prefer it to be possible to go through the hard fork and at least have consistent smart contract behavior prior to the upgrade.
I don't think this is a huge issue, but consistency feels safer here.
I'll ask @gpsanant to weigh in here as well

gpsanant: i agree with @ChaoticWalrus. i think we will need to go the (b) path in the future, lets keep the behavior consistent for now.

ChaoticWalrus:
Looks like there's some in-line comments that talk about us using an underestimation as well (around where we talk about hysteresis) -- would appreciate cleanup on those

docs/core/EigenPodManager.md
```diff
@@ -192,7 +184,7 @@ For each validator the Pod Owner wants to verify, the Pod Owner must supply:
     * `VALIDATOR_STATUS` moves from `INACTIVE` to `ACTIVE`
     * `validatorIndex` is recorded
     * `mostRecentBalanceUpdateTimestamp` is set to the `oracleTimestamp` used to fetch the beacon block root
-    * `restakedBalanceGwei` is set to `_calculateRestakedBalanceGwei(effectiveBalance)`
+    * `restakedBalanceGwei` is set to the validator's balance
```
ChaoticWalrus, line 187: in this particular instance we use the effective (rather than actual) balance, right?

src/contracts/pods/EigenPod.sol
```diff
@@ -483,7 +475,12 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         validatorInfo.status = VALIDATOR_STATUS.ACTIVE;
         validatorInfo.validatorIndex = validatorIndex;
         validatorInfo.mostRecentBalanceUpdateTimestamp = oracleTimestamp;
-        validatorInfo.restakedBalanceGwei = _calculateRestakedBalanceGwei(validatorEffectiveBalanceGwei);
+
+        if (validatorEffectiveBalanceGwei > MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR) {
```
ChaoticWalrus, line 479: this should be `>=`, not `>`

ChaoticWalrus: nvm, doesn't matter, deleting this comment

ChaoticWalrus (approved):
LGTM now. Thanks for responding to all my comments + nits @Sidu28 !

PR #298, opened 2023-10-26
chore: remove never-used storage at end of storage layout and increas…

ChaoticWalrus:
…e __gap size to compensate

This storage is at the end of the used slots in the storage layout, and was never used either on testnet or mainnet. Therefore, it should be able to be safety deleted without consequence. This commit also increases the size of the __gap variable to compensate for the removed storage.


Sidu28 (approved):
ah this is remnants of epm processed withdrawals.  RIP, LGTM

Issue #225, opened 2023-10-05
Request: Return the created EigenPod address from `createPod`


solimander:
This would be helpful when the creator is a contract as it would remove the need to call `ownerToPod` to fetch the pod address following creation.

PR #301, opened 2023-10-27
Return the created EigenPod address from createPod

Gajesh2007:
#225


Sidu28 (approved):
seems like a reasonable change

PR #307, opened 2023-10-30
EigenPod: Small Cleanup and Commenting

8sunyuan:
Reordered some of the constant indexes to be in-order in `BeaconChainProofs.sol`
for better readability while comparing with the consensus specs (example here https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#validator)

Also added some additional comments regarding the VERIFY_BALANCE_UPDATE_WINDOW_SECONDS checks


wadealexc (approved):
:+1:

Sidu28 (approved):
Awesome!! LGTM

PR #311, opened 2023-11-02
Update balance updates to use effective balances

ypatil12:
This draft refactor replaces balance update proofs with effective balance (or validator container) proofs. The end result is that validators can only prove balance updates on a per-epoch (6.4 minutes) level instead of on a per-slot basis.

Tests have not been refactored yet.


wadealexc:
Really nice! Seems like a good amount of code removed :) 

Do we have consensus on the downsides of the change?

ChaoticWalrus:
Seems like a good change, and it felt like people agreed it probably makes sense yesterday.
I like the consistency here, personally.
As I understand, the main downsides would be:
1) effective balance has both rounding and hysteresis applied (see [here](https://eth2book.info/capella/part2/incentives/balances/)), so it is simply "less accurate", which could lead to slightly more "overestimation" of validators' ETH by EigenLayer, in various (edge) cases
2) effective balance is updated once/epoch (~6min) instead of every slot (12 sec), so proofs would be against "staler" state, and there are perhaps edge cases where it would take EigenLayer longer to be able to get a balance update that a validator has been slashed.

Given the current oracle model, I think (2) is pretty much a non-issue, since it roughly just extends the update window by ~6min or less.
The proposal would be basically to make (1) well-known / well-understood. Over-estimation by EigenLayer is already possible due to the async nature of needing to prove consensus layer state (pull model, not push model), and practically, I doubt this change would significantly increase over-estimation by EigenLayer. Crucially, when proving withdrawals, we still get the actual amount, so users still should not ever be able to "over-withdraw".

Upsides I'm seeing:
- cheaper balance updates (since one less proof is being performed)
- code simplification
- consistency (we no longer use effective balance when proving withdrawal credentials vs. actual balance when proving a balance update, it's just effective balance for both)

In the balance, I think I'm in favor of this change.
@Sidu28 please point out if I've missed any key point(s) here?

src/contracts/pods/EigenPod.sol
```diff
@@ -213,7 +213,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
                 oracleTimestamp,
                 validatorIndices[i],
                 stateRootProof.beaconStateRoot,
-                balanceUpdateProofs[i],
```
Sidu28, line 216: can we add a comment here just explaining why we are using the fields proof here?  Just saying that the validator fields contains the effective balance

src/contracts/pods/EigenPod.sol
```diff
@@ -534,26 +533,18 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
             validatorIndex: validatorIndex
         });
 
         // Done with proofs! Now update the validator's balance and send to the EigenPodManager if needed
 
         uint64 currentRestakedBalanceGwei = validatorInfo.restakedBalanceGwei;
-         uint64 newRestakedBalanceGwei;
-        if (validatorBalance > MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR) {
+        uint64 newRestakedBalanceGwei;
+        if (validatorEffectiveBalanceGwei > MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR) {
```
Sidu28, line 544: we're setting MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR to 32 ETH now and effective balance is also capped at 32 ETH, so we can remove the if-else logic and just set newRestakedBalanceGwei = validatorEffectiveBalanceGwei

Sidu28:
> Seems like a good change, and it felt like people agreed it probably makes sense yesterday. I like the consistency here, personally. As I understand, the main downsides would be:
> 
> 1. effective balance has both rounding and hysteresis applied (see [here](https://eth2book.info/capella/part2/incentives/balances/)), so it is simply "less accurate", which could lead to slightly more "overestimation" of validators' ETH by EigenLayer, in various (edge) cases
> 2. effective balance is updated once/epoch (~6min) instead of every slot (12 sec), so proofs would be against "staler" state, and there are perhaps edge cases where it would take EigenLayer longer to be able to get a balance update that a validator has been slashed.
> 
> Given the current oracle model, I think (2) is pretty much a non-issue, since it roughly just extends the update window by ~6min or less. The proposal would be basically to make (1) well-known / well-understood. Over-estimation by EigenLayer is already possible due to the async nature of needing to prove consensus layer state (pull model, not push model), and practically, I doubt this change would significantly increase over-estimation by EigenLayer. Crucially, when proving withdrawals, we still get the actual amount, so users still should not ever be able to "over-withdraw".
> 
> Upsides I'm seeing:
> 
> * cheaper balance updates (since one less proof is being performed)
> * code simplification
> * consistency (we no longer use effective balance when proving withdrawal credentials vs. actual balance when proving a balance update, it's just effective balance for both)
> 
> In the balance, I think I'm in favor of this change. @Sidu28 please point out if I've missed any key point(s) here?

Yeah this captures it pretty much.  The only real downside is that:

- balance updates have to wait at most 6 minutes (but like you said, non issue)
- The balances in EL are can be underestimated by as much as 0.75 ETH

src/contracts/pods/EigenPod.sol
ypatil12: Is there any edge case if ETH increases the effective balance cap from 32 ETH? Should we keep this logic or would we adequately assess impacts/upgrade well in advance?

Sidu28: Perhaps, though at that point I'd imagine EL does the same, if the caps are raised to 64, for .  I think we should remove the if/else, makes it cleaner and more readable

ChaoticWalrus: Hmm...so we still need the `MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR` on withdrawals, and would be using it there.
So an upgrade would still be necessary in the event of a hard fork.

Definitely feeling like we shouldn't rush to just remove this -- we had a similar discussion in another PR, but I would rather have consistent behavior across EigenLayer and then change behavior consistently across-the-board in EigenLayer with an upgrade following/coordinated with any hard fork, rather than have a hard fork switch ~half of EigenLayer's behavior -- simply removing this feels like a recipe for weird behavior to happen in the event of a hard fork.

Sidu28: ah yeah good point, gives us some breathing room to do the upgrade.  Lets keep it for now then

ChaoticWalrus (approved):
PR is looking good to me.
We will need to coordinate a bit on associated UI + API changes, and possibly communicate about the change in UX (although this seems pretty small, overall).

Sidu28 (approved):

PR #316, opened 2023-11-06
eigenpod tests

Gajesh2007:


src/test/unit/EigenPodUnit.t.sol
```diff
@@ -532,4 +541,11 @@ contract EigenPodUnitTests is Test, ProofParsing {
         eigenPodManager.createPod();
         cheats.stopPrank();
     }
+
+    function _getLatestDelayedWithdrawalAmount(address recipient) internal view returns (uint256) {
```
ypatil12, line 545: Can you move this helper function in EigenPod.t.sol since no tests in this file use `_getLatestDelayedWithdrawalAmount`

src/test/unit/EigenPodUnit.t.sol
```diff
@@ -157,6 +157,15 @@ contract EigenPodUnitTests is Test, ProofParsing {
         fuzzedAddressMapping[address(0x0)] = true;
     }
 
+    function testDeployPodWithStateChecks() external {
+        cheats.startPrank(address(42000094992494));
```
ypatil12, line 161: Aren't we already testing this in `testCheckThatHasRestakedIsSetToTrue` ?

ypatil12 (commented):
Comments in unit test files

src/test/EigenPod.t.sol
```diff
@@ -290,6 +290,37 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
         );
     }
 
+    function testWithdrawBeforeRestakingWithTimestampChecks() public {
```
Sidu28, line 293: seems to me that this can be combines with the `testWithdrawBeforeRestaking` test

src/test/EigenPod.t.sol
```diff
@@ -366,6 +397,28 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
         cheats.stopPrank();
     }
 
+    function testWithdrawNonBeaconChainETHBalanceWei() public {
+        IEigenPod pod = testDeployAndVerifyNewEigenPod();
+        IEigenPod _pod = eigenPodManager.getPod(podOwner);
+
+        require(_pod == pod, "add check");
```
Sidu28, line 404: don't think this check is strictly necessary, since testDeployAndVerifyNewEigenPod() is returning the value of getPod() already.

src/test/EigenPod.t.sol
```diff
@@ -366,6 +397,28 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
...
+        cheats.deal(address(podOwner), 10 ether);
+        emit log_named_address("opd", address(pod));
+
+        (bool sent, ) = payable(address(pod)).call{value: 1 ether}("");
+
+        require(sent == true, "not sent");
+
+        cheats.startPrank(podOwner, podOwner);
+        pod.withdrawNonBeaconChainETHBalanceWei(
+            podOwner,
```
Sidu28, line 415: Also could add a check that the event is being emitted in `withdrawNonBeaconChainETHBalanceWei`

src/test/unit/EigenPodUnit.t.sol
Sidu28: This is checking it on a fresh deployed pod? Equivalent but this feels more contained as a unit test.

Gajesh2007: ohh. just found it's already implemented in `testDeployedPodIsRestaked`

Sidu28: Ah ok cool :)

src/test/EigenPod.t.sol
Sidu28: what did you decide on doing here?

Sidu28 (changes requested):
One last comment, otherwise LGTM

Sidu28:
@Gajesh2007 no no I didn't want you to remove the test...... The initial comment was to combine the tests, the checks you added were good.  So you have two options:

- So either bring over everything from the old test that your new test doesn't have and delete the old test
- Add the new checks you make in your new test to the old test and delete the new test.

Sidu28 (approved):

PR #318, opened 2023-11-06
update slasher poc

gpsanant:
Implement the slasher with 2 functions for the UI

### DIFF
--- a/src/contracts/core/Slasher.sol
+++ b/src/contracts/core/Slasher.sol
@@ -1,75 +1,18 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity =0.8.12;
 
-import "../interfaces/ISlasher.sol";
-import "../interfaces/IDelegationManager.sol";
-import "../interfaces/IStrategyManager.sol";
-import "../libraries/StructuredLinkedList.sol";
-import "../permissions/Pausable.sol";
-import "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
-import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
+import "../libraries/EIP1271SignatureUtils.sol";
+import "./SlasherStorage.sol";
 
 /**
  * @title The primary 'slashing' contract for EigenLayer.
  * @author Layr Labs, Inc.
  * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
- * @notice This contract specifies details on slashing. The functionalities are:
- * - adding contracts who have permission to perform slashing,
- * - revoking permission for slashing from specified contracts,
- * - tracking historic stake updates to ensure that withdrawals can only be completed once no middlewares have slashing rights
- * over the funds being withdrawn
+ * @notice This contract allows AVSs and operators to emit events to be viewed by offchain services before slashing is enabled.
  */
-contract Slasher is Initializable, OwnableUpgradeable, ISlasher, Pausable {
-    using StructuredLinkedList for StructuredLinkedList.List;
+contract Slasher is SlasherStorage {
 
-    uint256 private constant HEAD = 0;
-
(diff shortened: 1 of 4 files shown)


src/contracts/core/Slasher.sol
```diff
@@ -83,534 +26,45 @@ contract Slasher is Initializable, OwnableUpgradeable, ISlasher, Pausable {
...
-    function whitelistedContractDetails(
-        address operator,
-        address serviceContract
-    ) external view returns (MiddlewareDetails memory) {
-        return _whitelistedContractDetails[operator][serviceContract];
+    function registerOperatorWithoutSlashingWithAVS(address operator, uint256 expiry, bytes memory signature) external {
+        require(expiry > block.timestamp, "Slasher.registerOperatorWithoutSlashingWithAVS: signature expired");
+        require(!registeredWithoutSlashing[msg.sender][operator], "Slasher.registerOperatorWithoutSlashingWithAVS: operator already registered without slashing");
+        // check operator signature
+        EIP1271SignatureUtils.checkSignature_EIP1271(operator, calculateRegistrationWithoutSlashingDigestHash(operator, msg.sender, nonces[msg.sender]++, expiry), signature);
```
ChaoticWalrus, line 38: I think it would make more sense to use a 'salt' here instead of an incrementing nonce

src/contracts/core/Slasher.sol
```diff
@@ -83,534 +26,45 @@ contract Slasher is Initializable, OwnableUpgradeable, ISlasher, Pausable {
...
+     * @param expiry The desired expiry time of the operator's signature
+     * @param signature The signature of the operator
      */
-    function whitelistedContractDetails(
-        address operator,
-        address serviceContract
-    ) external view returns (MiddlewareDetails memory) {
-        return _whitelistedContractDetails[operator][serviceContract];
+    function registerOperatorWithoutSlashingWithAVS(address operator, uint256 expiry, bytes memory signature) external {
+        require(expiry > block.timestamp, "Slasher.registerOperatorWithoutSlashingWithAVS: signature expired");
```
wadealexc, line 35: Would prefer `registerOperatorNoSlashing` and `deregisterOperatorNoSlashing`

wadealexc: Oh and `calculateRegistrationNoSlashingDigest`

gpsanant: got it

src/contracts/core/Slasher.sol
gpsanant: salts feel cleaner fs

src/contracts/core/Slasher.sol
gpsanant: what do you think the storage should be called?

wadealexc: `registeredNoSlashing`, I suppose?

ChaoticWalrus: what if we just call it `registered` (or similar) and slashing-related things will live in slashing-specific storage?
I realize this is kind of getting into the future state model of slashing, but am imagining something like:
```
mapping(operator => AVS => bool) registeredForAVS
mapping(operator => AVS => slashingInfo) slashingPermissions
```
and that we limit transitions in a way where, e.g. `slashingPermissions[operator][AVS]` can only transition to non-empty if `registeredForAVS[operator][AVS] == true`

or we can make a storage struct right now, and combine the mappings? something like:
```
mapping(operator => AVS => RegistrationInfo) registrationInfo
struct RegistrationInfo {
bool operatorIsRegistered
slashingInfo (unspecified)
}
```

src/contracts/core/Slasher.sol
ChaoticWalrus: > sure, what's your deciding criteria though?

my thought process is simply "do these things need to be in order? (if yes, then use a nonce), or will enforcing order lead to undesired race conditions? (if yes, then use a salt)"

PR #320, opened 2023-11-07
chore: add correct M2 Goerli deployment info

ChaoticWalrus:


wadealexc (changes requested):
LGTM pending README update :)

ChaoticWalrus:
> Could you update the addresses in the README?

Feels like maybe we shouldn't update those until we actually upgrade the contracts?
Right now we just have the implementation contracts deployed

wadealexc:
> > Could you update the addresses in the README?
> 
> Feels like maybe we shouldn't update those until we actually upgrade the contracts? Right now we just have the implementation contracts deployed

Oh! That's fair, nevermind then :)

wadealexc (approved):

PR #323, opened 2023-11-08
DO NOT MERGE - feat: BROKEN implementation of prioritizing paying down "share debt" …

ChaoticWalrus:
…with partial withdrawals

note that this change introduces a path by which a user can unfairly increase their beacon chain ETH shares:
- queue withdrawal (of "beacon chain ETH shares") in EigenLayer
- get slashed on beacon chain
- prove balance update (decrease) to EigenLayer, end up with negative beacon chain ETH shares
- prove some partial withdrawals (could be old or new), which would then "pay off" the negative beacon chain ETH share balance
- complete the withdrawal "as shares", getting credited with the full, original amount of "beacon chain ETH shares"

end result is that the user has more "beacon chain ETH shares" than the combined balances of their validators


wadealexc:
thanks for putting this together. im not sold on its inclusion in mainnet if only because i think what we currently have works well and i dont want the headache of figuring out this edge case :smile:

ChaoticWalrus:
Closing out this PR as it seems we're agreed that it's at least not worth investing in trying to make this work right now.
Might revisit the idea in the future; we'll see.

PR #325, opened 2023-11-10
Succinct Integration

Sidu28:


src/contracts/interfaces/IEigenPod.sol
```diff
@@ -49,13 +49,25 @@ interface IEigenPod {
...
+    }
 
     enum PARTIAL_WITHDRAWAL_CLAIM_STATUS {
         REDEEMED,
         PENDING,
         FAILED
     }
 
+     enum REQUEST_STATUS {
+        PENDING, // request is pending
```
ChaoticWalrus, line 66: default status probably shouldn't be 'PENDING'

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -94,9 +106,21 @@ interface IEigenPod {
     /// @notice Emitted when ETH that was previously received via the `receive` fallback is withdrawn
     event NonBeaconChainETHWithdrawn(address indexed recipient, uint256 amountWithdrawn);
 
+    /// @notice Emitted when partial withdrawal proof is fulfilled
+    event PartialWithdrawalProven(uint256 nonce, uint256 partialWithdrawalSum);
```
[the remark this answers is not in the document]
ChaoticWalrus: or perhaps `PartialWithdrawalProofConfirmed` or similar, to parallel the naming of the other related events

src/contracts/pods/EigenPod.sol
```diff
@@ -92,7 +96,18 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     uint256 public nonBeaconChainETHBalanceWei;
 
      /// @notice This variable tracks the total amount of partial withdrawals claimed via merkle proofs prior to a switch to ZK proofs for claiming partial withdrawals
-    uint64 public sumOfPartialWithdrawalsClaimedGwei;
+    uint256 public sumOfPartialWithdrawalsClaimedWei;
```
ChaoticWalrus, line 99: noting that this change is **totally upgrade incompatible**.
might end up being OK depending on rollout plans though :/

src/contracts/pods/EigenPod.sol
```diff
@@ -306,6 +327,8 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         // ensure that caller has previously enabled restaking by calling `activateRestaking()`
         hasEnabledRestaking
     {
+        // initializes the  "timestampProvenUntil" for batched zk partial withdrawals to the GENESIS_TIME
+        timestampProvenUntil = GENESIS_TIME;
```
ChaoticWalrus, line 331: why is this inside the `verifyWithdrawalCredentials` function?
seems totally broken
probably we should just  use some sane `startTimestamp` as the default, if `timestampProvenUntil == 0` -- it could be set in the EigenPod constructor, and equal to the timestamp of the first EigenPod deployment, perhaps?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +451,96 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+            recipient: recipient,
+            status: REQUEST_STATUS.PENDING
+       });
+
+
+
+        emit PartialWithdrawalProofRequested(startTimestamp, endTimestamp, requestNonce);
+
+        //NOTE: it is not clear what callback data needs to passed here since succinct is going to make multiple callbacks for each request.
+        //These are the inputs to the callback function for this specific proof request.
```
ChaoticWalrus, line 479: is the splitting based upon amount of proofs?
seems like a notable design problem then, at minimum.
would be easy if the splitting was just based on a time/slot length

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +451,96 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+            * zk proof is fulfilled, we will subtract sumOfPartialWithdrawalsClaimedWei from the amount proven by the oracle
+            */
+            sumOfPartialWithdrawalsClaimedWei = 0;
+        }
+
+        //mark the partial withdrawal proof as being fulfilled
+        if(endSlot == request.endTimestamp){
+            _partialWithdrawalProofRequests[nonce].status = REQUEST_STATUS.FULFILLED;
+            lastRequestNonceProven = nonce;
+        }
```
ChaoticWalrus, line 530: ahh, so this is the current approach you have to batching?
I see what you were talking about now -- if batched callbacks aren't in order, then it's a major issue as partial withdrawals will get stuck!

src/contracts/pods/EigenPod.sol
```diff
@@ -792,5 +928,5 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[44] private __gap;
+    uint256[41] private __gap;
```
ChaoticWalrus, line 931: rn I think this size is off by 1
I count 4 new slots used with the gap decreasing in size by 3

src/contracts/pods/EigenPodManager.sol
```diff
@@ -314,12 +330,38 @@ contract EigenPodManager is
...
+
+    //make a request for a proof to the succinct gateway
+    function requestProofViaSuccinctGateway(
+        bytes32 FUNCTION_ID,
+        uint64 startSlot,
+        uint64 endSlot,
+        address podAddress,
+        bytes memory callbackData,
+        uint32 callbackGasLimit
+    ) external payable {
```
ChaoticWalrus, line 350: hmm, probably want this access-restricted to EigenPods?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +451,96 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+
+
+        eigenPodManager.requestProofViaSuccinctGateway(
+            RANGE_SPLITTER_FUNCTION_ID, 
+            _timestampToSlot(timestampProvenUntil),
+            _timestampToSlot(endTimestamp),
+            address(this),
+            callBackData,
+            callbackGasLimit
+        );
```
ChaoticWalrus, line 490: we need to forward msg.value in this call, right?

src/contracts/pods/EigenPod.sol
Sidu28: Several things I was considering here:
- In terms of setting the startTimestamp value, ideally it could be pod creation timestamp but that varies across all pods.  Is it feasible for us to run a script that calls this function on every pod? I was under the assumption that it probably isn't generally speaking (tho we do have 300 or so pods so may not be actually that bad).  Thats also the reason I set it in the verifyWithdrawalCredentials, is because none of the pods have called that function yet on mainnet.   Again the alternative is calling a setter fucntion on every pod.  Maybe that is actually viable

src/contracts/interfaces/IEigenPod.sol
Sidu28: ah good point

src/contracts/interfaces/IEigenPod.sol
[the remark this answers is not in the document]
Sidu28: PartialWithdrawalProofConfirmed is good

src/contracts/pods/EigenPod.sol
Sidu28: oh shoot ur right we already had this in the contract, fuck I forgot

src/contracts/pods/EigenPodManager.sol
Sidu28: yeah good call

src/contracts/pods/EigenPod.sol
Sidu28: fixed!

src/contracts/pods/EigenPod.sol
Sidu28: ah I forgot we upgraded the contract rip.  We didn;t on mainnet tho right?

Sidu28: ok just checked, we haven't on mainnet

src/contracts/pods/EigenPod.sol
Sidu28: exactly!  If we had separate nonces for each split, we could enforce they are made in order in the handleCallback function. This would be way better imo

src/contracts/pods/EigenPod.sol
ChaoticWalrus: 100% can't leave it as-is, since it'll keep resetting the value.
I think what I suggested makes more sense, but I'm sure there's other viable options.
In general, I'm not a fan of having to call something for all existing pods, especially if there's a risk of weird ordering problems (e.g. if it's an issue if we *don't* call some special function on some pods, prior to users making a proof request), since that feels like it could necessitate a sequencing of (1) upgrade, (2) call special function, (3) upgrade again -- that's a lot of operational work and review to make sure everything is correct!

Sidu28: Yeah this was an attempt to avoid anything involving calling all existing pods.  And yeah blanked on the fact that there are  multiple validators per pod, smh.  What if we did something stupid like this:

if timestampProvenUntil == 0:
        timestampProvenUntil = GENESIS_TIME;

podOwners are forced to call verifyWithdrawalCredentials in order to set themselves as active.  We just need to make sure that any function that timestampProvenUntil is set in, can never be called before verifyWithdrawalCredentials, which seems reasomable

Sidu28: Actually, we should just add this to the requestProof function, not verify withdrawal credentials lol

src/contracts/pods/EigenPod.sol
Sidu28: ah fixed, didn't update it after I added  lastRequestNonceProven;

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -49,13 +49,26 @@ interface IEigenPod {
         int256 sharesDeltaGwei;
     }
 
+    struct PartialWithdrawalProofRequest{
+        uint64 startTimestamp;
```
puma314, line 53: can we make these slots, not timestamps?

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -94,9 +107,21 @@ interface IEigenPod {
     /// @notice Emitted when ETH that was previously received via the `receive` fallback is withdrawn
     event NonBeaconChainETHWithdrawn(address indexed recipient, uint256 amountWithdrawn);
 
+    /// @notice Emitted when partial withdrawal proof is fulfilled
+    event PartialWithdrawalBatchProven(uint256 nonce, uint256 partialWithdrawalSum);
+
+    /// @notice Emitted when a partial withdrawal proof is requested from succinct
+    event PartialWithdrawalProofRequested(uint64 startTimestamp, uint64 endTimestamp, uint256 requestNonce);
```
puma314, line 114: again, possible to make these slots?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +450,104 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        require(endTimestamp > timestampProvenUntil, "EigenPod.submitPartialWithdrawalsBatchForVerification: endTimestamp must be greater than timestampProvenUntil");
+        
+        //the proofs are made [startTimestamp, endTimestamp), so the next startTimestamp = previous endTimestamp     
+        uint64 startTimestamp = timestampProvenUntil;
+        
+        //record the partial withdrawal proof request
+       _partialWithdrawalProofRequests[requestNonce] = PartialWithdrawalProofRequest({
+            startTimestamp: startTimestamp,
+            endTimestamp: endTimestamp,
+            recipient: recipient,
```
puma314, line 474: Unless I'm misunderstanding something, shouldn't `recipient` be `address(this)`? I'm pretty sure I've gotten something wrong lol

src/contracts/pods/EigenPodManager.sol
```diff
@@ -314,12 +330,39 @@ contract EigenPodManager is
...
         bytes32 stateRoot = beaconChainOracle.timestampToBlockRoot(timestamp);
         require(
             stateRoot != bytes32(0),
             "EigenPodManager.getBlockRootAtTimestamp: state root at timestamp not yet finalized"
         );
         return stateRoot;
     }
+
+    //make a request for a proof to the succinct gateway
+    function requestProofViaSuccinctGateway(
```
puma314, line 343: I'm confused--is this the entrypoint for the user requesting a withdrawal? If that's the case, then they shouldn't be able to pass in a FUNCTION_ID, it should be stored and we should just use that stored version (also the stored FUNCTION_ID should have upgradeability similar to the gateway)

src/contracts/pods/EigenPod.sol
```diff
@@ -65,6 +66,9 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     /// @notice This is the genesis time of the beacon state, to help us calculate conversions between slot and timestamp
     uint64 public immutable GENESIS_TIME;
 
+    bytes32 public immutable RANGE_SPLITTER_FUNCTION_ID = sha256(bytes("RANGE_SPLITTER_FUNCTION_ID"));
```
puma314, line 69: Similar to the Succinct Gateway address, I think it's best if these  can be changed by `onlyOwner`. We do have a way for people to update the function_id through our gateway, but it might be cleaner for you guys if it's just all under the same system in the same file.

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +450,104 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        PartialWithdrawalProofRequest memory request = _partialWithdrawalProofRequests[nonce];
+
+        require(request.status == REQUEST_STATUS.PENDING, "EigenPod.handleCallback: request nonce is either cancelled or fulfilled");
+        require(nonce == lastRequestNonceProven + 1, "EigenPod.handleCallback: must callback in order");
+        require(startSlot < endSlot, "EigenPod.handleCallback: invalid start and end slot values");
+        require(_slotToTimestamp(endSlot) <= request.endTimestamp, "EigenPod.handleCallback: endSlot must be less than the request's endTimestamp");
+        require(_slotToTimestamp(startSlot) == timestampProvenUntil, "EigenPod.handleCallback: startSlot must be equal to the timestampProvenUntil");
+
+        bytes32 beaconBlockRoot = eigenPodManager.getBlockRootAtTimestamp(oracleTimestamp);
+        bytes memory output = eigenPodManager.confirmProofVerification(WITHDRAWAL_FUNCTION_ID, abi.encodePacked(beaconBlockRoot, address(this), startSlot, endSlot));
```
puma314, line 509: why have this level of indirection through the eigenpod manager instead of directly calling the succinct gateway here?

PR #326, opened 2023-11-13
chore: fix broken links and clean up readme

wadealexc:


README.md
```diff
@@ -4,26 +4,18 @@
...
+EigenLayer is a set of smart contracts deployed on Ethereum that enable restaking of assets to secure new services.
 
-Click the links in the Table of Contents below to access more specific documentation. We recommend starting with the [EigenLayer Technical Specification](docs/EigenLayer-tech-spec.md) to get a better overview before diving into any of the other docs.  For contracts addresses deployed on Goerli, click [here](https://github.com/Layr-Labs/eigenlayer-contracts/blob/master/script/output/M1_deployment_goerli_2023_3_23.json).
-## Table of Contents  
-* [Introduction](#introduction)
-* [Installation and Running Tests / Analyzers](#installation)
-* [EigenLayer Technical Specification](docs/EigenLayer-tech-spec.md)
+We recommend starting with the [technical documentation](docs/README.md) to get an overview of the contracts before diving into the code.
```
samlaf, line 10: you prefer not to point to the EIgenLayer-tech-spec in docs/outdated?
Do we have plans to rewrite that at some point?

README.md
```diff
@@ -4,26 +4,18 @@
...
-Design Docs
-* [Withdrawals Design Doc](docs/Guaranteed-stake-updates.md)
-* [EigenPods Design Doc](docs/EigenPods.md)
+## Table of Contents  
 
-Flow Docs
-* [EigenLayer Withdrawal Flow](docs/EigenLayer-withdrawal-flow.md)
-* [EigenLayer Deposit Flow](docs/EigenLayer-deposit-flow.md)
-* [EigenLayer Delegation Flow](docs/EigenLayer-delegation-flow.md)
-* [Middleware Registration Flow for Operators](docs/Middleware-registration-operator-flow.md)
```
samlaf, line 26: why are we getting rid of all of these? Or we just let them navigate the docs/ folder by themselves?

samlaf (approved):
Thanks for this. LGTM, but should get @ChaoticWalrus to also review. 
Added two small questions, but shouldn't block merging.

README.md
wadealexc: Looking at that file, it's very outdated, so I definitely don't want to point to it for now.

A lot of the content there is now contained in the docs/README, or the top of each docs file.

I think the main content there that it might be nice to have is some of the conceptual concepts presented, but I don't currently have time to bring it up to date and I think the current docs do a good enough job otherwise.

README.md
wadealexc: None of those links function anymore. 

The README file in the docs folder serves as a good table of contents into the rest of the docs, but the specific files I'm removing here are really out of date. I think the material should be covered by the new docs - do you see gaps anywhere?

PR #333, opened 2023-11-14
Integration testing ideas

ChaoticWalrus:
Marking this as a draft initially.
Combines some cleanup of low-level deployment + helper functionality, and some ~scaffolding of ideas around abstract-ifying users.

Adding some thoughts after making a few more commits:

- I feel like it would be ~quite annoying to set up, and create perhaps a bit of a learning curve for using/writing tests, _but_ am feeling like this level of abstraction is closer to the “right” level, and that it would ultimately help to make our tests more modular + clearer, with more consistent (if redundant) checks on system properties + behavior.
- The present implementation stores everything in memory, which is just...super slow. We probably want to switch to using storage instead if we decide to go this route.
- This is also using a kind of awkward syntax at the moment -- it would be nice if we could support syntax like `Staker.updateState()` or `Staker.deposit(strategies, shares, withdrawer)` instead -- perhaps we should make 'Staker' a contract type instead of just a struct.
- my intuition is that this level of abstraction does not work the best when we want to check that reversion occurs. maybe a good, compatible solution is out there though.

Update (as of https://github.com/Layr-Labs/eigenlayer-contracts/pull/333/commits/674473377b8a68381a353aa7a17dae980accad50):
This now uses storage and abstracts users (a Staker and/or Operator) into a "User" contract. Feels better to me, at least.

### DIFF
--- a/src/test/Delegation.t.sol
+++ b/src/test/Delegation.t.sol
@@ -145,10 +145,9 @@ contract DelegationTests is EigenLayerTestHelper {
         _testDelegation(operator, staker, ethAmount, eigenAmount);
 
         (IStrategy[] memory strategyArray, uint256[] memory shareAmounts) = strategyManager.getDeposits(staker);
-        uint256[] memory strategyIndexes = new uint256[](strategyArray.length);
 
         // withdraw shares
-        _testQueueWithdrawal(staker, strategyIndexes, strategyArray, shareAmounts, staker /*withdrawer*/);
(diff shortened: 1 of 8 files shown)


ChaoticWalrus:
Update: planning on scrapping the bulk of this, will probably transform this PR into just a bit of a "cleanup" PR.

ChaoticWalrus:
closing this as stale. we sync'd on what makes sense here and got that done instead.

PR #335, opened 2023-11-16
wip: integration test framework

wadealexc:
See `integration/README.md` for notes on what's going on and how to use this framework.


src/test/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    function _newOperator() internal returns (User, IStrategy[] memory, uint[] memory) {
+        (User operator, IStrategy[] memory strategies, uint[] memory tokenBalances) = _randUser();
+        
+        operator.registerAsOperator();
+        // TODO - deposit into strats
+
+        assert_HasUnderlyingTokenBalances(operator, strategies, tokenBalances, "_newOperator: failed to award token balances");
+        assertTrue(delegationManager.isOperator(address(operator)), "_newOperator: operator should be registered");
```
ChaoticWalrus, line 37: could have a variant that changes the order of registering as an operator and depositing?
we can also add an `assert_HasOperatorShares` at the end 😄

src/test/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,367 @@
...
+                tokenBalance = address(user).balance;
+            } else {
+                tokenBalance = strat.underlyingToken().balanceOf(address(user));
+            }
+
+            assertEq(expectedBalance, tokenBalance, err);
+        }
+    }
+
+    function assert_HasNoUnderlyingTokenBalance(User user, IStrategy[] memory strategies, string memory err) internal {
```
ChaoticWalrus, line 75: should this just call `assert_HasUnderlyingTokenBalances` with an array of zeroes? little opportunity for de-duping

src/test/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,367 @@
...
+        uint prevQueuedWithdrawals = _getPrevCumulativeWithdrawals(staker);
+
+        assertEq(prevQueuedWithdrawals + withdrawals.length, curQueuedWithdrawals, err);
+    }
+
+    function assert_Snap_IncreasedTokenBalances(
+        User staker,
+        IERC20[] memory tokens,
+        uint[] memory addedTokens,
+        string memory err
```
ChaoticWalrus, line 236: I am feeling like perhaps some singlet versions of these functions make sense, but that feels like it could also be overkill to implement before trying out actually using these a bit more

src/test/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,367 @@
...
+            IStrategy strat = strategies[i];
+
+            uint tokenBalance = tokenBalances[i];
+            if (strat == BEACONCHAIN_ETH_STRAT) {
+                // TODO - need to calculate this
+                // expectedShares[i] = eigenPodManager.underlyingToShares(tokenBalance);
+                revert("_calculateExpectedShares: unimplemented for native eth");
+            } else {
+                expectedShares[i] = strat.underlyingToShares(tokenBalance);
+            }
```
ChaoticWalrus, line 269: pretty sure we just want to set `expectedShares[i] = tokenBalances[i]` in this case, but perhaps it depends a bit on the exact handling we use for 'beaconChainETH' elsewhere

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -0,0 +1,489 @@
...
+    DelayedWithdrawalRouter delayedWithdrawalRouter;
+
+    // Base strategy implementation in case we want to create more strategies later
+    StrategyBase baseStrategyImplementation;
+
+    // List of LST strategies
+    IStrategy[] _strategies;
+
+    // TODO - rename to time machine??
+    Global global;
```
ChaoticWalrus, line 48: yes, something like `TimeMachine` or even just `Snapshot` would be clearer

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -0,0 +1,489 @@
...
+        assetTypeToStr[HOLDS_ETH] = "HOLDS_ETH";
+        assetTypeToStr[HOLDS_MIX] = "HOLDS_MIX";
+        
+        signedTypeToStr[NO_SIGNED_METHODS] = "NO_SIGNED_METHODS";
+        signedTypeToStr[SIGNED_METHODS] = "SIGNED_METHODS";
+    }
+
+    function setUp() public virtual {
+        // Deploy ProxyAdmin
+        eigenLayerProxyAdmin = new ProxyAdmin();
```
ChaoticWalrus, line 138: I'd like to see us build this so we could also do fork testing w/ ~the same tests -- is that too extreme? seems very useful for, e.g., running the existing suite of tests against an upgrade.
We have something along these lines at present, which I tried to improve a bit in my PR -- see here:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/eb5ebea3987c97635fb5af39d5611119441a4cd7/src/test/EigenLayerDeployer.t.sol#L121-L127
(or for the present version, here: https://github.com/Layr-Labs/eigenlayer-contracts/blob/4bd42db27708138e9459de3744c146f977536e21/src/test/EigenLayerDeployer.t.sol#L126-L135)

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -0,0 +1,489 @@
...
+            )
+        );
+
+        // Push to list of tokens/strategies
+        _strategies.push(strategy);
+
+        // Whitelist strategy
+        IStrategy[] memory strategies = new IStrategy[](1);
+        strategies[0] = strategy;
+        strategyManager.addStrategiesToDepositWhitelist(strategies);
```
ChaoticWalrus, line 289: gonna need a `cheats.prank(strategyManager.strategyWhitelister())` just before the `strategyManager.addStrategiesToDepositWhitelist` call

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -0,0 +1,489 @@
...
+        ethStrats.push(BEACONCHAIN_ETH_STRAT);
+        mixedStrats.push(BEACONCHAIN_ETH_STRAT);
+
+        global = new Global();
+    }
+
+    /// @dev Deploy a strategy and its underlying token, push to global lists of tokens/strategies, and whitelist
+    /// strategy in strategyManager
+    function _newStrategyAndToken(string memory tokenName, string memory tokenSymbol, uint initialSupply, address owner) internal {
+        IERC20 underlyingToken = new ERC20PresetFixedSupply(tokenName, tokenSymbol, initialSupply, owner);
```
ChaoticWalrus, line 272: probably want to nix the tokenName and tokenSymbol inputs?
not sure about the 'owner' but maybe we should have a preset that makes some of these tests easier, like one that anyone can mint?

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -0,0 +1,489 @@
...
+        mixedStrats.push(strategy);
+    } 
+
+    function _configRand(
+        uint16 _randomSeed, 
+        uint _assetTypes,
+        uint _signedTypes
+    ) internal {
+        // Using uint16 for the seed type so that if a test fails, it's easier
+        // to manually use the seed to replay the same test.
```
ChaoticWalrus, line 302: not sure I'm understanding this comment. You should be able to run a fuzzed test with a particular input using syntax like:
```
forge script src/test/unit/DelegationUnit.t.sol:DelegationUnitTests --sig "testDecreaseDelegatedShares(address,address[],uint128,bool)" 0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f "[0x86294a1f20Afb94b41fbF27bd26930729C967dC2, 0x00000000000000000000000000000000000007D8, 0x02f460f68b8d2Bf4D66903FA7eb9F6a172b0FA20, 0x000000000000000000000000000000000000133C, 0x0000000000000000000000000000000000002616, 0x0000000000000000000000000000000000002925, 0x0000000000000000000000000000000000004baB, 0x27EA5aD96722c4D148c38Fb89BeE32f64F0e6065, 0x6D69979669F28Ac041dFBa0B2a781fe6f366E474, 0x00000000000000604A3ac11ce6251dCA61A05869]" 8843 true
```

I would think a uint256 seed would be best, to capture more scenarios / so it could be chopped up into more slices

src/test/integration/users/User.sol
```diff
@@ -0,0 +1,257 @@
...
+        DelegationManager _delegationManager,
+        StrategyManager _strategyManager,
+        EigenPodManager _eigenPodManager,
+        Global _global
+    ) {
+        delegationManager = _delegationManager;
+        strategyManager = _strategyManager;
+        eigenPodManager = _eigenPodManager;
+        global = _global;
+    }
```
ChaoticWalrus, line 39: should we perhaps just have the User contract read this info from its deployer (msg.sender), so the constructor doesn't need any inputs?

src/test/integration/IntegrationBase.t.sol
wadealexc: What we could do is update `User.registerAsOperator` to register -> deposit by default, and create like a `User_DepositFirst`-type of user variant... but i'm not sure where it's useful outside of this specific case.

I think we should talk about what additional variants we want to see during the call today!

For now I'm going to keep a static ordering for this, and add `assert_Snap_AddedOperatorShares` + `assert_Snap_AddedStakerShares` as checks

src/test/integration/IntegrationBase.t.sol
wadealexc: on it.

src/test/integration/IntegrationBase.t.sol
wadealexc: They might be useful, but so far I haven't needed em. Would be very easy to add though, so we can just wait till there's a need :)

src/test/integration/IntegrationBase.t.sol
wadealexc: i think some of the spots where i revert with "unimplemented for native eth" type things have easy answers, but i didn't want to leave some functional paths in when most of it doesnt work

src/test/integration/users/User.sol
ypatil12: +1 for this - much cleaner

src/test/integration/IntegrationDeployer.t.sol
wadealexc: i like TimeMachine. i'm trying really hard not to leave all kinds of jokes all over the place

wadealexc: but some stuff can be fun. i like having fun when i write tests

src/test/integration/IntegrationDeployer.t.sol
wadealexc: that would be cool! Im not familiar with how this works entirely, but if someone else wants to take it on - i think that's a good idea

src/test/integration/IntegrationDeployer.t.sol
wadealexc: i like these solely because i'm printing the token allocations in `_newUser`, in case a test fails and we want to see what sort of users we were generating

wadealexc: "anyone can mint" -- actually apparently `StdCheats.deal` just takes care of the minting question? although in retrospect that may be because we're calling as the owner? i have no idea tbh, i grabbed it from yash's tests

src/test/integration/IntegrationDeployer.t.sol
wadealexc: i think what i'm noticing is that when a test fails, it gives you the input it failed with - like the input was "0" or "1" or whatever. if the typing is a low-size uint, those values are readable and easy to quickly spin up a test with the same inputs. i was worried that a high-size input would require a lot more parsing and care?

we could definitely increase the size of `uint16`, but it's only running 256 scenarios regardless

src/test/integration/IntegrationDeployer.t.sol
wadealexc: [6ef08a6](https://github.com/Layr-Labs/eigenlayer-contracts/pull/335/commits/6ef08a676fe70389631c4627a9aa2d825a660884)

src/test/integration/IntegrationDeployer.t.sol
wadealexc: [6ef08a6](https://github.com/Layr-Labs/eigenlayer-contracts/pull/335/commits/6ef08a676fe70389631c4627a9aa2d825a660884)

src/test/integration/IntegrationBase.t.sol
wadealexc: [6ef08a6](https://github.com/Layr-Labs/eigenlayer-contracts/pull/335/commits/6ef08a676fe70389631c4627a9aa2d825a660884)

src/test/integration/IntegrationBase.t.sol
wadealexc: [6ef08a6](https://github.com/Layr-Labs/eigenlayer-contracts/pull/335/commits/6ef08a676fe70389631c4627a9aa2d825a660884)

src/test/integration/users/User.sol
wadealexc: would require a circular import, no?

src/test/integration/users/User.sol
```diff
@@ -0,0 +1,261 @@
...
+
+contract User is Test {
+
+    Vm cheats = Vm(HEVM_ADDRESS);
+
+    DelegationManager delegationManager;
+    StrategyManager strategyManager;
+    EigenPodManager eigenPodManager;
+
+    IStrategy[] depositedStrategies;
```
wadealexc, line 23: Do we need these? Top-level it's been easy to get by without them and I worry that too much stuff like this starts to accumulate logic in the User contracts

wadealexc: by these i mean `depositedStrategies`

src/test/integration/users/User.sol
ChaoticWalrus: you should be able to make an interface and then cast the caller to that interface
something like 
```
interface IUserDeployer {
function delegationManager() external view returns (DelegationManager);
...
}
```

ChaoticWalrus (approved):
Feeling really good about this overall.
We can continue to iterate and improve the framework as we build more tests 👍

src/test/integration/users/User.sol
wadealexc: [c99e847](https://github.com/Layr-Labs/eigenlayer-contracts/pull/335/commits/c99e847709852d7246c73b7d72d44bba368b760e)

src/test/integration/users/User.sol
wadealexc: [c99e847](https://github.com/Layr-Labs/eigenlayer-contracts/pull/335/commits/c99e847709852d7246c73b7d72d44bba368b760e)

src/test/integration/IntegrationDeployer.t.sol
wadealexc: [c99e847](https://github.com/Layr-Labs/eigenlayer-contracts/pull/335/commits/c99e847709852d7246c73b7d72d44bba368b760e)

increased to uint24

src/test/integration/IntegrationDeployer.t.sol
ypatil12: `StdCheats.deal` works by modifying storage manually :)

src/test/integration/IntegrationDeployer.t.sol
ypatil12: I'll add multi chain support in another PR

PR #337, opened 2023-11-18
Update AVS-Guide.md

pucedoteth:
when you click on link it takes you to 404 page, so i updated the link.


ChaoticWalrus:
Thanks for the PR!
Unfortunately multiple links are broken in this particular doc, and the doc which this _was_ linking to has been deleted.
We'll try to fix this soon though 😄

PR #340, opened 2023-11-20
support native eth in integration tests

wadealexc:
- fixes some error messages to display the correct function name
- added `BeaconChainMock`, a mock interface to the beacon chain which generates proofs for pod validators
- full support for native eth stakers using the `HOLDS_ETH` and `HOLDS_ALL` flags to generate users with one or more validators-worth of restaked eth. 
- eth granted to users is now a random multiple of 32 ether. users that are granted more than 1 validator-worth will create multiple validators in the same pod.

### DIFF
--- a/src/test/integration/IntegrationBase.t.sol
+++ b/src/test/integration/IntegrationBase.t.sol
@@ -22,7 +22,7 @@ abstract contract IntegrationBase is IntegrationDeployer {
     function _newRandomStaker() internal returns (User, IStrategy[] memory, uint[] memory) {
         (User staker, IStrategy[] memory strategies, uint[] memory tokenBalances) = _randUser();
         
-        assert_HasUnderlyingTokenBalances(staker, strategies, tokenBalances, "_newStaker: failed to award token balances");
+        assert_HasUnderlyingTokenBalances(staker, strategies, tokenBalances, "_newRandomStaker: failed to award token balances");
 
         return (staker, strategies, tokenBalances);
     }
@@ -33,9 +33,9 @@ abstract contract IntegrationBase is IntegrationDeployer {
         operator.registerAsOperator();
         operator.depositIntoEigenlayer(strategies, tokenBalances);
 
-        assert_Snap_AddedStakerShares(operator, strategies, tokenBalances, "_newOperator: failed to add delegatable shares");
-        assert_Snap_AddedOperatorShares(operator, strategies, tokenBalances, "_newOperator: failed to award shares to operator");
-        assertTrue(delegationManager.isOperator(address(operator)), "_newOperator: operator should be registered");
+        assert_Snap_AddedStakerShares(operator, strategies, tokenBalances, "_newRandomOperator: failed to add delegatable shares");
+        assert_Snap_AddedOperatorShares(operator, strategies, tokenBalances, "_newRandomOperator: failed to award shares to operator");
+        assertTrue(delegationManager.isOperator(address(operator)), "_newRandomOperator: operator should be registered");
 
         return (operator, strategies, tokenBalances);
     }
@@ -88,9 +88,13 @@ abstract contract IntegrationBase is IntegrationDeployer {
 
             uint actualShares;
             if (strat == BEACONCHAIN_ETH_STRAT) {
-                // TODO
-                // actualShares = eigenPodManager.podOwnerShares(address(user));
-                revert("unimplemented");
+                // TODO - is this the right way to handle this?
+                int shares = eigenPodManager.podOwnerShares(address(user));
+                if (shares < 0) {
+                    revert("assert_HasExpectedShares: negative shares");
+                }
+
+                actualShares = uint(shares);
             } else {
                 actualShares = strategyManager.stakerStrategyShares(address(user), strat);
             }
@@ -251,9 +255,7 @@ abstract contract IntegrationBase is IntegrationDeployer {
 
             uint tokenBalance = tokenBalances[i];
             if (strat == BEACONCHAIN_ETH_STRAT) {
-                // TODO - need to calculate this
-                // expectedShares[i] = eigenPodManager.underlyingToShares(tokenBalance);
-                revert("_calculateExpectedShares: unimplemented for native eth");
+                expectedShares[i] = tokenBalances[i];
             } else {
                 expectedShares[i] = strat.underlyingToShares(tokenBalance);
             }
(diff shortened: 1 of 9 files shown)


src/test/integration/IntegrationBase.t.sol
```diff
@@ -323,8 +323,13 @@ abstract contract IntegrationBase is IntegrationDeployer {
...
             if (strat == BEACONCHAIN_ETH_STRAT) {
-                // curShares[i] = eigenPodManager.podOwnerShares(address(staker));
-                revert("TODO: unimplemented");
+                // TODO - is this the right way to handle this?
+                int shares = eigenPodManager.podOwnerShares(address(staker));
+                if (shares < 0) {
+                    revert("_getStakerShares: negative shares");
+                }
+
+                curShares[i] = uint(shares);
```
ChaoticWalrus, line 332: to answer the question asked in the TODO here:
I think this is _OK_ as-is.
We should make sure we have some coverage for users having negative shares, but it is a bit of an "edge case", so I feel alright with it requiring a bit more special handling, and non-negative shares being the default.

ChaoticWalrus: other option is probably making every share amount an int256 and then casting back to uint256 as necessary, but that just sounds like wayyyy more of a pain

src/test/integration/IntegrationBase.t.sol
```diff
@@ -349,7 +354,11 @@ abstract contract IntegrationBase is IntegrationDeployer {
         uint[] memory balances = new uint[](tokens.length);
 
         for (uint i = 0; i < tokens.length; i++) {
-            balances[i] = tokens[i].balanceOf(address(staker));
+            if (address(tokens[i]) == address(0)) {
+                balances[i] = address(staker).balance;
```
ChaoticWalrus, line 358: does using the zero address for native ETH make sense here / could it be confusing? maybe we want to use the 0xbeaC0eeEeeeeEEeEeEEEEeeEEeEeeeEeeEEBEaC0 we're using elsewhere?
not too sure or strongly opinionated

ypatil12: On board with using `0xbeaC0..`

src/test/integration/mocks/BeaconChainMock.t.sol
```diff
@@ -0,0 +1,689 @@
+// SPDX-License-Identifier: BUSL-1.1
```
ypatil12, line 1: Nit: none of our mock files end in `.t.sol`

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -378,14 +385,20 @@ abstract contract IntegrationDeployer is Test, IUserDeployer {
...
             }
-
-            return (strategies, tokenBalances);
         } else if (assetType == HOLDS_ETH) {
-            revert("_getRandAssets: HOLDS_ETH unimplemented");
+            strategies = new IStrategy[](1);
+            tokenBalances = new uint[](1);
+
+            // Award the user 32 ETH
+            uint amount = 32 ether;
```
ypatil12, line 393: Should we award the user some value of ETH in the set {<32 ETH, 32 ETH, > 32ETH} ?

ypatil12 (commented):
Looks great! Minor comments

src/test/integration/IntegrationBase.t.sol
wadealexc: Good thoughts. 

Agreed that non-negative is a good default, since we should only see negative shares if we do a balance update after a withdrawal -- and IMO that deserves its own top-level test.

If/when we write that test we can write a balance handler that specifically allows negative balances. In the meantime, if we hit negative here failing is fine because that shouldn't happen.

src/test/integration/IntegrationDeployer.t.sol
wadealexc: Great question, I was wondering too. 

There is a slight hitch with just using random ETH values: `EigenPod.stake` requires that `msg.value == 32 ether`. Also, the deposit contract itself has a few requirements that I'd want the mock beacon chain to enforce:
* `depositValue >= 1 ether`
* `depositValue % 1 gwei == 0`
* `depositValue / 1 gwei <= type(u64).max`

The easy, immediate changes we could make:
* Have `BeaconChainMock.newValidator` enforce the deposit contract conditions I listed above
* Create users that are awarded some multiple of 32 ETH (1x, 2x, 3x, etc) - and automatically deploy more validators to the same pod.

I think these would be valuable additions, so will work on those now.

wadealexc: Your question about having validators with other nonstandard balances is a good one though. So here's the idea: since you don't actually need to call `stake` to deposit and verify withdrawal credentials - we have a `userType` that skips that step in favor of depositing directly into the beacon chain.

Here's my what the changelog could look like, adding onto the changes I described above:

1. We change the `SIGNED_METHODS` usertype: it's now called `ALT_METHODS`. The concept is a user type that always calls the "non-default, but equivalent" methods.
2. We change `User_SignedMethods` (now called `User_AltMethods`) to override `depositIntoEigenlayer`. The new version does NOT call `eigenPodManager.stake`. Instead, it creates a pod, then directly deposits and verifies credentials. And this would mean it could handle other values for ETH (those supported by the deposit contract, anyway)
3. When dealing random assets, if we're handling `assetType == HOLDS_ETH` and `userType == ALT_METHODS`, we give them a random multiple of 32 ETH, but we ALSO add or subtract a random amount of gwei, e.g: { -10 gwei, -1 gwei, -0 gwei, +1 gwei, +10 gwei }

WDYT?

src/test/integration/mocks/BeaconChainMock.t.sol
wadealexc: This is true, but I think maybe they should... though I'm not entirely sure what `.t.sol` actually does after looking into it.

I asked in the foundry support channel and our slack: https://eigenlabs.slack.com/archives/C032198EJ23/p1701271995541119

src/test/integration/IntegrationBase.t.sol
wadealexc: Funny, as it turns out I already had a `NATIVE_ETH` constant declared that did this - I just wasn't using it!

Good call, will make this fix :smile:

src/test/integration/IntegrationDeployer.t.sol
ypatil12: `User_AltMethods` additions make sense to me. Is the reasoning of subtracting a random amount of gwei to test  `deposit value % 1 Gwei == 0` invariant from the deposit contract?

src/test/integration/IntegrationBase.t.sol
wadealexc: [3eba0cc](https://github.com/Layr-Labs/eigenlayer-contracts/pull/340/commits/3eba0cc4f8f1c459995f6e857cdc7cffc0c36215)

src/test/integration/mocks/BeaconChainMock.t.sol
wadealexc: going to keep this as is for now, but happy to revisit later if we want to standardize

ypatil12 (approved):

PR #342, opened 2023-11-20
Operator AVS Mapping Support

ypatil12:
This WIP PR adds support in the DM to register an operator under a specific AVS. The expectation is that an AVS's registry coordinator will call the `registerOperatorWithAVS` and `deregisterOperatorFromAVS` functions to register and deregister an operator. AVSs can also use the `updateAVSMetadataURI` function to update a URI link. Full design doc is linked [here](https://docs.google.com/document/d/18iGFx7srXSlEt89E4nYLhWxm80R15kpG27cedAgxkTo/edit).

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -443,7 +443,7 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
     }
 
     /**
-     * @notice External function called by AVSs to register an operator with the AVS.
+     * @notice Called by AVSs to register an operator with the AVS.
      * @param operator The address of the operator to register.
      * @param signatureWithSaltAndExpiry The signature, salt, and expiry of the operator's signature.
      */
@@ -476,7 +476,7 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
     }
 
     /**
-     * @notice External function called by AVSs to deregister an operator with the AVS.
+     * @notice Called by AVSs to deregister an operator with the AVS.
      * @param operator The address of the operator to deregister.
      */
     function deregisterOperatorFromAVS(address operator) external {
@@ -486,6 +486,14 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         registeredWithAVS[msg.sender][operator] = false;
 
         emit OperatorRegistrationStatusUpdated(operator, msg.sender, OperatorRegistrationStatus.DEREGISTERED);
+    }
+
+    /**
+     * @notice Called by an AVS to emit an `OperatorMetadataURIUpdated` event indicating the information has updated.
+     * @param metadataURI The URI for metadata associated with an AVS
+     */
+    function updateAVSMetadataURI(string calldata metadataURI) external {
+        emit AVSMetadataURIUpdated(msg.sender, metadataURI);
     }
 
     /*******************************************************************************
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -475,6 +475,19 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         emit OperatorRegistrationStatusUpdated(operator, msg.sender, OperatorRegistrationStatus.REGISTERED);
     }
 
+    /**
+     * @notice External function called by AVSs to deregister an operator with the AVS.
+     * @param operator The address of the operator to deregister.
+     */
+    function deregisterOperatorFromAVS(address operator) external {
+        require(registeredWithAVS[msg.sender][operator], "DelegationManager.deregisterOperatorFromAVS: operator not registered");
+
+        // deregister operator
+        registeredWithAVS[msg.sender][operator] = false;
+
+        emit OperatorRegistrationStatusUpdated(operator, msg.sender, OperatorRegistrationStatus.DEREGISTERED);
+    }
+
     /*******************************************************************************
                             INTERNAL FUNCTIONS
     *******************************************************************************/
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -472,7 +472,7 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         // register operator
         registeredWithAVS[msg.sender][operator] = true;
 
-        emit OperatorRegistrationStatusUpdated(operator, msg.sender, OperatorRegistrationStatus.REGISTERED);
+        emit OperatorRegistrationStatusUpdatedForAVS(operator, msg.sender, OperatorRegistrationStatus.REGISTERED);
     }
 
     /**
@@ -485,7 +485,7 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         // deregister operator
         registeredWithAVS[msg.sender][operator] = false;
 
-        emit OperatorRegistrationStatusUpdated(operator, msg.sender, OperatorRegistrationStatus.DEREGISTERED);
+        emit OperatorRegistrationStatusUpdatedForAVS(operator, msg.sender, OperatorRegistrationStatus.DEREGISTERED);
     }
 
     /**
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -442,6 +442,39 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         }
     }
 
+    /**
+     * @notice External function called by AVSs to register an operator with the AVS.
+     * @param operator The address of the operator to register.
+     * @param signatureWithSaltAndExpiry The signature, salt, and expiry of the operator's signature.
+     */
+    function registerOperatorWithAVS(address operator, ISignatureUtils.SignatureWithSaltAndExpiry memory signatureWithSaltAndExpiry) external {
+        require(signatureWithSaltAndExpiry.expiry > block.timestamp, "DelegationManager.registerOperatorWithAVS: signature expired");
+        require(operatorSaltIsSpent[operator][signatureWithSaltAndExpiry.salt] == false, "DelegationManager.registerOperatorWithAVS: salt already used");
+        require(!registeredWithAVS[msg.sender][operator], "DelegationManager.registerOperatorWithAVS: operator already registered");
+
+        // Calculate the digest hash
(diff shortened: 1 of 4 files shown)


src/contracts/core/DelegationManager.sol
```diff
@@ -442,6 +442,60 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         }
     }
 
+    /**
+     * @notice Called by AVSs to register an operator with the AVS.
+     * @param operator The address of the operator to register.
+     * @param signatureWithSaltAndExpiry The signature, salt, and expiry of the operator's signature.
+     */
+    function registerOperatorWithAVS(address operator, ISignatureUtils.SignatureWithSaltAndExpiry memory signatureWithSaltAndExpiry) external {
```
wadealexc, line 450: Would prefer this is a multi-line declaration, since it's quite long.

Additionally, I'm a big fan of just calling these signature parameters `signature` rather than the super long rewording of the struct name.

wadealexc: Or even better, `operatorSignature` because that communicates important info.

src/contracts/core/DelegationManager.sol
```diff
@@ -939,6 +993,28 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+     * @param avs The AVS the operator is registering with
+     * @param salt A unique and single use value associated with the approver signature.
+     * @param expiry Time after which the approver's signature becomes invalid
+     */
+    function calculateOperatorRegistrationDigestHash(
+        address operator,
+        address avs,
+        bytes32 salt,
+        uint256 expiry
+    ) public view returns (bytes32) {
```
wadealexc, line 1008: Request to shorten this function name. I see it's similar to other methods in this contract... not sure whether we need a second pass to just shorten all the names (similar to the middleware renaming PRs), or what.

src/contracts/core/DelegationManager.sol
```diff
@@ -442,6 +442,60 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        registeredWithAVS[msg.sender][operator] = true;
+
+        emit OperatorRegistrationStatusUpdatedForAVS(operator, msg.sender, OperatorRegistrationStatus.REGISTERED);
+    }
+
+    /**
+     * @notice Called by AVSs to deregister an operator with the AVS.
+     * @param operator The address of the operator to deregister.
+     */
+    function deregisterOperatorFromAVS(address operator) external {
```
wadealexc, line 482: Questions:
1. Are we ok with this being unsigned?
2. Do we want the operator to be able to deregister themselves, or should this only be something the AVS does?

src/contracts/core/DelegationManager.sol
```diff
@@ -442,6 +442,60 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        // deregister operator
+        registeredWithAVS[msg.sender][operator] = false;
+
+        emit OperatorRegistrationStatusUpdatedForAVS(operator, msg.sender, OperatorRegistrationStatus.DEREGISTERED);
+    }
+
+    /**
+     * @notice Called by an AVS to emit an `OperatorMetadataURIUpdated` event indicating the information has updated.
+     * @param metadataURI The URI for metadata associated with an AVS
+     */
```
wadealexc, line 494: Comments are incorrect; references an OperatorMetadataURI

src/contracts/core/DelegationManager.sol
```diff
@@ -442,6 +442,60 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
     }
 
+    /**
+     * @notice Called by AVSs to register an operator with the AVS.
+     * @param operator The address of the operator to register.
+     * @param signatureWithSaltAndExpiry The signature, salt, and expiry of the operator's signature.
+     */
+    function registerOperatorWithAVS(address operator, ISignatureUtils.SignatureWithSaltAndExpiry memory signatureWithSaltAndExpiry) external {
+        require(signatureWithSaltAndExpiry.expiry > block.timestamp, "DelegationManager.registerOperatorWithAVS: signature expired");
+        require(operatorSaltIsSpent[operator][signatureWithSaltAndExpiry.salt] == false, "DelegationManager.registerOperatorWithAVS: salt already used");
```
ChaoticWalrus, line 452: very tiny nit, but you can just use the boolean here instead of checking equivalence, like:
` require(!operatorSaltIsSpent[operator][signatureWithSaltAndExpiry.salt], ...)`

src/contracts/core/DelegationManager.sol
ChaoticWalrus: I _personally think_:

- for (1) yes, the AVS should be able to unilaterally do this
- we should aspire to also accomplish (2), but I'm unclear on if this is required (as part of the upcoming release, at least). will flag this to people for input

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -92,6 +96,14 @@ abstract contract DelegationManagerStorage is IDelegationManager {
     /// @notice the address of the StakeRegistry contract to call for stake updates when operator shares are changed
     IStakeRegistryStub public stakeRegistry;
 
+    /// @notice Mapping: AVS => operator => whether or not the operator is currently registered with the AVS
+    mapping(address => mapping(address => bool)) public registeredWithAVS;
```
ChaoticWalrus, line 100: feels maybe a bit odd to define the `OperatorRegistrationStatus` enum but not use it in this mapping / instead use a boolean here.
I think it would be more "harmonious" to use the enum here as well -- clarity seems similar but I think it'd be nice if we end up adding more status in the future?
not feeling strongly / definitely open to other ideas here though

src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -147,6 +153,15 @@ interface IDelegationManager is ISignatureUtils {
     /// @notice Emitted when the `withdrawalDelayBlocks` variable is modified from `previousValue` to `newValue`.
     event WithdrawalDelayBlocksSet(uint256 previousValue, uint256 newValue);
 
+    /// @notice Emitted when an operator's registration status for an AVS is updated
+    event OperatorRegistrationStatusUpdatedForAVS(address indexed operator, address indexed avs, OperatorRegistrationStatus status);
+
+    /**
+     * @notice Emitted when @param operator indicates that they are updating their MetadataURI string
+     * @dev Note that these strings are *never stored in storage* and are instead purely emitted in events for off-chain indexing
```
ChaoticWalrus, line 161: typo from copy-pasting(?):
`operator` => `avs`

src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -147,6 +153,15 @@ interface IDelegationManager is ISignatureUtils {
     /// @notice Emitted when the `withdrawalDelayBlocks` variable is modified from `previousValue` to `newValue`.
     event WithdrawalDelayBlocksSet(uint256 previousValue, uint256 newValue);
 
+    /// @notice Emitted when an operator's registration status for an AVS is updated
+    event OperatorRegistrationStatusUpdatedForAVS(address indexed operator, address indexed avs, OperatorRegistrationStatus status);
```
ChaoticWalrus, line 157: maybe just name this something like `AVSRegistrationStatusUpdated`? idk if that's actually that much better tho

src/contracts/core/DelegationManager.sol
```diff
@@ -442,6 +442,60 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        registeredWithAVS[msg.sender][operator] = false;
+
+        emit OperatorRegistrationStatusUpdatedForAVS(operator, msg.sender, OperatorRegistrationStatus.DEREGISTERED);
+    }
+
+    /**
+     * @notice Called by an AVS to emit an `OperatorMetadataURIUpdated` event indicating the information has updated.
+     * @param metadataURI The URI for metadata associated with an AVS
+     */
+    function updateAVSMetadataURI(string calldata metadataURI) external {
```
shrimalmadhur, line 495: who calls it? as in how does AVS update their metadata, via a frontend or something else?

src/contracts/core/DelegationManager.sol
ChaoticWalrus: I suspect it's just fine to keep (2) as aspirational for the moment, since it's a significant expansion of scope with unclear need (we then have to figure out how to communicate this deregistration to AVSs, at least).

ChaoticWalrus (approved):
Seems quite good. I only have a few nits which I'd like to see addressed

src/contracts/core/DelegationManager.sol
ypatil12: 1. Agree with this. Also note that AVS' can eject operators, so it makes sense that an AVS can force deregister an operator
2. #2 doesn't seem necessary in the short term and adds complexity

src/contracts/core/DelegationManager.sol
ypatil12: This is called by an AVS' registry coordinator contract:

https://github.com/Layr-Labs/eigenlayer-middleware/blob/5e695c1d6c898c394d8810c111777881ee0fa61d/src/BLSRegistryCoordinatorWithIndices.sol#L376-L383

src/contracts/core/DelegationManager.sol
samlaf: Talked to yash about this yesterday. Unclear that we need the signature. Can't the contract just check that the tx originator is the operator address?

src/contracts/core/DelegationManager.sol
ChaoticWalrus: > What do you guys mean by "AVS can eject operators"? Is that based on the contracts in this repo? Or is this a general requirement that we have? Also what's the difference between ejecting and deregistering? Is "deregistering" just the name you use to refer to the registry in the DM?

I think we use the term "ejection" to more specifically mean an AVS deregistering an operator _without the operator initiating the action_. While not a strict requirement, I would say AVSs SHOULD have some kind of ejection mechanism like what's in this repo

> Is there a difference between deregistering and just updating my stake to be 0? Seems like if operators can't either deregister, or remove their stake from an AVS, then they are stuck there forever..?

ideally, EigenLayer would know + index when an AVS no longer considers an operator to be opted-into/serving the service.
I think aspirationally there would be a high level of agreement/continuous consent in this relationship, where either party can sever the relationship (subject to some rules).
I agree right now it feels like operators could kind of "get stuck" in AVSs, _but_ I'm not seeing specific downsides to that, so I feel OK deferring development of some kind of operator escape hatch mechanism until later.

PR #345, opened 2023-11-22
made forceUndelegate queue a withdrawal for each strategy

Gajesh2007:


src/contracts/core/DelegationManager.sol
```diff
@@ -241,32 +252,36 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
-        // emit an event if this action was not initiated by the staker themselves
         if (msg.sender != staker) {
             emit StakerForceUndelegated(staker, operator);
         }
 
-        // undelegate the staker
         emit StakerUndelegated(staker, operator);
         delegatedTo[staker] = address(0);
 
-        // if no delegatable shares, return zero root, and don't queue a withdrawal
```
ChaoticWalrus, line 257: I'm not sure why some of these in-line comments are being deleted -- I find them helpful for clarity / splitting up the execution into "logical blocks", personally

ChaoticWalrus (changes requested):
This PR looks really good!
The change has pluses and minuses, but I do like how it gives stakes more flexibility in how they complete these withdrawals.
I'm surprised this PR doesn't require more changes to tests -- this does flag to me a potential area with low(ish) test coverage.

Requesting one main change here: this PR should be targeting the `m2-mainnet` branch instead of the `master` branch. I have a question about in-line comments too (bit of a small nit, but I think I'd like to see them restored if you lack a reason for deleting them).

ChaoticWalrus:
> quick question - should i branch out `m2-mainnet`? and open new PR?

I think it would have been good to do this, but I'm pretty sure (like 95%+?) that you can just change the "target" branch of the PR and it should all be just fine, and you avoid having to duplicate work.

Gajesh2007:
done.

Gajesh2007:
oh wait. there are a ton of changes. do we need this?

ChaoticWalrus:
> oh wait. there are a ton of changes. should we merge the other changes?

fuuuuuu guess I was wrong, we should have duplicated the PR, I'm sorry!

ChaoticWalrus:
Changed the base branch back to make this PR readable

src/contracts/core/DelegationManager.sol
```diff
@@ -245,17 +244,26 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
-            // Remove all strategies/shares from staker and operator and place into queue
-            return _removeSharesAndQueueWithdrawal({
-                staker: staker,
-                operator: operator,
-                withdrawer: staker,
-                strategies: strategies,
-                shares: shares
-            });
+            withdrawalRoots = new bytes32[](strategies.length);
+            for (uint i = 0; i < strategies.length; i++) {
```
wadealexc, line 250: `uint256`

ChaoticWalrus:
Looking into getting this into shape rn

wadealexc (approved):

PR #346, opened 2023-11-28
Generalized Proofs

Sidu28:


src/contracts/interfaces/IEigenPod.sol
```diff
@@ -50,6 +50,17 @@ interface IEigenPod {
     }
 
 
+    struct ProofService {
+        address caller;
+        // whether or not the proof fulfiller is allowed to fulfill proofs
+        bool permission;
+        // whether or not the proof fulfiller has been added
+        uint256 maxFee;
```
gpsanant, line 58: i feel like this should be in the user's signature?

src/contracts/pods/EigenPod.sol
```diff
@@ -94,6 +99,17 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
     uint64 public sumOfPartialWithdrawalsClaimedGwei;
 
+    /// @notice This is the offchain proving service
+    ProofService public proofService;
+
+    /// @notice This variable tracks the timestamp at which the last partial withdrawal was proven via ZK proofs
+    uint64 public withdrawalProvenUntilTimestamp;
+
+    /// @notice Switch to turn off partial withdrawal merkle proofs and turn on offchain proofs as a service
+    bool public partialWithdrawalProofSwitch;
```
gpsanant, line 107: shouldn't we make this an immutable

src/contracts/pods/EigenPod.sol
```diff
@@ -94,6 +97,17 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
      /// @notice This variable tracks the total amount of partial withdrawals claimed via merkle proofs prior to a switch to ZK proofs for claiming partial withdrawals
     uint64 public sumOfPartialWithdrawalsClaimedGwei;
 
+    /// @notice This is the offchain proving service
+    ProofService public proofService;
+
+    /// @notice This variable tracks the timestamp at which the last partial withdrawal was proven via ZK proofs
+    uint64 public withdrawalProvenUntilTimestamp;
```
gpsanant, line 104: cant we overload `mostRecentWithdrawalTimestamp` with this?

src/contracts/pods/EigenPod.sol
gpsanant: or set in EigenPodManager rather?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +466,49 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        uint64 startTimestamp,
+        uint64 endTimestamp,
+        uint256 provenPartialWithdrawalSumWei,
+        uint256 fee
+    ) external onlyProofService partialWithdrawalProofSwitchOn {
+        //TODO: figure if we wanna do the first proof from genesis time or not
+        if(withdrawalProvenUntilTimestamp == 0){
+            withdrawalProvenUntilTimestamp = GENESIS_TIME;
+        }
+        require(startTimestamp < endTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: startTimestamp must precede endTimestamp");
```
gpsanant, line 495: please check the doc for the actual checks to make

src/contracts/interfaces/IEigenPod.sol
Sidu28: it is?

src/contracts/pods/EigenPodManagerStorage.sol
```diff
@@ -64,6 +64,12 @@ abstract contract EigenPodManagerStorage is IEigenPodManager {
      */
     mapping(address => int256) public podOwnerShares;
 
+    /// @notice This is the offchain proving service
+    ProofService public proofService;
+
+    /// @notice Switch to turn off partial withdrawal merkle proofs and turn on offchain proofs as a service
+    bool public partialWithdrawalProofSwitch;
```
gpsanant, line 71: partialWithdrawalProofServiceEnabled

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +225,37 @@ contract EigenPodManager is
...
+            partialWithdrawalProofSwitch = false;
+        } else {
+            partialWithdrawalProofSwitch = true;
+        }
+    }
+
+    function updateProofService(address caller, uint256 maxFee, address feeRecipient) external onlyOwner {
+        proofService = ProofService({
+            caller: caller,
+            maxFee: maxFee,
```
gpsanant, line 253: this should be in the user signature, not global

src/contracts/pods/EigenPod.sol
```diff
@@ -467,46 +436,29 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        require(withdrawalCallbackInfo.startTimestamp < withdrawalCallbackInfo.endTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: startTimestamp must precede endTimestamp");
+        require(withdrawalCallbackInfo.startTimestamp == mostRecentWithdrawalTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: startTimestamp must match mostRecentWithdrawalTimestamp");
+        provenPartialWithdrawalSumWei -= withdrawalCallbackInfo.fee;
         //send proof service their fee
-        AddressUpgradeable.sendValue(payable(proofService.feeRecipient), fee);
+        AddressUpgradeable.sendValue(payable(feeRecipient), withdrawalCallbackInfo.fee);
 
         //subtract an partial withdrawals that may have been claimed via merkle proofs
         if(provenPartialWithdrawalSumWei > sumOfPartialWithdrawalsClaimedGwei * GWEI_TO_WEI) {
             provenPartialWithdrawalSumWei -= sumOfPartialWithdrawalsClaimedGwei * GWEI_TO_WEI;
```
gpsanant, line 457: we need to update `sumOfPartialWithdrawalsClaimedGwei`?

gpsanant: also this needs to include the fee

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -215,4 +214,9 @@ interface IEigenPod {
 
     /// @notice called by owner of a pod to remove any ERC20s deposited in the pod
     function recoverTokens(IERC20[] memory tokenList, uint256[] memory amountsToWithdraw, address recipient) external;
+
+    function fulfillPartialWithdrawalProofRequest(
```
gpsanant, line 218: needs natspec comment

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -39,6 +42,22 @@ interface IEigenPodManager is IPausable {
         bytes32 withdrawalRoot
     );
 
+    //info for each withdrawal called back by proof service
+    struct WithdrawalCallbackInfo {
+        address podOwner;
```
gpsanant, line 47: prefer comments on each var if possible

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +225,37 @@ contract EigenPodManager is
         ownerToPod[podOwner].withdrawRestakedBeaconChainETH(destination, shares);
     }
 
+    function proofServiceCallback(
```
gpsanant, line 228: natspec comment here and in interface

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +225,37 @@ contract EigenPodManager is
...
+        bytes32 blockRoot,
+        uint64 oracleTimestamp,
+        uint256 maxFee;
```
gpsanant, line 231: this is per callback?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +225,37 @@ contract EigenPodManager is
...
+            partialWithdrawalProofSwitch = true;
+        }
+    }
+
+    function updateProofService(address caller, address feeRecipient) external onlyOwner {
+        proofService = ProofService({
+            caller: caller,
+            feeRecipient: feeRecipient
+        });
+        emit ProofServiceUpdated(proofService.caller);
```
gpsanant, line 256: emit struct in event

src/contracts/pods/EigenPod.sol
```diff
@@ -124,6 +127,11 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         _;
     }
 
+    modifier partialWithdrawalProofSwitchOff() {
```
gpsanant, line 130: can delete?

src/contracts/pods/EigenPod.sol
```diff
@@ -162,6 +162,12 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         _;
     }
 
+    modifier partialWithdrawalProofSwitchOn() {
```
emagici: nit: `whenPartialWithdrawalProofSwitchOn` for better readability and clarification

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +435,35 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+    ) external onlyEigenPodManager  {
+        uint256 provenPartialWithdrawalSumWei = withdrawalCallbackInfo.provenPartialWithdrawalSumWei;
+
+        require(withdrawalCallbackInfo.startTimestamp < withdrawalCallbackInfo.endTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: startTimestamp must precede endTimestamp");
+        require(withdrawalCallbackInfo.startTimestamp == mostRecentWithdrawalTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: startTimestamp must match mostRecentWithdrawalTimestamp");
+        provenPartialWithdrawalSumWei -= withdrawalCallbackInfo.fee;
+        //send proof service their fee
+        AddressUpgradeable.sendValue(payable(feeRecipient), withdrawalCallbackInfo.fee);
+
+        //subtract an partial withdrawals that may have been claimed via merkle proofs
```
emagici, line 455: nit: // subtract

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +435,35 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+    function fulfillPartialWithdrawalProofRequest(
+        IEigenPodManager.WithdrawalCallbackInfo calldata withdrawalCallbackInfo,
+        address feeRecipient
+    ) external onlyEigenPodManager  {
+        uint256 provenPartialWithdrawalSumWei = withdrawalCallbackInfo.provenPartialWithdrawalSumWei;
+
+        require(withdrawalCallbackInfo.startTimestamp < withdrawalCallbackInfo.endTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: startTimestamp must precede endTimestamp");
+        require(withdrawalCallbackInfo.startTimestamp == mostRecentWithdrawalTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: startTimestamp must match mostRecentWithdrawalTimestamp");
+        provenPartialWithdrawalSumWei -= withdrawalCallbackInfo.fee;
+        //send proof service their fee
```
emagici, line 452: nit: // send

src/contracts/pods/EigenPod.sol
Sidu28: good catch

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +225,37 @@ contract EigenPodManager is
...
 
+    function proofServiceCallback(
+        bytes32 blockRoot,
+        uint64 oracleTimestamp,
+        uint256 maxFee;
+        WithdrawalCallbackInfo[] calldata callbackInfo
+    ) external onlyProofService partialWithdrawalProofSwitchOn {
+        require(blockRoot == getBlockRootAtTimestamp(oracleTimestamp), "EigenPodManager.proofServiceCallback: block root does not match oracleRoot for that timestamp");
+        for(uint256 i = 0; i < callbackInfo.length; i++) {
+            require(oracleTimestamp >= callbackInfo[i].endTimestamp, "EigenPodManager.proofServiceCallback: oracle timestamp must be greater than or equal to callback timestamp");
```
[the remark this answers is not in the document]
Sidu28: yeah purely a sanity check, will add comment

src/contracts/pods/EigenPodManagerStorage.sol
wadealexc: IMO just `proofServiceEnabled` is cleaner

src/contracts/pods/EigenPodManager.sol
```diff
@@ -44,6 +44,16 @@ contract EigenPodManager is
...
 
+    modifier onlyProofService() {
+        require(msg.sender == proofService.caller, "EigenPod.onlyProofService: not a permissioned fulfiller");
+        _;
+    }
+
+    modifier partialWithdrawalProofSwitchOn() {
+        require(partialWithdrawalProofSwitch, "EigenPod.partialWithdrawalProofSwitchOn: partial withdrawal proof switch is off");
+        _;
+    }
```
wadealexc, line 55: Rename: `proofServiceEnabled`

src/contracts/pods/EigenPod.sol
```diff
@@ -711,7 +746,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         uint64 withdrawalTimestamp,
         address recipient,
         uint64 partialWithdrawalAmountGwei
-    ) internal returns (VerifiedWithdrawal memory) {
+    ) internal partialWithdrawalProofSwitchOff returns (VerifiedWithdrawal memory) {
```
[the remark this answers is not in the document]
wadealexc: (in `_verifyAndProcessWithdrawal`)

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +225,36 @@ contract EigenPodManager is
...
 
+    /// @notice Called by proving service to fulfill a partial withdrawal proof request
+    function proofServiceCallback(
+        bytes32 blockRoot,
+        uint64 oracleTimestamp,
+        WithdrawalCallbackInfo[] calldata callbackInfo
+    ) external onlyProofService partialWithdrawalProofSwitchOn {
+        require(blockRoot == getBlockRootAtTimestamp(oracleTimestamp), "EigenPodManager.proofServiceCallback: block root does not match oracleRoot for that timestamp");
+        for(uint256 i = 0; i < callbackInfo.length; i++) {
+            //these checks are verified in the snark, we add them here again as a sanity check
```
wadealexc, line 236: "// these"

smh

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +225,36 @@ contract EigenPodManager is
...
+            require(oracleTimestamp >= callbackInfo[i].endTimestamp, "EigenPodManager.proofServiceCallback: oracle timestamp must be greater than or equal to callback timestamp");
+            require(callbackInfo[i].fee <= callbackInfo[i].maxFee, "EigenPod.fulfillPartialWithdrawalProofRequest: fee must be less than or equal to maxFee");
+            IEigenPod pod = ownerToPod[callbackInfo[i].podOwner];
+            pod.fulfillPartialWithdrawalProofRequest(callbackInfo[i], proofService.feeRecipient);
+        }
+    }
```
wadealexc, line 242: Please add a space here between methods

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +225,36 @@ contract EigenPodManager is
...
+        bytes32 blockRoot,
+        uint64 oracleTimestamp,
+        WithdrawalCallbackInfo[] calldata callbackInfo
+    ) external onlyProofService partialWithdrawalProofSwitchOn {
+        require(blockRoot == getBlockRootAtTimestamp(oracleTimestamp), "EigenPodManager.proofServiceCallback: block root does not match oracleRoot for that timestamp");
+        for(uint256 i = 0; i < callbackInfo.length; i++) {
+            //these checks are verified in the snark, we add them here again as a sanity check
+            require(oracleTimestamp >= callbackInfo[i].endTimestamp, "EigenPodManager.proofServiceCallback: oracle timestamp must be greater than or equal to callback timestamp");
+            require(callbackInfo[i].fee <= callbackInfo[i].maxFee, "EigenPod.fulfillPartialWithdrawalProofRequest: fee must be less than or equal to maxFee");
+            IEigenPod pod = ownerToPod[callbackInfo[i].podOwner];
```
wadealexc, line 239: Should we have a zero-address check with explicit revert?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +432,37 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+    *******************************************************************************/
+
+    /// @notice Called by the EigenPodManager to fulfill a partial withdrawal proof request
+    function fulfillPartialWithdrawalProofRequest(
+        IEigenPodManager.WithdrawalCallbackInfo calldata withdrawalCallbackInfo,
+        address feeRecipient
+    ) external onlyEigenPodManager  {
+        uint256 provenPartialWithdrawalSumWei = withdrawalCallbackInfo.provenPartialWithdrawalSumWei;
+
+        require(withdrawalCallbackInfo.startTimestamp < withdrawalCallbackInfo.endTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: startTimestamp must precede endTimestamp");
```
wadealexc, line 447: Is `<=` valid?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +432,37 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        require(withdrawalCallbackInfo.startTimestamp == mostRecentWithdrawalTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: startTimestamp must match mostRecentWithdrawalTimestamp");
```
wadealexc, line 448: What happens if we haven't processed any withdrawals? `mostRecentWithdrawalTimestamp` is zero in this case right? Will this function correctly?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +432,37 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+
+        //subtract an partial withdrawals that may have been claimed via merkle proofs
```
wadealexc, line 450: // subtract

smh smh

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +432,37 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        if(provenPartialWithdrawalSumWei > sumOfPartialWithdrawalsClaimedGwei * GWEI_TO_WEI) {
```
wadealexc, line 451: Should this be `>=`?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +225,36 @@ contract EigenPodManager is
         ownerToPod[podOwner].withdrawRestakedBeaconChainETH(destination, shares);
     }
 
+    /// @notice Called by proving service to fulfill a partial withdrawal proof request
+    function proofServiceCallback(
+        bytes32 blockRoot,
+        uint64 oracleTimestamp,
+        WithdrawalCallbackInfo[] calldata callbackInfo
+    ) external onlyProofService partialWithdrawalProofSwitchOn {
```
wadealexc, line 233: Let's make this nonReentrant since we're calling an untrusted address in the EigenPod

src/contracts/pods/EigenPod.sol
Sidu28: We're doing  non inclusice end timestamp, ie, (start, end] so we want "<"

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +432,37 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+            provenPartialWithdrawalSumWei -= sumOfPartialWithdrawalsClaimedGwei * GWEI_TO_WEI;
```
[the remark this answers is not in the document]
Sidu28: not sure I follow the question - if the try to "double prove" via merkle proofs then zk proofs, they would get nothing the second time aorund

src/contracts/pods/EigenPod.sol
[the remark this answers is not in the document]
Sidu28: is there a reason for this?  Just readability or is it something else?

.husky/commit-msg
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-npx --no -- commitlint --edit ${1}
```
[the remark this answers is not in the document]
Sidu28: no fucking idea.... somethign to do with the linter?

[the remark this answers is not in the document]
wadealexc: Yeah, it's the commit linter. Could you get rid of this change? We need our commit linter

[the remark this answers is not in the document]
wadealexc: (and the package.json change)

src/contracts/pods/EigenPod.sol
```diff
@@ -124,6 +124,11 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         _;
     }
 
+    modifier partialWithdrawalProofSwitchOff() {
```
wadealexc, line 127: After you make the `require` change I mentioned in another comment, this modifier should be removed

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +432,37 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+            provenPartialWithdrawalSumWei -= sumOfPartialWithdrawalsClaimedGwei * GWEI_TO_WEI;
+            sumOfPartialWithdrawalsClaimedGwei = 0;
+            _sendETH_AsDelayedWithdrawal(podOwner, provenPartialWithdrawalSumWei);
+        } else {
+            sumOfPartialWithdrawalsClaimedGwei -= uint64(provenPartialWithdrawalSumWei / GWEI_TO_WEI);
+        }
+
+        provenPartialWithdrawalSumWei -= withdrawalCallbackInfo.fee;
+        //send proof service their fee
+        AddressUpgradeable.sendValue(payable(feeRecipient), withdrawalCallbackInfo.fee);
```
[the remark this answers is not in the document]
wadealexc: Wait actually I'm realizing that we're subtracting from `provenPartialWithdrawalSumWei` but we don't even use the result anywhere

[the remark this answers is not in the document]
wadealexc: Okay so what actually needs to happen is that `sumOfPartialWithdrawalsClaimedGwei` needs to be decreased by `provenPartialWithdrawalSumWei + fee`?

I'm unsure now but the current impl doesn't look right

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -39,6 +45,29 @@ interface IEigenPodManager is IPausable {
...
+        // lower bound of the withdrawal period
+        uint64 startTimestamp;
+        // upper bound of the withdrawal period
+        uint64 endTimestamp;
+        // amount being proven for withdrawal
+        uint256 provenPartialWithdrawalSumWei;
+        // prover fee
+        uint256 fee;
+        // user-signed maxFee
```
wadealexc, line 60: Do we not want to validate this signature in the contracts? If the proof service is compromised they can make this up entirely

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +225,38 @@ contract EigenPodManager is
...
+    }
+
+    /// @notice enables partial withdrawal proving via offchain proofs
+    function enableProofService() external onlyOwner {
+        require(!partialWithdrawalProofSwitch);
+        partialWithdrawalProofSwitch = true;
+        emit ProofServiceEnabled();
+    }
+
+    function updateProofService(address caller, address feeRecipient) external onlyOwner {
```
wadealexc, line 252: Needs natspec comment.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +225,38 @@ contract EigenPodManager is
...
+            require(callbackInfo[i].fee <= callbackInfo[i].maxFee, "EigenPod.fulfillPartialWithdrawalProofRequest: fee must be less than or equal to maxFee");
+            IEigenPod pod = ownerToPod[callbackInfo[i].podOwner];
+            require(address(pod) != address(0), "EigenPodManager.proofServiceCallback: pod does not exist");
+            pod.fulfillPartialWithdrawalProofRequest(callbackInfo[i], proofService.feeRecipient);
+        }
+    }
+
+    /// @notice enables partial withdrawal proving via offchain proofs
+    function enableProofService() external onlyOwner {
+        require(!partialWithdrawalProofSwitch);
```
wadealexc, line 247: Do we want an error message here? I'm fine without it but we do typically have one.

src/contracts/pods/EigenPodManagerStorage.sol
wadealexc: Gonna bump this request again - see other comments for rationale, but I want to avoid "switch"

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -143,4 +172,19 @@ interface IEigenPodManager is IPausable {
      * @dev Reverts if `shares` is not a whole Gwei amount
      */
     function withdrawSharesAsTokens(address podOwner, address destination, uint256 shares) external;
+
+
+    /// @notice Returns the status of the proof switch
+    function partialWithdrawalProofSwitch() external view returns (bool);
```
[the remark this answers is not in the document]
wadealexc: Actually i guess this is to mirror the public state var -- so this should be `proofServiceEnabled() -> (bool)`

wadealexc (changes requested):
Copying some stuff over from Slack -- generally not a fan of these changes. I left a lot of comments in the PR, but also have some open questions and concerns:

1. Can we also have the proof service fees go through the delayed withdrawal router? Im not a fan of how if the proof service get compromised, it's basically an instant backdoor that bypasses our withdrawal delays in every way possible

2. Can we have some sort of max fee or other upper bound as a state variable in the EPM? It's weird that this service basically gets to pass in whatever numbers it wants and we just trust it entirely

3. The google docs around this mention that the user provides a signature at some point -- can we validate that signature in the contracts somewhere? As implemented, this can be sorta forced on a user (from the contract's perspective). Is there a reason we don't want to validate this signature here?

4. There are literally only 3 tests for the new functionality, and not a single one that checks the happy path - that the user gets the withdrawal and the proof service gets their fee. Looking at that part of the impl, it doesn't seem to work (see my comments above).

src/contracts/pods/EigenPod.sol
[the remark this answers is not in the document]
Sidu28: oh I see, we're always gonna have "startTimestamp == mostRecentWithdrawalTimestamp" when we submit the request to bonsai.  So if there is any claimed partial withdrawals, the starting timestamp provided is such that it counts that

src/contracts/pods/EigenPodManager.sol
```diff
@@ -44,6 +44,16 @@ contract EigenPodManager is
         _;
     }
 
+    modifier onlyProofService() {
+        require(msg.sender == proofService.caller, "EigenPod.onlyProofService: not a permissioned fulfiller");
+        _;
+    }
+
+    modifier proofServiceEnabled() {
+        require(partialWithdrawalProofSwitch, "EigenPod.proofServiceEnabled: partial withdrawal proof switch is off");
```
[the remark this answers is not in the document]
Sidu28: agreed, fixed! I also just made these explicit require statements like you asked before

src/contracts/interfaces/IEigenPodManager.sol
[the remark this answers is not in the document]
Sidu28: removed in favor of an explicit check in the function

src/contracts/pods/EigenPod.sol
Sidu28: yup already removed in favor of an explicit check

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +432,37 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+
+    /*******************************************************************************
+                    EXTERNAL FUNCTIONS CALLABLE BY PERMISSIONED SERVICES
+    *******************************************************************************/
+
+    /// @notice Called by the EigenPodManager to fulfill a partial withdrawal proof request
+    function fulfillPartialWithdrawalProofRequest(
+        IEigenPodManager.WithdrawalCallbackInfo calldata withdrawalCallbackInfo,
+        address feeRecipient
+    ) external onlyEigenPodManager {
```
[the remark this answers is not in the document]
Sidu28: good point, added!

src/contracts/pods/EigenPod.sol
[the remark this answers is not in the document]
Sidu28: this is a tragic copy paste error, pls look at the code now

src/contracts/pods/EigenPodManager.sol
Sidu28: added

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,36 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+    /// @notice Called by the EigenPodManager to fulfill a partial withdrawal proof request
+    function fulfillPartialWithdrawalProofRequest(
+        IEigenPodManager.WithdrawalCallbackInfo calldata withdrawalCallbackInfo,
+        address feeRecipient
+    ) external onlyEigenPodManager onlyWhenNotPaused(PAUSED_EIGENPODS_VERIFY_WITHDRAWAL) {
+        uint256 provenPartialWithdrawalSumWei = withdrawalCallbackInfo.provenPartialWithdrawalSumWei;
+        provenPartialWithdrawalSumWei -= withdrawalCallbackInfo.fee;
+                //send proof service their fee
+        AddressUpgradeable.sendValue(payable(feeRecipient), withdrawalCallbackInfo.fee);
+        require(mostRecentWithdrawalTimestamp < withdrawalCallbackInfo.endTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: mostRecentWithdrawalTimestamp must precede endTimestamp");
```
wadealexc, line 444: At some point you had a check that `startTimestamp == mostRecentWithdrawalTimestamp`

Do we no longer need that?

Sidu28: I just changed "startTimestamp" to "mostRecentWithdrawalTimestamp".  Technically startTimestamp is incorrect because we're doing (startTimestamp, endTimestamp].

src/contracts/pods/EigenPod.sol
```diff
@@ -712,6 +743,10 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         address recipient,
         uint64 partialWithdrawalAmountGwei
     ) internal returns (VerifiedWithdrawal memory) {
+        require(
```
gpsanant, line 746: why do we need this?

gpsanant: won't it be enabled during the upgrade and paused via regular pausing method?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,38 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+    /*******************************************************************************
+                    EXTERNAL FUNCTIONS CALLABLE BY PERMISSIONED SERVICES
+    *******************************************************************************/
+
+    /// @notice Called by the EigenPodManager to fulfill a partial withdrawal proof request
+    function fulfillPartialWithdrawalProofRequest(
+        IEigenPodManager.Journal calldata journal,
+        uint64 feeGwei,
+        address feeRecipient
+    ) external onlyEigenPodManager onlyWhenNotPaused(PAUSED_EIGENPODS_VERIFY_WITHDRAWAL) {
```
gpsanant, line 440: can we create another pause index for zk pws?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,38 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+
+        require(journal.mostRecentWithdrawalTimestamp == mostRecentWithdrawalTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: proven mostRecentWithdrawalTimestamp must match mostRecentWithdrawalTimestamp in the EigenPod");
+        require(mostRecentWithdrawalTimestamp < journal.endTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: mostRecentWithdrawalTimestamp must precede endTimestamp");
```
gpsanant, line 443: already checked in snark?

gpsanant: also should make it 

`journal.mostRecentWithdrawalTimestamp < journal.endTimestamp`

if you want to keep it for sanity

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,38 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+
+
```
gpsanant, line 445: image

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,38 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        uint256 provenPartialWithdrawalSumGwei = journal.provenPartialWithdrawalSumGwei;
+        require(sumOfPartialWithdrawalsClaimedGwei <= provenPartialWithdrawalSumGwei - feeGwei, "EigenPod.fulfillPartialWithdrawalProofRequest: sumOfPartialWithdrawalsClaimedGwei must be less than or equal to provenPartialWithdrawalSumGwei + feeGwei");
```
gpsanant, line 447: `sumOfPartialWithdrawalsClaimedGwei + feeGwei <= provenPartialWithdrawalSumGwei`

to avoid underflow

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,38 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        IEigenPodManager.Journal calldata journal,
+        uint64 feeGwei,
+        address feeRecipient
+    ) external onlyEigenPodManager onlyWhenNotPaused(PAUSED_EIGENPODS_VERIFY_WITHDRAWAL) {
+
+        require(journal.mostRecentWithdrawalTimestamp == mostRecentWithdrawalTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: proven mostRecentWithdrawalTimestamp must match mostRecentWithdrawalTimestamp in the EigenPod");
+        require(mostRecentWithdrawalTimestamp < journal.endTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: mostRecentWithdrawalTimestamp must precede endTimestamp");
+
+
+        uint256 provenPartialWithdrawalSumGwei = journal.provenPartialWithdrawalSumGwei;
```
gpsanant, line 446: no need for this extra mstore?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,38 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        require(sumOfPartialWithdrawalsClaimedGwei <= provenPartialWithdrawalSumGwei - feeGwei, "EigenPod.fulfillPartialWithdrawalProofRequest: sumOfPartialWithdrawalsClaimedGwei must be less than or equal to provenPartialWithdrawalSumGwei + feeGwei");
+        provenPartialWithdrawalSumGwei -= feeGwei;
+
+
+        // subtract an partial withdrawals that may have been claimed via merkle proofs
+        provenPartialWithdrawalSumGwei -= sumOfPartialWithdrawalsClaimedGwei;
```
gpsanant, line 452: rename `sumOfPartialWithdrawalsClaimedGwei` to `sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei` or someth for clarity pls

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,38 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        sumOfPartialWithdrawalsClaimedGwei = 0;
+        _sendETH_AsDelayedWithdrawal(podOwner, provenPartialWithdrawalSumGwei);
+    
+        mostRecentWithdrawalTimestamp = journal.endTimestamp;
```
gpsanant, line 456: comment

also rename to `withdrawalsProvenUntilTimestamp` or similar

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +224,53 @@ contract EigenPodManager is
         ownerToPod[podOwner].withdrawRestakedBeaconChainETH(destination, shares);
     }
 
+    /// @notice Called by proving service to fulfill partial withdrawal proof requests
+    function proofServiceCallback(
+        WithdrawalCallbackInfo[] calldata callbackInfo
+    ) external onlyProofService nonReentrant {
+        require(proofServiceEnabled, "EigenPodManager.proofServiceCallback: offchain partial withdrawal proofs are not enabled");
+        for(uint256 i = 0; i < callbackInfo.length; i++) {
```
gpsanant, line 232: wait what.

we should be verifying a single proof that proves the pws for many pods against the same block root.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +224,53 @@ contract EigenPodManager is
...
+
+    /// @notice enables partial withdrawal proving via offchain proofs
+    function enableProofService() external onlyOwner {
+        require(!proofServiceEnabled, "EigenPodManager.enableProofService: proof service already enabled");
+        proofServiceEnabled = true;
+        emit ProofServiceEnabled();
+    }
+
+    /// @notice changes the proof service related information
+    function updateProofService(address caller, address feeRecipient, address verifier) external onlyOwner {
```
gpsanant, line 265: just pass a proof service in as param

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +224,53 @@ contract EigenPodManager is
...
+        }
+    }
+
+    function parsePayload(bytes calldata payload) public returns(bytes32, Journal memory, bytes memory){
+       (Journal memory journal, bytes32 imageId) = abi.decode(payload, (Journal, bytes32));
+        return (imageId, journal, payload[0:payload.length - 32]);
+    }
+
+    /// @notice enables partial withdrawal proving via offchain proofs
+    function enableProofService() external onlyOwner {
```
gpsanant, line 258: this should just happen when we upgrade no?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,38 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         _sendETH(recipient, amountWei);
     }
 
+
```
gpsanant, line 430: we should disable pw merkle proofs

src/contracts/pods/EigenPodManager.sol
```diff
@@ -226,34 +224,37 @@ contract EigenPodManager is
...
+        
+        for (uint256 i = 0; i < journal.podOwners.length; i++) {
 
-            require(journal.blockRoot == getBlockRootAtTimestamp(callbackInfo[i].oracleTimestamp), "EigenPodManager.proofServiceCallback: block root does not match oracleRoot for that timestamp");
-            
             // these checks are verified in the snark, we add them here again as a sanity check
-            require(callbackInfo[i].oracleTimestamp >= journal.endTimestamp, "EigenPodManager.proofServiceCallback: oracle timestamp must be greater than or equal to callback timestamp");
-            require(callbackInfo[i].feeGwei <= journal.maxFeeGwei, "EigenPodManager.proofServiceCallback: fee must be less than or equal to maxFee");
+            require(callbackInfo.oracleTimestamp >= journal.endTimestamps[i], "EigenPodManager.proofServiceCallback: oracle timestamp must be greater than or equal to callback timestamp");
+            require(callbackInfo.feeGwei <= journal.maxFeesGwei[i], "EigenPodManager.proofServiceCallback: fee must be less than or equal to maxFee");
```
gpsanant, line 240: super confused, how can all of the journals have the same fee  and oracle timestamp? the callback struct needs its internal fields arrayified

src/contracts/pods/EigenPodManager.sol
```diff
@@ -226,34 +224,37 @@ contract EigenPodManager is
...
+            IEigenPod pod = ownerToPod[journal.podOwners[i]];
             require(address(pod) != address(0), "EigenPodManager.proofServiceCallback: pod does not exist");
-            require(address(pod) == journal.podAddress, "EigenPodManager.proofServiceCallback: pod address does not match");
+            require(address(pod) == journal.podAddresses[i], "EigenPodManager.proofServiceCallback: pod address does not match");
+
+            IEigenPod.VerifiedPartialWithdrawal memory partialWithdrawal = IEigenPod.VerifiedPartialWithdrawal({
+                provenPartialWithdrawalSumGwei:  journal.provenPartialWithdrawalSumsGwei[i],
+                mostRecentWithdrawalTimestamp: journal.mostRecentWithdrawalTimestamps[i],
+                endTimestamp: journal.endTimestamps[i]
+
```
gpsanant, line 251: space

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -57,25 +57,27 @@ interface IEigenPodManager is IPausable {
...
+        address[] podAddresses;
         // the address of the pod owner
-        address podOwner;
+        address[] podOwners;
         // the latest timestamp proven until
-        uint64 mostRecentWithdrawalTimestamp;
+        uint64[] mostRecentWithdrawalTimestamps;
         // upper bound of the withdrawal period
-        uint64 endTimestamp;
+        uint64[] endTimestamps;
```
gpsanant, line 78: another approach could be to have a consistent end timestamp for all proofs determined by backend (i.e. next end timestamp is every 12 hrs). ask @taekyunggg about the UX of each

src/contracts/pods/EigenPod.sol
```diff
@@ -754,7 +753,7 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
             partialWithdrawalAmountGwei
         );
 
-        sumOfPartialWithdrawalsClaimedGwei += partialWithdrawalAmountGwei;
+        sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei += partialWithdrawalAmountGwei;
```
gpsanant, line 756: this function should be removed, partial withdrawal processing in general outside of zk. if you think otherwise, can you write something for discussion?

src/contracts/pods/EigenPod.sol
```diff
@@ -434,26 +434,25 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        require(sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei + feeGwei <= verifiedPartialWithdrawal.provenPartialWithdrawalSumGwei, "EigenPod.fulfillPartialWithdrawalProofRequest: proven sum must be less than or equal to provenPartialWithdrawalSumGwei + feeGwei");
+        uint64 provenPartialWithdrawalSumGwei = verifiedPartialWithdrawal.provenPartialWithdrawalSumGwei;
         provenPartialWithdrawalSumGwei -= feeGwei;
 
 
         // subtract an partial withdrawals that may have been claimed via merkle proofs
-        provenPartialWithdrawalSumGwei -= sumOfPartialWithdrawalsClaimedGwei;
-        sumOfPartialWithdrawalsClaimedGwei = 0;
+        provenPartialWithdrawalSumGwei -= sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei;
+        sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei = 0;
```
gpsanant, line 452: we should only be doing this logic if `sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei != 0`

src/contracts/interfaces/IEigenPodManager.sol
gpsanant: oops nvm this doesn't work because they don't know how to set the min proven sum in this case

src/contracts/pods/EigenPod.sol
Sidu28: we check journal.mostRecentWithdrawalTimestamp == mostRecentWithdrawalTimestamp right before that........

src/contracts/pods/EigenPod.sol
Sidu28: I mean wouldn't really help either way......

src/contracts/pods/EigenPodManager.sol
Sidu28: thats incorrect, fixed

src/contracts/pods/EigenPod.sol
Sidu28: ah I removed that logic from before for some reason smh, added back

src/contracts/pods/EigenPod.sol
Sidu28: added

src/contracts/pods/EigenPod.sol
Sidu28: Sure where do you want me to write this?  I think its pretty simple.  In the event Risc0 goes down, we jjust have to make a very simple transaction (can be done via etherscan) vs an upgrade.  I agree having non functional code in the contract isn't good but having Risc0 go down and being able to react really quick I think is important too

src/contracts/pods/EigenPod.sol
```diff
@@ -437,20 +437,24 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         IEigenPod.VerifiedPartialWithdrawal calldata verifiedPartialWithdrawal,
```
gpsanant, line 437: would name this struct something different... like VerifiedPartialWithdrawalBatch?

src/contracts/pods/EigenPod.sol
```diff
@@ -437,20 +437,24 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        require(sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei <= verifiedPartialWithdrawal.provenPartialWithdrawalSumGwei - feeGwei, "EigenPod.fulfillPartialWithdrawalProofRequest: proven sum must be less than or equal to provenPartialWithdrawalSumGwei + feeGwei");
         uint64 provenPartialWithdrawalSumGwei = verifiedPartialWithdrawal.provenPartialWithdrawalSumGwei;
         provenPartialWithdrawalSumGwei -= feeGwei;
 
 
         // subtract an partial withdrawals that may have been claimed via merkle proofs
-        provenPartialWithdrawalSumGwei -= sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei;
-        sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei = 0;
-        _sendETH_AsDelayedWithdrawal(podOwner, provenPartialWithdrawalSumGwei);
+        if(provenPartialWithdrawalSumGwei > sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei){
```
gpsanant, line 451: doesnt this need to include the fee?

src/contracts/pods/EigenPod.sol
```diff
@@ -437,20 +437,24 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+            provenPartialWithdrawalSumGwei -= sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei;
```
gpsanant, line 452: again, a lot of this is conditional on `sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei == 0`, we should condition on that IMO to save gas

.husky/commit-msg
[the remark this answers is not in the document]
ChaoticWalrus: bump on this @Sidu28

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,40 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+    function fulfillPartialWithdrawalProofRequest(
+        IEigenPod.VerifiedPartialWithdrawalBatch calldata verifiedPartialWithdrawalBatch,
+        uint64 feeGwei,
+        address feeRecipient
+    ) external onlyEigenPodManager onlyWhenNotPaused(PAUSED_EIGENPODS_FULFILL_PARTIAL_WITHDRAWAL_PROOF_REQUEST) {
+
+        require(verifiedPartialWithdrawalBatch.mostRecentWithdrawalTimestamp == mostRecentWithdrawalTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: proven mostRecentWithdrawalTimestamp must match mostRecentWithdrawalTimestamp in the EigenPod");
+        require(mostRecentWithdrawalTimestamp < verifiedPartialWithdrawalBatch.endTimestamp, "EigenPod.fulfillPartialWithdrawalProofRequest: mostRecentWithdrawalTimestamp must precede endTimestamp");
+
+        require(sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei <= verifiedPartialWithdrawalBatch.provenPartialWithdrawalSumGwei - feeGwei, "EigenPod.fulfillPartialWithdrawalProofRequest: proven sum must be less than or equal to provenPartialWithdrawalSumGwei + feeGwei");
```
ChaoticWalrus, line 445: presumably your revert message is wrong here. seems you're using `provenPartialWithdrawalSumGwei` instead of `sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei` in it ?

ChaoticWalrus: also -- is this require statement necessary at all? seems...overly strict?

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -49,6 +49,15 @@ interface IEigenPod {
         int256 sharesDeltaGwei;
     }
 
+    struct VerifiedPartialWithdrawalBatch{
+        // amount being proven for withdrawal
+        uint64 provenPartialWithdrawalSumGwei;
+        // the latest timestamp proven until
+        uint64 mostRecentWithdrawalTimestamp;
+        // upper bound of the withdrawal period
+        uint64 endTimestamp;
```
ChaoticWalrus, line 58: these 2 fields seem redundant, and I think(?) I'm not seeing the `mostRecentWithdrawalTimestamp` field being used in the EigenPod contract, at least. Please correct me if I'm wrong

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,40 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+            provenPartialWithdrawalSumGwei -= sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei;
+            sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei = 0;
+            _sendETH_AsDelayedWithdrawal(podOwner, provenPartialWithdrawalSumGwei);
+        } else {
+            sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei -= provenPartialWithdrawalSumGwei;
+        }
+    
+        mostRecentWithdrawalTimestamp = verifiedPartialWithdrawalBatch.endTimestamp;
+
+        provenPartialWithdrawalSumGwei -= feeGwei;
```
ChaoticWalrus, line 459: *potential bug* shouldn't this happen before the above? I think you are double-dipping if this line in particular isn't before the if-else block

src/contracts/pods/EigenPod.sol
ChaoticWalrus: to elaborate, it seems like you only need to check that `verifiedPartialWithdrawalBatch.provenPartialWithdrawalSumGwei > feeGwei` ?

src/contracts/pods/EigenPod.sol
```diff
@@ -428,6 +427,40 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        // subtract an partial withdrawals that may have been claimed via merkle proofs
+        if(provenPartialWithdrawalSumGwei > sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei && sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei > 0){
+            provenPartialWithdrawalSumGwei -= sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei;
+            sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei = 0;
+            _sendETH_AsDelayedWithdrawal(podOwner, provenPartialWithdrawalSumGwei);
+        } else {
+            sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei -= provenPartialWithdrawalSumGwei;
+        }
+    
+        mostRecentWithdrawalTimestamp = verifiedPartialWithdrawalBatch.endTimestamp;
```
ChaoticWalrus, line 457: also seems proper to move this above the if-else block, to better observe checks-effects-interactions style

src/contracts/pods/EigenPod.sol
ChaoticWalrus: bump on this @Sidu28 
I had the same questions.

src/contracts/pods/EigenPod.sol
ChaoticWalrus: I do think leaving this code in reduces the trust in the system, at least in some cases.
I can see the potential for bugs in the interactions between this code and the new zk-batched proving code, but it does seem like we at least have to deal with _most_ of these issues simply due to the upgrade happening at all.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -215,6 +222,54 @@ contract EigenPodManager is
         ownerToPod[podOwner].withdrawRestakedBeaconChainETH(destination, shares);
     }
 
+    /// @notice Called by proving service to fulfill partial withdrawal proof requests
+    function proofServiceCallback(
+        WithdrawalCallbackInfo calldata callbackInfo
+    ) external onlyProofService nonReentrant {
+        require(proofServiceEnabled, "EigenPodManager.proofServiceCallback: offchain partial withdrawal proofs are not enabled");
```
ChaoticWalrus, line 229: again seems like this should probably be just another pausing flag to me.

I think in this function is the more appropriate place to check this parameter, as it is at the start of the chain of calls.

src/contracts/pods/EigenPodPausingConstants.sol
```diff
@@ -21,4 +21,6 @@ abstract contract EigenPodPausingConstants {
     uint8 internal constant PAUSED_EIGENPODS_VERIFY_BALANCE_UPDATE = 3;
     /// @notice Index for flag that pauses the `verifyBeaconChainFullWithdrawal` function *of the EigenPods* when set. see EigenPod code for details.
     uint8 internal constant PAUSED_EIGENPODS_VERIFY_WITHDRAWAL = 4;
+    /// @notice Index for flag that pauses `fulfillPartialWithdrawalProofRequest` function *of the EigenPods* when set. see EigenPod code for details.
+    uint8 internal constant PAUSED_EIGENPODS_FULFILL_PARTIAL_WITHDRAWAL_PROOF_REQUEST = 5;
```
ChaoticWalrus, line 25: feels like it might be more appropriate to pause at the proofServiceCallback level?
I think doing so would be the same, but just cause calls to revert earlier when things are paused.
Feeling like we can eliminate the `proofServiceEnabled` flag then as well.

src/test/unit/EigenPodManagerUnit.t.sol
```diff
@@ -550,3 +562,104 @@ contract EigenPodManagerUnitTests_ShareAdjustmentCalculationTests is EigenPodMan
         assertEq(sharesDelta, sharesAfter - sharesBefore, "Shares delta must be equal to the difference between sharesAfter and sharesBefore");
     }
 }
+
+contract EigenPodManagerUnitTests_OffchainProofGenerationTests is EigenPodManagerUnitTests {
```
ChaoticWalrus, line 566: the tests you've added so far look good, but I'm only seeing coverage for the "unhappy" paths (revert cases) and nothing for the "happy" paths (non-reverting cases)

ChaoticWalrus (changes requested):
The bones of this look good; thanks for taking it on.
I feel like there might be a bug to fix, think some cleanup can be done, and I'd like to see coverage added for the non-reverting cases, but it's headed in the right direction 👍

src/contracts/interfaces/IEigenPod.sol
Sidu28: check line 442 - we're basically checking that the offchain proven mostRecentWithdrawalTimestamp == mostRecentWithdrawalTimestamp in the contract

src/contracts/pods/EigenPod.sol
Sidu28: Ah yeah this is missing a check. 
Initially, I had it before the if/else statement.  This was my original design:  I imagined that risc0 always gets its fee first and we do not subtract it from "proven amounts".  So essentially it would take longer for the podowner to "pay off" the proven sum.  

G seems to disagree here, ie, if there is already a merkle proven balance, risc0 may not get a fee for proving it (probably a one time thing).

Sidu28: I've fixed it I think, take a look at it.  I think I still like the idea of separating out the fee before the if/else block

src/contracts/pods/EigenPodManager.sol
Sidu28: We check it in _processPartialWithdrawal() which is called in _verifyAndProcessWithdrawal()l, which is in turn called in verifyAndProcessWithdrawals() which is called directly from the pod  - that's why I added it as a separate check there.

src/contracts/pods/EigenPodPausingConstants.sol
Sidu28: Yeah good point, I will add it as a flag at the EPM level - I think Im misunderstanding your point about proofServiceEnabled checks in the EigenPod (see my response above).

src/contracts/pods/EigenPod.sol
Sidu28: oh ur right, fixed

Sidu28: fixed, ur right!

src/contracts/interfaces/IEigenPod.sol
ChaoticWalrus: ahhhhhh OK nevermind then, resolving.

src/contracts/pods/EigenPod.sol
ChaoticWalrus: I think you may have misinterpreted my recommendation.
All I wanted to do was move the line that deducts the fee from the proven amount to above the if-else block, as then the fee would be paid to Risc0 correctly _in all cases_ and your accounting would be fixed, since it was essentially just not accounting for the fee before (i.e. it was giving the "fee" amount _both_ to the user _and_ to Risc0).
The current version you have definitely looks wrong to me. why would you subtract from `sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei` before removing the fee amount from `provenPartialWithdrawalSumGwei`?
will follow up with a comment of exactly what I'm suggesting.

src/contracts/pods/EigenPodPausingConstants.sol
ChaoticWalrus: I think I was viewing the `proofServiceEnabled` checks and the way you're using this `PAUSED_EIGENPODS_FULFILL_PARTIAL_WITHDRAWAL_PROOF_REQUEST` flag as redundant.
Perhaps I read wrong, but it seems like:
a user calls the proof service,
which calls `EigenPodManager.proofServiceCallback`,
which checks the `proofServiceEnabled` flag (AND  now checks the (redundant?) `PAUSED_EIGENPODS_FULFILL_PARTIAL_WITHDRAWAL_PROOF_REQUEST` flag),
which calls `EigenPod.fulfillPartialWithdrawalProofRequest`, (which previously checked the (redundant?) `PAUSED_EIGENPODS_FULFILL_PARTIAL_WITHDRAWAL_PROOF_REQUEST` flag)

src/contracts/pods/EigenPod.sol
Sidu28: This was my initial design [exactly](https://github.com/Layr-Labs/eigenlayer-contracts/pull/346/commits/eb5c18e1cec4e1c6de7732d840d3800ac17afd8e#diff-bb16ee06e48bbb1b75ec8e08586f88e0a73765c73d98b51060096657174d235f).  G's design suggestion that we subtract the fee after, thus the current implementation: https://github.com/Layr-Labs/eigenlayer-contracts/pull/346#discussion_r1440670548

src/contracts/pods/EigenPodPausingConstants.sol
ChaoticWalrus: I see what you're saying about pausing Merkle proof-based partial withdrawals now though! but I think if you want that functionality to remain, it should be paused with a separate flag.
Definitely feels like either we should leave the functionality in and design a system where both proof types can be unpaused/enabled simultaneously (which I believe the accounting model I suggested satisfies) OR just get rid of Merkle-proofs in this upgrade.
The `proofServiceEnabled` flag current doubles as a `partialWithdrawalsViaMerkleProofsDisabled` flag, which makes the whole thing effectively just an "upgrade switch" -- I feel like if we're doing that, we might as well just not have the flag or Merkle based partial withdrawal proofs at all!

src/contracts/pods/EigenPod.sol
ChaoticWalrus: not to get too pedantic about what the code looked like before, but perhaps G was trying to point out the same bug as I did, as it appears it [was present at the time](https://github.com/Layr-Labs/eigenlayer-contracts/blob/eb5c18e1cec4e1c6de7732d840d3800ac17afd8e/src/contracts/pods/EigenPod.sol#L449-L461) as well.

I feel very confident that the present form is incorrect, and believe my suggestion is the simplest way to keep this accounting consistent + follow proper code conventions like CEI

Sidu28: ok wait I think I'm in agreement with you here - G's suggestion is that in the check, remainingSum should include the fee, i.e., don't subtract it before.  Here's the state of the file when he made the [comment](https://github.com/Layr-Labs/eigenlayer-contracts/pull/346/files/8ea334d74deaf9152aec11a039f8ce22c9fe9f4c..4e1663c7233f5eebbb072e78417b4c34e0e09f66#diff-bb16ee06e48bbb1b75ec8e08586f88e0a73765c73d98b51060096657174d235f). Sorry if im completely muddling this but this seems to be what you're suggesting now?  I think either way the accounting wouldn't be necessarily correct its just whether or not we want to pay risc0 their fee from for that very first proof where sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei is nonzero.  After that sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei is always 0 and so this shouldn't be an issue at all

src/test/unit/EigenPodManagerUnit.t.sol
Sidu28: added some happy path cases!

src/contracts/pods/EigenPod.sol
ChaoticWalrus: ahhh OK.
I would think Risc0 would want to get paid regardless, as they are doing work/computation here -- doesn't really feel fair for them to not pay them, tbh.

Ah, but I guess there's a risk where the funds to pay the fee aren't actually in the EigenPod in the event that the proven amount doesn't exceed `sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei` ?
was that a worry before / reason for the require check specifically checking that the proven amount exceeds `sumOfPartialWithdrawalsClaimedViaMerkleProvenGwei + feeGwei`? Because now that check actually makes sense to me lol -- I hadn't considered this risk before.
perhaps we could just add a check that `address(this).balance >= feeGwei` ?
then I think we are OK in all instances, the system seems flexible, and Risc0 gets paid any time a proof is submitted successfully (which seems proper to me, at least).

ChaoticWalrus:
@Sidu28 if we leave in the ability to prove partial withdrawals via Merkle proofs, then I believe the `_processPartialWithdrawal` function needs an added check that the withdrawalTimestamp is *after* the the `mostRecentWithdrawalTimestamp`, as otherwise (as long as if either Merkle proofs and zk proofs are possible at the same time, or a transition is made Merkle->zk->Merkle) a user can zk prove a withdrawal and then Merkle-proof it afterwords, to "double dip".

This is definitely also a mark in favor of just fully removing Merkle-proving of partial withdrawals in this upgrade (as it's an example of the added complexity / potential dangers of having both proof methods active).

Sidu28:
> @Sidu28 if we leave in the ability to prove partial withdrawals via Merkle proofs, then I believe the `_processPartialWithdrawal` function needs an added check that the withdrawalTimestamp is _after_ the the `mostRecentWithdrawalTimestamp`, as otherwise (as long as if either Merkle proofs and zk proofs are possible at the same time, or a transition is made Merkle->zk->Merkle) a user can zk prove a withdrawal and then Merkle-proof it afterwords, to "double dip".
> 
> This is definitely also a mark in favor of just fully removing Merkle-proving of partial withdrawals in this upgrade (as it's an example of the added complexity / potential dangers of having both proof methods active).

Discussed offline! We both agree that the current check being made [here](https://github.com/Layr-Labs/eigenlayer-contracts/blob/8d5058c9c08c8d98126cf9a6c4a9d6667a8b962c/src/contracts/pods/EigenPod.sol#L583) should suffice in allowing a switch back to merkle proving.

ChaoticWalrus:
Closing this PR as stale.

PR #338, opened 2023-11-18
added read functions for eigenpods

Sidu28:
Making this PR based on requests to access these functions.  Since the pubkey hash is not simply sha256(pubkey) but rather sha256(pubkey + bytes(16)), adding these would make it easier for users to call these functions with data they have on hand. Specifically, Figment requested this.

Note:  maybe worht just deleting the other functions that take in the pubkey hash if we dont use them actively for the FE?

PR #348, opened 2023-11-29
Add Read Functions to M2 Mainnet

Sidu28:
Adding some read functions to EP contract that take in bls pubkey as input not pubkeyhash.  Duplicate of this PR but into m2-mainnet instead of master: https://github.com/Layr-Labs/eigenlayer-contracts/pull/338


ChaoticWalrus (approved):
This is recreating a PR but based off `m2-mainnet` instead of `master`.
LGTM
Only question is if we should expose the helper for calculating the validator pubkey hash? Feels fine to leave it, just hadn't occurred to me until now.

PR #350, opened 2023-11-30
Integration Tests: Scenarios 2-7

ypatil12:


src/test/integration/IntegrationBase.t.sol
```diff
@@ -122,19 +122,41 @@ abstract contract IntegrationBase is IntegrationDeployer {
 
     function assert_AllWithdrawalsPending(bytes32[] memory withdrawalRoots, string memory err) internal {
         for (uint i = 0; i < withdrawalRoots.length; i++) {
-            assertTrue(delegationManager.pendingWithdrawals(withdrawalRoots[i]), err);
+            assert_withdrawalPending(withdrawalRoots[i], err);
```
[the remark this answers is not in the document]
wadealexc: (nit applies to each helper method below)

src/test/integration/User.t.sol
```diff
@@ -123,6 +123,16 @@ contract User is Test {
...
 
+    /// @dev Undelegate from operator
+    function undelegate() public createSnapshot virtual returns(bytes32){
+        return delegationManager.undelegate(address(this));
+    }
+
+    /// @dev Force undelegate staker
+    function undelegate(User staker) public createSnapshot virtual returns(bytes32){
+        return delegationManager.undelegate(address(staker));
+    }
```
wadealexc, line 134: Could we rename this `forceUndelegate`?

It matches the dmgr interface less, but I think it's clearer at the test level

src/test/integration/tests/Deposit_Delegate_Queue_Complete.t.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity =0.8.12;
 import "src/test/integration/IntegrationBase.t.sol";
 import "src/test/integration/User.t.sol";
 
-contract Deposit_Delegate_Queue_Complete is IntegrationBase {
+contract Integration_Deposit_Delegate_Queue_Complete is IntegrationBase {
```
wadealexc, line 7: I'm fine with making this change, but had avoided adding anything extra because I forsee these names getting very long

Although I will say it's nice to be able to `forge test --match-contract Integration` to run all the integration tests.

src/test/integration/tests/Deposit_Delegate_Redelegate_Complete.t.sol
```diff
@@ -0,0 +1,112 @@
...
+import "src/test/integration/IntegrationBase.t.sol";
+import "src/test/integration/User.t.sol";
+import "src/test/integration/tests/utils.t.sol";
+
+contract Integration_Deposit_Delegate_Redelegate_Complete is IntegrationTestUtils {
+    /// Randomly generates a user with different held assets. Then:
+    /// 1. deposit into strategy
+    /// 2. delegate to an operator
+    /// 3. undelegates from the operator
+    /// 4. complete queued withdrawal as shares
```
wadealexc, line 13: comment: "as tokens"

src/test/integration/tests/Deposit_Delegate_Redelegate_Complete.t.sol
```diff
@@ -0,0 +1,112 @@
...
+        assertDepositState(staker, strategies, shares);
+
+        // 2. Delegate to an operator
+        staker.delegateTo(operator1);
+        assertDelegationState(staker, operator1, strategies, shares);
+
+        // 3. Undelegate from an operator
+        IDelegationManager.Withdrawal memory expectedWithdrawal = _getExpectedWithdrawalStruct(staker);
+        bytes32 withdrawalRoot = staker.undelegate();
+        assertUndelegateState(staker, operator1, expectedWithdrawal, withdrawalRoot, strategies, shares);
```
wadealexc, line 54: What I've been doing in my version is that `staker.queueWithdrawal(strategies, shares)` returns an array of these "expected withdrawal" structs, rather than the withdrawal roots.

I think this will be more flexible for future tests, because if we want `ALT_METHODS` to use the plural versions of `queueWithdrawal` and `undelegate` under the hood, we'll need to make sure our tests expect more than one withdrawal from these methods.

Also, I like creating the expected withdrawal struct in the `User` contract, because the user is the one actually performing the withdrawal and if they end up performing multiple withdrawals, they need to be able to increment each subsequent withdrawal's nonce - something that gets messier if you're calculating the expected withdrawal separately.

src/test/integration/User.t.sol
```diff
@@ -123,6 +123,16 @@ contract User is Test {
         delegationManager.delegateTo(address(operator), emptySig, bytes32(0));
     }
 
+    /// @dev Undelegate from operator
+    function undelegate() public createSnapshot virtual returns(bytes32){
+        return delegationManager.undelegate(address(this));
+    }
```
wadealexc, line 129: Mentioned this in another comment, but I'd recommend we have these return `Withdrawal[] memory`

We can query the user's current strategies/shares to figure out what the withdrawal would look like -- i think `delegationManager.getDelegatableShares` should work

src/test/integration/tests/Deposit_Delegate_Redelegate_Complete.t.sol
wadealexc: nevermind, this is indeed shares

src/test/integration/tests/Deposit_Delegate_Redelegate_Complete.t.sol
```diff
@@ -0,0 +1,112 @@
...
+
+        // 4. Complete withdrawal as shares
+        // Fast forward to when we can complete the withdrawal
+        cheats.roll(block.number + delegationManager.withdrawalDelayBlocks());
+        staker.completeQueuedWithdrawal(expectedWithdrawal, false);
```
wadealexc, line 59: The more I see `completeQueuedWithdrawal(x, bool)`, the more I wonder if (for clarity) we should just split this into:
* `User.completeWithdrawalAsShares`
* `User.completeWithdrawalAsTokens`

wadealexc:
I love the `assertXState` methods in your utils file. Bundling the assertions is much cleaner and we should do that from now on. I do want to figure out the naming conventions though, because right now it's not immediately apparent that this is different from one of the base-level assertions. My idea:

1. `assert_Statement` and `assert_Snap_Delta_Thing`

These are basic assertions that check one condition or state variable. e.g. `assert_HasNoDelegatableShares` should specifically check only the user's delegatable shares. We know it's not doing anything complicated and so we can use it all over the place comfortably.

2. `check_Action_State`:

These are the bundled assertions you're using in the utils file. They contain a bunch of the basic assertions mentioned above and are specific to one of the high-level user actions we're testing. e.g. `check_Deposit_State` contains a handful of post-deposit checks.

Could also shorten to just `check_Deposit` or something.

wadealexc:
One more thing - I would like to move the bundled assertions outside of the `/tests` folder and up into the base folder. Maybe `IntegrationChecks.t.sol` or something?

src/test/integration/tests/Deposit_Delegate_Queue_Complete.t.sol
wadealexc: after some consideration i like the renaming for the reason i mentioned above :smile:

src/test/integration/IntegrationBase.t.sol
```diff
@@ -367,4 +439,21 @@ abstract contract IntegrationBase is IntegrationDeployer {
 
         return balances;
     }
+
+    function _getPrevStrategyShares(IStrategy[] memory strategies) internal timewarp() returns (uint256[] memory) {
+        return _getStrategyShares(strategies);
+    }
+
+    function _getStrategyShares(IStrategy[] memory strategies) internal view returns (uint256[] memory) {
```
ChaoticWalrus, line 447: probably want to put 'Total' in the name here somewhere?

src/test/integration/tests/Deposit_Delegate_Redelegate_Complete.t.sol
ChaoticWalrus: another option could be to use the named param syntax
`staker.completeQueuedWithdrawal(withdrawal: expectedWithdrawal, withdrawAsTokens: false);`
but I think separate methods would be clearer + easier to be consistent with.

ChaoticWalrus (approved):
This is 🔥 🔥 🔥 
Agreed with Alex's suggestions, maybe added a couple of my own, but I'm mostly just stoked to see this coming together.
The new `assert_...` methods feel nice and clean in particular

src/test/integration/IntegrationBase.t.sol
```diff
@@ -225,6 +247,17 @@ abstract contract IntegrationBase is IntegrationDeployer {
         assertEq(prevQueuedWithdrawals + withdrawals.length, curQueuedWithdrawals, err);
     }
 
+    function assert_Snap_IncrementQueuedWithdrawals(
+        User staker,
+        string memory err
+    ) internal {
```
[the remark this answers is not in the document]
ypatil12: Kept increment for increasing nonces, but changed rest

src/test/integration/tests/Deposit_Delegate_Redelegate_Complete.t.sol
ypatil12: Added separate methods

wadealexc (approved):
Amazing work, no more notes! Approved.

PR #354, opened 2023-12-01
feat: enable AVS update metadata uri

bowenli86:
enable AVS to update their metadata uri by adding `function updateAVSMetadataURI` and `event AVSMetadataURIUpdated`

borrowed many from Yash's draft, and further polished

i'm not sure what unit tests to add, as there seems not much to test. If there is any, can you advise?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ broadcast
 
 # Deployment tools
 /data
+.idea/
 
 # Certora Outputs
 .certora_internal/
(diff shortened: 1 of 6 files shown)


ChaoticWalrus:
> i'm not sure what unit tests to add, as there seems not much to test. If there is any, can you advise?

we can add a really simple unit test, let me link you to the operator URI update one, as I would think a check for the event emission could make sense:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/b91a743b9099762ef190ef855eac059df04e9dcf/src/test/unit/DelegationUnit.t.sol#L572-L583

bowenli86:
test added and passed, will merge

very good experience of my first solidity PR. thanks @ChaoticWalrus @ypatil12 !

PR #355, opened 2023-12-01
Feat: Remove set withdrawal delay

8sunyuan:
Description:
Removes the ability to modify `delegation.withdrawDelayBlocks()` and only settable in `initialize` as the BLSSignatureChecker in middleware depends on this withdrawal window and modifying `withdrawDelayBlocks` could make previous invalid signatures valid. Since we don't have plans on modifying this value the plan is to just remove the setter entirely.

Changes:
- `Delegation.initialize` and `Delegation._initializeWithdrawalDelayBlocks`
- Fixed all scripts and compile errors
- added a regression test for initializing <= MAX_WITHDRAWAL_DELAY_BLOCKS

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -758,6 +758,7 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
             _withdrawalDelayBlocks <= MAX_WITHDRAWAL_DELAY_BLOCKS,
             "DelegationManager._initializeWithdrawalDelayBlocks: _withdrawalDelayBlocks cannot be > MAX_WITHDRAWAL_DELAY_BLOCKS"
         );
+        emit WithdrawalDelayBlocksSet(withdrawalDelayBlocks, _withdrawalDelayBlocks);
         withdrawalDelayBlocks = _withdrawalDelayBlocks;
     }
 
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -65,15 +65,18 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
 
     /**
      * @dev Initializes the addresses of the initial owner, pauser registry, and paused status.
+     * withdrawalDelayBlocks is set only once here
      */
     function initialize(
         address initialOwner,
         IPauserRegistry _pauserRegistry,
-        uint256 initialPausedStatus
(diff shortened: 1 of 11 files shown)


src/contracts/core/DelegationManager.sol
```diff
@@ -763,6 +753,14 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         }
     }
 
+    function _initializeWithdrawalDelayBlocks(uint256 _withdrawalDelayBlocks) internal {
+        require(
+            _withdrawalDelayBlocks <= MAX_WITHDRAWAL_DELAY_BLOCKS,
+            "DelegationManager._initializeWithdrawalDelayBlocks: _withdrawalDelayBlocks cannot be > MAX_WITHDRAWAL_DELAY_BLOCKS"
+        );
+        withdrawalDelayBlocks = _withdrawalDelayBlocks;
+    }
```
ChaoticWalrus, line 762: would be good to still emit the `WithdrawalDelayBlocksSet` event, I think?
feels like we're not committing to this variable never changing, just that it changing it will _require_ an upgrade now

src/test/unit/DelegationUnit.t.sol
```diff
@@ -43,6 +43,8 @@ contract DelegationManagerUnitTests is EigenLayerUnitTestSetup, IDelegationManag
     address defaultStaker = cheats.addr(uint256(123_456_789));
     address defaultOperator = address(this);
 
+    uint256 initializedWithdrawalDelayBlocks = 0;
```
ChaoticWalrus, line 46: should this test or any others set this to a nonzero value?
feels harder to test that the functionality is correct if we are setting this to zero, I think.

ChaoticWalrus (approved):
LGTM. Shared a couple minor thoughts

src/contracts/core/DelegationManager.sol
8sunyuan: Good call, added the event here https://github.com/Layr-Labs/eigenlayer-contracts/pull/355/commits/810b2a115b80d011a2077000a70369abf11b3464

src/test/unit/DelegationUnit.t.sol
8sunyuan: I left it as 0 since we never changed it from its default 0 value in the unit tests. 
Updated now to a nonzero value but tests still seem to pass because we never added back the unit tests for the withdrawals. Will add them back to `DelegationUnit.t.sol` in a followup PR

PR #358, opened 2023-12-03
Delegation integration tests

Gajesh2007:


src/test/integration/IntegrationChecks.t.sol
```diff
@@ -127,4 +127,10 @@ contract IntegrationCheckUtils is IntegrationBase {
         assert_Snap_Unchanged_OperatorShares(operator, "operator should have shares unchanged");
         assert_Snap_Unchanged_StrategyShares(strategies, "strategies should have total shares unchanged");
     }
+
+    function assertRegisteredAsOperatorState(User staker) internal {
```
ypatil12, line 131: I'd argue we don't need a function for this since it's only a one line check - cleaner to just keep it in original test file

src/test/integration/tests/Deposit_Delegate_Redelegate_Complete.t.sol
```diff
@@ -56,6 +123,77 @@ contract Integration_Deposit_Delegate_Redelegate_Complete is IntegrationCheckUti
...
+        withdrawalRoots = _getWithdrawalHashes(withdrawals);
+        check_QueuedWithdrawal_State(staker, operator2, strategies, shares, withdrawals, withdrawalRoots);
+
+        // 7. Complete withdrawal
+        // Fast forward to when we can complete the withdrawal
+        cheats.roll(block.number + delegationManager.withdrawalDelayBlocks());
+
+        // Complete all but last withdrawal as tokens
+        for (uint i = 0; i < withdrawals.length - 1; i++) {
+            staker.completeWithdrawalAsTokens(withdrawals[i]);
```
ypatil12, line 145: No checks are being made on these withdrawals, only the last withdrawal below

src/test/integration/tests/Deposit_Queue_Complete.t.sol
```diff
@@ -0,0 +1,79 @@
...
+
+        // 2. Queue Withdrawal
+        IDelegationManager.Withdrawal[] memory withdrawals = staker.queueWithdrawals(strategies, shares);
+
+        // 3. Complete Queued Withdrawal
+        cheats.roll(block.number + delegationManager.withdrawalDelayBlocks());
+        for (uint i = 0; i < withdrawals.length; i++) {
+            uint[] memory expectedTokens = _calculateExpectedTokens(strategies, shares);
+            IERC20[] memory tokens = staker.completeWithdrawalAsTokens(withdrawals[i]);
+            check_Withdrawal_AsTokens_State(staker, address(0), withdrawals[i], strategies, shares, tokens, expectedTokens);
```
ypatil12, line 40: I added address(0) for now, so the file won't compile. Can you overload the `check_WithdrawalAstokens_State` to accept a non-user for the operator so that these tests will pass?

src/test/integration/tests/Deposit_Queue_Complete.t.sol
```diff
@@ -0,0 +1,79 @@
...
+        assertFalse(delegationManager.isDelegated(address(staker)), "Staker should not be delegated after deposit");
+
+        // 2. Queue Withdrawal
+        IDelegationManager.Withdrawal[] memory withdrawals = staker.queueWithdrawals(strategies, shares);
+
+        // 3. Complete Queued Withdrawal
+        cheats.roll(block.number + delegationManager.withdrawalDelayBlocks());
+        for (uint i = 0; i < withdrawals.length; i++) {
+            staker.completeWithdrawalAsShares(withdrawals[i]); 
+            check_Withdrawal_AsShares_State(staker, address(0), withdrawals[i], strategies, shares);
```
ypatil12, line 73: Same as above, please create helper function for this edge case

src/test/integration/tests/Deposit_Register_QueueWithdrawal_Complete.t.sol
```diff
@@ -0,0 +1,70 @@
...
+        uint[] memory shares = _calculateExpectedShares(strategies, tokenBalances);
+        check_Deposit_State(staker, strategies, shares);
+
+        // 2. Staker registers as an operator
+        staker.registerAsOperator();
+        assertRegisteredAsOperatorState(staker);
+
+        // 3. Queue Withdrawal
+        IDelegationManager.Withdrawal[] memory withdrawals = staker.queueWithdrawals(strategies, shares);
+
```
ypatil12, line 30: Nit: add check for asserting withdrawal

src/test/integration/tests/Deposit_Register_QueueWithdrawal_Complete.t.sol
```diff
@@ -0,0 +1,70 @@
...
```
ypatil12, line 61: Add check for queued withdrawal

src/test/integration/tests/Deposit_Delegate_Redelegate_Complete.t.sol
```diff
@@ -56,6 +123,77 @@ contract Integration_Deposit_Delegate_Redelegate_Complete is IntegrationCheckUti
...
+        }
+
+        // Complete last withdrawal as shares
+        staker.completeWithdrawalAsTokens(withdrawals[withdrawals.length - 1]);
+        check_Withdrawal_AsTokens_State(staker, operator2, withdrawals[withdrawals.length - 1], strategies, shares);
+    }
+
+    function testFuzz_deposit_delegate_reDelegate_withAdditionalDepositBefore(uint24 _random) public {
```
ypatil12, line 153: I think the "additional deposit before" part of this test may have been removed from the rebases... can you add back?

ypatil12: Still think it hasn't been added

src/test/integration/tests/Delegate_Deposit_Queue_Complete.t.sol
```diff
@@ -0,0 +1,81 @@
...
+        // 1. Delegate to operator
+        staker.delegateTo(operator);
+        check_Delegation_State(staker, operator, strategies, new uint256[](strategies.length)); // Initial shares are zero
+
+        // 2. Deposit into strategy
+        staker.depositIntoEigenlayer(strategies, tokenBalances);
+        uint[] memory shares = _calculateExpectedShares(strategies, tokenBalances);
+
+        // Check that the deposit increased operator shares the staker is delegated to
+        check_Delegation_State(staker, operator, strategies, shares);
```
ypatil12, line 30: Looks like everything passes except the two tests in this file. Reason is that this check helper function specifically assumes that the Staker has deposited and then delegated. As a result, the test checks that the Staker's shares _have not_ changed when they should have. Would recommend creating a new helper method that isolates this specific test case where the Staker's shares do increase after a deposit (because of previously delegating).

ypatil12 (changes requested):

ypatil12 (approved):
LGTM!

PR #362, opened 2023-12-04
Fix: Update rpc on withdrawal migration test

ypatil12:


ypatil12:
Works locally and test passes in CI

ChaoticWalrus (approved):
LGTM. ty ty

PR #363, opened 2023-12-04
feat: support avs<>operator mapping with new APIs

bowenli86:
support avs<>operator mapping with new APIs

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -422,6 +422,77 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         }
     }
 
+    /**
+     * @notice Called by an avs to register an operator with the avs.
+     * @param operator The address of the operator to register.
+     * @param operatorSignature The signature, salt, and expiry of the operator's signature.
+     */
+    function registerOperatorToAVS(
+        address operator,
+        ISignatureUtils.SignatureWithSaltAndExpiry memory operatorSignature) external {
+
+        require(
+            operatorSignature.expiry >= block.timestamp,
+            "DelegationManager.registerOperatorToAVS: operator signature expired"
+        );
+        require(
+            avsOperatorStatus[msg.sender][operator] != OperatorAVSRegistrationStatus.REGISTERED,
+            "DelegationManager.registerOperatorToAVS: operator already registered"
+        );
+        require(
+            !operatorSaltIsSpent[operator][operatorSignature.salt],
+            "DelegationManager.registerOperatorToAVS: salt already spent"
+        );
+
+        // Calculate the digest hash
+        bytes32 operatorRegistrationDigestHash = calculateOperatorAVSRegistrationDigestHash({
+            operator: operator,
+            avs: msg.sender,
+            salt: operatorSignature.salt,
+            expiry: operatorSignature.expiry
+        });
+
+        // Check that the signature is valid
+        EIP1271SignatureUtils.checkSignature_EIP1271(
+            operator,
+            operatorRegistrationDigestHash,
+            operatorSignature.signature
+        );
+
+        // Set the operator as registered
+        avsOperatorStatus[msg.sender][operator] = OperatorAVSRegistrationStatus.REGISTERED;
+
+        // Mark the salt as spent
+        operatorSaltIsSpent[operator][operatorSignature.salt] = true;
+
+        emit OperatorAVSRegistrationStatusUpdated(operator, msg.sender, OperatorAVSRegistrationStatus.REGISTERED);
+    }
+
+    /**
+     * @notice Called by an avs to deregister an operator with the avs.
+     * @param operator The address of the operator to deregister.
+     */
+    function deregisterOperatorFromAVS(address operator) external {
+        require(
+            avsOperatorStatus[msg.sender][operator] == OperatorAVSRegistrationStatus.REGISTERED,
+            "DelegationManager.deregisterOperatorFromAVS: operator not registered"
+        );
+
+        // Set the operator as deregistered
+        avsOperatorStatus[msg.sender][operator] = OperatorAVSRegistrationStatus.UNREGISTERED;
+
+        emit OperatorAVSRegistrationStatusUpdated(operator, msg.sender, OperatorAVSRegistrationStatus.UNREGISTERED);
+    }
+
+    /**
+     * @notice Returns whether or not an operator is registered to an avs.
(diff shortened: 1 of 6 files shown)


src/test/unit/DelegationUnit.t.sol
```diff
@@ -578,6 +602,77 @@ contract DelegationManagerUnitTests_RegisterModifyOperator is DelegationManagerU
...
+        address operator = cheats.addr(delegationSignerPrivateKey);
+        cheats.assume(operatorSignature.expiry < block.timestamp);
+
+        cheats.expectRevert("DelegationManager.registerOperatorToAVS: operator signature expired");
+        delegationManager.registerOperatorToAVS(operator, operatorSignature);
+    }
+
+    // @notice Verifies an operator registers fails when it's already registered to the avs
+    function testFuzz_revert_whenOperatorAlreadyRegisteredToAVS(bytes32 salt) public {
+        cheats.prank(defaultAVS);
```
ypatil12, line 659: Nit: Move prank to be right above `registerOperatorToAVS` and use `cheats.startPrank(defaultAVS)` since we are pranking over multiple calls. Reason why is passes regardless is that the defaultAVS is the contract address. Technically pranking isn't needed but I like it as it is clear to the reader of tests what's going on.

src/contracts/core/DelegationManager.sol
```diff
@@ -422,6 +422,77 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+
+        emit OperatorAVSRegistrationStatusUpdated(operator, msg.sender, OperatorAVSRegistrationStatus.UNREGISTERED);
+    }
+
+    /**
+     * @notice Returns whether or not an operator is registered to an avs.
+     * @param operator The address of the operator.
+     * @param avs The address of the avs.
+     */
+    function isRegisteredToAVS(address operator, address avs) external view returns (bool) {
```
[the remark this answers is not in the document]
bowenli86: not used anywhere, removed

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -93,6 +97,13 @@ abstract contract DelegationManagerStorage is IDelegationManager {
     /// See conversation here: https://github.com/Layr-Labs/eigenlayer-contracts/pull/365/files#r1417525270
     address private __deprecated_stakeRegistry;
 
+    /// @notice Mapping: AVS => operator => enum of operator status to the AVS
+    mapping(address => mapping(address => OperatorAVSRegistrationStatus)) public avsOperatorStatus;
+
+    /// @notice Mapping: operator => 32-byte salt => whether or not the salt has already been used by the operator.
+    /// @dev Salt is used in the `registerOperatorToAVS` function.
+    mapping(address => mapping(bytes32 => bool)) public operatorSaltIsSpent;
```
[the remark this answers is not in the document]
ypatil12: Adding it to the CI is in my TODOs...

src/contracts/core/DelegationManager.sol
```diff
@@ -422,6 +422,68 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
         }
     }
 
+    /**
+     * @notice Called by an avs to register an operator with the avs.
+     * @param operator The address of the operator to register.
+     * @param operatorSignature The signature, salt, and expiry of the operator's signature.
+     */
+    function registerOperatorToAVS(
```
wadealexc, line 430: Shouldn't this method check that `operator` is an operator? IMO this method should continue the flow that `registerAsOperator` starts

wadealexc (changes requested):
See previous comment

src/contracts/core/DelegationManager.sol
ChaoticWalrus: this is a good point; we should add the check and test coverage which makes sure the function reverts appropriately when called by an address that has never registered as an operator.
there's an `isOperator` method which we use for the same check elsewhere in the DelegationManager, e.g. here: https://github.com/Layr-Labs/eigenlayer-contracts/blob/783797b307ad49b51acac36a6837ed8fc664601b/src/contracts/core/DelegationManager.sol#L120 -- so you can copy-paste that check + just update the revert string.

ypatil12: Good idea, makes sense to me

bowenli86:
> Oh, one more thing -- we should probably make the register/deregister methods pausable, like we do with most state changing methods.

how to do that? can you give or link some examples?

src/contracts/core/DelegationManager.sol
bowenli86: done

src/contracts/core/DelegationManager.sol
```diff
@@ -422,6 +425,71 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        address operator,
+        ISignatureUtils.SignatureWithSaltAndExpiry memory operatorSignature) external onlyWhenNotPaused(PAUSED_OPERATOR_REGISTER_DEREGISTER_TO_AVS) {
```
wadealexc, line 435: Oh, not blocking but would request that the `external` portion of the method declaration is on a new line. Check out our other multi-line method declarations for an example -- like `_delegate` below

src/contracts/core/DelegationManager.sol
```diff
@@ -422,6 +425,71 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        operatorSaltIsSpent[operator][operatorSignature.salt] = true;
+
+        emit OperatorAVSRegistrationStatusUpdated(operator, msg.sender, OperatorAVSRegistrationStatus.REGISTERED);
+    }
+
+    /**
+     * @notice Called by an avs to deregister an operator with the avs.
+     * @param operator The address of the operator to deregister.
+     */
+    function deregisterOperatorFromAVS(address operator) external onlyWhenNotPaused(PAUSED_OPERATOR_REGISTER_DEREGISTER_TO_AVS) {
```
8sunyuan, line 481: This function is callable by anyone once the operator is registered for an AVS.
Shouldn't this only be permissioned by the AVS?

src/contracts/core/DelegationManager.sol
bowenli86: ok will do before merging

bowenli86:
@ypatil12 @wadealexc @ChaoticWalrus thanks for reviewing!

PR #364, opened 2023-12-04
test: support random balance updates in integration tests

wadealexc:
- Adds a new helper, `_randBalanceUpdate` which returns:
  - `int[] tokenDeltas`: change in LST/native ETH tokens
  - `int[] stakerShareDeltas`: expected change in staker's "delegatable" shares
  - `int[] operatorShareDeltas`: expected change in operator's delegated shares
- Adds a User method, `User.updateBalances(strategies, tokenDeltas)`:
  - For native ETH, this creates a balance update proof and verifies it in the staker's `EigenPod`
  - For LSTs, this deposits tokens into the `StrategyManager`
- Adds some logging to User methods for easier debugging
  - Note: may want to add user identifiers to make log output easier to read. Could also consider a debug flag
---

`_randBalanceUpdate` generates a random balance update for any strategy type. 
* For LSTs, the balance update is in the positive direction only, and includes minting tokens to the staker to later deposit into the strategy manager.
* For ETH, the balance update is either positive or negative, and simulates a beacon chain balance update in one of the staker's validators.

The implementation is a bit messy because we have to do a lot of complicated delta math on operator shares, pod owner shares, validator restaked balances, etc.


src/test/integration/User.t.sol
```diff
@@ -78,6 +80,7 @@ contract User is Test {
 
     /// @dev For each strategy/token balance, call the relevant deposit method
     function depositIntoEigenlayer(IStrategy[] memory strategies, uint[] memory tokenBalances) public createSnapshot virtual {
+        emit log("- User.depositIntoEigenlayer");
```
ypatil12, line 83: These logs are helpful!

src/test/integration/tests/Deposit_Delegate_Queue_Complete.t.sol
```diff
@@ -10,6 +10,70 @@ contract Deposit_Delegate_Queue_Complete is IntegrationBase {
...
+            IStrategy[] memory strategies, 
+            uint[] memory tokenBalances
+        ) = _newRandomStaker();
+        (User operator, ,) = _newRandomOperator();
+        uint[] memory shares = _calculateExpectedShares(strategies, tokenBalances);
+
+        assert_HasNoDelegatableShares(staker, "staker should not have delegatable shares before depositing");
+        assertFalse(delegationManager.isDelegated(address(staker)), "staker should not be delegated");
+
+        staker.depositIntoEigenlayer(strategies, tokenBalances);
```
ypatil12, line 31: Nit: Missing first step comment

src/test/integration/tests/Deposit_Delegate_Queue_Complete.t.sol
```diff
@@ -10,6 +10,70 @@ contract Deposit_Delegate_Queue_Complete is IntegrationBase {
...
+        // The staker will queue one or more withdrawals for all strategies and shares
+        //
+        // ... check that each withdrawal was successfully enqueued, that the returned withdrawals
+        //     match now-pending withdrawal roots, and that the staker and operator have
+        //     reduced shares.
+        withdrawals = staker.queueWithdrawals(strategies, shares);
+        withdrawalRoots = _getWithdrawalHashes(withdrawals);
+
+        assert_ValidWithdrawalHashes(withdrawals, withdrawalRoots, "calculated withdrawals should match returned roots");
+        assert_AllWithdrawalsPending(withdrawalRoots, "staker's withdrawals should now be pending");
```
ypatil12, line 59: Nit: I've been trying to not use apostrophes since the highlighting below gets messed up

src/test/integration/IntegrationBase.t.sol
```diff
@@ -499,6 +641,31 @@ abstract contract IntegrationBase is IntegrationDeployer {
         return curShares;
     }
 
+    /// @dev Uses timewarp modifier to get staker shares at the last snapshot
```
ypatil12, line 644: Line 631 above: https://github.com/Layr-Labs/eigenlayer-contracts/blob/36203255b143550de64b7e093a55a48f4e9a2fcc/src/test/integration/IntegrationBase.t.sol#L626-L633

Should we now remove this revert message? Perhaps we just use ints instead of uints across the board when asserting shares? - `assert_Snap_Unchanged/Added/Removed_StrategyShares` could just pass in zero/positive/negative share deltas into `assert_Snap_Delta_StakerShares`

ypatil12 (commented):
LGTM, minor comment on now using ints instead of uints on share assertions

src/test/integration/IntegrationBase.t.sol
ypatil12: Actually DM only uses uint256...

wadealexc: I have no idea... I want to avoid having to do a ton of casting on the test level

wadealexc: If the casting is under the hood I think that'd be fine

ypatil12: ya it's fine for now

PR #366, opened 2023-12-05
Fix: Update balance update not pod revert test

ypatil12:


wadealexc (approved):
works on my end; LGTM

PR #368, opened 2023-12-05
feat: more formal verification for eigenpods

ChaoticWalrus:
Adds an EigenPod-specific spec (in addition to the EigenPodManager spec) and several rules.
There are some TODOs + commented-out drafts left in right now, but this is passing at the moment + feels useful to me, at least.
Commits are definitely rather messy; this should 100% be squash-merged.


ChaoticWalrus:
merging this now to help with CI efficiency

PR #369, opened 2023-12-06
chore: migrate to certora prover v5

ChaoticWalrus:
- switch from fixed v4.13.1 to floating / latest version

- implement minimal changes that (hopefully) make existing specs work with v5

This is a minimal PR to migrate from v4.13.1 to v5 _without breaking our specs_.
We will need to do more work to actually take full advantage of the new features in v5.


Kirkeelee:
@ChaoticWalrus Is there a way to contact you? I cannot find the munged contracts in your repo.

PR #371, opened 2023-12-08
chore: update README, docs, and prune old files

wadealexc:


docs/README.md
```diff
@@ -1,62 +1,102 @@
...
 
 These contracts work together to enable native ETH restaking:
 * Users deploy `EigenPods` via the `EigenPodManager`, which contain beacon chain state proof logic used to verify a validator's withdrawal credentials, balance, and exit. An `EigenPod's` main role is to serve as the withdrawal address for one or more of a user's validators.
 * The `EigenPodManager` handles `EigenPod` creation and accounting+interactions between users with restaked native ETH and the `DelegationManager`.
-* The `DelayedWithdrawalRouter` imposes a 7-day delay on completing certain withdrawals from an `EigenPod`. This is primarily to add a stopgap against a hack being able to instantly withdraw funds (note that most withdrawals are processed via the `DelegationManager`, not the `DelayedWithdrawalRouter`).
+* The `DelayedWithdrawalRouter` imposes a 7-day delay on completing certain withdrawals from an `EigenPod`. This is primarily to add a stopgap against a hack being able to instantly withdraw funds (note users initiate withdrawals via the `DelegationManager`).
```
ChaoticWalrus, line 30: I think this wording change may a step back?
maybe instead:
note that all withdrawals from EigenLayer -- other than those of "partial withdrawals" earned by validators -- are initiated via the `DelegationManager`

ChaoticWalrus (approved):
Changes LGTM.
Noted one small critique of some wording.
I think the only things in the deleted docs that it might make sense to re-include are:

- clarifying the trust assumptions (particularly between stakers and operators)
- explaining a bit about the share-based accounting model (had a question about this recently)

but these suggested additions are totally non-blocking for this PR -- we can just do them later.

stevennevins (approved):

docs/README.md
wadealexc: good call

wadealexc: made a slightly different change: [cda4c51](https://github.com/Layr-Labs/eigenlayer-contracts/pull/371/commits/cda4c5113f83693fd390a93313ed3b24960d78b0)

wadealexc: it does omit the other 2 cases where the withdrawal router is being used (funds sent directly to pod and pre-"restaking activated" funds), but i think this captures the high level concept and the EPM docs can be more specific.

PR #372, opened 2023-12-11
fix: update docs and ensure important state changing methods are pausable

wadealexc:
Docs:
* Updates EPM docs with `verifyBalanceUpdates` and `createPod` method signature change
* Adds docs for #363 

Changes:
* EPM.createPod and EPM.stake now pause at the top level. Previously, stake could be called during a pause if the pod owner had a pod deployed already
* EP.recoverTokens and EP.withdrawNonBeaconChainETHBalanceWei now have a pause flag

### DIFF
--- a/docs/core/DelegationManager.md
+++ b/docs/core/DelegationManager.md
@@ -51,6 +51,10 @@ Operators interact with the following functions to become an Operator:
 * [`DelegationManager.modifyOperatorDetails`](#modifyoperatordetails)
 * [`DelegationManager.updateOperatorMetadataURI`](#updateoperatormetadatauri)
 
+Once registered as an Operator, Operators can use an AVS's contracts to register with a specific AVS. The AVS will call these functions on the Operator's behalf:
+* [`DelegationManager.registerOperatorToAVS`](#registeroperatortoavs)
+* [`DelegationManager.deregisterOperatorFromAVS`](#deregisteroperatorfromavs)
+
 #### `registerAsOperator`
 
 ```solidity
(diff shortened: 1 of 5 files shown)


docs/core/DelegationManager.md
```diff
@@ -51,6 +51,10 @@ Operators interact with the following functions to become an Operator:
 * [`DelegationManager.modifyOperatorDetails`](#modifyoperatordetails)
 * [`DelegationManager.updateOperatorMetadataURI`](#updateoperatormetadatauri)
 
+Once registered as an Operator, Operators can use an AVS's contracts to register with a specific AVS. The AVS will call these functions on the Operator's behalf:
+* [`DelegationManager.registerOperatorToAVS`](#registeroperatortoavs)
```
8sunyuan, line 55: In the context an AVS is using our middleware contracts, should we also specify that the "AVS" calling these functions refers to the RegistryCoordinator?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -235,7 +239,7 @@ contract EigenPodManager is
 
     // INTERNAL FUNCTIONS
 
-    function _deployPod() internal onlyWhenNotPaused(PAUSED_NEW_EIGENPODS) returns (IEigenPod) {
+    function _deployPod() internal returns (IEigenPod) {
```
ChaoticWalrus, line 242: did you move this to the external functions just for clarity or something?
seems these changes mildly change the behavior of the `stake` function -- it will now revert if `PAUSED_NEW_EIGENPODS` is set, even if the caller already has a deployed EigenPod.
I don't care strongly either way, but think I have a slight preference for the previous behavior.

wadealexc: no, i moved it so that we could pause calls to `EigenPod.stake` in all scenarios - since that's a potentially dangerous state changing method.

I want to ensure that if we find out something is wrong with `stake` and interaction with the deposit contract that we can prevent its use temporarily. IMO this is much more important than preventing the deployment of pods specifically, since that's a really simple process

ChaoticWalrus: works for me 👍

ChaoticWalrus (approved):
LGTM, thanks!

PR #378, opened 2023-12-16
chore: fix tests to work with modified behavior

ChaoticWalrus:
integration tests in particular are now slightly more flexible

NOTE: this fix doesn't actually work.
Something weird is going on, where the expected queued withdrawals calculated in `_getExpectedWithdrawalStructsForStaker` don't match the actually queued withdrawals.

It seems like things go one of two directions: withdrawals are missing for beacon chain ETH, or all withdrawals have the last strategy in the staker's array as their strategy, rather than the i-th strategy.

I can't figure out quite what the root cause is here. Tried adding some logging to the `_getExpectedWithdrawalStructsForStaker` function, which is how I got this deep.

### DIFF
--- a/docs/core/DelegationManager.md
+++ b/docs/core/DelegationManager.md
@@ -229,14 +229,14 @@ function undelegate(
 ) 
     external 
     onlyWhenNotPaused(PAUSED_ENTER_WITHDRAWAL_QUEUE)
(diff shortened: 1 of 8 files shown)


wadealexc (approved):
skrrrt skrrt

PR #380, opened 2023-12-24
Update AVS-Guide.md

Akare123:
I suggested some improvements for this section. Those are:

Original: This grace period is configurable on an operator level.
Suggested: This grace period is configurable at the operator level.

Original: The following figure illustrates the above flow:
Suggested: The figure below illustrates this flow:

Original: For an AVS to be able to slash an operator in an objective manner, the AVS needs to deploy a DisputeResolution contract which anyone can call to raise a challenge against an EigenLayer operator for its adversarial action.
Suggested: For an AVS to slash an operator objectively, the AVS needs to deploy a DisputeResolution contract that anyone can call to raise a challenge against an EigenLayer operator for its adversarial action.

### DIFF
--- a/docs/experimental/AVS-Guide.md
+++ b/docs/experimental/AVS-Guide.md
@@ -58,7 +58,7 @@ The following figure illustrates the above flow:
 
 A staker does not restake into AVSs. A staker delegates to an operator and it is the operator that registers for new AVSs (with the staker having option to opt-out).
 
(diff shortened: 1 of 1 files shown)


ChaoticWalrus:
Feels like some pretty opinionated changes where the grammar already appears to be acceptable, on a doc that is fairly out-dated.
Thank you for your suggestions, but I'm closing this out without merging.

PR #381, opened 2024-01-02
chore: fix typo

ylmin:


ChaoticWalrus:
feels like some weird gaming to try to get commits -- this is 4 commits that introduce typos and then "fix" them.
please don't open things like this, it's dishonest and unprofessional.

PR #383, opened 2024-01-02
Fix: Update registration documentation & type hash

ypatil12:


src/contracts/core/DelegationManager.sol
```diff
@@ -434,7 +434,7 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
     }
 
     /**
-     * @notice Called by an avs to register an operator with the avs.
+     * @notice Called by the AVS's service manager contract to register an operator with the avs.
```
samlaf, line 437: if its called by the service manager, do we really need to pass the address of the service manager as a parameter?

ypatil12: We don't pass the address in as a parameter?

docs/core/DelegationManager.md
```diff
@@ -126,7 +126,7 @@ Allows the caller (an AVS) to register an `operator` with itself, given the prov
 * Pause status MUST NOT be set: `PAUSED_OPERATOR_REGISTER_DEREGISTER_TO_AVS`
 * `operator` MUST already be a registered Operator
 * `operator` MUST NOT already be registered with the AVS
-* `operatorSignature` must be a valid, unused, unexpired signature from the `operator`
+* `operatorSignature` must be a valid, unused, unexpired signature from the `operator`. The signature is an ECDSA signature by the operator over the [`OPERATOR_AVS_REGISTRATION_TYPEHASH`](../../src/contracts/core/DelegationManagerStorage.sol#L29). Expiry is a utc timestamp in seconds. Salt is used only once per signature to prevent replay attacks.
```
wadealexc, line 129: Good change, but I'd request that you remove the specific line number link, since that can get outdated quickly. I've just been referring to the constants without a link - it's CTRL-F-able at least

wadealexc:
Reposting my request from Slack - the changes look fine, but this needs to target a new branch as SigP requested we don't make contract changes directly to `m2-mainnet` while they work.

Let's make a new branch, maybe `m2-mainnet-fixes`?

src/contracts/core/DelegationManager.sol
samlaf: ehhh my bad I confused 2 functions.

docs/core/DelegationManager.md
```diff
@@ -126,7 +126,7 @@ Allows the caller (an AVS) to register an `operator` with itself, given the prov
 * Pause status MUST NOT be set: `PAUSED_OPERATOR_REGISTER_DEREGISTER_TO_AVS`
 * `operator` MUST already be a registered Operator
 * `operator` MUST NOT already be registered with the AVS
-* `operatorSignature` must be a valid, unused, unexpired signature from the `operator`
+* `operatorSignature` must be a valid, unused, unexpired signature from the `operator`. The signature is an ECDSA signature by the operator over the [`OPERATOR_AVS_REGISTRATION_TYPEHASH`](../../src/contracts/core/DelegationManagerStorage). Expiry is a utc timestamp in seconds. Salt is used only once per signature to prevent replay attacks.
```
wadealexc, line 129: Needs a `.sol` if you're gonna keep the link :)

wadealexc (approved):
LGTM

PR #392, opened 2024-01-18
feat: strategy configs

8sunyuan:


src/contracts/core/StrategyManager.sol
```diff
@@ -138,6 +138,10 @@ contract StrategyManager is
         uint256 expiry,
         bytes memory signature
     ) external onlyWhenNotPaused(PAUSED_DEPOSITS) nonReentrant returns (uint256 shares) {
+        require(
+            !creditTransfersDisabled[strategy],
+            "StrategyManager.depositIntoStrategyWithSignature: credit transfers disabled"
+        );
```
wadealexc, line 144: I think this variable name is confusing, but i don't know what I'd call it...

src/contracts/core/StrategyManager.sol
```diff
@@ -202,6 +206,17 @@ contract StrategyManager is
...
+    /**
+     * If true for a strategy, a user cannot depositIntoStrategyWithSignature into that strategy for another staker
+     * and also when performing DelegationManager.queueWithdrawals, a staker can only withdraw to themselves.
+     * Defaulted to false for all existing strategies.
+     * @param strategy The strategy to set `creditTransfersDisabled` value to
+     * @param value bool value to set `creditTransfersDisabled` to
+     */
+    function setCreditTransfersDisabled(IStrategy strategy, bool value) external onlyOwner {
+        _setCreditTransfersDisabled(strategy, value);
+    }
```
wadealexc, line 218: Should this be an `onlyOwner` method, or a `onlyStrategyWhitelister` method? I'm thinking the latter:

The strategy whitelister can set transfers disabled by removing an existing strategy, then re-adding it with transfers disabled.

And if the strategy isn't already whitelisted, they can of course add it with transfers disabled.

wadealexc: Additional thing: this method should probably check that the strategy is whitelisted

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -79,12 +79,10 @@ abstract contract DelegationManagerStorage is IDelegationManager {
     mapping(address => mapping(bytes32 => bool)) public delegationApproverSaltIsSpent;
 
     /**
-     * @notice Minimum delay enforced by this contract for completing queued withdrawals. Measured in blocks, and adjustable by this contract's owner,
-     * up to a maximum of `MAX_WITHDRAWAL_DELAY_BLOCKS`. Minimum value is 0 (i.e. no delay enforced).
-     * @dev Note that the withdrawal delay is not enforced on withdrawals of 'beaconChainETH', as the EigenPods have their own separate delay mechanic
-     * and we want to avoid stacking multiple enforced delays onto a single withdrawal.
+     * @notice Deprecated from an old Goerli release, setting withdrawalDelayBLocks on per strategy basis
```
wadealexc, line 82: Clarify comment:

"Deprecated from an old Goerli release, which had a global withdrawal delay for all withdrawals. We now configure withdrawal delays on a per strategy basis"

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -104,6 +102,14 @@ abstract contract DelegationManagerStorage is IDelegationManager {
     /// @dev Salt is used in the `registerOperatorToAVS` function.
     mapping(address => mapping(bytes32 => bool)) public operatorSaltIsSpent;
 
+    /**
+     * @notice Minimum delay enforced by this contract per Strategy for completing queued withdrawals. Measured in blocks, and adjustable by this contract's owner,
+     * up to a maximum of `MAX_WITHDRAWAL_DELAY_BLOCKS`. Minimum value is 0 (i.e. no delay enforced).
+     * @dev Note that the withdrawal delay is not enforced on withdrawals of 'beaconChainETH', as the EigenPods have their own separate delay mechanic
+     * and we want to avoid stacking multiple enforced delays onto a single withdrawal.
```
wadealexc, line 109: I don't think this is true -- are we changing this?

Native ETH currently does use the withdrawal delay.

wadealexc: > EigenPods have their own separate delay mechanic

They do, mainly for partial withdrawals, which don't need to go through the DMgr queue anyway. 

For a full withdrawal, which does go through the queue, the completed withdrawal is sent directly (not via the delayed withdrawal router)

src/contracts/core/DelegationManager.sol
```diff
@@ -620,11 +620,6 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
             "DelegationManager.completeQueuedAction: action is not in queue"
         );
 
-        require(
-            withdrawal.startBlock + withdrawalDelayBlocks <= block.number, 
-            "DelegationManager.completeQueuedAction: withdrawalDelayBlocks period has not yet passed"
-        );
-
```
wadealexc, line 627: This was enforcing a withdrawal delay on EigenPod shares and is no longer doing that -- that's not an intended change, unless I'm misunderstanding!

src/contracts/core/DelegationManager.sol
```diff
@@ -682,6 +683,11 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
                         });
                     }
                 } else {
+                    require(
+                        withdrawal.startBlock + strategyWithdrawalDelayBlocks[withdrawal.strategies[i]] <= block.number, 
+                        "DelegationManager.completeQueuedAction: withdrawalDelayBlocks period has not yet passed for this strategy"
```
wadealexc, line 688: While we're at it, let's change the error text here to `completeQueuedWithdrawal` :sweat: 

The `completeQueuedAction` text is a holdover from my very first chicken scratch refactor of the withdrawal queues

src/contracts/core/DelegationManager.sol
```diff
@@ -797,13 +807,36 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
 
-    function _initializeWithdrawalDelayBlocks(uint256 _withdrawalDelayBlocks) internal {
+    /**
+     * @notice Sets the withdrawal delay blocks for each strategy in `_strategiesToSetDelayBlocks` to `_withdrawalDelayBlocks`.
+     * only gets called once when initializing the contract.
+     */
+    function _initializeStrategyWithdrawalDelayBlocks(
+        IStrategy[] calldata _strategiesToSetDelayBlocks,
+        uint256[] calldata _withdrawalDelayBlocks
+    ) internal {
```
wadealexc, line 817: I wonder if it's worth it to add a function that allows us to update withdrawal delay blocks per strategy?

I think now that we're not doing a global delay, this might be a good idea, in case we misconfigure a strategy?

src/contracts/core/DelegationManager.sol
```diff
@@ -797,13 +807,36 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
         );
-        emit WithdrawalDelayBlocksSet(withdrawalDelayBlocks, _withdrawalDelayBlocks);
-        withdrawalDelayBlocks = _withdrawalDelayBlocks;
+        uint256 numStrats = _strategiesToSetDelayBlocks.length;
+        for (uint256 i = 0; i < numStrats; ++i) {
+            IStrategy strategy = _strategiesToSetDelayBlocks[i];
+            uint256 prevStrategyWithdrawalDelayBlocks = strategyWithdrawalDelayBlocks[strategy];
+            uint256 newStrategyWithdrawalDelayBlocks = _withdrawalDelayBlocks[i];
+            require(
+                newStrategyWithdrawalDelayBlocks <= MAX_WITHDRAWAL_DELAY_BLOCKS,
```
wadealexc, line 828: Do we need to change this value? I think the proposal wants a different max delay.

wadealexc (changes requested):
See comments above!

src/contracts/core/StrategyManager.sol
ChaoticWalrus: `depositsToDifferentAddressForbidden`, perhaps?
ideally something equivalent but shorter 😅

src/contracts/core/StrategyManager.sol
ChaoticWalrus: I agree this should probably be controlled by the `strategyWhitelister` role.

> this method should probably check that the strategy is whitelisted

I think I actually disagree with this. seems fine to allow setting this prior to whitelisting.

src/contracts/core/DelegationManager.sol
ChaoticWalrus: I thought we still wanted a global "minimum withdrawal delay" as well, which it still seems appropriate to check here.

src/contracts/core/DelegationManager.sol
ChaoticWalrus: yes. need to lengthen this to I think 4 weeks.

src/contracts/core/StrategyManager.sol
ChaoticWalrus: ahh but it's not for deposits only, as it's also used when queuing withdrawals...
`transfersWithinEigenLayerForbidden` ? idk.

src/contracts/core/StrategyManager.sol
wadealexc: > I think I actually disagree with this. seems fine to allow setting this prior to whitelisting.

ok fair, that would make the ops around whitelisting easier

src/contracts/core/StrategyManager.sol
8sunyuan: `thirdPartyTransfersForbidden`, `transferOnBehalfForbidden`, `enableDelegatedTransfers`?
Personally think something like `thirdPartyTransfersForbidden` is most clear

src/contracts/core/DelegationManagerStorage.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/323fa4e387941179468d541a4c9d9470c2382bd1#diff-4d12f7b9016c76f80ea4d90dc4e49dbddda5177ec6d7c8cd4055b2f3f857a20aR84

src/contracts/core/StrategyManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/323fa4e387941179468d541a4c9d9470c2382bd1#diff-31bf779120d094208ed8d64498e5f0c894da92b024a34723b583bef49524a877R219
Updated to `onlyStrategyWhitelister`

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/235957282bc952638c45078ae1ba28638f78263b#diff-d323417ded59a593f4ea410225a9167b24198c4d2c05cbc0fe9af47d34b88810R641
Added back and named it `minWithdrawalDelayBlocks`. Min number of blocks to pass to withdraw a strategy will be 
max(`minWithdrawalDelayBlocks`, `strategyWithdrawalDelayBlocks[strategy]`)

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/323fa4e387941179468d541a4c9d9470c2382bd1#diff-d323417ded59a593f4ea410225a9167b24198c4d2c05cbc0fe9af47d34b88810R646

src/contracts/core/DelegationManager.sol
```diff
@@ -751,6 +757,10 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
                  */
                 eigenPodManager.removeShares(staker, shares[i]);
             } else {
+                require(
+                    !strategyManager.creditTransfersDisabled(strategies[i]) || staker == withdrawer,
```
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/323fa4e387941179468d541a4c9d9470c2382bd1#diff-d323417ded59a593f4ea410225a9167b24198c4d2c05cbc0fe9af47d34b88810R781

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/323fa4e387941179468d541a4c9d9470c2382bd1#diff-d323417ded59a593f4ea410225a9167b24198c4d2c05cbc0fe9af47d34b88810R511
Added setter here

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/323fa4e387941179468d541a4c9d9470c2382bd1#diff-4d12f7b9016c76f80ea4d90dc4e49dbddda5177ec6d7c8cd4055b2f3f857a20aR49

src/contracts/core/DelegationManagerStorage.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/files#diff-d323417ded59a593f4ea410225a9167b24198c4d2c05cbc0fe9af47d34b88810R640
Added back min global delay

src/contracts/core/StrategyManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/a3471c4bb10919a8e8eb6c8feed36e394b182d4d

docs/core/StrategyManager.md
```diff
@@ -23,6 +23,7 @@ This document organizes methods according to the following themes (click each to
 * `mapping(address => IStrategy[]) public stakerStrategyList`: Maintains a list of the strategies a Staker holds a nonzero number of shares in.
     * Updated as needed when Stakers deposit and withdraw: if a Staker has a zero balance in a Strategy, it is removed from the list. Likewise, if a Staker deposits into a Strategy and did not previously have a balance, it is added to the list.
 * `mapping(IStrategy => bool) public strategyIsWhitelistedForDeposit`: The `strategyWhitelister` is (as of M2) a permissioned role that can be changed by the contract owner. The `strategyWhitelister` has currently whitelisted 3 `StrategyBaseTVLLimits` contracts in this mapping, one for each supported LST.
+* `mapping(IStrategy => bool) public thirdPartyTransfersForbidden`: The `strategyWhitelister` can disable third party transfers for a given strategy. This means that if `thirdPartyTransfersForbidden[strategy] == true`, then you cannot deposit on behalf of another Staker and award them shares. You also cannot withraw shares on behalf of another Staker.  (see [`DelegationManager.queueWithdrawals`](./DelegationManager#queueWithdrawals.md))
```
ChaoticWalrus, line 26: typo: `withraw` instead of `withdraw`.

src/contracts/core/DelegationManager.sol
```diff
@@ -68,18 +68,21 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
 
     /**
      * @dev Initializes the addresses of the initial owner, pauser registry, and paused status.
-     * withdrawalDelayBlocks is set only once here
+     * strategyWithdrawalDelayBlocks is set only once here
```
ChaoticWalrus, line 71: small nit: seems like the key thing set once is actually the `minWithdrawalDelayBlocks` variable

8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/5d8e5ff14d1eee680465fa4a712eef41033cb9a6

docs/core/StrategyManager.md
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/5d8e5ff14d1eee680465fa4a712eef41033cb9a6

wadealexc (commented):
One thing to point out - if we disable third party transfers after someone has already queued a withdrawal with a different withdrawer address, there's currently no way for that withdrawal to be completed.

Should we include a way for the original staker to complete a queued withdrawal to prevent this?

wadealexc (commented):
I think we should have a function `DelegationManager.getWithdrawalDelay(IStrategy[] strategies)` that computes the minimum time you'd have to wait to queue a withdrawal that includes all the listed strategies.

src/contracts/core/DelegationManager.sol
```diff
@@ -797,13 +828,45 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        emit MinWithdrawalDelayBlocksSet(minWithdrawalDelayBlocks, _minWithdrawalDelayBlocks);
+        minWithdrawalDelayBlocks = _minWithdrawalDelayBlocks;
+    }
+
+    /**
+     * @notice Sets the withdrawal delay blocks for each strategy in `_strategiesToSetDelayBlocks` to `_withdrawalDelayBlocks`.
+     * gets called when initializing contract or by calling `setStrategyWithdrawalDelayBlocks`
+     */
+    function _setStrategyWithdrawalDelayBlocks(
+        IStrategy[] calldata _strategiesToSetDelayBlocks,
```
wadealexc, line 845: Could just call this param `_strategies`

The context makes it pretty obvious what the list is for, so the current name is super long

ChaoticWalrus:
> One thing to point out - if we disable third party transfers after someone has already queued a withdrawal with a different withdrawer address, there's currently no way for that withdrawal to be completed.
> 
> Should we include a way for the original staker to complete a queued withdrawal to prevent this?

Oh, I thought this was just on queuing withdrawals! I think that's the correct place to enforce this instead of on withdrawal completion

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/cfaf96f9d8ee310f65643697ffd77ad66c323c91

8sunyuan:
> I think we should have a function `DelegationManager.getWithdrawalDelay(IStrategy[] strategies)` that computes the minimum time you'd have to wait to queue a withdrawal that includes all the listed strategies.
> 
> Maybe the naming isn't the best? something similar though

https://github.com/Layr-Labs/eigenlayer-contracts/pull/392/commits/cfaf96f9d8ee310f65643697ffd77ad66c323c91
Added right here, a separate PR for the sigchecker in middleware needs to account for these changes and call this function instead!

PR #393, opened 2024-01-19
capella-to-deneb with governance switch

Sidu28:
Changes EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT from 4 to 5 with a governance switch.  This would be for M2-mainnet assuming that we deploy prior to deneb hard fork on mainnet

### DIFF
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -213,7 +213,7 @@ library BeaconChainProofs {
         bytes32 beaconStateRoot,
         bytes32[] calldata withdrawalFields,
         WithdrawalProof calldata withdrawalProof,
-        bool denebForkTimestamp
+        uint64 denebForkTimestamp
     ) internal view {
         require(
             withdrawalFields.length == 2 ** WITHDRAWAL_FIELD_TREE_HEIGHT,
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -234,7 +234,7 @@ library BeaconChainProofs {
             "BeaconChainProofs.verifyWithdrawal: historicalSummaryIndex is too large"
         );
 
-        if(block.timestamp >= denebForkTimestamp){
+        if(block.timestamp <= denebForkTimestamp){
             require(
                 withdrawalProof.withdrawalProof.length ==
                     32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT_CAPELLA + WITHDRAWALS_TREE_HEIGHT + 1),
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -213,7 +213,7 @@ library BeaconChainProofs {
         bytes32 beaconStateRoot,
         bytes32[] calldata withdrawalFields,
         WithdrawalProof calldata withdrawalProof,
-        bool isDeneb
+        bool denebForkTimestamp
     ) internal view {
         require(
             withdrawalFields.length == 2 ** WITHDRAWAL_FIELD_TREE_HEIGHT,
@@ -234,7 +234,7 @@ library BeaconChainProofs {
             "BeaconChainProofs.verifyWithdrawal: historicalSummaryIndex is too large"
         );
 
-        if(!isDeneb){
+        if(block.timestamp >= denebForkTimestamp){
             require(
                 withdrawalProof.withdrawalProof.length ==
                     32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT_CAPELLA + WITHDRAWALS_TREE_HEIGHT + 1),
@@ -258,7 +258,7 @@ library BeaconChainProofs {
             "BeaconChainProofs.verifyWithdrawal: slotProof has incorrect length"
         );
 
-        if(!isDeneb){
+        if(block.timestamp >= denebForkTimestamp){
             require(
             withdrawalProof.timestampProof.length == 32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT_CAPELLA),
             "BeaconChainProofs.verifyWithdrawal: timestampProof has incorrect length"
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -258,7 +258,7 @@ library BeaconChainProofs {
             "BeaconChainProofs.verifyWithdrawal: slotProof has incorrect length"
         );
 
-        if(block.timestamp >= denebForkTimestamp){
+        if(block.timestamp <= denebForkTimestamp){
             require(
             withdrawalProof.timestampProof.length == 32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT_CAPELLA),
             "BeaconChainProofs.verifyWithdrawal: timestampProof has incorrect length"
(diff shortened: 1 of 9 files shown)


src/contracts/pods/EigenPodManagerStorage.sol
```diff
@@ -64,6 +64,8 @@ abstract contract EigenPodManagerStorage is IEigenPodManager {
      */
     mapping(address => int256) public podOwnerShares;
 
+    bool public isDeneb;
+
```
ChaoticWalrus, line 68: gotta fix the gap size from this new variable

ChaoticWalrus (changes requested):
Seems reasonable as a fix, just kind of ugly. 

Since the EigenPodManager inherits storage in a specific order, fixing the storage gap is absolutely blocking for merging this.

This PR is currently targeting the `m2-mainnet` branch -- if we upgrade on Goerli are there any other breaking changes from the currently-deployed code?

src/contracts/pods/EigenPodManagerStorage.sol
Sidu28: fixed!

Sidu28:
> Seems reasonable as a fix, just kind of ugly.
> 
> Since the EigenPodManager inherits storage in a specific order, fixing the storage gap is absolutely blocking for merging this.
> 
> This PR is currently targeting the `m2-mainnet` branch -- if we upgrade on Goerli are there any other breaking changes from the currently-deployed code?

This PR would be for mainnet.  For goerli, the plan is to just upgrade the library directly.

ChaoticWalrus:
> This PR would be for mainnet. For goerli, the plan is to just upgrade the library directly.

Ahhh, gotcha. I am more interested in getting the Goerli fix in then, but I think this makes sense.
Will give it a little thought because there isn't a rush with this if it's for mainnet.

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -232,11 +234,20 @@ library BeaconChainProofs {
...
-            withdrawalProof.withdrawalProof.length ==
-                32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT + WITHDRAWALS_TREE_HEIGHT + 1),
-            "BeaconChainProofs.verifyWithdrawal: withdrawalProof has incorrect length"
-        );
+        if(block.timestamp >= denebForkTimestamp){
+            require(
+                withdrawalProof.withdrawalProof.length ==
+                    32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT_CAPELLA + WITHDRAWALS_TREE_HEIGHT + 1),
+                "BeaconChainProofs.verifyWithdrawal: withdrawalProof has incorrect length"
+            );
```
wadealexc, line 242: Surely this is the wrong constant to use here? It should be the deneb constant right?

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -232,11 +234,20 @@ library BeaconChainProofs {
             "BeaconChainProofs.verifyWithdrawal: historicalSummaryIndex is too large"
         );
 
-        require(
-            withdrawalProof.withdrawalProof.length ==
-                32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT + WITHDRAWALS_TREE_HEIGHT + 1),
-            "BeaconChainProofs.verifyWithdrawal: withdrawalProof has incorrect length"
-        );
+        if(block.timestamp >= denebForkTimestamp){
```
wadealexc, line 237: The if statement with 2 identical requires is kinda messy. Can we separate this out into an if statement + a single require? Something like this:

```
uint256 withdrawalProofLen;
if (block.timestamp >= denebForkTimestamp) {
  withdrawalProofLen = /**deneb calc here*/;
} else {
  withdrawalProofLen = /**capella calc here*/;
}

require(withdrawalProof.withdrawalProof.length == withdrawalProofLen, "err here");
```

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -246,10 +257,18 @@ library BeaconChainProofs {
             withdrawalProof.slotProof.length == 32 * (BEACON_BLOCK_HEADER_FIELD_TREE_HEIGHT),
             "BeaconChainProofs.verifyWithdrawal: slotProof has incorrect length"
         );
-        require(
-            withdrawalProof.timestampProof.length == 32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT),
+
+        if(block.timestamp >= denebForkTimestamp){
+            require(
+            withdrawalProof.timestampProof.length == 32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT_CAPELLA),
```
wadealexc, line 263: Wrong constant for this branch?

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -246,10 +257,18 @@ library BeaconChainProofs {
             withdrawalProof.slotProof.length == 32 * (BEACON_BLOCK_HEADER_FIELD_TREE_HEIGHT),
             "BeaconChainProofs.verifyWithdrawal: slotProof has incorrect length"
         );
-        require(
-            withdrawalProof.timestampProof.length == 32 * (EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT),
+
+        if(block.timestamp >= denebForkTimestamp){
```
wadealexc, line 261: Same comment about the if branches being messy (see above)

src/contracts/pods/EigenPodManager.sol
```diff
@@ -233,6 +233,14 @@ contract EigenPodManager is
         _updateBeaconChainOracle(newBeaconChainOracle);
     }
 
+    /**
+    * @notice Sets the denebTimestamp
+    * @dev Callable only by the owner of this contract (i.e. governance)
+    */
+    function setDenebForkTimestamp(uint64 timestamp) external onlyOwner {
+        denebForkTimestamp = timestamp;
+    }
```
wadealexc, line 242: Should this be an "only set this once" situation?

Should we emit an event?

wadealexc (changes requested):
Requested some changes, see comments above. Additionally:
- Please PR against `m2-mainnet-fixes`, not `m2-mainnet`
- Please fix failing tests
- Please add tests that check expected behavior in deneb

wadealexc:
> Seems reasonable as a fix, just kind of ugly.

It's unfortunate we need to do something like this, but... this does match the fork-compatibility code changes I've seen in other codebases. Haven't actually seen this in smart contracts, but in Filecoin's core codebase they frequently have things like `if (networkVersion >= NV.17) { } else {}`

I'm not sure if there's a cleaner way to do it other than a separate upgrade post-fork.

I guess we can excise the old fork code in an unrelated upgrade down the line.

Sidu28:
> > Seems reasonable as a fix, just kind of ugly.
> 
> It's unfortunate we need to do something like this, but... this does match the fork-compatibility code changes I've seen in other codebases. Haven't actually seen this in smart contracts, but in Filecoin's core codebase they frequently have things like `if (networkVersion >= NV.17) { } else {}`
> 
> I'm not sure if there's a cleaner way to do it other than a separate upgrade post-fork.
> 
> I guess we can excise the old fork code in an unrelated upgrade down the line.

@wadealexc I think it is still undecided if we want to wait till the network hard forks on mainnet and then deploy m2.  In that case we can wait without this ugly switch logic.

ChaoticWalrus:
Discussed some offline, but one option is to also switch to native block roots at the time of the Dencun fork.
This might require switching to a new "oracle" contract which just fetches this root, as we're unsure on Solidity support for fetching this root.

PR #394, opened 2024-01-19
EXECUTION_PAYLOAD_HEADER_FIELD_TREE_HEIGHT changed from 4->5

Sidu28:


ChaoticWalrus:
> it is branched off this commit hash as per the deployment tracking spreadsheet: [af8d8755fdad94705479679c7c63d454c4ed9a8e](https://github.com/Layr-Labs/eigenlayer-contracts/tree/af8d8755fdad94705479679c7c63d454c4ed9a8e)

Ahh, I see what happened. That is the commit hash _of the proxy deployment_. The implementations are all at https://github.com/Layr-Labs/eigenlayer-contracts/tree/1d926f25f3bd401c4d6552f43aa561782982cb00

> As per deneb upgrade, an additional 2 fields are added to the ExecutionPayloadHeader container which means that it merkelizes differently (theres an extra layer to the tree).

OK so just another layer to the tree? No need to add any fields to struct or anything?
If you are confident this is sufficient, then please regen this based off the correct commit, and we can get deployment done.

PR #395, opened 2024-01-19
Deneb Mainnet Patch

Sidu28:
As per deneb upgrade, an additional 2 fields are added to the ExecutionPayloadHeader container which means that it merkelizes differently (theres an extra layer to the tree). These two fields however are not connected to our proofs and do not change them.  See the deneb updates [here](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/beacon-chain.md#executionpayloadheader). 

This PR adds a switch that allows us to set a deneb fork timestamp twice - once from 0 -> to type(uint64).max and then a second time type(uint64).max -> actual_deneb_fork_timestamp.

Currently, I've modified some of the unit tests to just use the capella values.  I'm gonna make a ticket to add/update all the proofs to deneb spec at some point but not sure now is the time.


ChaoticWalrus:
Looks correct to me now 👍

wadealexc:
Marked this as a draft so we don't accidentally merge.

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -27,7 +27,8 @@ library BeaconChainProofs {
     uint256 internal constant VALIDATOR_FIELD_TREE_HEIGHT = 3;
 
     uint256 internal constant NUM_EXECUTION_PAYLOAD_HEADER_FIELDS = 15;
```
gpsanant, line 29: this changes to 17?

ChaoticWalrus: ah, you are correct!

ChaoticWalrus: this constant isn't used anywhere though 🤷

ChaoticWalrus:
Was asked to review the latest changes. Copy-pasting my response here:
since we know the fork timestamp for Goerli, can’t we just hard-code the value?
I believe this will work as-is though, so am personally OK with the approach you’re suggesting 👍

src/contracts/pods/EigenPodManager.sol
```diff
@@ -233,6 +233,14 @@ contract EigenPodManager is
         _updateBeaconChainOracle(newBeaconChainOracle);
     }
 
+    /**
+    * @notice Sets the denebTimestamp
+    * @dev Callable only by the owner of this contract (i.e. governance)
+    */
+    function setDenebForkTimestamp(uint64 timestamp) external onlyOwner {
+        denebForkTimestamp = timestamp;
```
gpsanant, line 241: event

src/contracts/pods/EigenPodManagerStorage.sol
```diff
@@ -64,6 +64,8 @@ abstract contract EigenPodManagerStorage is IEigenPodManager {
      */
     mapping(address => int256) public podOwnerShares;
 
+    uint64 public constant denebForkTimestamp = 1705473120;
```
ChaoticWalrus, line 67: copy-pasting my comment from elsewhere:
just validated this myself
Teku has the Goerli fork epoch as 231680 here https://github.com/Consensys/teku/pull/7857/files#diff-dc8fb07d95769aadd81ea69e5f034ece1fd19cd7803a09042d08bdb20483cd66R45 and 7413760 is the first slot of this epoch, which I checked with the [beaconcha.in](http://beaconcha.in/) explorer (https://goerli.beaconcha.in/slot/7413760 vs https://goerli.beaconcha.in/epoch/231679)
and `1705473120` is indeed the timestamp for slot `7413760`

src/contracts/pods/EigenPodManager.sol
```diff
@@ -44,6 +44,21 @@ contract EigenPodManager is
...
+    */
+    modifier denebForkEnabled(uint64 newDenebForkTimestamp) {
+        require(newDenebForkTimestamp != 0, "EigenPodManager.denebForkEnabled: cannot set newDenebForkTimestamp to 0");
+        if(newDenebForkTimestamp == type(uint64).max){
+            require(denebForkTimestamp == 0, "EigenPodManager.denebForkEnabled: denebForkTimestamp must not be set yet");
+        } else {
+            require(denebForkTimestamp ==type(uint64).max, "EigenPodManager.denebForkEnabled: Deneb fork timestamp cannot be set");
+        }
+        _;
+    }
```
ChaoticWalrus, line 60: since this modifier is only used once, can't it just be a 'require' statement inside the `setDenebForkTimestamp` function?
I'm also confused about the code structure here -- right now the contract owner can transition `denebForkTimestamp`:
`0 -> type(uint64).max`
or `type(uint64).max -> nonzero`
but not `0 -> nonzero` (which is actually what we'd need on Goerli, at least).

I was imagining just doing:
```
require(newDenebForkTimestamp != 0, "EigenPodManager.denebForkEnabled: cannot set newDenebForkTimestamp to 0");
require(denebForkTimestamp == 0 || denebForkTimestamp == type(uint64).max)
```
is there a reason we can't just do that instead? I think it'd be easier to understand + reason through.

ChaoticWalrus: to clarify: I think what's there would _technically_ meet the need. I just find it confusing and am seeing a particular case in which it seems to raise the complexity unnecessarily, so would prefer a simpler + clearer implementation.

Sidu28: My idea is that the modifer is going to be used exactly twice, once to set it to unit64.max and once to set it to the actual timestamp.  Maybe a simpler way to do it is just make it callable twice with an incrementor.

ChaoticWalrus: perhaps we're talking past each other a bit here.
by "used once" I meant "appear in exactly one place"

I get your intention, but was trying to illustrate my own reasoning here.

I'm going to just mark this as acceptable; as I said above it technically meets the needs here.

Sidu28: oh oh I totally misunderstood,  I will remove the modifier and add it as logic

wadealexc: +1, let's add requires to the method this is used in. This doesn't need to be a modifier.

PR #396, opened 2024-01-19
Feat: add event for beacon chain balance update

ypatil12:
We need to handle the following indexing cases for LST & Native ETH
- podSharesUpdate on balance updates/proving WC
- podSharesUpdate on withdrawing as tokens
- podSharesUpdate on withdrawing as shares
- Deposit event when withdrawing as shares

This is needed for monitoring + will be useful for external parties to properly index shares

Companion doc: https://docs.google.com/document/d/1vjT_i9V1bAclJSS62YpCLAo3KDP2IsHVWCYjdMcfV5w/edit#heading=h.958kgw1yie6h

### DIFF
--- a/src/contracts/pods/EigenPodManager.sol
+++ b/src/contracts/pods/EigenPodManager.sol
@@ -211,11 +211,11 @@ contract EigenPodManager is
             if (shares > currentShareDeficit) {
                 podOwnerShares[podOwner] = 0;
                 shares -= currentShareDeficit;
-                emit PodSharesUpdated(podOwner, currentShareDeficit);
+                emit PodSharesUpdated(podOwner, int256(currentShareDeficit));
             // otherwise get rid of as much deficit as possible, and return early, since there is nothing left over to forward on
             } else {
                 podOwnerShares[podOwner] += int256(shares);
-                emit PodSharesUpdated(podOwner, shares);
+                emit PodSharesUpdated(podOwner, int256(shares));
                 return;
             }
         }
--- a/src/contracts/pods/EigenPodManager.sol
+++ b/src/contracts/pods/EigenPodManager.sol
@@ -211,9 +211,11 @@ contract EigenPodManager is
             if (shares > currentShareDeficit) {
                 podOwnerShares[podOwner] = 0;
                 shares -= currentShareDeficit;
+                emit PodSharesUpdated(podOwner, currentShareDeficit);
             // otherwise get rid of as much deficit as possible, and return early, since there is nothing left over to forward on
             } else {
                 podOwnerShares[podOwner] += int256(shares);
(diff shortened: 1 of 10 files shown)


src/contracts/pods/EigenPodManager.sol
```diff
@@ -140,6 +140,7 @@ contract EigenPodManager is
                 });
             }
         }
+        emit PodBalanceUpdated(podOwner, sharesDelta);
```
ChaoticWalrus, line 143: is this the right place to have this event?
trying to understand if this is all that's needed to fully track shares changes with indexing of events -- do we have a good doc on this? if so, where? if not, can you please make one?

bowenli86: interesting, the event only emits deltas, so to get accurate balance users need to replay and stack all deltas?

is it possible to emit balance directly instead?

ChaoticWalrus: token transfer events similar just log in/out amounts, but I hear you on this @bowenli86 
 this isn't the only place in the code where these share amounts change, so a single event here won't be sufficient regardless. we have other events elsewhere but I'd have to review to remind myself of the details.

Sidu28: Shouldn't we be ok just emitting an event with "updatedPodOwnerShares"/"sharesDelta? Looks like there are events in the DM for changes in delegated shares: OperatorSharesIncreased, OperatorSharesDecreased

ypatil12: I'm compiling a doc on the other events we need. There are some edge cases we are missing, particularly with withdrawals as shares (this is for both LST and Native ETH)

ChaoticWalrus (approved):
LGTM.
Good catches all-around.

Would personally really like to see a bit more test coverage added in a follow-up PR.

ChaoticWalrus:
Ahhh, actually looks like there's a compiler error on this branch?

wadealexc (changes requested):
Please fix compiler errors :heart:

src/test/events/IEigenPodManagerEvents.sol
```diff
@@ -10,4 +10,7 @@ interface IEigenPodManagerEvents {
 
     /// @notice Emitted when `maxPods` value is updated from `previousValue` to `newValue`
     event MaxPodsUpdated(uint256 previousValue, uint256 newValue);
+
+    /// @notice Emitted when the balance of an EigenPod is updated
+    event PodBalanceUpdated(address indexed podOwner, int256 sharesDelta);
```
wadealexc, line 15: Wrong event name

wadealexc (approved):
1. Make sure to rebase!
2. One failing test, see comment above

Otherwise, LGTM.

PR #403, opened 2024-01-24
refactor: decouple AVS<>Operator mapping from DelegationManager

8sunyuan:
Due to contract bytecode size of the DelegationManager exceeding the 24KB limit, this is where I created a draft PR to decouple the operatorToAVS registration mapping in the DM and separate it into the AVSDirectory contract. I also removed the "DelegationManager.<error message>" portion out of the require strings to save bytes as well.

Initially marked as a draft to get more thoughts on this and tests need to be fixed.

image

### DIFF
--- a/src/contracts/core/AVSDirectory.sol
+++ b/src/contracts/core/AVSDirectory.sol
@@ -1,0 +1,179 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
+
+import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
+import "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
+import "@openzeppelin-upgrades/contracts/security/ReentrancyGuardUpgradeable.sol";
+import "../permissions/Pausable.sol";
+import "../libraries/EIP1271SignatureUtils.sol";
+import "./AVSDirectoryStorage.sol";
+
+contract AVSDirectory is
+    Initializable,
+    OwnableUpgradeable,
+    Pausable,
+    AVSDirectoryStorage,
+    ReentrancyGuardUpgradeable
+{
+    // @dev Index for flag that pauses operator register/deregister to avs when set.
+    uint8 internal constant PAUSED_OPERATOR_REGISTER_DEREGISTER_TO_AVS = 0;
+
+    // @dev Chain ID at the time of contract deployment
+    uint256 internal immutable ORIGINAL_CHAIN_ID;
+
+    /*******************************************************************************
+                            INITIALIZING FUNCTIONS
+    *******************************************************************************/
+
+    /**
+     * @dev Initializes the immutable addresses of the strategy mananger, delegationManager, slasher, 
+     * and eigenpodManager contracts
+     */
+    constructor(
+        IDelegationManager _delegation,
+        IStrategyManager _strategyManager,
+        ISlasher _slasher,
+        IEigenPodManager _eigenPodManager
+    ) AVSDirectoryStorage(_delegation, _strategyManager, _slasher, _eigenPodManager) {
+        _disableInitializers();
+        ORIGINAL_CHAIN_ID = block.chainid;
+    }
+
+    /**
+     * @dev Initializes the addresses of the initial owner, pauser registry, and paused status.
+     * minWithdrawalDelayBlocks is set only once here
+     */
+    function initialize(
+        address initialOwner,
+        IPauserRegistry _pauserRegistry,
+        uint256 initialPausedStatus
+    ) external initializer {
+        _initializePauser(_pauserRegistry, initialPausedStatus);
+        _DOMAIN_SEPARATOR = _calculateDomainSeparator();
+        _transferOwnership(initialOwner);
+    }
+
+    /*******************************************************************************
+                            EXTERNAL FUNCTIONS 
+    *******************************************************************************/
+
+
+    /**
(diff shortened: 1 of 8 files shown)


ypatil12:
I am OK with these changes. The AVS<>Operator mapping functionality was added to the DM as we had nowhere else to place it. We could have placed it in the slasher, but we wanted to keep it a stub for now. Separating it out into the AVS Directory makes sense. Keep in mind that the middleware ServiceManager will have to be updated as well

ypatil12:
Since we've added the AVSDirectory, can we bring back the `DelegationManager` in the require strings? What is the byte code size diff if we do?

8sunyuan:
> Since we've added the AVSDirectory, can we bring back the `DelegationManager` in the require strings? What is the byte code size diff if we do?

image
Its about 0.5KB in savings

src/test/unit/DelegationUnit.t.sol
```diff
@@ -1,3236 +1,3236 @@
-// SPDX-License-Identifier: BUSL-1.1
-pragma solidity =0.8.12;
```
ypatil12, line 2: Nit: think most of this file was not meant to be modified?

Can we potentially add tests to a separate `AVSDirectory.unit.t.sol` file?

ypatil12 (commented):
Nit comments. We should also open up a PR in middleware repo to bring up changes to the ServiceManager contract. Happy to do that, just lmk @8sunyuan

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -99,13 +95,6 @@ abstract contract DelegationManagerStorage is IDelegationManager {
     /// See conversation here: https://github.com/Layr-Labs/eigenlayer-contracts/pull/365/files#r1417525270
     address private __deprecated_stakeRegistry;
 
-    /// @notice Mapping: AVS => operator => enum of operator status to the AVS
-    mapping(address => mapping(address => OperatorAVSRegistrationStatus)) public avsOperatorStatus;
-
-    /// @notice Mapping: operator => 32-byte salt => whether or not the salt has already been used by the operator.
-    /// @dev Salt is used in the `registerOperatorToAVS` function.
-    mapping(address => mapping(bytes32 => bool)) public operatorSaltIsSpent;
```
wadealexc, line 107: Do we need to deprecate these variables instead? if they're internal and unused, i don't think it would contribute to bytecode size, though maybe i should double check that.

wadealexc: i'm wondering what happens if we upgrade the current m2 deployment to something that doesn't have these variables when they've already been used -- if we have an m3 down the line that reintroduces variables at these slots, we might find storage collisions?

wadealexc: OTOH maybe it doesnt matter since we'll deprecate Goerli well before m3

8sunyuan: These storage variables have not been set in the DM yet on mainnet so should be safe to remove. For second goerli upgrade however I need to doublecheck...

wadealexc (approved):
TMNT :turtle:

src/test/unit/DelegationUnit.t.sol
8sunyuan: Done, moved old tests to AVSDirectoryUnit.t.sol and left commented out for now
https://github.com/Layr-Labs/eigenlayer-contracts/pull/403/commits/a2dafdc78d22a375f48e4919478c153d24722a62

src/contracts/core/AVSDirectory.sol
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * @dev Initializes the immutable addresses of the strategy mananger, delegationManager, slasher, 
+     * and eigenpodManager contracts
+     */
+    constructor(
+        IDelegationManager _delegation,
+        IStrategyManager _strategyManager,
+        ISlasher _slasher,
+        IEigenPodManager _eigenPodManager
```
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/403/commits/9a08e42bd4d4e27525c280a6c58a81d09888ad87
Kept DelegationManager as we use it to check if an operator is registered

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -0,0 +1,63 @@
...
+    /// @notice The EIP-712 typehash for the contract's domain
+    bytes32 public constant DOMAIN_TYPEHASH =
+        keccak256("EIP712Domain(string name,uint256 chainId,address verifyingContract)");
+
+    /// @notice The EIP-712 typehash for the `Registration` struct used by the contract
+    bytes32 public constant OPERATOR_AVS_REGISTRATION_TYPEHASH =
+        keccak256("OperatorAVSRegistration(address operator,address avs,bytes32 salt,uint256 expiry)");
+
+    /// @notice The DelegationManager contract for EigenLayer
+    IDelegationManager public immutable delegation;
```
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/403/commits/9a08e42bd4d4e27525c280a6c58a81d09888ad87

src/contracts/core/DelegationManagerStorage.sol
ChaoticWalrus: These variables weren't in the initial Goerli deployment either, at least if I recall correctly.
so that would mean super safe to delete.

PR #407, opened 2024-01-26
fix: operator<>AVS mapping tests

8sunyuan:
Added back and fixed AVSDirectory unit tests, added unit test for pausing register/deregister functions as well

Fixed some tests failing on `--fuzz-runs 4096`


wadealexc (approved):
LGTM

PR #410, opened 2024-01-30
fix: failing certora-ci

8sunyuan:


ChaoticWalrus (approved):
Weird, did we change this function signature?
Ah, was perhaps for the events changes from Yash.
Regardless, the changes LGTM

PR #415, opened 2024-02-01
feat: simplify logic

ChaoticWalrus:
discussed this some offline
I think the logic here is significantly less confusing and meets the needs here well


ChaoticWalrus:
something got totally fucked with this when the branch it was targeting got merged. going to nuke this PR and this branch.

PR #422, opened 2024-02-06
Fixed comments

Sidu28:
Fixed comment describing withdrawal index calculation

### DIFF
--- a/src/contracts/libraries/BeaconChainProofs.sol
+++ b/src/contracts/libraries/BeaconChainProofs.sol
@@ -325,11 +325,9 @@ library BeaconChainProofs {
 
         {
             /**
-             * Next we verify the withdrawal fields against the blockRoot:
-             * First we compute the withdrawal_index relative to the blockRoot by concatenating the indexes of all the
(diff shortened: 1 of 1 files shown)


src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -325,11 +325,9 @@ library BeaconChainProofs {
...
         {
             /**
-             * Next we verify the withdrawal fields against the blockRoot:
-             * First we compute the withdrawal_index relative to the blockRoot by concatenating the indexes of all the
-             * intermediate root indexes from the bottom of the sub trees (the withdrawal container) to the top, the blockRoot.
-             * Then we calculate merkleize the withdrawalFields container to calculate the the withdrawalRoot.
-             * Finally we verify the withdrawalRoot against the executionPayloadRoot.
+             * Next we verify the withdrawal fields against the executionPayloadRoot:
+             * First we compute the withdrawal_index, then we merkleize the 
+             * withdrawalFields container to calculate the the withdrawalRoot.
```
wadealexc, line 330: Extra "the" here

otherwise LGTM

wadealexc (approved):
just fix the extra word, then we're good!

PR #423, opened 2024-02-06
feat: Operator shares view for avs-sync

8sunyuan:
Used to reduce gas costs for `_weightOfOperator` calls for avs-syncs. Below is the middleware PR that uses this addition.
https://github.com/Layr-Labs/eigenlayer-middleware/pull/170


wadealexc (approved):
Clean, LGTM!

PR #424, opened 2024-02-06
fix: add staker address to DEPOSIT typehash

ChaoticWalrus:
This provides additional signature replay protection for the `StrategyManager.depositIntoStrategyWithSignature` method

Specifically, it addresses the issue outlined in
https://mirror.xyz/curiousapple.eth/pFqAdW2LiJ-6S4sg_u1z08k4vK6BCJ33LcyXpnNb8yU where some ERC1271 wallets might be vulnerable to "replays" of signatures

While the theoretical "damage" would be ~zero
(allowing someone to deposit and credit the deposit to a user), adding this field to the typehash seems to be best practice, at least.

### DIFF
--- a/src/contracts/core/StrategyManager.sol
+++ b/src/contracts/core/StrategyManager.sol
@@ -146,7 +146,7 @@ contract StrategyManager is
         require(expiry >= block.timestamp, "StrategyManager.depositIntoStrategyWithSignature: signature expired");
         // calculate struct hash, then increment `staker`'s nonce
         uint256 nonce = nonces[staker];
-        bytes32 structHash = keccak256(abi.encode(DEPOSIT_TYPEHASH, strategy, token, amount, nonce, expiry));
+        bytes32 structHash = keccak256(abi.encode(DEPOSIT_TYPEHASH, staker, strategy, token, amount, nonce, expiry));
         unchecked {
             nonces[staker] = nonce + 1;
         }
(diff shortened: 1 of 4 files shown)


ChaoticWalrus:
Note that the churn approver signature in the middleware repo (https://github.com/Layr-Labs/eigenlayer-middleware/blob/m2-mainnet-fixes/src/RegistryCoordinator.sol#L907) exhibits similar behavior, but since in that instance there can be only one signing party, it does not appear to present the same kind of potential issue.
All other instances in our codebase appear to use the signer's address as part of the digest hash (see https://github.com/Layr-Labs/eigenlayer-contracts/blob/m2-mainnet-fixes/src/contracts/core/DelegationManager.sol#L189-L193 and https://github.com/Layr-Labs/eigenlayer-contracts/blob/m2-mainnet-fixes/src/contracts/core/AVSDirectory.sol#L83-L88, in particular)

src/contracts/core/StrategyManager.sol
```diff
@@ -146,7 +146,7 @@ contract StrategyManager is
         require(expiry >= block.timestamp, "StrategyManager.depositIntoStrategyWithSignature: signature expired");
         // calculate struct hash, then increment `staker`'s nonce
         uint256 nonce = nonces[staker];
-        bytes32 structHash = keccak256(abi.encode(DEPOSIT_TYPEHASH, strategy, token, amount, nonce, expiry));
+        bytes32 structHash = keccak256(abi.encode(DEPOSIT_TYPEHASH, staker, strategy, token, amount, nonce, expiry));
```
wadealexc, line 149: Should we add msg.sender as well?

PR #427, opened 2024-02-08
fix: updated beacon chain proof docs and correct error message

Sidu28:
Updated BeaconChainProofs.md to reflect new effective balance proofs.  Seems like the docs are updated in some places but not in others.  Also fixed an incorrect proof diagram and comment in the BeaconChainProofs.sol


wadealexc (approved):
added some fixes for formatting. thanks for the change!

PR #428, opened 2024-02-08
feat: slight refactor to make better use of strategybase hooks

wadealexc:


src/contracts/strategies/StrategyBase.sol
8sunyuan, line 1: Maybe note in the natspec @dev comments in deposit() and withdraw() that token validation is done in the _beforeDeposit and _beforeWithdraw hooks. These functions only explicitly handle shares now w/o the hooks.

8sunyuan (approved):
LGTM

gpsanant (approved):

src/contracts/strategies/StrategyBase.sol
wadealexc: [651d22a](https://github.com/Layr-Labs/eigenlayer-contracts/pull/428/commits/651d22aa944336e4fb53520aecfcf7b697d4a4ce)

PR #430, opened 2024-02-09
Update implementation addresses for goerliv2 deployment

0x0aa0:
Update implementation addresses for goerliv2 deployment


wadealexc:
Closing since we're now pointing at the public docs site :(

NimaVaziri:
@wadealexc if we're pointing at the public docs site should we have a disclaimer around these tables? or maybe even better yet, not include the addresses here?

wadealexc:
> @wadealexc if we're pointing at the public docs site should we have a disclaimer around these tables? or maybe even better yet, not include the addresses here?

Yes, that's what I meant -- on the `m2-mainnet-fixes` branch, the README has been edited so the tables/addresses are not included, and a link points to the docs site instead.

Those changes will get merged onto this branch sometime this week -- just waiting to wrap up some audit stuff!

PR #432, opened 2024-02-09
docs: update README to point to deployment info

wadealexc:
* we are now using the main eigenlayer docs site for deployment info: https://docs.eigenlayer.xyz/eigenlayer/deployed-contracts


ChaoticWalrus (approved):
RIP contract addresses on GH 😢

PR #433, opened 2024-02-12
test: Fork Integration Testing refactor

8sunyuan:
With the different upgrade paths between Goerli and Mainnet, this is an attempt at having more upgrade testing by using the utils parser `ExistingDeploymentParser` and deploying in `IntegrationDeployer.setUp()` depending on the chainid that tests are being run on.

To run upgrade integration tests

1. Setup anvil node
`anvil -f $RPC_GOERLI`
2. Run integration tests 
`forge test --match-contract Integration -f http://127.0.0.1:8545`

### DIFF
--- a/src/test/integration/IntegrationDeployer.t.sol
+++ b/src/test/integration/IntegrationDeployer.t.sol
@@ -18,6 +18,8 @@ import "src/contracts/pods/EigenPod.sol";
 import "src/contracts/pods/DelayedWithdrawalRouter.sol";
 import "src/contracts/permissions/PauserRegistry.sol";
 
+import "script/utils/ExistingDeploymentParser.sol";
+
 import "src/test/mocks/EmptyContract.sol";
 import "src/test/mocks/ETHDepositMock.sol";
 import "src/test/integration/mocks/BeaconChainOracleMock.t.sol";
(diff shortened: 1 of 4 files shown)


wadealexc (commented):
Would be nice to have something that can target mainnet, if possible

wadealexc:
> Would be nice to have something that can target mainnet, if possible

Wait nvm there's a setupMainnet method...

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -131,13 +121,27 @@ abstract contract IntegrationDeployer is Test, IUserDeployer {
...
+        string memory deploymentInfoPath;
+        if (block.chainid == 1) {
+            _setUpMainnet();
+        } else if (block.chainid == 5) {
+            deploymentInfoPath = "script/utils/Goerli_current_deploy.config.json";
+            _setUpGoerli(deploymentInfoPath);
+        } else if (block.chainid == 31337) {
+            _setUpLocal();
+        } else {
+            revert();
```
wadealexc, line 133: Probably good to have a message here, otherwise this would be a meganightmare to debug

8sunyuan:
> > Would be nice to have something that can target mainnet, if possible
> 
> Wait nvm there's a setupMainnet method...

Yeah not sure if this is the best place/method to do the upgrade testing so haven't added that in yet. But it made the most sense as there are no mocks setup(outside of beacon stuff) and this will make us take more advantage of our integration testing.

ypatil12 (commented):
Awesome start! I think there are two parts to this idea of integration tests post-upgrade

1. Normal interaction on M1 -> Upgrade -> Continue interaction on M2
2. Upgrade, and ensure that all interaction on M1 can be done on M2

I think this framework does 2, but not necessarily 1? Couple of example scenarios:
- queueing a withdrawal on M1 and completing on M2 (tested via `WithdrawalMigration.t.sol`)
- depositing on M1 and then doing the entire LST lifecycle on M2 (delegate, queue, complete)
- creating a pod in M1 and pointing WC and then going through the entire pod lifecycle on M2 (verifying WC, balance updates, partial/full withdrawals)
- setting an admin var in M1 and then resetting on M2

For 2, I think the main things we should test are normal interaction, which we can easily do with this framework and the already written integration tests, and setting all the admin variables, which could catch the setting of `withdrawalDelayBlocks`

ypatil12 (commented):
Also, think we should update the CI to run the integration and off of a fork in addition to the local execution!

PR #434, opened 2024-02-12
feat: cancel AVS registration salt 

0x0aa0:
Adds a function for an operator to mark a salt as spent

### DIFF
--- a/src/contracts/core/AVSDirectory.sol
+++ b/src/contracts/core/AVSDirectory.sol
@@ -127,6 +127,14 @@ contract AVSDirectory is
         emit AVSMetadataURIUpdated(msg.sender, metadataURI);
     }
 
+    /**
+     * @notice Called by an operator to cancel a salt that has been used to register with an AVS.
+     * @param salt A unique and single use value associated with the approver signature.
+     */
+    function cancelSalt(bytes32 salt) external {
+        operatorSaltIsSpent[msg.sender][salt] = true;
+    }
+
     /*******************************************************************************
                             VIEW FUNCTIONS
     *******************************************************************************/


src/contracts/core/AVSDirectory.sol
```diff
@@ -127,6 +127,14 @@ contract AVSDirectory is
         emit AVSMetadataURIUpdated(msg.sender, metadataURI);
     }
 
+    /**
+     * @notice Called by an operator to cancel a salt that has been used to register with an AVS.
+     * @param salt A unique and single use value associated with the approver signature.
+     */
+    function cancelSalt(bytes32 salt) external {
```
wadealexc, line 134: Please add a require that the salt is not already spent

wadealexc (changes requested):
Needs some docs. I can push those when the PR is looking finalized

Also needs some unit tests:
- Check that a salt cannot be cancelled twice
- Check that a salt cannot be cancelled if an operator has registered using it
- Check that after a salt has been cancelled, an operator cannot be registered using it

src/contracts/core/AVSDirectory.sol
wadealexc: [2f921f0](https://github.com/Layr-Labs/eigenlayer-contracts/pull/434/commits/2f921f0845762c40dc5dbe73efe177a0dea94936)

wadealexc (approved):
LGTM, but would like someone to double-check my changes!

8sunyuan (approved):
Tests LGTM! Could be good to fuzz the operator address but looks comprehensive enough.

PR #435, opened 2024-02-12
fix: correct eip712 typehash

ChaoticWalrus:
`delegationApprover` was missing from the `DELEGATION_APPROVAL_TYPEHASH` definition


wadealexc (approved):
LGTM

PR #439, opened 2024-02-13
fix: Add back setMinWithdrawalDelayBlocks

8sunyuan:
`minWithdrawalDelayBLocks` only being set in initialize function. Adding back the admin setter here


wadealexc (approved):
LGTM

PR #443, opened 2024-02-14
updated m2_deploy_from_scratch to use avsDirectory

samlaf:
Fix the m2_deploy_from_scratch script which wasn't updated with the m2-mainnet-fixes
We are using this in eigencert. Perhaps we shouldn't and should be using DeployOpenEigenlayer @stevennevins ?
In the meantime we need these fixes in eigencert though.


samlaf:
@wadealexc any idea what happened to this PR when you renamed the branches? It says "base automatically changed to m2-mainnet", but now it's saying "merging into dev", and this was a 1-3 file change PR but now there's 75 files being merged.

wadealexc:
I'm guessing it's a very old PR that needs to be rebased

samlaf:
Closing in favor of https://github.com/Layr-Labs/eigenlayer-contracts/pull/449
Branches got messed up, so cherry-picked last commit on top of dev in the other commit.
Only problem is we won't have this fix in the testnet-goerli branch now... if needed I could cherry-pick I guess but this would break all the other PRs on top, which would need to be rebased. Would be nice to clean the stuff on top of testnet-goerli branch at some point.
image

PR #458, opened 2024-02-25
feat: add initial interfaces

gpsanant:
Let's agree on the initial payment interfaces.

TODO:
- multiproofs of token tree

### DIFF
--- a/src/contracts/interfaces/IClaimingManager.sol
+++ b/src/contracts/interfaces/IClaimingManager.sol
@@ -1,0 +1,111 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+/**
+ * @title Interface for the `IClaimingManager` contract.
+ * @author Layr Labs, Inc.
+ * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
+ */
+interface IClaimingManager {
+
+    /// STRUCTS ///
+
+    struct DistributionRoot {
+        uint32 activatedAfter; // timestamp after which the root can be claimed against
+        bytes32 root; // merkle root of the distribution
+    }
+
+    struct PaymentMerkleClaim {
+        IERC20 token; 
+        uint256 amount;
+        address recipient; // Explicit recipient of the claim
+
+        uint32 rootIndex; // The index of the root in the list of roots for this token
+        uint32 leafIndex; // The index of the leaf in the merkle tree for this root
+
+        bytes proof;
+    }
+
+    /// EVENTS /// 
+
+    event PaymentUpdaterSet(address oldPaymentUpdater, address newPaymentUpdater);
+    event ActivationDelaySet(uint32 oldActivationDelay, uint32 newActivationDelay);
+    event GlobalCommissionBipsSet(uint16 oldGlobalCommissionBips, uint16 newGlobalCommissionBips);
+    event RecipientSet(address account, address recipient);
+    event RootSubmitted(bytes32 root, uint32 paymentsCalculatedUntilTimestamp, uint32 activatedAfter);
+    event PaymentClaimed(IERC20 token, address recipient, uint256 amount);
+    
+    /// VIEW FUNCTIONS ///
+
+    /// @notice The address of the entity that can update the contract with new merkle roots
+    function paymentUpdater() external view returns (address);
+
+    /// @notice Delay in timestamp before a posted root can be claimed against
+    function activationDelay() external view returns (uint32);
+
+    /// @notice Mapping: account => the address of the entity that can claim payments on behalf of the accounts
+    function recipients(address account) external view returns (address);
+
+    /// @notice Mapping: recipient => token => total amount claimed
+    function cumulativeClaimed(address recipient, IERC20 token) external view returns (uint256);
+
+    /// @notice the commission for all operators across all avss
+    function globalCommissionBips() external view returns (uint16);
+
+    /// EXTERNAL FUNCTIONS ///
+
+    /**
+     * @notice Initializes the contract
+     * @param initialOwner The address of the initial owner of the contract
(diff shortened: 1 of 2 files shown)


src/contracts/interfaces/IClaimingManager.sol
```diff
@@ -0,0 +1,111 @@
...
+    event ActivationDelaySet(uint32 oldActivationDelay, uint32 newActivationDelay);
+    event GlobalCommissionBipsSet(uint16 oldGlobalCommissionBips, uint16 newGlobalCommissionBips);
+    event RecipientSet(address account, address recipient);
+    event RootSubmitted(bytes32 root, uint32 paymentsCalculatedUntilTimestamp, uint32 activatedAfter);
+    event PaymentClaimed(IERC20 token, address recipient, uint256 amount);
+    
+    /// VIEW FUNCTIONS ///
+
+    /// @notice The address of the entity that can update the contract with new merkle roots
+    function paymentUpdater() external view returns (address);
```
ChaoticWalrus, line 43: owner() could be added to this interface, but is probably unnecessary

src/contracts/interfaces/IClaimingManager.sol
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /// EVENTS /// 
+
+    event PaymentUpdaterSet(address oldPaymentUpdater, address newPaymentUpdater);
+    event ActivationDelaySet(uint32 oldActivationDelay, uint32 newActivationDelay);
+    event GlobalCommissionBipsSet(uint16 oldGlobalCommissionBips, uint16 newGlobalCommissionBips);
+    event RecipientSet(address account, address recipient);
+    event RootSubmitted(bytes32 root, uint32 paymentsCalculatedUntilTimestamp, uint32 activatedAfter);
+    event PaymentClaimed(IERC20 token, address recipient, uint256 amount);
```
ChaoticWalrus, line 38: I think consistent ordering would be good. seems you are using (token, amount, recipient) in the `PaymentMerkleClaim` struct definition above, at least

ChaoticWalrus, line 38: likely want to mark all the addresses + perhaps the token as 'indexed'

src/contracts/interfaces/IClaimingManager.sol
```diff
@@ -0,0 +1,111 @@
...
+    
+    /// VIEW FUNCTIONS ///
+
```
ChaoticWalrus, line 41: a couple convenience functions + struct which I could see us adding:
```
struct MerkleLeaf {
        IERC20 token; 
        uint256 amount;
        address recipient;
}
``` 
^MerkleLeaf could then be used inside the `PaymentMerkleClaim` struct definition
```
function findLeafHash(MerkleLeaf calldata leaf) external view returns (bytes32)
function checkClaim(PaymentMerkleClaim calldata claim) external view returns (bool)
```

these might be overkill, but I could see them being somewhat useful, at least.

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+    /// EVENTS ///
+
+    event RangePaymentCreated(RangePayment rangePayment);
+
+    /// VIEW FUNCTIONS ///
+
+    /// @notice The address of the entity that can update the contract with new merkle roots
+    function claimingManager() external view returns (IClaimingManager);
```
ChaoticWalrus, line 33: I think this comment is inaccurate?
the usage here would be the address where payments are forwarded, right?

src/contracts/interfaces/IClaimingManager.sol
```diff
@@ -0,0 +1,111 @@
...
+     */
+    function setRecipient(address account, address recipient) external;
+
+    /**
+     * @notice Creates a new distribution root
+     * @param root The merkle root of the distribution
+     * @param paymentsCalculatedUntilTimestamp The timestamp until which payments have been calculated
+     * @dev Only callable by the paymentUpdater
+     */
+    function submitRoot(bytes32 root, uint32 paymentsCalculatedUntilTimestamp) external;
```
ChaoticWalrus, line 104: might also want a only-owner function for cancelling roots before they activate or something?

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,61 @@
...
+
+    /// @notice The interval in seconds at which the calculation for range payment distribution is done. ranges must be aligned to multiples of this interval
+    function CALCULATION_INTERVAL_SECONDS() external view returns (uint32);
```
ChaoticWalrus, line 36: if this might change, perhaps it doesn't make sense to be a constant / all caps?

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,61 @@
...
+     */
+    function initialize(address initialOwner) external;
+    
+    /**
+     * @notice Creates a new range payment on behalf of an AVS
+     * @param rangePayment The range payment being created
+     * @dev The end time of the range must be at most 3 months in the future
+     * @dev The start time of the range must be after the configured lower bound
+     * @dev The tokens are sent to the claiming manager contract
+     */
```
ChaoticWalrus, line 59: you may want to refer to the constants instead of their values in this comment. e.g. `3 months` => `MAX_FUTURE_RANGE_END seconds` or similar

src/contracts/interfaces/IClaimingManager.sol
gpsanant: tbh prefer pausing here

ChaoticWalrus: pausing as-in pausing of claiming?
seems like then you'd be forced to upgrade to recover, so I'm not seeing how that's really better.
we could do both?

ChaoticWalrus:
closing this in favor of https://github.com/Layr-Labs/eigenlayer-contracts/pull/470

PR #462, opened 2024-03-01
test: fork testing framework

wadealexc:
draft/wip - do not merge!


src/test/integration/mocks/BeaconChainMock.t.sol
```diff
@@ -191,17 +198,21 @@ contract BeaconChainMock is Test {
         proof.validatorFields[0][BeaconChainProofs.VALIDATOR_BALANCE_INDEX] = 
             _toLittleEndianUint64(validator.effectiveBalanceGwei);
 
+        emit log("10");
```
wadealexc, line 201: oh man, whatever led you to having to do this, i feel for you haha

wadealexc: is it fixed now? we should remove if so

8sunyuan: fixed now, much pain

src/test/integration/users/User.t.sol
```diff
@@ -268,6 +276,17 @@ contract User is Test {
         return _completeQueuedWithdrawal(withdrawal, false);
     }
 
+    /// @notice We set the proof generation start time to be after the timestamp that pod restaking is activated
+    /// We do this to prevent proofIsForValidTimestamp modifier from reverting
+    function activateRestaking() public {
+        pod.activateRestaking();
+        // To ensure oracle proof timestamp is greater than pod.mostRecentWithdrawalTimestamp
+        cheats.warp(block.timestamp + 10);
+        timeMachine.createSnapshot();
```
wadealexc, line 285: Why do you snapshot here?

8sunyuan: TimeMachine will jump back before the timestamp that pod.activateRestaking() is called and then the `proofIsForValidTimestamp` modifier will revert. I added this here to work around that

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -23,24 +24,22 @@ import "src/test/mocks/ETHDepositMock.sol";
...
-    Slasher slasher;
-    IBeacon eigenPodBeacon;
-    EigenPod pod;
-    DelayedWithdrawalRouter delayedWithdrawalRouter;
+    // Fork ids for specific fork tests
+    bool isUpgraded;
+    uint256 mainnetForkBlock = 19_280_000;
+    uint256 mainnetForkId;
+    uint256 goerliForkBlock = 10_575_000;
+    uint256 goerliForkId;
```
wadealexc, line 41: Let's get rid of the goerli references, since that's unstable (and also slightly out of date with the dev branch)

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -269,16 +326,206 @@ abstract contract IntegrationDeployer is Test, IUserDeployer {
...
         eigenPodManager.setDenebForkTimestamp(type(uint64).max);
     }
 
+    /**
+     * @notice deploy current implementation contracts and upgrade the existing proxy EigenLayer contracts
+     * on Goerli. Setup for integration tests on Goerli fork.
+     * 
+     * Note that beacon chain oracle and eth deposit contracts are mocked and pointed to different addresses for these tests.
+     */
+    function _upgradeGoerliContracts() public virtual {
```
wadealexc, line 342: Let's nix this method (see above comment about goerli)

wadealexc (commented):
Initial pass comments:

- needs to actually upgrade the strategy implementation all the LST strats are using
- needs to make sure that `initialize` methods in the upgraded contracts are "respected" (we can't call `initialize` again, but we should make sure their corresponding variables are set in the upgraded code)

wadealexc:
Oh, and we also need to fetch the strategy contracts from mainnet! Right now it looks like we're just deploying `_newStrategyAndToken` within `_deployOrFetchContracts` (i change that method's name from `_configEigenLayerContracts` because it was a bit unclear what it was doing)

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -330,8 +493,105 @@ abstract contract IntegrationDeployer is Test, IUserDeployer {
...
+        } else if (forkType == MAINNET) {
+            cheats.selectFork(mainnetForkId);
+            string memory deploymentInfoPath = "script/configs/mainnet/Mainnet_current_deploy.config.json";
+            _parseDeployedContracts(deploymentInfoPath);
+
+            // Unpause to enable deposits and withdrawals for initializing random user state
+            cheats.prank(eigenLayerPauserReg.unpauser());
+            strategyManager.unpause(0);
+
+            _newStrategyAndToken("Strategy1Token", "str1", 10e50, address(this)); // initialSupply, owner
```
wadealexc, line 531: ^ is there a reason we don't fetch these from mainnet?

8sunyuan:
Closed, updated PR is here
https://github.com/Layr-Labs/eigenlayer-contracts/pull/483

PR #463, opened 2024-03-01
chore: remove maxPods

8sunyuan:
We are no longer limiting maximum number of deployed eigenpods. EigenPodManager on mainnet has `maxPods` set to uint256 max https://etherscan.io/address/0x91E677b07F7AF907ec9a428aafA9fc14a0d3A338#readProxyContract

This is removing unused `maxPods` code


wadealexc:
While we're at it, may as well deprecate the other state variables right?

wadealexc:
Oh I see, the diff is somewhat large. makes sense to split the changes up in that case

ypatil12:
Just flagging - this is something we want to document as a non breaking change in the release notes. There isn't a dependency on the BE, but the SDK should take an update with the new ABI

PR #467, opened 2024-03-06
feat: poc for checkpointed partial withdrawals

wadealexc:
Pod Owners create a partial withdrawal checkpoint by calling `EigenPod.startNonRestakedWithdrawal`. This sets `EigenPod.currentCheckpoint` to the current block timestamp, and records the pod's current:
* `activeValidatorCount`: the number of validators with proven withdrawal credentials in the ACTIVE state in the pod. This becomes the `proofsRemaining` required to complete the checkpoint.
* `eligibleBalance`: the unaccounted for balance in the pod as of the checkpoint's creation (`this.balance - nonBeaconChainETHBalanceWei - withdrawableRestakedExecutionLayerWei`)

A checkpoint's progress is tracked using the following state variables:
```solidity
/// (optimized for a single SLOAD)
struct Checkpoint {
    uint24 proofsRemaining;
    uint232 eligibleBalance;
}

/// @dev The current non-restaked withdrawal checkpoint, and the timestamp used to prove
/// the checkpoint's validator set.
///
/// Note: If there is no currently-active checkpoint, `currentCheckpointTimestamp` will be 0
Checkpoint currentCheckpoint;
uint64 currentCheckpointTimestamp;
    
/// @dev Tracks whether a validator's balance has been accounted for at a checkpoint
/// timestamp. We keep track of this to ensure that a checkpoint's `proofsRemaining`
/// is only decremented for proofs on validators who were active when the checkpoint
/// was created.
///
/// checkpointTimestamp -> validator pubkeyHash -> valid proof submitted
mapping(uint64 => mapping(bytes32 => bool)) validatorProven;
```

`currentCheckpoint` keeps track of the `Checkpoint` being proven at the oracle timestamp given by `currentCheckpointTimestamp`.

`validatorProven` keeps track of any validators whose balance we've accounted for at specific oracle timestamps.

When any proof is submitted where the proof's `oracleTimestamp` matches this checkpoint timestamp, the checkpoint is updated. Users can use both `verifyBalanceUpdates` or `verifyAndProcessWithdrawals` to make progress on the `proofsRemaining` for the current checkpoint. When any validator is proven at a given timestamp, `validatorProven[oracleTimestamp][pubkeyHash]` is set to true so that only one proof can be submitted for that validator at that timestamp.

Additional notes:
* Adds a new state variable `activeValidatorCount`. This is incremented when a new set of withdrawal credentials are proven, and decremented when a full withdrawal is proven. When a checkpoint is created, the current value of `activeValidatorCount` is used to determine how many proofs are needed to finish the checkpoint and withdraw the eligible balance.
* Adds new state variable `cachedStateRoots`, which caches state root proofs for any of the 3 proving methods. If a user is submitting a proof using an oracle timestamp that has already been used, they won't need to submit another `StateRootProof` (or read from the beacon chain oracle).
* This PR gets rid of some of the timing restrictions for balance update proofs, because the checkpointing system implies we allow "stale" proofs on unaccounted-for balance. To account for this, I used the same "staleness check" to instead determine whether a balance update should calculate a share delta and send that to the EigenPodManager.
* Partial withdrawal proofs have been removed from `verifyAndProcessWithdrawals`, since those can now be accounted for using the checkpoint system (even if a checkpoint is created after a full withdrawal proof is submitted!)

Finally - the withdrawal flow right now is the biggest WIP, and I need to think a bit more about beacon state changes and transitions!

### DIFF
--- a/src/contracts/pods/EigenPod.sol
+++ b/src/contracts/pods/EigenPod.sol
@@ -94,6 +94,31 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
      /// @notice This variable tracks the total amount of partial withdrawals claimed via merkle proofs prior to a switch to ZK proofs for claiming partial withdrawals
     uint64 public sumOfPartialWithdrawalsClaimedGwei;
 
+    /// @notice number of validators proven to have withdrawal credentials pointed at this pod, minus
+    /// the number of validators who have already proven full withdrawals from this pod.
+    uint256 activeValidatorCount;
+    
+    /// @dev oracleTimestamp -> cached proven beaconStateRoot
+    /// NOTE: We could take this even further, and cache all the way down to the validator tree
+    mapping(uint64 => bytes32) cachedStateRoots;
+
+    struct Checkpoint {
+        uint24 proofsRemaining;
+        uint232 eligibleBalance;
+    }
+
+    /// @dev If we have an active partial withdrawal checkpoint, `currentCheckpoint`
+    ///      is the oracle timestamp being used to prove that checkpoint's validator set
+    ///
+    ///      This value can be used to look up additional checkpoint info in 
+    ///      the mappings below.
+    uint64 currentCheckpoint;
+
+    /// @dev oracleTimestamp -> partial withdrawal checkpoint
+    mapping(uint64 => Checkpoint) partialWithdrawCheckpoints;
+    /// @dev oracleTimestamp -> validator pubkeyHash -> valid proof
+    mapping(uint64 => mapping(bytes32 => bool)) validatorProven;
+
     modifier onlyEigenPodManager() {
         require(msg.sender == address(eigenPodManager), "EigenPod.onlyEigenPodManager: not eigenPodManager");
         _;
@@ -173,9 +198,42 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     }
 
     /**
-     * @notice This function records an update (either increase or decrease) in a validator's balance.
+     * @dev Begin the partial withdrawal process by checkpointing the pod's current active validators
+     *      and unaccounted-for balance.
+     */
+    function startPartialWithdrawal() external onlyEigenPodOwner() /**onlyWhenNotPaused*/ {
+        uint256 eligibleBalance = 
+            address(this).balance 
+                - nonBeaconChainETHBalanceWei 
+                - withdrawableRestakedExecutionLayerGwei;
+        require(eligibleBalance != 0, "No eligible balance to withdraw");
+
+        // If we have no active validators, there are no additional steps needed!
+        // Withdraw the eligible balanace as a delayed withdrawal.
+        if (activeValidatorCount == 0) {
+            _sendETH_AsDelayedWithdrawal(podOwner, eligibleBalance);
+
+            // emit event
+        } else {
+            // Otherwise, create a new partial withdrawal checkpoint, overwriting
+            // a previous block's checkpoint if it exists.
+            uint64 oracleTimestamp = uint64(block.timestamp);
+            require(currentCheckpoint != oracleTimestamp, "checkpoint already created this block");
+            
+            currentCheckpoint = oracleTimestamp;
+            partialWithdrawCheckpoints[oracleTimestamp] = Checkpoint({
+                proofsRemaining: uint24(activeValidatorCount),
+                eligibleBalance: uint232(eligibleBalance)
+            });
+
+            // emit event
+        }
+    }
+
+    /**
+     * @notice This function processes a proof for one or more validators' balances.
+     *         TODO - more comments
      * @param oracleTimestamp The oracleTimestamp whose state root the proof will be proven against.
-     *        Must be within `VERIFY_BALANCE_UPDATE_WINDOW_SECONDS` of the current block.
      * @param validatorIndices is the list of indices of the validators being proven, refer to consensus specs 
      * @param stateRootProof proves a `beaconStateRoot` against a block root fetched from the oracle
      * @param validatorFieldsProofs proofs against the `beaconStateRoot` for each validator in `validatorFields`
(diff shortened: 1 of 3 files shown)


wadealexc:
Fix needed for this order of operations:
- checkpoint created at time K
- validator exited and proven using oracle timestamp from prior to K

src/contracts/pods/EigenPod.sol
```diff
@@ -194,34 +252,67 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-                stateRootProof.beaconStateRoot,
+                beaconStateRoot,
                 validatorFieldsProofs[i], // Use validator fields proof because contains the effective balance
                 validatorFields[i]
             );
+
+            // If we're working on a partial withdrawal checkpoint and we haven't seen
+            // this validator yet, mark the validator as proven
+            if (updateCheckpoint && !validatorProven[oracleTimestamp][pubkeyHash]) {
+                validatorProven[oracleTimestamp][pubkeyHash] = true;
```
gpsanant, line 289: we have to make sure that this validator was, in fact, active when the checkpoint was created

src/contracts/pods/EigenPod.sol
```diff
@@ -194,34 +252,67 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+                checkpoint.proofsRemaining--;
+            }
+        }
+
+        // If we have updated validator balances, send them to the EigenPodManager
+        if (sharesDeltaGwei != 0) {
+            eigenPodManager.recordBeaconChainETHBalanceUpdate(podOwner, sharesDeltaGwei * int256(GWEI_TO_WEI));
```
gpsanant, line 296: we need to make sure that balance updates cannot be proven out of order

src/contracts/pods/EigenPod.sol
```diff
@@ -244,36 +335,59 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-        // DelayedWithdrawalRouter
-        if (withdrawalSummary.amountToSendGwei != 0) {
-            _sendETH_AsDelayedWithdrawal(podOwner, withdrawalSummary.amountToSendGwei * GWEI_TO_WEI);
+            // If we created a checkpoint after this validator was exited from the beacon chain
+            // and we haven't seen a proof for this validator, mark them as proven and decrement
+            // their `amountToQueueGwei` from the checkpoint's eligible balance
+            if (updateCheckpoint && !validatorProven[oracleTimestamp][pubkeyHash]) {
+                validatorProven[oracleTimestamp][pubkeyHash] = true;
+                checkpoint.proofsRemaining--;
+                checkpoint.eligibleBalance -= verifiedWithdrawal.amountToQueueGwei;
```
gpsanant, line 369: hmm i think this might open up the

1. repoint val w 32 eth
2. fully withdraw it from beaconchain for 32 eth
3. deposit to it again for 1 eth
4. fully withdraw again from beacon for 1 eth
5. start checkpoint
6. prove 1 eth full withdrawal

now u get 32 eth partial withdrawal for free unstaked. whereas before you would eventually have to restake the 32 eth again.

src/contracts/pods/EigenPod.sol
```diff
@@ -244,36 +335,59 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
             "EigenPod.verifyAndProcessWithdrawals: inputs must be same length"
         );
 
-        // Verify passed-in beaconStateRoot against oracle-provided block root:
-        BeaconChainProofs.verifyStateRootAgainstLatestBlockRoot({
-            latestBlockRoot: eigenPodManager.getBlockRootAtTimestamp(oracleTimestamp),
-            beaconStateRoot: stateRootProof.beaconStateRoot,
-            stateRootProof: stateRootProof.proof
-        });
+        // Check for a relevant partial withdrawal checkpoint using the oracle timestamp
```
gpsanant, line 338: active validator count should be decremented somewhere here

src/contracts/pods/EigenPod.sol
```diff
@@ -584,15 +706,21 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-         * Withdrawal processing should only be performed for "ACTIVE" or "WITHDRAWN" validators.
-         * (WITHDRAWN is allowed because technically you can deposit to a validator even after it exits)
+         * Withdrawal processing should only be performed for "ACTIVE" validators
+         * WITHDRAWN validators will have their ETH automatically sent to this contract
+         * and can exit these funds via the partial withdrawal flow.
          */
         require(
-            _validatorPubkeyHashToInfo[validatorPubkeyHash].status != VALIDATOR_STATUS.INACTIVE,
-            "EigenPod._verifyAndProcessWithdrawal: Validator never proven to have withdrawal credentials pointed to this contract"
+            _validatorPubkeyHashToInfo[validatorPubkeyHash].status == VALIDATOR_STATUS.ACTIVE,
```
gpsanant, line 716: oh i think this might get rid of my edge case above. the issue now is that its unclear what happens to pods which have validators that have never been repointed but have fully withdrawn. 

i think this is a common enough case that we can't say "don't do that" like we would for redeposits...

src/contracts/pods/EigenPod.sol
wadealexc: Correct! In `verifyWithdrawalCredentials`, I mark a validator proven if there's a checkpoint timestamp after their oracle timestamp.

So when they hit this line, they'll already be marked `true` in `validatorProven`.

src/contracts/pods/EigenPod.sol
wadealexc: Indeed - I modified `_verifyBalanceUpdate` to return a `sharesDeltaGwei` of 0 if `oracleTimestamp < validatorInfo.mostRecentBalanceUpdateTimestamp`.

So it's possible to submit balance proofs out of order, but they won't actually result in a share delta.

src/contracts/pods/EigenPod.sol
wadealexc: I'm a little confused about the order of operations - at what point are withdrawal credentials proven? Because you need to do that to prove a withdrawal.

src/contracts/pods/EigenPod.sol
wadealexc: I'm gonna take "never been repointed" to mean that they haven't verified withdrawal credentials?

The answer is that that ether is "unaccounted for" and will be included in the `eligibleBalance` calculated when you start a partial withdrawal.

So they can be withdrawn through the checkpoint process.

src/contracts/pods/EigenPod.sol
gpsanant: step 1

src/contracts/pods/EigenPod.sol
gpsanant: ah yep, sounds good

src/contracts/pods/EigenPod.sol
wadealexc: oh you're talking about the "current issue" i found with eigenpods?

checkpointing doesn't "open up" the issue again. it's always been there and this doesn't change that

src/contracts/pods/EigenPod.sol
gpsanant: ah didn't see that. great

src/contracts/pods/EigenPod.sol
gpsanant: oooh nice

src/contracts/pods/EigenPod.sol
gpsanant: But it's sort of amplified here. In the current system you must restake the 32 eth again, whereas you would be able to withdraw it in this scenario

wadealexc: > you must restake the 32 eth again

by "restake" you mean that you need to provide another withdrawal proof, causing the shares to be re-awarded (and only then withdrawing)

i'm not convinced that's much better. i think the root problem is the same for both, and that's how both of these interact with how slashing is supposed to function with the withdrawal queue

If you can suddenly change the reported backing amount, you can reduce your slashable amount on a whim.

wadealexc: Whether you need to "reenter the system" or the eth just leaves the pod, IMO not the crux of the system. Both would need to be accounted for when we build slashing.

ChaoticWalrus:
Closing this PR as stale. It has been superseded by https://github.com/Layr-Labs/eigenlayer-contracts/pull/515

PR #470, opened 2024-03-08
feat: add contract interface for Range Payments

ChaoticWalrus:


src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+ * @author Layr Labs, Inc.
+ * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
+ */
+interface IPaymentCoordinator {
+
+    /// STRUCTS ///
+	struct RangePayment {
+        IStrategy[] strategies;
+        // relative weights of shares in the strategies
+        uint256[] weights;
```
wadealexc, line 18: IIRC we decided to do 1 array of StrategyParams, rather than 2 (less input validation required)

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * @title Interface for the `IPaymentCoordinator` contract.
+ * @author Layr Labs, Inc.
+ * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
+ */
+interface IPaymentCoordinator {
+
+    /// STRUCTS ///
+	struct RangePayment {
```
gpsanant, line 15: indent

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    struct MerkleLeaf {
+        IERC20 token; 
+        uint256 amount;
+        // Explicit recipient of the claim
+        address recipient;
+    }
+
+    struct PaymentMerkleClaim {
```
gpsanant, line 40: This should likely be 

```solidity
struct PaymentMerkleClaim {
        MerkleLeaf[] leaves;
        // The index of the root in the list of roots
        uint32 rootIndex;
        bytes accountTreeProof;
        // The index of the leaf in the merkle tree for this root
        uint32[] leafIndices;
        bytes[] tokenTreeProofs;
    }
```

or something to enable ammortizing the account tree proof when claiming multiple tokens

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+    struct DistributionRoot {
+        // timestamp after which the root can be claimed against
+        uint32 activatedAfter;
+        // merkle root of the distribution
+        bytes32 root;
+    }
+
+    struct MerkleLeaf {
```
gpsanant, line 33: `ClaimsTreeMerkleLeaf`

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+    function owner() external view returns (address);
+
+    /**
+     * @notice The interval in seconds at which the calculation for range payment distribution is done.
+     * @dev Payment durations must be multiples of this interval.
+     */
+    function calculationIntervalSeconds() external view returns (uint32);
+
+    /// @notice The maximum amount of time that a range payment can end in the future
+    function MAX_PAYMENT_DURATION() external view returns (uint32);
```
gpsanant, line 74: name doesn't really describe what this is...

what about

`MAX_RANGE_PAYMENT_FUTURE_OFFSET()`

my naming skills undefeated

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+    function recipientOf(address account) external view returns (address);
+
+    /// @notice Mapping: recipient => token => total amount claimed
+    function cumulativeClaimed(address recipient, IERC20 token) external view returns (uint256);
+
+    /// @notice the commission for all operators across all avss
+    function globalOperatorCommissionBips() external view returns (uint16);
+
+    /// @notice returns the hash of the leaf
+    function findLeafHash(MerkleLeaf calldata leaf) external view returns (bytes32);
```
gpsanant, line 92: `calculateLeafHash`

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+     * @notice Creates a new range payment on behalf of an AVS, to be split amongst the
+     * set of stakers delegated to operators who are registered to the `avs`
+     * @param rangePayment The range payment being created
+     * @dev Expected to be called by the ServiceManager of the AVS on behalf of which the payment is being made
+     * @dev The duration of the `rangePayment` cannot exceed `MAX_PAYMENT_DURATION`
+     * @dev The tokens are sent to the `claimingManager` contract
+     * @dev This function will revert if the `rangePayment` is malformed,
+     * e.g. if the `strategies` and `weights` arrays are of non-equal lengths
+     */
+	function payForRange(RangePayment calldata rangePayment) external;
```
gpsanant, line 118: indent

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * @notice similar to `payForRange` except the payment is split amongst *all* stakers
+     * rather than just those delegated to operators who are registered to a single avs
+     */
+    function payAllForRange(RangePayment calldata rangePayment) external;
+
```
gpsanant, line 125: `setPaymentUpdater`

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+     * @param paymentsCalculatedUntilTimestamp The timestamp until which payments have been calculated
+     * @dev Only callable by the paymentUpdater
+     */
+    function submitRoot(bytes32 root, uint32 paymentsCalculatedUntilTimestamp) external;
+
+    /**
+     * @notice Claims payments for the given claims
+     * @param claims The claims to be processed
+     */
+    function processClaims(PaymentMerkleClaim[] calldata claims) external;
```
gpsanant, line 161: would need to change this to account for the proof change i suggested

src/contracts/interfaces/IPaymentCoordinator.sol
gpsanant: Oh interested, we're doing start/duration. this works then

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+
+    /// EVENTS ///
+
+    event RangePaymentCreated(RangePayment rangePayment);
+    event PaymentUpdaterSet(address indexed oldPaymentUpdater, address indexed newPaymentUpdater);
+    event ActivationDelaySet(uint32 oldActivationDelay, uint32 newActivationDelay);
+    event GlobalCommissionBipsSet(uint16 oldGlobalCommissionBips, uint16 newGlobalCommissionBips);
+    event RecipientSet(address indexed account, address indexed recipient);
+    event RootSubmitted(bytes32 root, uint32 paymentsCalculatedUntilTimestamp, uint32 activatedAfter);
+    event PaymentClaimed(MerkleLeaf leaf);
```
gpsanant, line 57: needs a root index

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+     */
+    function setRecipient(address account, address recipient) external;
+
+    /**
+     * @notice Creates a new distribution root
+     * @param root The merkle root of the distribution
+     * @param paymentsCalculatedUntilTimestamp The timestamp until which payments have been calculated
+     * @dev Only callable by the paymentUpdater
+     */
+    function submitRoot(bytes32 root, uint32 paymentsCalculatedUntilTimestamp) external;
```
gpsanant, line 155: might.... get rid of the `paymentsCalculatedUntilTimestamp`. Also my spec includes a check against beacon state roots. are we going to do that?

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+     * @notice The interval in seconds at which the calculation for range payment distribution is done.
+     * @dev Payment durations must be multiples of this interval.
+     */
+    function calculationIntervalSeconds() external view returns (uint32);
+
+    /// @notice The maximum amount of time that a range payment can end in the future
+    function MAX_PAYMENT_DURATION() external view returns (uint32);
+
+    /// @notice The lower bound for the start of a range payment
+    function LOWER_BOUND_START_RANGE() external view returns (uint32);
```
gpsanant, line 77: looks like this will be a max amount of intervals that this can be in the past

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+        // The index of the root in the list of roots
+        uint32 rootIndex;
+        // The index of the leaf in the merkle tree for this root
+        uint32 leafIndex;
+        bytes proof;
+    }
+
+    /// EVENTS ///
+
+    event RangePaymentCreated(RangePayment rangePayment);
```
gpsanant, line 51: needs to include AVS

src/contracts/interfaces/IPaymentCoordinator.sol
gpsanant: we should call thsee multipliers

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: looks fine on my machine

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: I'm fine with this for now
will talk to you about the beacon state root part -- I definitely didn't see that, myself

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: https://github.com/Layr-Labs/eigenlayer-contracts/pull/470/commits/1d438f42e18ff6e494ed572df71b024a4e965de6

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: https://github.com/Layr-Labs/eigenlayer-contracts/pull/470/commits/1d438f42e18ff6e494ed572df71b024a4e965de6
I'm fine with this for now

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: idk, I'm fine with changing it later, doesn't feel super clear rn

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: added in https://github.com/Layr-Labs/eigenlayer-contracts/pull/470/commits/1d438f42e18ff6e494ed572df71b024a4e965de6

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: addressed both in https://github.com/Layr-Labs/eigenlayer-contracts/pull/470/commits/1d438f42e18ff6e494ed572df71b024a4e965de6

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: sure, change made in https://github.com/Layr-Labs/eigenlayer-contracts/pull/470/commits/1d438f42e18ff6e494ed572df71b024a4e965de6

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: https://github.com/Layr-Labs/eigenlayer-contracts/pull/470/commits/1d438f42e18ff6e494ed572df71b024a4e965de6

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,162 @@
...
+    /// EXTERNAL FUNCTIONS ///
+
+    /**
+     * @notice Initializes the contract
+     * @param initialOwner The address of the initial owner of the contract
+     * @param _activationDelay Delay in timestamp before a posted root can be claimed against
+     * @param _globalCommissionBips The commission for all operators across all avss
+     * @dev Only callable once
+     */
+    function initialize(address initialOwner, uint32 _activationDelay, uint16 _globalCommissionBips) external;
```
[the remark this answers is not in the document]
ChaoticWalrus: added in https://github.com/Layr-Labs/eigenlayer-contracts/pull/470/commits/1d438f42e18ff6e494ed572df71b024a4e965de6

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: ahh, missed this, please make the change yourself

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: added in https://github.com/Layr-Labs/eigenlayer-contracts/pull/470/commits/1d438f42e18ff6e494ed572df71b024a4e965de6

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /// EVENTS ///
+
+    event RangePaymentCreated(address indexed avs, RangePayment rangePayment);
+    event PaymentUpdaterSet(address indexed oldPaymentUpdater, address indexed newPaymentUpdater);
+    event ActivationDelaySet(uint32 oldActivationDelay, uint32 newActivationDelay);
+    event GlobalCommissionBipsSet(uint16 oldGlobalCommissionBips, uint16 newGlobalCommissionBips);
+    event RecipientSet(address indexed account, address indexed recipient);
+    event RootSubmitted(bytes32 root, uint32 paymentsCalculatedUntilTimestamp, uint32 activatedAfter);
```
ypatil12, line 61: Would make sense to specify the exact window for which this root was submitted, especially since we want these roots to be verified by 3rd parties. So something like:
`event RootSubmitted(bytes 32root, uint32 paymentCalculationStartTimestamp, uint32 paymentCalculationEndTimestamp, uint32 activatedAfter)`

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
...
+     */
+    function setGlobalOperatorCommission(uint16 _globalCommissionBips) external;
+
+    /**
+     * @notice Sets the address of the entity that can claim payments on behalf of the account
+     * @param account The account whose recipient is being set
+     * @param recipient The address of the entity that can claim payments on behalf of the account
+     * @dev Only callable by the account or their paymentRecipient
+     */
+    function setRecipient(address account, address recipient) external;
```
ypatil12, line 158: I think there is a thread somewhere discussing _only_ the `account` being able to update the recipient, and not the `paymentRecipient`. An argument could be made for either but just want to make sure we are all aligned

src/contracts/interfaces/IPaymentCoordinator.sol
ypatil12: I guess the assumption is that there will be no gaps in calculation periods & payments begin whenever payments are activated, but would be useful for convenience

src/contracts/interfaces/IPaymentCoordinator.sol
wadealexc: yeah the indent is totally screwy for me too

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
...
+ * @title Interface for the `IPaymentCoordinator` contract.
+ * @author Layr Labs, Inc.
+ * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
+ */
+interface IPaymentCoordinator {
+
+    /// STRUCTS ///
+    struct StrategyAndMultiplier {
+        IStrategy strategy;
+        uint96 multiplier;
```
scotthconner, line 17: Inline docs about what the multiplier is used for.

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "./IStrategy.sol";
+
+/**
+ * @title Interface for the `IPaymentCoordinator` contract.
+ * @author Layr Labs, Inc.
+ * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
```
scotthconner, line 10: Probably some context as to what role this plays in the protocol. It'd be nice to be able to read the top of a file to understand what it does. Just a nit-pick.

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
...
+interface IPaymentCoordinator {
+
+    /// STRUCTS ///
+    struct StrategyAndMultiplier {
+        IStrategy strategy;
+        uint96 multiplier;
+    }
+	struct RangePayment {
+        // Strategies & relative weights of shares in the strategies
+        StrategyAndMultiplier strategiesAndMultlipliers;
```
scotthconner, line 21: Does this need to be an array?

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
...
+        IERC20 token;
+        uint256 amount;
+        uint256 startTimestamp;
+        // duration in seconds
+        uint256 duration;
```
scotthconner, line 26: Would it make sense to have this number be "weeks"? If it was, we wouldn't have to check the input and we could eliminate an error condition completely.

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
...
+        uint32 activatedAfter;
+        // merkle root of the distribution
+        bytes32 root;
+    }
+
+    struct ClaimsTreeMerkleLeaf {
+        IERC20 token; 
+        uint256 amount;
+        // Explicit recipient of the claim
+        address recipient;
```
scotthconner, line 40: This we want the leaves to be specific to the owner, and we want the active claimer to be able to claim it. That is, the payments are attached to the owner and not the claimer.

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
...
+
+    struct PaymentMerkleClaim {
+        ClaimsTreeMerkleLeaf[] leaves;
+        // The index of the root in the list of roots
+        uint32 rootIndex;
+        // proof of the claimaint's account root in the Merkle tree
+        bytes accountTreeProof;
+        // The indices and proofs of the leafs in the claimaint's merkle tree for this root
+        uint32[] leafIndices;
+        bytes[] tokenTreeProofs;
```
scotthconner, line 51: This struct will likely also need the "owner" (operator/staker) address as input if the claimer (aka msg.sender) is not the address used in the merkle creation.

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /// EVENTS ///
+
+    event RangePaymentCreated(address indexed avs, RangePayment rangePayment);
+    event PaymentUpdaterSet(address indexed oldPaymentUpdater, address indexed newPaymentUpdater);
```
scotthconner, line 57: Probably worth logging the governor address for this too.

src/contracts/interfaces/IPaymentCoordinator.sol
scotthconner: lets be explicit. I agree

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
...
+    event RootSubmitted(bytes32 root, uint32 paymentsCalculatedUntilTimestamp, uint32 activatedAfter);
+    event PaymentClaimed(ClaimsTreeMerkleLeaf leaf);
+
+    /// VIEW FUNCTIONS ///
+
+    /// @notice The address of the entity that can update the contract with new merkle roots
+    function paymentUpdater() external view returns (address);
+
+    // @notice The owner of this contract
+    function owner() external view returns (address);
```
scotthconner, line 70: do we simply use Ownable from OZ for this?

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
...
+    function activationDelay() external view returns (uint32);
+
+    /// @notice Mapping: account => the address of the entity to which new payments are directed on behalf of the account
+    function recipientOf(address account) external view returns (address);
+
+    /// @notice Mapping: recipient => token => total amount claimed
+    function cumulativeClaimed(address recipient, IERC20 token) external view returns (uint256);
+
+    /// @notice the commission for all operators across all avss
+    function globalOperatorCommissionBips() external view returns (uint16);
```
scotthconner, line 94: If we name it this way, then we are baking in the assumption that commissions are global, would we want something like operatorComissionBips(address operator) and then simply return the same value for all inputs for now?

In this second case, we dont have the assumption that its global and always the same for every operator. The code that uses this interface can then be ignorant of it, and more flexible.

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,173 @@
...
+     * @dev This function will revert if the `rangePayment` is malformed,
+     * e.g. if the `strategies` and `weights` arrays are of non-equal lengths
+     */
+    function payForRange(RangePayment calldata rangePayment) external;
+
+    /**
+     * @notice similar to `payForRange` except the payment is split amongst *all* stakers
+     * rather than just those delegated to operators who are registered to a single avs
+     */
+    function payAllForRange(RangePayment calldata rangePayment) external;
```
scotthconner, line 130: I think we decided to remove this explicitly from the use case support for MVP.

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: yes lol good catch

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: I think it depends if we think the `calculationIntervalSeconds` is fixed (at one week) or will change.
using a duration in seconds + doing input sanitization would more fluidly support changing this interval; as long as the old new interval is equally divisible by the new interval, you get total backwards compatibility while enabling "new" (finer grain) control.
I'm OK switching this to weeks, just summarizing what I think is the best reasoning for not doing so.

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: is this a comment about what to call the field?
I think `earner` would be most appropriate, given the other discussions I've had about this. it would be the staker/operator address, not necessarily the (only) address capable of making the claim.

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: this should also have an `accountTreeIndex` field, I think.

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: I would think the 'owner' or 'earner' address should be in the leaves themselves. The access control for claims can then function as either:

A. ensure all leaves have same 'earner', and check that `msg.sender = earner` or `msg.sender = claimerOf(earner)` OR
B. ensure that for each leaf, `msg.sender = earner` or `msg.sender = claimerOf(earner)`

technically the 'earner' is duplicate data, so there's likely a way to structure the data differently that is somewhat more efficient, but having the leaves contain _not_ 64 bytes of data is useful for preventing the ability to prove internal nodes in the Merkle tree

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: like, the current owner?
guess I am assuming this contract inherits from OZ's Ownable or similar

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: works for me, will make the change.

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: I certainly think so!
this doesn't have to go in the interface, just added it for "completeness" / since the interface refers to the owner elsewhere.

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: > If we name it this way, then we are baking in the assumption that commissions are global, would we want something like operatorComissionBips(address operator) and then simply return the same value for all inputs for now?

instead, or in addition?
current interface implies that this value could change, at minimum
I'm fine with adding `operatorComissionBips(address operator)`; I think it's a good idea for future compatibility, as you suggest, but wouldn't be useful until `globalOperatorCommissionBips` is deprecated, or at least only used as a default.

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: I didn't think so?
that would be great; my understanding was only on agreement that this should be a separate function if it's needed.

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: yes, we agreed on only the `account` being able to change this. I'll make the change to the comment, thanks for noting this.

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: added in https://github.com/Layr-Labs/eigenlayer-contracts/pull/470/commits/f4a3836b095ff5e45823afe0cec3b42106f9c109
resolving this thread; a new one can be opened for further changes

src/contracts/interfaces/IPaymentCoordinator.sol
ChaoticWalrus: wording fixed inhttps://github.com/Layr-Labs/eigenlayer-contracts/pull/470/commits/f4a3836b095ff5e45823afe0cec3b42106f9c109

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,179 @@
...
+
+    /// EVENTS ///
+
+    event RangePaymentCreated(address indexed avs, RangePayment rangePayment);
+    event PaymentUpdaterSet(address indexed oldPaymentUpdater, address indexed newPaymentUpdater);
+    event ActivationDelaySet(uint32 oldActivationDelay, uint32 newActivationDelay);
+    event GlobalCommissionBipsSet(uint16 oldGlobalCommissionBips, uint16 newGlobalCommissionBips);
+    event RecipientSet(address indexed account, address indexed recipient);
+    event RootSubmitted(bytes32 root, uint32 paymentCalculationStartTimestamp, uint32 paymentCalculationEndTimestamp, uint32 activatedAt);
+    event PaymentClaimed(ClaimsTreeMerkleLeaf leaf);
```
gpsanant, line 68: need `address claimer, uint256 rootIndex` here

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,179 @@
...
+        // The staker or operator who earned these tokens
+        address earner;
+    }
+
+    struct PaymentMerkleClaim {
+        ClaimsTreeMerkleLeaf[] leaves;
+        // The index of the root in the list of roots
+        uint32 rootIndex;
+        // proof of the claimaint's account root in the Merkle tree
+        uint32 accountIndex;
```
gpsanant, line 53: prob change `account -> earner` after recent discussions

src/contracts/interfaces/IPaymentCoordinator.sol
gpsanant: I don't think we've removed it either. It's still allowed for permissioned entities

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,179 @@
...
+     * @dev This function will revert if the `rangePayment` is malformed,
+     * e.g. if the `strategies` and `weights` arrays are of non-equal lengths
+     */
+    function payForRange(RangePayment calldata rangePayment) external;
+
+    /**
+     * @notice similar to `payForRange` except the payment is split amongst *all* stakers
+     * rather than just those delegated to operators who are registered to a single avs
+     */
+    function payAllForRange(RangePayment calldata rangePayment) external;
```
gpsanant, line 136: Does this emit a range payment with `avs = address(0)`?

ChaoticWalrus: you're talking about the event, right?
`avs` is no longer a field in the RangePayment struct
Perhaps we should have a 2nd event here?

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,179 @@
...
+    /// STRUCTS ///
+    struct StrategyAndMultiplier {
+        IStrategy strategy;
+        // weight used to compare shares in multiple strategies against one another
+        uint96 multiplier;
+    }
+    struct RangePayment {
+        // Strategies & relative weights of shares in the strategies
+        StrategyAndMultiplier[] strategiesAndMultlipliers;
+        IERC20 token;
```
8sunyuan, line 27: Could probably be good to have a from address here for the transfer so AVSs aren't forced into putting all their payments into the ServiceManager first.

ChaoticWalrus: funny you should mention that.
an earlier version actually had an 'avs' input parameter, which was removed, but would serve the purpose of not having to route through the ServiceManager.
with the way token approvals function, I don't think there's a safe way to take a 'from' parameter here, unless we have very different ideas on how this would work.

ChaoticWalrus:
Closing this stale PR in favor of https://github.com/Layr-Labs/eigenlayer-contracts/pull/514

PR #474, opened 2024-03-12
feat: track active validator count in pods

wadealexc:


src/contracts/pods/EigenPod.sol
```diff
@@ -91,9 +91,12 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
     /// @notice This variable tracks any ETH deposited into this contract via the `receive` fallback function
     uint256 public nonBeaconChainETHBalanceWei;
 
-     /// @notice This variable tracks the total amount of partial withdrawals claimed via merkle proofs prior to a switch to ZK proofs for claiming partial withdrawals
+    /// @notice This variable tracks the total amount of partial withdrawals claimed via merkle proofs prior to a switch to ZK proofs for claiming partial withdrawals
     uint64 public sumOfPartialWithdrawalsClaimedGwei;
 
+    /// @notice Number of validators with proven withdrawal credentials, who do not have proven full withdrawals
+    uint256 activeValidatorCount;
```
ChaoticWalrus, line 98: we should make visibility explicit
I would think 'public' for sake of simplicity

src/contracts/pods/EigenPod.sol
```diff
@@ -697,9 +701,12 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
          * Finally, the validator is fully withdrawn. Update their status and place in state:
          */
 
-        validatorInfo.restakedBalanceGwei = 0;
-        validatorInfo.status = VALIDATOR_STATUS.WITHDRAWN;
+        if (validatorInfo.status != VALIDATOR_STATUS.WITHDRAWN) {
```
ChaoticWalrus, line 704: this is equivalent to `if (validatorInfo.status == VALIDATOR_STATUS.ACTIVE)`, because we require that the status is not INACTIVE above.
seems OK 👍

ChaoticWalrus (approved):
My only nit is the variable visibility; please address that and then I'm cool with merging.
Thanks for putting this together.

src/contracts/pods/EigenPod.sol
wadealexc: public adds bytecode

i'm fine with this but... i figured we should stop defaulting to public

idk maybe this is important?

wadealexc: actually ive decided

public is worse because if people use it we can't ever get rid of it

let's make it public IFF we release the partial withdrawal feature that uses this var.

PR #477, opened 2024-03-18
fix: certora prover ci

8sunyuan:
Updates to certora-cli uses cli flag `--optimistic_fallback` instead of `prover_args` with -optimisticFallback as value.
Here is example of error https://github.com/Layr-Labs/eigenlayer-contracts/actions/runs/8330391177/job/22794927969


ChaoticWalrus (approved):
LGTM, thanks, will merge immediately.

PR #479, opened 2024-03-21
fix: off-by-one error

ChaoticWalrus:
Fixes an off-by-one error that could have caused withdrawals to get stuck in EigenPods.
This is because withdrawals are processed at the _end_ of the block in which they are posted (rather than at the top of the block), so the strict inequality of `require(timestamp > mostRecentWithdrawalTimestamp)` could cause withdrawals that are processed in the same block as `mostRecentWithdrawalTimestamp` is set to get stuck.
The fix is simply changing the inequality to be non-strict. PR contains a new regression test for the behavior.

### DIFF
--- a/src/contracts/pods/EigenPod.sol
+++ b/src/contracts/pods/EigenPod.sol
@@ -118,11 +118,11 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         _;
     }
 
-    /// @notice Checks that `timestamp` is strictly greater than the value stored in `mostRecentWithdrawalTimestamp`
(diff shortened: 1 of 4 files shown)


wadealexc (commented):
Scenario I'm thinking through --

- I have an m1 validator with credentials pointed at the pod
- I exit the validator
- The same block where the ETH hits the pod, i hit activateRestaking/withdrawBeforeRestaking

I then attempt to verify withdrawal credentials for this validator, using the above block's timestamp as the oracle timestamp

It seems like I can do this, as long as the oracle timestamp is within `VERIFY_BALANCE_UPDATE_WINDOW_SECONDS` of my WC proof.

wadealexc (approved):
yep

ChaoticWalrus:
ty for the comment, agreed it's very confusing, so it's helpful to have it worked through in this PR thread.

PR #480, opened 2024-03-21
feat: m2 mainnet upgrade script

8sunyuan:
**M2-mainnet LFGGGG**

Script for deploying and doing sanity checks found under `script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol`
Run script locally with 
```
anvil --fork-url $RPC_MAINNET
forge script script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol --rpc-url http://127.0.0.1:8545 --private-key $PRIVATE_KEY --broadcast -vvvv
```

Uses two config json files:
`script/output/mainnet/M1_deployment_mainnet_2023_6_9.json` 
Used for existing mainnet contract addresses (slightly modified to fix json parse errors)

`script/configs/mainnet/M2_mainnet_upgrade.config.json`
Used for expected initalization values. Sanity checks against these config values after calling `_deployImplementationContracts()` and `_simulateUpgrade()`


script/configs/mainnet/M2_mainnet_upgrade.config.json
```diff
@@ -0,0 +1,46 @@
...
+  "eigenPodManager": {
+    "init_paused_status": 0,
+    "deneb_fork_timestamp": "1710338135"
+  },
+  "delayedWithdrawalRouter": {
+    "init_paused_status": 0,
+    "init_withdrawalDelayBlocks": 50400
+  },
+  "ethPOSDepositAddress": "0x00000000219ab540356cBB839Cbe05303d7705Fa",
+  "beaconOracleAddress": "0x343907185b71aDF0eBa9567538314396aa985442"
```
ChaoticWalrus, line 45: just reviewed all params in this config; LGTM

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
```diff
@@ -0,0 +1,130 @@
...
+        _deployImplementationContracts();
+
+        // STOP RECORDING TRANSACTIONS FOR DEPLOYMENT
+        vm.stopBroadcast();
+
+        // Simulate upgrade of contracts to new implementations
+        _simulateUpgrade();
+
+        // Sanity Checks
+        _verifyContractPointers();
```
ChaoticWalrus, line 33: I guess we could also add the AVSDirectory.delegationManager() to this?

ChaoticWalrus: oh, this is already done, incredible.

script/output/mainnet/M1_deployment_mainnet_2023_6_9.json
```diff
@@ -1,6 +1,9 @@
 {
   "addresses": {
+    "avsDirectory": "0x0000000000000000000000000000000000000000",
+    "avsDirectoryImplementation": "0x0000000000000000000000000000000000000000",
     "baseStrategyImplementation": "0xdfdA04f980bE6A64E3607c95Ca26012Ab9aA46d3",
+    "beaconOracleAddress": "0x0000000000000000000000000000000000000000",
```
ChaoticWalrus, line 6: should this be non-zero? I get that the `avsDirectory` is set to the zero address because it is not deployed yet.
I would guess that this doesn't matter, just searching for anything, honestly.

ChaoticWalrus (changes requested):
This is looking excellent!
I'm marking this as 'request changes' for the moment, but purely because I want to ensure this gets multiple reviews before it's merged; I couldn't find anything amiss, myself.

script/output/mainnet/M1_deployment_mainnet_2023_6_9.json
8sunyuan: Ah yes I added this because the `_parseDeployedContracts` helper will throw an error when trying to parse it out. It was used for the holesky deployment so I added these 0 addresses here based on the current mainnet state.
The beaconOracle address gets overwritten when `_parseInitialDeploymentParams` is called right after.

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
```diff
@@ -0,0 +1,308 @@
...
+pragma solidity =0.8.12;
+
+import "../../utils/ExistingDeploymentParser.sol";
+import "../../utils/TimelockEncoding.sol";
+import "../../utils/Multisend.sol";
+
+/**
+ * @notice Script used for the first deployment of EigenLayer core contracts to Holesky
+ * anvil --fork-url $RPC_MAINNET
+ * forge script script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol --rpc-url http://127.0.0.1:8545 --private-key $PRIVATE_KEY --broadcast -vvvv
```
ypatil12, line 11: The scripts LGTM, but when running the command locally (without broadcast), I get `Multiple contracts in the target path. Please specify the contract name with `--tc ContractName`.

wadealexc (commented):
Do we still need `script/deploy/mainnet/M2Deploy.s.sol`?

script/utils/ExistingDeploymentParser.sol
```diff
@@ -116,6 +117,7 @@ contract ExistingDeploymentParser is Script, Test {
         operationsMultisig = stdJson.readAddress(existingDeploymentData, ".parameters.operationsMultisig");
         communityMultisig = stdJson.readAddress(existingDeploymentData, ".parameters.communityMultisig");
         pauserMultisig = stdJson.readAddress(existingDeploymentData, ".parameters.pauserMultisig");
+        timelock = stdJson.readAddress(existingDeploymentData, ".parameters.timelock");
```
wadealexc, line 120: I see this is added here, but don't see a matching line added to `_parseInitialDeploymentParams`. Is that important?

script/configs/mainnet/M2_mainnet_upgrade.config.json
```diff
@@ -0,0 +1,46 @@
...
+    "init_minWithdrawalDelayBlocks": 50400
+  },
+  "avsDirectory": {
+    "init_paused_status": 0
+  },
+  "slasher": {
+    "init_paused_status": 115792089237316195423570985008687907853269984665640564039457584007913129639935
+  },
+  "eigenPod": {
+    "MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR": 32000000000000000000,
```
wadealexc, line 33: Is this correct? it looks to be denominated in wei, not gwei

script/configs/mainnet/M2_mainnet_upgrade.config.json
```diff
@@ -0,0 +1,46 @@
...
+    "GENESIS_TIME": 1606824000
```
wadealexc, line 34: Where are we getting this? Not that it's wrong, but I want a source to double check this.

script/configs/mainnet/M2_mainnet_upgrade.config.json
```diff
@@ -0,0 +1,46 @@
...
+  },
+  "eigenPodManager": {
+    "init_paused_status": 0,
+    "deneb_fork_timestamp": "1710338135"
```
wadealexc, line 38: Where are we getting this? Not that it's wrong, but I want a source to double check this.

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
```diff
@@ -0,0 +1,308 @@
...
+            IETHPOSDeposit(ETHPOSDepositAddress),
+            eigenPodBeacon,
+            strategyManager,
+            slasher,
+            delegationManager
+        );
+        delayedWithdrawalRouterImplementation = new DelayedWithdrawalRouter(eigenPodManager);
+    }
+
+    function _simulateUpgrade() internal {
```
wadealexc, line 85: Where does migrating queued withdrawals come in? Do we do that separately?

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
```diff
@@ -0,0 +1,308 @@
...
+
+    function test_queueUpgrade() external {
+        _parseDeployedContracts("script/output/mainnet/M1_deployment_mainnet_2023_6_9.json");
+        _parseInitialDeploymentParams("script/configs/mainnet/M2_mainnet_upgrade.config.json");
+
+        // TODO: fill in correct addresses
+        // simulate deploying contracts
+        _deployImplementationContracts();
+        // confirmed correct address
+        address beaconChainOracle = 0x343907185b71aDF0eBa9567538314396aa985442;
```
wadealexc, line 149: I don't think this is needed? The deployment parser picks this up from the JSON config

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
ChaoticWalrus: yes, that would be completed separately, post-upgrade

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
```diff
@@ -0,0 +1,308 @@
...
+            0, // value
+            abi.encodeWithSelector(Pausable.unpause.selector, 0)
+        );
+
+        // set the min withdrawal delay blocks on the DelegationManager
+        txs[7] = Tx(
+            address(delegationManager), 
+            0, // value
+            abi.encodeWithSelector(DelegationManager.setMinWithdrawalDelayBlocks.selector, DELEGATION_MANAGER_MIN_WITHDRAWAL_DELAY_BLOCKS)
+        );
```
wadealexc, line 223: I assume these transactions are executed en masse, but in case they aren't (or just to avoid footgun) - let's swap the order of these two transactions and make sure the withdrawal delay is configured before we unpause the contract.

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
```diff
@@ -0,0 +1,308 @@
...
+            0, // value
+            abi.encodeWithSelector(EigenPodManager.updateBeaconChainOracle.selector, IBeaconChainOracle(beaconChainOracle))
+        );
+
+        // set Deneb fork timestamp on EigenPodManager
+        txs[10] = Tx(
+            address(eigenPodManager), 
+            0, // value
+            abi.encodeWithSelector(EigenPodManager.setDenebForkTimestamp.selector, EIGENPOD_MANAGER_DENEB_FORK_TIMESTAMP)
+        );
```
wadealexc, line 244: Same request here - configure, then unpause.

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
ChaoticWalrus: good point, will resolve

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
ChaoticWalrus: they are all completed as part of a single transaction. I can change the order though

wadealexc (changes requested):
Have left several questions and some requested changes

Should be fine once those are addressed.

I glossed over some of the multisig interactions/queue upgrade test -- will make time to review that stuff more closely when it's time to do the actual upgrade. 

The deploy stuff looks good otherwise.

script/configs/mainnet/M2_mainnet_upgrade.config.json
ChaoticWalrus: The Geth release has this https://github.com/ethereum/go-ethereum/releases/tag/v1.13.12

8sunyuan: https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/fork.md#configuration
https://beaconcha.in/epoch/269568
Fork epoch is `269568` and I converted to the timestamp

script/configs/mainnet/M2_mainnet_upgrade.config.json
ChaoticWalrus: the phase0 spec has this https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md but it appears to be the 'minimum' time, not the set time.
looking at the explorer https://beaconcha.in/slot/0 I see `1606824023` instead, and there is a blog post confirming it.
so this should be `1606824023`; good catch

script/configs/mainnet/M2_mainnet_upgrade.config.json
8sunyuan: Found the actual config in the specs here https://github.com/ethereum/consensus-specs/blob/68dceb7cda9f7745689fda007d712e6b128f94ac/configs/mainnet.yaml#L52

script/configs/mainnet/M2_mainnet_upgrade.config.json
8sunyuan: https://github.com/ethereum/consensus-specs/blob/68dceb7cda9f7745689fda007d712e6b128f94ac/configs/mainnet.yaml#L52

ChaoticWalrus: fixed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/480/commits/02f7aff812fa5e73c6f0fdfe056b622444112797

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
ChaoticWalrus: fixed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/480/commits/9b7a0c33c94c20f2e5bb8abfacf115e15bf24426

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
ChaoticWalrus: adjusted the ordering in https://github.com/Layr-Labs/eigenlayer-contracts/pull/480/commits/9b7a0c33c94c20f2e5bb8abfacf115e15bf24426

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
ChaoticWalrus: adjusted the ordering in https://github.com/Layr-Labs/eigenlayer-contracts/pull/480/commits/9b7a0c33c94c20f2e5bb8abfacf115e15bf24426

script/configs/mainnet/M2_mainnet_upgrade.config.json
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/480/commits/888d3c6b02a8f0f2261ca7590b0852f3c98cccd0

8sunyuan: Added a modulo check in case this value gets overflowed again

script/configs/mainnet/M2_mainnet_upgrade.config.json
wadealexc: Interesting -- do we know if the holesky value is correct?

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
wadealexc: thanks, i feel better about this :)

script/configs/mainnet/M2_mainnet_upgrade.config.json
8sunyuan: https://github.com/eth-clients/holesky
https://holesky.beaconcha.in/epoch/0
I checked the holesky client page initially, seems to align with the explorer

wadealexc (approved):
Fixes look good!

script/configs/mainnet/M2_mainnet_upgrade.config.json
ChaoticWalrus: LGTM, thank you

script/deploy/mainnet/M2_Mainnet_Upgrade.s.sol
ChaoticWalrus: this was addressed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/480/commits/888d3c6b02a8f0f2261ca7590b0852f3c98cccd0 😄

ChaoticWalrus (approved):
seems we have agreement this is looking good. going to merge it.

PR #482, opened 2024-03-22
feat: add m2 upgrade output file

ChaoticWalrus:
- Add output file with implementation addresses for M2 upgrade
- Fix a few minor naming inconsistencies with writing to + reading from JSON
- Parse implementation addresses in upgrade simulation, rather than simulating deploying the implementations


ChaoticWalrus:
note: upgrade queuing will still require an appropriate timestamp to be set.

8sunyuan (approved):
Implementations look good to me as well as AVSDirectory TUPS deployment

wadealexc (commented):
I get `[FAIL. Reason: revert: You are on the wrong chain for this config] test_queueUpgrade() (gas: 16134)` when running `forge test` on this branch by default 

i can get this to work if i pass in the valid fork url, but is there a way we can adjust this so it doesn't auto-fail `forge test`? We'll have a red line when we merge.

Edit: actually it appears tests are running and passing in the CI? Do I need to do some extra config in my local env then?

ChaoticWalrus:
> i can get this to work if i pass in the valid fork url, but is there a way we can adjust this so it doesn't auto-fail `forge test`? We'll have a red line when we merge.

I think CI doesn't run tests outside of the /test/ folder? Happy to make changes if necessary though!

wadealexc (commented):
We have (somewhere) a json file floating around that can be used by a pauser to trigger a blanket pause.

Could we update that json file to include the AVSDirectory, and also add the file to a script or some other file here?

script/output/mainnet/M2_mainnet_upgrade.output.json
```diff
@@ -0,0 +1,35 @@
+{
+  "addresses": {
+    "avsDirectory": "0x135DDa560e946695d6f155dACaFC6f1F25C1F5AF",
+    "avsDirectoryImplementation": "0xdAbdB3Cd346B7D5F5779b0B614EdE1CC9DcBA5b7",
+    "baseStrategyImplementation": "0xdfdA04f980bE6A64E3607c95Ca26012Ab9aA46d3",
```
wadealexc, line 5: This is the old strategy implementation - are we going to deploy + upgrade to the new one?

ChaoticWalrus: because there have been zero logical or storage changes, I believe it would be unnecessary to do so.
I acknowledge that the DelayedWithdrawalRouter is in a similar situation, but I let that one go 🤷

wadealexc (approved):
Would be good to get some updated pause JSON, but this LGTM otherwise.

ChaoticWalrus:
going to go ahead and merge this; will make a note that I should also update the pausing JSON file 👍

PR #484, opened 2024-03-25
test: fork test new implementation contracts

8sunyuan:
Small draft to test that upgrading to new implementation contracts still passes all our existing integration tests.
Note that **eigenPodImplementation** doesn't get used because we need to use the ETHPOSDepositMock in tests.

Can double check the updated implementations here in the config matches what we have just deployed.


ChaoticWalrus:
This LGTM; I'm not sure what's wrong with the backend to cause these failures in the CI though.

8sunyuan:
> This LGTM; I'm not sure what's wrong with the backend to cause these failures in the CI though.

It all passes locally running forge test --match-contract Integration. I think the RPC URLs are getting rate limited.

PR #487, opened 2024-03-26
test: clean up fork testing

wadealexc:
* complete user migration en masse rather than passing in individual addresses
* improve logging around user creation and upgrading
* fix eigenpod withdrawal timestamps


8sunyuan (approved):
LGTM

PR #488, opened 2024-03-27
feat: initial PaymentCoordinator Implementation

8sunyuan:
Building off of Range Payment interface here https://github.com/Layr-Labs/eigenlayer-contracts/pull/470

Some changes to the existing interface.
- Adding a from address in Struct RangePayment otherwise AVSs will have to first transfer token payments to their ServiceManager.
- Allowing array of range payments for `payForRange`
- Adding pausing to initalize
- Changing all timestamps and time intervals to uint64 to be consistent.
- Some added events for setters


8sunyuan:
Noting @ChaoticWalrus reply here for the from parameter in the interface discussions
https://github.com/Layr-Labs/eigenlayer-contracts/pull/470#discussion_r1539950575

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,286 @@
...
+        );
+        require(
+            rangePayment.duration % calculationIntervalSeconds == 0,
+            "PaymentCoordinator.payForRange: duration must be a multiple of calculationIntervalSeconds"
+        );
+        // If retroactive payments enabled must at least be past LOWER_BOUND_START_RANGE,
+        // otherwise should start earliest at current block timestamp
+        require(
+            (retroactivePaymentsEnabled && rangePayment.startTimestamp >= LOWER_BOUND_START_RANGE) ||
+                rangePayment.startTimestamp >= block.timestamp,
```
scotthconner, line 146: Make sure this logic isn't ambiguous to the reader. One additional set of parens really helps not introducing a bug later when this is touched.

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,286 @@
...
+            "PaymentCoordinator.payForRange: invalid startTimestamp set"
+        );
+
+        // Require rangePayment is for whitelisted strategies
+        for (uint256 j = 0; j < rangePayment.strategiesAndMultlipliers.length; j++) {
+            IStrategy strategy = rangePayment.strategiesAndMultlipliers[j].strategy;
+            require(
+                strategyManager.strategyIsWhitelistedForDeposit(strategy),
```
scotthconner, line 154: Could we determine if there is anything deposited in the strategy manager for that? Basing payments on what is whitelisted for deposit could change over time, or maybe deposits are disabled but there is still active delegation in the system?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,286 @@
...
+                "PaymentCoordinator.payForRange: strategy not whitelisted for deposit"
+            );
+        }
+
+        // Set hash of rangePayment in mapping
+        bytes32 rangePaymentHash = keccak256(abi.encode(msg.sender, rangePayment));
```
scotthconner, line 160: This isn't guaranteed to be unique. I think this is because you want to prevent duplicates?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,286 @@
...
+        isRangePaymentHash[msg.sender][rangePaymentHash] = true;
```
scotthconner, line 161: keying off the message sender is needed later for potential withdrawals?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,286 @@
...
+
+        rangePayment.token.safeTransferFrom(msg.sender, address(this), rangePayment.amount);
+        return rangePaymentHash;
+    }
+
+    /**
+     * @notice Claim payments for the given claim
+     * @param claim The claims to be processed
+     */
```
scotthconner, line 170: Is this method more thoroughly documented in the interface?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,286 @@
...
+    function processClaims(PaymentMerkleClaim calldata claim) external onlyWhenNotPaused(PAUSED_CLAIM_PAYMENTS) {}
+
+    /**
+     * @notice Sets the address of the entity that can claim payments on behalf of the account
+     * @param account The account whose recipient is being set
+     * @param recipient The address of the entity that can claim payments on behalf of the account
+     * @dev Only callable by the `account`
+     */
+    function setRecipient(address account, address recipient) external {}
```
scotthconner, line 179: recipient is likely misleading.

setClaimer is I had in mind. Regulatory-wise, "recipient" sounds like they will passively receive which isn't the case.

src/contracts/core/PaymentCoordinatorStorage.sol
```diff
@@ -0,0 +1,103 @@
...
+    /// Slot 3
+    /// @notice Delay in timestamp before a posted root can be claimed against
+    uint64 public activationDelay;
+    /// @notice the commission for all operators across all avss
+    uint16 public globalOperatorCommissionBips;
+    /// @notice Payments must start at earliest now except unless retroactive payments enabled
+    bool public retroactivePaymentsEnabled;
+
+    /// @notice Mapping: account => the address of the entity to which new payments are directed on behalf of the account
+    mapping(address => address) public recipientOf;
```
scotthconner, line 70: delegatedClaimer?

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+    event CalculationIntervalSecondsSet(uint64 oldCalculationIntervalSeconds, uint64 newCalculationIntervalSeconds);
+    event GlobalCommissionBipsSet(uint16 oldGlobalCommissionBips, uint16 newGlobalCommissionBips);
+    event RecipientSet(address indexed account, address indexed recipient);
+    event RootSubmitted(
+        bytes32 root,
+        uint32 paymentCalculationStartTimestamp,
+        uint32 paymentCalculationEndTimestamp,
+        uint32 activatedAt
+    );
+    event PaymentClaimed(ClaimsTreeMerkleLeaf leaf);
```
scotthconner, line 77: I suppose the thought is these are "self" documenting? Might be good to @dev or @info these a bit, especially if they are emitted for multiple operations.

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+    function recipientOf(address account) external view returns (address);
+
+    /// @notice Mapping: recipient => token => total amount claimed
+    function cumulativeClaimed(address recipient, IERC20 token) external view returns (uint256);
+
+    /// @notice the commission for all operators across all avss
+    function globalOperatorCommissionBips() external view returns (uint16);
+
+    /// @notice returns the hash of the leaf
+    function calculateLeafHash(ClaimsTreeMerkleLeaf calldata leaf) external view returns (bytes32);
```
scotthconner, line 109: One thing that helps here is *why* does this method exist? Who would use it?

I think its because clients use it to prepare the request for claiming their rewards. Right?

src/contracts/interfaces/IStrategyManager.sol
```diff
@@ -127,6 +127,9 @@ interface IStrategyManager {
     /// @notice Returns the address of the `strategyWhitelister`
     function strategyWhitelister() external view returns (address);
 
+    /// @notice Returns bool for whether or not `strategy` is whitelisted for deposit
+    function strategyIsWhitelistedForDeposit(IStrategy strategy) external view returns (bool);
+
```
scotthconner, line 132: How we handle this wasn't part of the design or the use cases. What's the point of adding this? If the strategy is invalid the rows in the pipeline come out as zero. Is this to prevent bad payments from being made?

src/test/mocks/StrategyManagerMock.sol
```diff
@@ -84,6 +84,8 @@ contract StrategyManagerMock is
         thirdPartyTransfersForbidden[strategy] = value;
     }
 
+    function strategyIsWhitelistedForDeposit(IStrategy strategy) external view returns (bool) {}
```
scotthconner, line 87: Does this defer to false?

scotthconner (changes requested):
Looks really good. Couple of questions about docs, naming a few things, and the dependency on the strategy manager for validation.

The dependency and assumption we are making with the strategy manager is the only thing that really stuck out to me.

I didn't double check logic but some of the requires are ambiguous. I am curious why we aren't using typed errors instead of strings as well (I'm sure theres a reason).

The two things that need to be considered are:
1) Technical: dependency and assumption on strategy manager
2) Wording: Are we sure we want to go with payments coordinator, payments, etc. Just get a gut check from Jeff here.

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,286 @@
...
+
+    /// @dev Index for flag that pauses
+    uint8 internal constant PAUSED_CLAIM_PAYMENTS = 2;
+
+    /// @dev Chain ID at the time of contract deployment
+    uint256 internal immutable ORIGINAL_CHAIN_ID;
+
+    modifier onlyPaymentUpdater() {
+        require(msg.sender == paymentUpdater, "PaymentCoordinator: caller is not the paymentUpdater");
+        _;
```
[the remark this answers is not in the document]
8sunyuan: I am also a pro typed errors for the gas savings. Think we stuck with requires for readability versus doing multiple if blocks

src/contracts/core/PaymentCoordinator.sol
8sunyuan: Asked the same thing yesterday and AVSs can simply increase the amount by 1 wei. Happy to add in a nonce though.

src/contracts/core/PaymentCoordinator.sol
8sunyuan: We could have a check for non zero value of `strategy.totalShares`. Imo I don't think removing certain strategies from whitelist will occur frequently if at all except for potential exploit vulnerabilities.

src/test/mocks/StrategyManagerMock.sol
8sunyuan: Can ignore this for now, will add testing once we have enough agreement on implementation.

src/contracts/core/PaymentCoordinatorStorage.sol
8sunyuan: Would prefer to limit the word usage of delegation for purely delegated stake. Maybe `claimerFor` is better going off of changing setRecipient to setClaimer.

src/contracts/core/PaymentCoordinator.sol
8sunyuan: agreed

src/contracts/interfaces/IStrategyManager.sol
8sunyuan: This is just input sanitization. We can assume offchain that we can read staker/operator shares for whitelisted strats and calculate payments proportionally.

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: Yes and allow forward-compatability if this is being used as part of offchain merkleization. Will add a comment for that.

src/contracts/core/PaymentCoordinatorStorage.sol
8sunyuan: `claimerForEarner`

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+        uint64 activatedAfter;
+        // merkle root of the distribution
+        bytes32 root;
+    }
+
+    struct ClaimsTreeMerkleLeaf {
+        IERC20 token;
+        uint256 amount;
+        // The staker or operator who earned these tokens
+        address earner;
```
gpsanant, line 46: remove

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    struct PaymentMerkleClaim {
+        ClaimsTreeMerkleLeaf[] leaves;
```
gpsanant, line 50: add `address earner`

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+        // The index of the root in the list of roots
+        uint32 rootIndex;
+        // proof of the claimaint's account root in the Merkle tree
+        uint32 accountIndex;
+        bytes accountTreeProof;
```
gpsanant, line 55: account -> earner

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+    event PayAllForRangeSubmitterSet(address indexed payAllForRangeSubmitter, bool indexed oldValue, bool indexed newValue);
+    event ActivationDelaySet(uint64 oldActivationDelay, uint64 newActivationDelay);
+    event CalculationIntervalSecondsSet(uint64 oldCalculationIntervalSeconds, uint64 newCalculationIntervalSeconds);
+    event GlobalCommissionBipsSet(uint16 oldGlobalCommissionBips, uint16 newGlobalCommissionBips);
+    event RecipientSet(address indexed account, address indexed recipient);
+    event RootSubmitted(
+        bytes32 root,
+        uint32 paymentCalculationStartTimestamp,
+        uint32 paymentCalculationEndTimestamp,
+        uint32 activatedAt
```
gpsanant, line 75: activatedAt or activatedAfter?

src/contracts/interfaces/IPaymentCoordinator.sol
gpsanant, line 77: need the root index here, I think. cc @ypatil12 for the frontend indexing...

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+    function paymentUpdater() external view returns (address);
+
+    /**
+     * @notice The interval in seconds at which the calculation for range payment distribution is done.
+     * @dev Payment durations must be multiples of this interval.
+     */
+    function calculationIntervalSeconds() external view returns (uint64);
+
+    /// @notice The maximum amount of time that a range payment can end in the future
+    function MAX_PAYMENT_DURATION() external view returns (uint64);
```
gpsanant, line 91: mmm ok. so this means that retroactivate payments might need to broken up into ranges of `MAX_PAYMENT_DURATION` until they cover the required retroactive range. that's fine, just something to keep track of

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+     */
+    function setGlobalOperatorCommission(uint16 _globalCommissionBips) external;
+
+    /**
+     * @notice Sets the address of the entity that can claim payments on behalf of the account
+     * @param account The account whose recipient is being set
+     * @param recipient The address of the entity that can claim payments on behalf of the account
+     * @dev Only callable by the `account`
+     */
+    function setRecipient(address account, address recipient) external;
```
gpsanant, line 160: did we decide on recipient vs claimer?

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * @notice Creates a new distribution root
+     * @param root The merkle root of the distribution
+     * @param paymentsCalculatedUntilTimestamp The timestamp until which payments have been calculated
+     * @dev Only callable by the paymentUpdater
+     */
+    function submitRoot(bytes32 root, uint32 paymentsCalculatedUntilTimestamp) external;
```
gpsanant, line 168: not sure if we want to check against a beacon block root to track reorgs here. i put this in my spec, but not sure if too tryhard...

cc @ypatil12 @ChaoticWalrus

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * @notice Claim payments for the given claim
+     * @param claim The claims to be processed
+     */
+    function processClaims(PaymentMerkleClaim calldata claim) external;
```
gpsanant, line 174: `processClaim`?

src/contracts/core/PaymentCoordinatorStorage.sol
```diff
@@ -0,0 +1,103 @@
...
+    IDelegationManager public immutable delegationManager;
+
+    /// @notice The EigenPodManager contract for EigenLayer
+    IEigenPodManager public immutable eigenPodManager;
+
+    /// @notice The StrategyManager contract for EigenLayer
+    IStrategyManager public immutable strategyManager;
+
+    /// @notice The Slasher contract for EigenLayer
+    ISlasher public immutable slasher;
```
gpsanant, line 41: dont need?

src/contracts/core/PaymentCoordinatorStorage.sol
```diff
@@ -0,0 +1,103 @@
...
+    /// @dev Payment durations must be multiples of this interval.
+    uint64 public calculationIntervalSeconds;
+
+    /// Slot 3
+    /// @notice Delay in timestamp before a posted root can be claimed against
+    uint64 public activationDelay;
+    /// @notice the commission for all operators across all avss
+    uint16 public globalOperatorCommissionBips;
+    /// @notice Payments must start at earliest now except unless retroactive payments enabled
+    bool public retroactivePaymentsEnabled;
```
gpsanant, line 67: would we ever disable this? I think it's fine to remove this

src/contracts/core/PaymentCoordinatorStorage.sol
gpsanant: agree w @8sunyuan. claimer is my preference

src/contracts/core/PaymentCoordinatorStorage.sol
```diff
@@ -0,0 +1,103 @@
...
+     * Use the getter function `domainSeparator` to get the current domain separator for this contract.
+     */
+    bytes32 internal _DOMAIN_SEPARATOR;
+
+    /// Slot 2
+    /// @notice The address of the entity that can update the contract with new merkle roots
+    address public paymentUpdater;
+    /// @notice The interval in seconds at which the calculation for range payment distribution is done.
+    /// @dev Payment durations must be multiples of this interval.
+    uint64 public calculationIntervalSeconds;
```
gpsanant, line 59: would be nice to make this immutable... do you see an issue with that @ypatil12 @ChaoticWalrus ?

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/2a9e9d5e0f3c92f77bb06cef994ab21210310cfe
Nonce added

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/2a9e9d5e0f3c92f77bb06cef994ab21210310cfe

src/contracts/core/PaymentCoordinatorStorage.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/2a9e9d5e0f3c92f77bb06cef994ab21210310cfe

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/2a9e9d5e0f3c92f77bb06cef994ab21210310cfe

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,287 @@
...
+    function _payForRange(RangePayment calldata rangePayment) internal returns (bytes32) {
+        require(rangePayment.strategiesAndMultlipliers.length > 0, "PaymentCoordinator.payForRange: no strategies set");
+        require(rangePayment.amount > 0, "PaymentCoordinator.payForRange: amount cannot be 0");
+        require(
+            rangePayment.duration <= MAX_PAYMENT_DURATION,
+            "PaymentCoordinator.payForRange: duration exceeds MAX_PAYMENT_DURATION"
+        );
+        require(
+            rangePayment.duration % calculationIntervalSeconds == 0,
+            "PaymentCoordinator.payForRange: duration must be a multiple of calculationIntervalSeconds"
```
gpsanant, line 140: `._payForRange`?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,287 @@
...
+/**
+ * @title PaymentCoordinator
+ * @author Eigen Labs Inc.
+ * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
+ * @notice  This is the contract for payments in EigenLayer. The main functionalities of this contract are
+ * - enabling any ERC20 payments from AVSs to their operators and stakers for a given time range
+ * - allowing stakers and operators to claim their earnings including a commission bips for operators
+ * - allowing the protocol to provide ERC20 tokens to stakers over a specified time range
+ */
+contract PaymentCoordinator is
```
gpsanant, line 22: I think we're calling this the claimingManager? not sure...

src/contracts/core/PaymentCoordinator.sol
gpsanant: for migrating to SecurityPayments

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,287 @@
...
+            require(
+                strategyManager.strategyIsWhitelistedForDeposit(strategy),
+                "PaymentCoordinator.payForRange: strategy not whitelisted for deposit"
+            );
+        }
+
+        // Set hash of rangePayment in mapping
+        bytes32 rangePaymentHash = keccak256(abi.encode(msg.sender, paymentNonce, rangePayment));
+        paymentNonce++;
+        isRangePaymentHash[msg.sender][rangePaymentHash] = true;
```
gpsanant, line 162: We need to check that `isRangePaymentHash[msg.sender][rangePaymentHash] == false` before this?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,287 @@
...
+
+        rangePayment.token.safeTransferFrom(msg.sender, address(this), rangePayment.amount);
```
gpsanant, line 164: prefer to do this after the require checks for cosmentics

src/contracts/core/PaymentCoordinatorStorage.sol
8sunyuan: I can see this being used to get AVSs to pay ahead of time. I'd think this is a nice-to-have

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+        // The staker or operator who earned these tokens
+        address earner;
+    }
+
+    struct PaymentMerkleClaim {
+        ClaimsTreeMerkleLeaf[] leaves;
+        // The index of the root in the list of roots
+        uint32 rootIndex;
+        // proof of the claimaint's account root in the Merkle tree
+        uint32 accountIndex;
```
ypatil12, line 54: Can we standardize using earner instead of account everywhere across the PR too?

8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a3aab21914ae5548b63100b39ecdec463212b9dd

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,175 @@
...
+    /// EVENTS ///
+
+    event RangePaymentCreated(address indexed avs, bytes32 rangePaymentHash, RangePayment rangePayment);
+    event RangePaymentForAllCreated(address indexed submitter, bytes32 rangePaymentHash, RangePayment rangePayment);
+    event PaymentUpdaterSet(address indexed oldPaymentUpdater, address indexed newPaymentUpdater);
+    event PayAllForRangeSubmitterSet(address indexed payAllForRangeSubmitter, bool indexed oldValue, bool indexed newValue);
+    event ActivationDelaySet(uint64 oldActivationDelay, uint64 newActivationDelay);
+    event CalculationIntervalSecondsSet(uint64 oldCalculationIntervalSeconds, uint64 newCalculationIntervalSeconds);
+    event GlobalCommissionBipsSet(uint16 oldGlobalCommissionBips, uint16 newGlobalCommissionBips);
+    event ClaimerForSet(address indexed account, address indexed claimer);
```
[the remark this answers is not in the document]
8sunyuan: Left as `ClaimerForSet` because storage is `claimerFor`

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,287 @@
...
+            isPayAllForRangeSubmitter[msg.sender],
+            "PaymentCoordinator: caller is not a valid payAllForRange submitter"
+        );
+        _;
+    }
+
+    /// @dev Sets the immutable variables for the contract
+    constructor(
+        IAVSDirectory _avsDirectory,
+        IDelegationManager _delegationManager,
```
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a3aab21914ae5548b63100b39ecdec463212b9dd
Removed, left the DM in to check for beaconChainETH strat

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a3aab21914ae5548b63100b39ecdec463212b9dd#diff-2a28246efa54c09dff5f0afef55f490d130b1a13c5b7e30efb51d5647c52a1a2R138

src/contracts/core/PaymentCoordinator.sol
8sunyuan: You mean the transfer call? Not sure what you mean

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/4341741e9d7d6855c2b46ae67dca3970eb240f8d
Added some extra comments for some of these events.

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a3aab21914ae5548b63100b39ecdec463212b9dd
Removed and added to `PaymentMerkleClaim`

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a3aab21914ae5548b63100b39ecdec463212b9dd#diff-a9495834f8b68265d0d625d94e217cace2813c0c4334cdaa645f8ae6e7e5e5bcL49

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a3aab21914ae5548b63100b39ecdec463212b9dd#diff-a9495834f8b68265d0d625d94e217cace2813c0c4334cdaa645f8ae6e7e5e5bcL49

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a3aab21914ae5548b63100b39ecdec463212b9dd#diff-a9495834f8b68265d0d625d94e217cace2813c0c4334cdaa645f8ae6e7e5e5bcR88

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a3aab21914ae5548b63100b39ecdec463212b9dd
Added the distributionRoot and the earnerTokenRoot here

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a3aab21914ae5548b63100b39ecdec463212b9dd#diff-a9495834f8b68265d0d625d94e217cace2813c0c4334cdaa645f8ae6e7e5e5bcR95
Fixed

src/contracts/core/PaymentCoordinatorStorage.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/commit/a3aab21914ae5548b63100b39ecdec463212b9dd

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/commit/a3aab21914ae5548b63100b39ecdec463212b9dd

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,287 @@
...
+     * @notice similar to `payForRange` except the payment is split amongst *all* stakers
+     * rather than just those delegated to operators who are registered to a single avs
+     */
+    function payAllForRange(
+        RangePayment[] calldata rangePayments
+    ) external onlyWhenNotPaused(PAUSED_PAY_ALL_FOR_RANGE) onlyPayAllForRangeSubmitter {
+        for (uint256 i = 0; i < rangePayments.length; i++) {
+            RangePayment calldata rangePayment = rangePayments[i];
+            bytes32 rangePaymentHash = _payForRange(rangePayment);
+            emit RangePaymentForAllCreated(msg.sender, rangePaymentHash, rangePayment);
```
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/commit/a3aab21914ae5548b63100b39ecdec463212b9dd

src/contracts/core/PaymentCoordinator.sol
8sunyuan: Just broke it into an if block here https://github.com/Layr-Labs/eigenlayer-contracts/commit/a3aab21914ae5548b63100b39ecdec463212b9dd#diff-2a28246efa54c09dff5f0afef55f490d130b1a13c5b7e30efb51d5647c52a1a2R151

src/contracts/core/PaymentCoordinator.sol
8sunyuan: Also adding beaconChainETHStrat check here https://github.com/Layr-Labs/eigenlayer-contracts/commit/a3aab21914ae5548b63100b39ecdec463212b9dd#diff-2a28246efa54c09dff5f0afef55f490d130b1a13c5b7e30efb51d5647c52a1a2R172

Can check it from the DM here https://github.com/Layr-Labs/eigenlayer-contracts/blob/dev/src/contracts/core/DelegationManager.sol#L38

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/59612913dd174b2cc932c15b5e9ff1b950f68f55
Added some additional documentation

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -33,48 +32,73 @@ interface IPaymentCoordinator {
...
         bytes32 root;
+        // The timestamp which calculated payments started
+        uint64 paymentCalculationStartTimestamp;
+        // The timestamp until which payments have been calculated
+        uint64 paymentCalculationEndTimestamp;
+        // timestamp at which the root can be claimed against
+        uint64 activatedAt;
     }
 
     struct ClaimsTreeMerkleLeaf {
```
gpsanant, line 45: should we literally do a TokenTreeMerkleLeaf and a EarnerTreeMerkleLeaf?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -110,88 +106,328 @@ contract PaymentCoordinator is
...
+        uint64 paymentCalculationStartTimestamp,
+        uint64 paymentCalculationEndTimestamp,
+        uint64 activatedAt
+    ) external onlyPaymentUpdater {
+        uint32 rootIndex = uint32(distributionRoots.length);
+        distributionRoots.push(
+            DistributionRoot({
+                root: root,
+                activatedAt: activatedAt,
+                paymentCalculationStartTimestamp: paymentCalculationStartTimestamp,
```
gpsanant, line 159: dont need start

src/contracts/core/PaymentCoordinator.sol
gpsanant: prefer to do the transfer call on line 219ish

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -110,88 +106,328 @@ contract PaymentCoordinator is
...
         require(
             rangePayment.duration % calculationIntervalSeconds == 0,
-            "PaymentCoordinator.payForRange: duration must be a multiple of calculationIntervalSeconds"
+            "PaymentCoordinator._payForRange: duration must be a multiple of calculationIntervalSeconds"
         );
-        // If retroactive payments enabled must at least be past LOWER_BOUND_START_RANGE,
-        // otherwise should start earliest at current block timestamp
+        if (retroactivePaymentsEnabled) {
+            require(
+                block.timestamp - LOWER_BOUND_START_RANGE <= rangePayment.startTimestamp,
```
gpsanant, line 196: Can we name this to `RETROACTIVE_WINDOW` or something?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -110,88 +106,328 @@ contract PaymentCoordinator is
...
+        bytes32 root,
+        uint64 paymentCalculationStartTimestamp,
+        uint64 paymentCalculationEndTimestamp,
+        uint64 activatedAt
+    ) external onlyPaymentUpdater {
+        uint32 rootIndex = uint32(distributionRoots.length);
+        distributionRoots.push(
+            DistributionRoot({
+                root: root,
+                activatedAt: activatedAt,
```
gpsanant, line 158: activatedAt should be set to `block.timestamp + activationDelay`

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -110,88 +106,328 @@ contract PaymentCoordinator is
...
-     * @param claim The claims to be processed
+     * @notice Process a payment claim.
+     * @param claim PaymentMerkleClaim struct containing the claim details for proof verification
+     * and payment transfer.
+     * @dev only callable by claimerFor[claim.earner]
+     */
+    function _processClaim(PaymentMerkleClaim calldata claim) internal onlyWhenNotPaused(PAUSED_CLAIM_PAYMENTS) {
+        DistributionRoot memory root = distributionRoots[claim.rootIndex];
+        require(
+            block.timestamp > root.activatedAt + activationDelay,
```
gpsanant, line 237: should just be `block.timestamp > root.activatedAt`

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -110,88 +106,328 @@ contract PaymentCoordinator is
...
+            "PaymentCoordinator._processClaim: root not activated yet"
+        );
+
+        // If claimerFor earner is not set, claimer is by default the earner. Else set to claimerFor
+        address claimer = claimerFor[claim.earner] == address(0) ? claim.earner : claimerFor[claim.earner];
```
gpsanant, line 242: this will do 2 sloads if im not wrong
```solidity
address claimer = claimerFor[claim.earner];
if (claimer == address(0)) {
      claimer = claim.earner;
}
```

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -110,88 +106,328 @@ contract PaymentCoordinator is
...
+    }
+
+    /// @notice return the hash of the earner's leaf
+    function calculateEarnerLeafHash(address earner, bytes32 earnerTokenRoot) public pure returns (bytes32) {
+        return sha256(abi.encode(earner, earnerTokenRoot));
+    }
+
+    /// @notice returns the hash of the earner's token leaf
+    function calculateTokenLeafHash(ClaimsTreeMerkleLeaf calldata leaf) public pure returns (bytes32) {
+        return sha256(abi.encode(leaf));
```
gpsanant, line 388: use keccak tbh

gpsanant: way less gas

src/contracts/core/PaymentCoordinator.sol
gpsanant: Ok I actually think we should literally set a LOWER_BOUND on when they can send payments and a MAX_DURATION

The way things are currently coded, if AVSs dont implement payments before `m2.timestamp +  LOWER_BOUND_START_RANGE`, the cant pay operators from the beginning

gpsanant: cc @ypatil12 @ChaoticWalrus

ChaoticWalrus: are we OK with payments having a like, year-long duration?
should we have both of these things? 

I think we need to agree on desired behavior, rather than keep tweaking what's here haphazardly.
naming should be clear + also come from the specified behavior.
in my mind, it's something like:

- MAX_RETROACTIVE_LENGTH = max amount of time that a payment can start in the past
- PAYMENTS_GENESIS_TIME = absolute minimum start time for a payment

I can see arguments for both, but am not seeing clarity on what we need (or want).

likewise, MAX_DURATION should perhaps be MAX_FUTURE_LENGTH or something similar, if that's actually how we _want_ it to be used.

my 2c: do PAYMENTS_GENESIS_TIME, MAX_RETROACTIVE_LENGTH, and MAX_FUTURE_LENGTH -- the max duration is MAX_FUTURE_LENGTH + MAX_RETROACTIVE_LENGTH, and AVSs have to start making payments within MAX_RETROACTIVE_LENGTH of PAYMENTS_GENESIS_TIME in order to make payments that start from the genesis time.

I'm open to other ideas, I just want to clarify what is desired + why.

gpsanant: Ok that actually seems pretty good

This allows for deprecating earlier versions and deleting some of our database due to MAX_RETROACTIVE_LENGTH

It allows for upgrading to newer versions due to MAX_FUTURE_LENGTH

it does place the bound on AVSs to integrate within MAX_RETROACTIVE_LENGTH of PAYMENTS_GENESIS_TIME. It also means that MAX_RETROACTIVE_LENGTH > paymentLaunch.timestamp - PAYMENTS_GENESIS_TIME. That seems ok tbh

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -131,6 +158,31 @@ interface IPaymentCoordinator {
...
+     * @notice Creates a new distribution root
+     * @param root The merkle root of the distribution
+     * @param paymentCalculationStartTimestamp The start timestamp which payments have been calculated from
+     * @param paymentCalculationEndTimestamp The timestamp until which payments have been calculated
+     * @param activatedAt timestamp at which that the root can be claimed against
+     * @dev Only callable by the paymentUpdater
+     */
+    function submitRoot(
+        bytes32 root,
+        uint64 paymentCalculationStartTimestamp,
```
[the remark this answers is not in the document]
ypatil12: Prefer to keep for clarity

src/contracts/core/PaymentCoordinator.sol
8sunyuan: Seems a bit unintuitive. So do we also emit block.timestamp + activationDelay? 
The activation delay is only relevant for the paymentUpdater so if we wanted to ensure a later claiming timestamp we could have just submitted a root with a later activatedAt timestamp. Not entirely sure the point of this delay unless you want to have a require like 
```
require(activatedAt - block.timestamp > activationDelay)
```
To ensure a submitted root has at minimum activationDelay number of seconds until the root can be active.

gpsanant: the point of enforcing the delay at the contract level is that, if the payment updater posts an incorrect root, then users can withdraw and governance can pause within the delay.

we can do this either by adding the delay when the root is posted or adding the delay in every claim (like we do in DM withdrawals). the former is less gas than the latter.

src/contracts/core/PaymentCoordinatorStorage.sol
gpsanant: i don't think we're ever going to deprecate retroactive payments tho?

src/contracts/interfaces/IPaymentCoordinator.sol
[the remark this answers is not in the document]
gpsanant: What's clearer? it's just for logging purposes are this point or?

If we really need it, can we store it in the contract and update on root submission and emit in event

src/contracts/interfaces/IStrategyManager.sol
ypatil12: We aligned offline here, but we basically want to prevent bad payments being made. Eg. AVS sends a 6 figure+ payment to a non-existent strategy and we then have to update the contract to claw back these funds.

src/contracts/interfaces/IPaymentCoordinator.sol
ypatil12: Aligns best with tree design IMO

src/contracts/interfaces/IPaymentCoordinator.sol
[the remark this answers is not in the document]
ypatil12: Primarily for logging purposes. At the least would want an event with (root, range_start, range_end).

However, I also do think it's easier to "reason" about this function by passing in the window. There is no guarantee anywhere in the contracts that the ranges are contiguous, though that is what we expect in practice.

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,523 @@
...
+    }
+
+    /// @notice returns 'true' if the claim would currently pass the check in `processClaims`
+    function checkClaim(PaymentMerkleClaim calldata claim) external view returns (bool) {
+        DistributionRoot memory root = distributionRoots[claim.rootIndex];
```
ypatil12, line 393: Nit: This logic is essentially repeated twice - here and in `_verifyClaim`. Would prefer having it in one place. Recommend adding all of this functionality in a single function `_checkClaim`, which will also be called by `processClaim`. The `_processTokenClaim` function will then only need to transfer tokens, as the `_checkClaim` function will check earner and token roots.

src/contracts/interfaces/IPaymentCoordinator.sol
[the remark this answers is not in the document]
gpsanant: are u onboard with storing last timestamp in the contract?

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,217 @@
...
+
+    struct PaymentMerkleClaim {
+        // The index of the root in the list of roots
+        uint32 rootIndex;
+        // proof of the earner's account root in the Earner Merkle tree
+        uint32 earnerIndex;
+        bytes earnerTreeProof;
+        EarnerTreeMerkleLeaf earnerLeaf;
+        // The indices and proofs of the leafs in the claimaint's merkle tree for this root
+        uint32[] leafIndices;
```
gpsanant, line 61: should we make index/proof/leaf another struct which we have an array of?

gpsanant, line 61: tokenIndices

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,217 @@
...
+    /**
+     * @notice Creates a new distribution root
+     * @param root The merkle root of the distribution
+     * @param paymentCalculationEndTimestamp The timestamp until which payments have been calculated
+     * @param activatedAt timestamp at which that the root can be claimed against
+     * @dev Only callable by the paymentUpdater
+     */
+    function submitRoot(
+        bytes32 root,
+        uint64 paymentCalculationEndTimestamp,
```
gpsanant, line 186: uint32

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    struct TokenTreeMerkleLeaf {
+        IERC20 token;
+        uint256 cumulativeEarnings;
+    }
+
+    struct PaymentMerkleClaim {
+        // The index of the root in the list of roots
+        uint32 rootIndex;
```
gpsanant, line 55: can we add a view function to get the rootIndex from the root for the backend? could be inefficient reverse looping like middleware contracts

8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/blob/37b0406649f4056b78253278d75b49aef8d3a290/src/contracts/core/PaymentCoordinator.sol#L465

Added here

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: but overflow in 100 years :))
Is this a necessity? Will have to change across multiple files and keeping all timestamps and durations consistent is nice QOL to avoid casting

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/files#diff-a9495834f8b68265d0d625d94e217cace2813c0c4334cdaa645f8ae6e7e5e5bcR61

8sunyuan: Think this looks good as is

src/contracts/core/PaymentCoordinator.sol
8sunyuan: Done. `_processClaim` uses `_checkClaim` now. Originally didn't want to loop over tokenIndices again and do another storage read for claimerFor. I guess it shouldn't be too much more expensive though

src/contracts/core/PaymentCoordinator.sol
8sunyuan: Done and using `verifyInclusionKeccak`

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/blob/8bcbaf830523cf49949011d507b4750cd221ad67/src/contracts/core/PaymentCoordinator.sol#L236

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/blob/8bcbaf830523cf49949011d507b4750cd221ad67/src/contracts/core/PaymentCoordinator.sol#L151

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/blob/8bcbaf830523cf49949011d507b4750cd221ad67/src/contracts/core/PaymentCoordinator.sol#L275
Made it `block.timestamp >= root.activatedAt` so it actually activates at the timestamp

src/contracts/core/PaymentCoordinator.sol
8sunyuan: To summarize, we have a sliding window essentially configured by 
```
    /// @notice max amount of time that a payment can start in the past
    uint64 public immutable MAX_RETROACTIVE_LENGTH;
    /// @notice max amount of time that a payment can start in the future
    uint64 public immutable MAX_FUTURE_LENGTH;
```
But there is a absolute minimum defined by below where a startTimestamp cannot be before this.
```
    /// @notice absolute min timestamp that a payment can start at
    uint64 public immutable GENESIS_PAYMENT_TIMESTAMP;
```

And there is also a max on duration of a payment
```
    uint64 public immutable MAX_PAYMENT_DURATION;
```

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/blob/8bcbaf830523cf49949011d507b4750cd221ad67/src/contracts/core/PaymentCoordinator.sol#L160

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/files#diff-a9495834f8b68265d0d625d94e217cace2813c0c4334cdaa645f8ae6e7e5e5bcR48

src/contracts/interfaces/IPaymentCoordinator.sol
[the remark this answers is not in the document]
8sunyuan: Could enforce contiguous ranges if we store RangePayments onchain

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -161,12 +163,7 @@ contract PaymentCoordinator is
                 paymentCalculationEndTimestamp: paymentCalculationEndTimestamp
             })
         );
-        emit DistributionRootSubmitted(
-            rootIndex,
-            root,
-            paymentCalculationEndTimestamp,
-            activatedAt
-        );
+        emit DistributionRootSubmitted(rootIndex, root, paymentCalculationEndTimestamp, activatedAt);
```
gpsanant, line 166: @ypatil12 we also need to index events these offchain for the payment updater

gpsanant: on order for them to know the timestamp to lookup in the payments table

ypatil12: We need an event to know when `GENESIS_PAYMENT_TIMESTAMP` is so we can figure out what the first range of DistributionRoot is. This is partially why I'm for adding the entire window in the DistributionRootSubmitted event. cc @8sunyuan

ypatil12: Another option is to just use RPC and store for the very first update

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,287 @@
...
+            bytes32 rangePaymentHash = _payForRange(rangePayment);
+            emit RangePaymentCreated(msg.sender, rangePaymentHash, rangePayment);
+        }
+    }
+
+    /**
+     * @notice similar to `payForRange` except the payment is split amongst *all* stakers
+     * rather than just those delegated to operators who are registered to a single avs
+     */
+    function payAllForRange(
```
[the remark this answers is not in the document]
gpsanant: it will be to operators too

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+        require(rangePayment.amount > 0, "PaymentCoordinator._payForRange: amount cannot be 0");
+        require(
+            rangePayment.duration <= MAX_PAYMENT_DURATION,
+            "PaymentCoordinator._payForRange: duration exceeds MAX_PAYMENT_DURATION"
+        );
+        require(
+            rangePayment.duration % calculationIntervalSeconds == 0,
+            "PaymentCoordinator._payForRange: duration must be a multiple of calculationIntervalSeconds"
+        );
+        if (retroactivePaymentsEnabled) {
```
gpsanant, line 190: let's just take off this toggle imo

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+            );
+        }
+        require(
+            rangePayment.startTimestamp <= block.timestamp + MAX_FUTURE_LENGTH,
+            "PaymentCoordinator._payForRange: startTimestamp too far in the future"
+        );
+
+        // Require rangePayment is for whitelisted strategy or beaconChainETHStrategy
+        for (uint256 j = 0; j < rangePayment.strategiesAndMultipliers.length; j++) {
+            IStrategy strategy = rangePayment.strategiesAndMultipliers[j].strategy;
```
gpsanant, line 209: maybe make sure no duplicates? i.e. it must be an ordered list

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+        tokenLeaf.token.safeTransfer(claimer, claimAmount);
+        emit PaymentClaimed(root, tokenLeaf);
+    }
+
+    function _checkClaim(PaymentMerkleClaim calldata claim) internal view returns (bool) {
+        DistributionRoot memory root = distributionRoots[claim.rootIndex];
+        require(block.timestamp >= root.activatedAt, "PaymentCoordinator._checkClaim: root not activated yet");
+
+        // If claimerFor earner is not set, claimer is by default the earner. Else set to claimerFor
+        address claimer = claimerFor[claim.earnerLeaf.earner];
```
gpsanant, line 280: we do this twice in `_processClaim` also, let's only do it in `_processClaim`

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+     */
+    function setCalculationIntervalSeconds(uint64 _calculationIntervalSeconds) external onlyOwner {
+        _setCalculationIntervalSeconds(_calculationIntervalSeconds);
+    }
+
+    /**
+     * @notice Set a new value for retroactivePaymentsEnabled. Only callable by owner
+     * @param _retroactivePaymentsEnabled The new value for retroactivePaymentsEnabled
+     */
+    function setRetroactivePaymentsEnabled(bool _retroactivePaymentsEnabled) external onlyOwner {
```
gpsanant, line 405: delete

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,1613 @@
...
+    StrategyBase strategyImplementation;
+    uint256 mockTokenInitialSupply = 10e50;
+    IPaymentCoordinator.StrategyAndMultiplier[] defaultStrategyAndMultipliers;
+
+    // Config Variables
+    /// @notice Max duration is 5 epochs (2 weeks * 5 = 10 weeks in seconds)
+    uint64 MAX_PAYMENT_DURATION = 86400 * 70;
+
+    /// @notice Lower bound start range is 3 months into the past
+    uint64 MAX_RETROACTIVE_LENGTH = 86400 * 90;
```
gpsanant, line 50: you can do things like `90 days`

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,1613 @@
...
+                        calculationIntervalSeconds,
+                        globalCommissionBips
+                    )
+                )
+            )
+        );
+
+        // Deploy mock token and strategy
+        token1 = new ERC20PresetFixedSupply("dog wif hat", "MOCK1", mockTokenInitialSupply, address(this));
+        token1 = new ERC20PresetFixedSupply("jeo boden", "MOCK2", mockTokenInitialSupply, address(this));
```
gpsanant, line 116: token2?

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,1613 @@
...
+
+        return balances;
+    }
+}
+
+contract PaymentCoordinatorUnitTests_initializeAndSetters is PaymentCoordinatorUnitTests {
+    function testFuzz_setClaimerFor(address earner, address claimer) public filterFuzzedAddressInputs(earner) {
+        cheats.startPrank(earner);
+        cheats.expectEmit(true, true, true, true, address(paymentCoordinator));
+        emit ClaimerForSet(earner, paymentCoordinator.claimerFor(earner), claimer);
```
gpsanant, line 248: won't `paymentCoordinator.claimerFor(earner)` be 0 here?

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,1613 @@
...
+        cheats.startPrank(paymentCoordinator.owner());
+        cheats.expectEmit(true, true, true, true, address(paymentCoordinator));
+        emit PaymentUpdaterSet(paymentCoordinator.paymentUpdater(), newPaymentUpdater);
+        paymentCoordinator.setPaymentUpdater(newPaymentUpdater);
+        assertEq(newPaymentUpdater, paymentCoordinator.paymentUpdater(), "paymentUpdater not set");
+        cheats.stopPrank();
+    }
+
+    function testFuzz_setPayAllForRangeSubmitter(address submitter, bool newValue) public {
+        cheats.startPrank(paymentCoordinator.owner());
```
gpsanant, line 309: would be good to have sanity testes for "NotOwnerCannotXYZ"

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,1613 @@
...
+
+    /// @notice Claim against latest submitted root, rootIndex 3 but modify some of the leaf values
+    /// forge-config: default.fuzz.runs = 10
+    function testFuzz_processClaim_Revert_WhenInvalidEarnerClaim(
+        bool setClaimerFor,
+        address claimerFor,
+        address invalidEarner
+    ) public filterFuzzedAddressInputs(claimerFor) {
+        // if setClaimerFor is true, set the earners claimer to the fuzzed address
+        address claimer;
```
gpsanant, line 1498: putting the starting block (up to 1513) into an internal, reused function would be good

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,1613 @@
...
+        uint256 startTimestamp,
+        uint256 duration,
+        uint256 amount
+    ) public filterFuzzedAddressInputs(avs) {
+        cheats.assume(avs != address(0));
+        cheats.prank(paymentCoordinator.owner());
+        paymentCoordinator.setRetroactivePaymentsEnabled(retroactivePaymentsEnabled);
+
+        // 1. Bound fuzz inputs to valid ranges and amounts
+        IERC20 paymentToken = new ERC20PresetFixedSupply("dog wif hat", "MOCK1", mockTokenInitialSupply, avs);
```
gpsanant, line 418: you can reuse the starting block of a lot of these

src/contracts/core/PaymentCoordinator.sol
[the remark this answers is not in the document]
ypatil12: or `payAllEarnersForRange`

src/contracts/core/PaymentCoordinatorStorage.sol
```diff
@@ -0,0 +1,105 @@
...
+
+    /// @notice Mapping: claimer => token => total amount claimed
+    mapping(address => mapping(IERC20 => uint256)) public cumulativeClaimed;
+
+    /// @notice Timestamp for latest submitted payment range
+    uint64 public latestPaymentRangeTimestamp;
+    /// @notice Used for unique rangePaymentHashes
+    uint256 public paymentNonce;
+    /// @notice Mapping: avs => rangePaymentHash => bool to check if range payment hash has been submitted
+    mapping(address => mapping(bytes32 => bool)) public isRangePaymentHash;
```
gpsanant, line 79: we should be able to distinguish payForAll from pay in these hashes...

src/contracts/core/PaymentCoordinator.sol
ypatil12: Another thing to think about - we are currently assuming the the window for the first range distribution root starts at `GENESIS_PAYMENT_TIMESTAMP`. That effectively means that the "0th" distribution root ends at this timestamp minus 1. This is expected, but just a call out

We may have to add an event for this if the off-chain logic becomes wonky here, but for now its fine

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+    function payAllForRange(
+        RangePayment[] calldata rangePayments
+    ) external onlyWhenNotPaused(PAUSED_PAY_ALL_FOR_RANGE) onlyPayAllForRangeSubmitter nonReentrant {
+        for (uint256 i = 0; i < rangePayments.length; i++) {
+            RangePayment calldata rangePayment = rangePayments[i];
+            bytes32 rangePaymentHash = _payForRange(rangePayment);
+            // paymentNonce - 1 as the nonce is incremented in _payForRange
+            emit RangePaymentForAllCreated(msg.sender, paymentNonce - 1, rangePaymentHash, rangePayment);
+        }
+        latestPaymentRangeTimestamp = uint64(block.timestamp);
```
gpsanant, line 131: why?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+        uint64 activatedAt
+    ) external onlyPaymentUpdater {
+        require(activatedAt >= block.timestamp, "PaymentCoordinator.submitRoot: activatedAt cannot be in the past");
+        activatedAt = activatedAt + activationDelay;
+        uint32 rootIndex = uint32(distributionRoots.length);
+        distributionRoots.push(
+            DistributionRoot({
+                root: root,
+                activatedAt: activatedAt,
+                paymentCalculationEndTimestamp: paymentCalculationEndTimestamp
```
gpsanant, line 165: didn't we agree to store `paymentCalculationEndTimestamp` and make sure it was only increasing?

src/test/test-data/paymentCoordinator/processClaimProofs_Root1.json
```diff
@@ -0,0 +1,54 @@
+{
+  "Root": "0x5a17c863bab9249dd3463e6f942c51e3a885b46632c3e127666b5c6d9efbefda",
```
stevennevins, line 2: When we generate the roots, we need to confirm that the library we use generates the tree the same way as openzeppelin's merkle-tree js library.  There is a note in our merkle.sol library that relates to the extra precautions OZ takes

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+    function _checkClaim(PaymentMerkleClaim calldata claim) internal view returns (bool) {
+        DistributionRoot memory root = distributionRoots[claim.rootIndex];
+        require(block.timestamp >= root.activatedAt, "PaymentCoordinator._checkClaim: root not activated yet");
+
+        // If claimerFor earner is not set, claimer is by default the earner. Else set to claimerFor
+        address claimer = claimerFor[claim.earnerLeaf.earner];
+        if (claimer == address(0)) {
+            claimer = claim.earnerLeaf.earner;
+        }
+        require(msg.sender == claimer, "PaymentCoordinator._checkClaim: caller is not valid claimer");
```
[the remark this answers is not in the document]
8sunyuan: or the 0 address, to allow earners who never call `setClaimerFor` to claim themselves still

src/contracts/core/PaymentCoordinator.sol
8sunyuan: Also this msg.sender check means it would revert if calling the view `checkClaim` function. Even if the claim is valid and passes all the merkle proofs. Going to move this out of the _checkClaim internal function checks.

src/test/test-data/paymentCoordinator/processClaimProofs_Root1.json
gpsanant: good callout, the leaves we use are

(address, root) = 52 bytes

and

(token address, amount = 52 bytes

stevennevins: it still depends on the implementation for creating the leaves, whether we encode or encodePacked
Screenshot 2024-04-12 at 8 50 55 AM

src/contracts/core/PaymentCoordinator.sol
8sunyuan: had this originally confused, see below

src/contracts/core/PaymentCoordinatorStorage.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a27911bceb4c60db7ad1a53551301dfb4615d020
Added `isRangePaymentForAllHash`

src/test/unit/PaymentCoordinatorUnit.t.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a27911bceb4c60db7ad1a53551301dfb4615d020

src/test/unit/PaymentCoordinatorUnit.t.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a27911bceb4c60db7ad1a53551301dfb4615d020
Cleaned up

src/test/unit/PaymentCoordinatorUnit.t.sol
8sunyuan: added

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+    }
+
+    /**
+     * @notice Sets the address of the entity that can claim payments on behalf of the earner
+     * @param earner The earner whose claimer is being set
+     * @param claimer The address of the entity that can claim payments on behalf of the earner
+     * @dev Only callable by the `earner`
+     */
+    function setClaimerFor(address earner, address claimer) external {
+        require(msg.sender == earner, "PaymentCoordinator.setClaimer: caller is not the earner");
```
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a27911bceb4c60db7ad1a53551301dfb4615d020#diff-2a28246efa54c09dff5f0afef55f490d130b1a13c5b7e30efb51d5647c52a1a2L375

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a27911bceb4c60db7ad1a53551301dfb4615d020#diff-2a28246efa54c09dff5f0afef55f490d130b1a13c5b7e30efb51d5647c52a1a2L190

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/488/commits/a27911bceb4c60db7ad1a53551301dfb4615d020#diff-2a28246efa54c09dff5f0afef55f490d130b1a13c5b7e30efb51d5647c52a1a2L279
Now checkClaim will pass if proofs are valid. Assuming the caller is the valid claimer

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+     * @notice Creates a new distribution root
+     * @param root The merkle root of the distribution
+     * @param paymentCalculationEndTimestamp The timestamp until which payments have been calculated
+     * @param activatedAt timestamp at which that the root can be claimed against
+     * @dev Only callable by the paymentUpdater
+     */
+    function submitRoot(
+        bytes32 root,
+        uint64 paymentCalculationEndTimestamp,
+        uint64 activatedAt
```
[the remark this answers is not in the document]
8sunyuan: we'll go with using params timestamp

src/test/unit/PaymentCoordinatorUnit.t.sol
8sunyuan: You mean `startTimestamp` for the rangePayment right? Think its better to fuzz the inputs

PR #490, opened 2024-03-28
chore: fork test via environment and custom foundry profile

wadealexc:
* regular, non-forking integration tests are run with `forge t --mc Integration`
* mainnet fork tests are run with `env FOUNDRY_PROFILE=forktest forge t --mc Integration`
* mainnet fork tests use the `forktest` foundry profile, which i've defined to only have 20 fuzz runs to limit RPC spam
* i added some notes to the integration test README that describe future directions for reducing RPC spam so we can bump the mainnet fork test intensity up, but for now this works

### DIFF
--- a/.github/workflows/testinparallel.yml
+++ b/.github/workflows/testinparallel.yml
@@ -41,7 +41,7 @@ jobs:
       id: build
 
     - name: Run unit tests
-      run: forge test --no-match-contract FFI --no-match-contract Integration
+      run: forge test --no-match-contract Integration
       env:
         RPC_MAINNET: ${{ secrets.RPC_MAINNET }}
         RPC_HOLESKY: ${{ secrets.RPC_HOLESKY }}
--- a/.github/workflows/testinparallel.yml
+++ b/.github/workflows/testinparallel.yml
@@ -42,6 +42,14 @@ jobs:
         forge build --sizes
       id: build
 
+    - name: Run integration mainnet fork tests
+      run: forge test --match-contract Integration
+      env:
+        FOUNDRY_PROFILE: "forktest"
+        RPC_MAINNET: ${{ secrets.RPC_MAINNET }}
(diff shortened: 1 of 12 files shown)


.github/workflows/testinparallel.yml
```diff
@@ -43,9 +42,24 @@ jobs:
...
+
+    - name: Run integration tests
+      run: forge test --match-contract Integration
       env:
         RPC_MAINNET: ${{ secrets.RPC_MAINNET }}
         RPC_HOLESKY: ${{ secrets.RPC_HOLESKY }}
         CHAIN_ID: ${{ secrets.CHAIN_ID }}
+
+    - name: Run integration mainnet fork tests
+      run: forge test --match-contract Integration
```
8sunyuan, line 60: Perhaps we don't need to run the mainnet fork tests too frequently. Maybe on PR merges into `dev`?
```
      if: github.event_name == 'pull_request' && github.event.action == 'closed' && github.event.pull_request.merged && github.ref == 'refs/heads/dev'
```

src/test/unit/DelegationUnit.t.sol
```diff
@@ -2757,6 +2757,8 @@ contract DelegationManagerUnitTests_Undelegate is DelegationManagerUnitTests {
      * Emits a `StakerUndelegated` event
      */
     function testFuzz_undelegate_noDelegateableShares(address staker) public filterFuzzedAddressInputs(staker) {
+        cheats.assume(staker != defaultOperator);
```
8sunyuan, line 2760: flaky fuzz test?

8sunyuan (approved):
Changes seem good to me. Checked the run and it seems the unit tests completed in ~1min. So I guess we didn't have to run those in parallel with the matrix setup.

.github/workflows/testinparallel.yml
wadealexc: good flag if we run into more issues, but i think this should be fine for now

PR #495, opened 2024-04-04
fix: m2_deploy_from_scratch script for devnet

samlaf:
This PR:
1. fixes the deploy from scratch script for anvil
2. moves the script into a /devnet subfolder (not sure why it wasn't like the others in goerli, holesky, miannet)
3. adds a github action to run this deploy script to make sure its always kept up-to-date

For 3.... should we be running this inside the `run parallel` github action to save on checkout/forge build costs?

### DIFF
--- a/script/deploy/devnet/M2_Deploy_From_Scratch.s.sol
+++ b/script/deploy/devnet/M2_Deploy_From_Scratch.s.sol
@@ -6,24 +6,24 @@ import "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
 import "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
 import "@openzeppelin/contracts/proxy/beacon/UpgradeableBeacon.sol";
 
-import "../../src/contracts/interfaces/IETHPOSDeposit.sol";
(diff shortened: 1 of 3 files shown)


script/deploy/devnet/M2_Deploy_From_Scratch.s.sol
```diff
@@ -87,28 +87,26 @@ contract Deployer_M2 is Script, Test {
...
 
-    function run(string memory configFile) external {
+    function run(string memory configFileName) external {
         // read and log the chainID
         uint256 chainId = block.chainid;
         emit log_named_uint("You are deploying on ChainID", chainId);
 
         // READ JSON CONFIG DATA
-        deployConfigPath = string(bytes(string.concat("script/testing/", configFile)));
+        deployConfigPath = string(bytes(string.concat("script/configs/devnet/", configFileName)));
```
samlaf, line 102: Actually wondering if we shouldn't get rid of the `script/configs/devnet/` prefix and just let the user pass in an absolute path. This way any other avs repo that uses this script as a submodule could create its own deploy config file and pass it in, as opposed to having to go create a PR in eigenlayer-contracts to add a new config file, which doesn't make sense.

stevennevins (approved):

PR #496, opened 2024-04-05
feat: add script for unpausing deposits

ChaoticWalrus:


gpsanant (approved):
lgtm

would be nice to add a deposit test for sanity, but ok to not

let's merge after execution

PR #502, opened 2024-04-07
fix: confusing name for the slot timestamp on the PodManager interface

aimxhaisse:


aimxhaisse:
This is quite confusing today if you rely on the interface as most humans, you think you can pass the withdrawal slot to check if a withdrawal proof was sent in this slot, however you actually need to pass the timestamp of this slot.

PR #504, opened 2024-04-08
chore: merge dev into mainnet

wadealexc:
* fixes the massive commit history disparity between the two branches


ChaoticWalrus (approved):
works for me! but this unfortunately has some tiny merge conflict

PR #514, opened 2024-04-15
feat: payments v1 

8sunyuan:
`PaymentCoordinator` Contract w/ unit tests and deploy scripts for Holesky
Will need to add documentation to /docs/core to followup.


src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,214 @@
...
+        address earner;
+        bytes32 earnerTokenRoot;
+    }
+
+    struct TokenTreeMerkleLeaf {
+        IERC20 token;
+        uint256 cumulativeEarnings;
+    }
+
+    struct PaymentMerkleClaim {
```
stevennevins, line 53: A high-level definition of what a claim is would be helpful. Is there a distinction between quorums? ie if I'm in two quorums would i have one or two claims in the window I was a staker/operator

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,491 @@
...
+
+    /**
+     * @notice Claim payments against a given root (read from distributionRoots[claim.rootIndex])
+     * @param claim The PaymentMerkleClaim to be processed.
+     * Contains the root index, earner, payment leaves, and required proofs
+     * @dev only callable by the valid claimer, that is
+     * if claimerFor[claim.earner] is address(0) then only the earner can claim, otherwise only
+     * claimerFor[claim.earner] can claim the payments.
+     */
+    function processClaim(PaymentMerkleClaim calldata claim) external onlyWhenNotPaused(PAUSED_CLAIM_PAYMENTS) {
```
stevennevins, line 154: What was the reason we didn't revert for duplicate claims?

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,214 @@
...
+ */
+interface IPaymentCoordinator {
+    /// STRUCTS ///
+    struct StrategyAndMultiplier {
+        IStrategy strategy;
+        // weight used to compare shares in multiple strategies against one another
+        uint96 multiplier;
+    }
+
+    struct RangePayment {
```
stevennevins, line 25: The term range is a little confusing here.  I'm assuming it's for a time range in the past the Operator/Staker served?  It would be helpful to clarify

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,214 @@
...
+    event ClaimerForSet(address indexed earner, address indexed oldClaimer, address indexed claimer);
+    /// @notice rootIndex is the specific array index of the newly created root in the storage array 
+    event DistributionRootSubmitted(
+        uint32 indexed rootIndex,
+        bytes32 indexed root,
+        uint32 paymentCalculationEndTimestamp,
+        uint32 activatedAt
+    );
+    /// @notice root is one of the submitted distribution roots that was claimed against
+    event PaymentClaimed(bytes32 indexed root, TokenTreeMerkleLeaf leaf);
```
stevennevins, line 101: Can index the token address

`event PaymentClaimed(bytes32 indexed root, address indexed token, uint256 amount)`

src/contracts/core/PaymentCoordinator.sol
8sunyuan: They don't have to revert, subsequent claims will just result in 0 tokens transferred. But if its better for UX I'm good to add it in

stevennevins: Yeah, it makes sense. That's what I observed. I don't know; I prefer reverting for something that would be a no-op, but I think this is fine the way it is

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: Do we want to update? @ypatil12 @gpsanant

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/519
Added comments here

src/contracts/interfaces/IPaymentCoordinator.sol
8sunyuan: No distinction of quorums. Seems we'll have some further discussion on this today

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,491 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
```
wadealexc, line 2: Need to update pragmas!

src/contracts/core/PaymentCoordinatorStorage.sol
```diff
@@ -0,0 +1,107 @@
...
+ */
+abstract contract PaymentCoordinatorStorage is IPaymentCoordinator {
+
+    /*******************************************************************************
+                               CONSTANTS AND IMMUTABLES 
+    *******************************************************************************/
+
+    /// @notice The EIP-712 typehash for the contract's domain
+    bytes32 public constant DOMAIN_TYPEHASH =
+        keccak256("EIP712Domain(string name,uint256 chainId,address verifyingContract)");
```
wadealexc, line 24: I know these are constant in our other contracts, but I'd like to start moving away from public constants. They bloat bytecode size and aren't needed.

src/contracts/core/PaymentCoordinatorStorage.sol
```diff
@@ -0,0 +1,107 @@
...
+    uint32 public immutable MAX_PAYMENT_DURATION;
+    /// @notice max amount of time that a payment can start in the past
+    uint32 public immutable MAX_RETROACTIVE_LENGTH;
+    /// @notice max amount of time that a payment can start in the future
+    uint32 public immutable MAX_FUTURE_LENGTH;
+    /// @notice absolute min timestamp that a payment can start at
+    uint32 public immutable GENESIS_PAYMENT_TIMESTAMP;
+
+    /// @notice The elegationManager contract for EigenLayer
```
wadealexc, line 33: `elegationManager`

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,214 @@
...
+        uint32 duration;
+    }
+
+    struct DistributionRoot {
+        // merkle root of the distribution
+        bytes32 root;
+        // The timestamp until which payments have been calculated
+        uint32 paymentCalculationEndTimestamp;
+        // timestamp at which the root can be claimed against
+        uint32 activatedAt;
```
wadealexc, line 40: `claimableAt` may be more clear?

(this is fine though)

src/contracts/core/PaymentCoordinatorStorage.sol
```diff
@@ -0,0 +1,107 @@
...
+    uint16 public globalOperatorCommissionBips;
+
+    /// @notice Mapping: earner => the address of the entity to which new payments are directed on behalf of the earner
+    mapping(address => address) public claimerFor;
+
+    /// @notice Mapping: claimer => token => total amount claimed
+    mapping(address => mapping(IERC20 => uint256)) public cumulativeClaimed;
+
+    // /// @notice Timestamp for last submitted 
+    // uint32 public currPaymentCalculationEndTimestamp;
```
wadealexc, line 75: Can be removed

wadealexc: [741ef4f](https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/741ef4f93c2a564b07e77ca97ac595e5eb1436c1)

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+            "PaymentCoordinator._payForRange: startTimestamp too far in the future"
+        );
+
+        // Require rangePayment is for whitelisted strategy or beaconChainETHStrategy
+        address currAddress = address(0);
+        for (uint256 j = 0; j < rangePayment.strategiesAndMultipliers.length; j++) {
+            IStrategy strategy = rangePayment.strategiesAndMultipliers[j].strategy;
+            require(
+                strategyManager.strategyIsWhitelistedForDeposit(strategy) ||
+                    strategy == delegationManager.beaconChainETHStrategy(),
```
wadealexc, line 229: Could we just make this an immutable?

Technically this won't do an external call every iteration because of short circuiting, but I see no reason to call in the first place

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+        for (uint256 i = 0; i < rangePayments.length; i++) {
+            RangePayment calldata rangePayment = rangePayments[i];
+            bytes32 rangePaymentHash = keccak256(abi.encode(msg.sender, paymentNonce, rangePayment));
+            require(
+                !isRangePaymentHash[msg.sender][rangePaymentHash],
+                "PaymentCoordinator._payForRange: range payment hash already submitted"
+            );
+            isRangePaymentHash[msg.sender][rangePaymentHash] = true;
+
+            _payForRange(rangePayment);
```
wadealexc, line 123: Not a huge fan of calling this `_payForRange`. I think this method does two things, neither of which I would call `_payForRange`:

`_validateRangePayment`

and

`_depositRangePayment`

Could we either rename this, or split it into 2 methods?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+            "PaymentCoordinator._payForRange: startTimestamp too far in the past"
+        );
+        require(
+            rangePayment.startTimestamp <= block.timestamp + MAX_FUTURE_LENGTH,
+            "PaymentCoordinator._payForRange: startTimestamp too far in the future"
+        );
+
+        // Require rangePayment is for whitelisted strategy or beaconChainETHStrategy
+        address currAddress = address(0);
+        for (uint256 j = 0; j < rangePayment.strategiesAndMultipliers.length; j++) {
```
wadealexc, line 225: Loop index starting with `j`? Criminal! :joy:

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+     * Earnings are cumulative so earners don't have to claim against all distribution roots they have earnings for,
+     * they can simply claim against the latest root and the contract will calculate the difference between
+     * their cumulativeEarnings and cumulativeClaimed. This difference is then transferred to claimerFor[claim.earner]
+     * @param claim The PaymentMerkleClaim to be processed.
+     * Contains the root index, earner, payment leaves, and required proofs
+     * @dev only callable by the valid claimer, that is
+     * if claimerFor[claim.earner] is address(0) then only the earner can claim, otherwise only
+     * claimerFor[claim.earner] can claim the payments.
+     */
+    function processClaim(PaymentMerkleClaim calldata claim) external onlyWhenNotPaused(PAUSED_CLAIM_PAYMENTS) {
```
wadealexc, line 164: Would prefer to not immediately call an internal function here - unless there's a need I'm missing? This seems to be the only place we call `_processClaim`

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+        rangePayment.token.safeTransferFrom(msg.sender, address(this), rangePayment.amount);
+    }
+
+    /**
+     * @notice Process a payment claim.
+     * @param claim PaymentMerkleClaim struct containing the claim details for proof verification
+     * and payment transfer.
+     * @dev only callable by claimerFor[claim.earner]
+     */
+    function _processClaim(PaymentMerkleClaim calldata claim) internal onlyWhenNotPaused(PAUSED_CLAIM_PAYMENTS) {
```
wadealexc, line 248: Not a fan of `internal` methods with a modifier. Could we move all this up to the external method?

(I assume there were two external methods that called this at some point?)

wadealexc (commented):
Noting for many of the methods here -- we have `nonReentrant` on `payForRange` and `payAllForRange` - but we don't use this modifier on any other methods.

I think if we're concerned about reentrancy, we need to cover all our methods with the modifier.

Edit: Addressed in [6f3d65e](https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/6f3d65e2dc454dcc0a3a57451e003455a69fa0fb)

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+        require(_checkClaim(claim), "PaymentCoordinator._processClaim: claim does not pass the check");
+        // If claimerFor earner is not set, claimer is by default the earner. Else set to claimerFor
+        address claimer = claimerFor[claim.earnerLeaf.earner];
+        if (claimer == address(0)) {
+            claimer = claim.earnerLeaf.earner;
+        }
+        require(msg.sender == claimer, "PaymentCoordinator._checkClaim: caller is not valid claimer");
+        for (uint256 i = 0; i < claim.tokenIndices.length; i++) {
+            _processTokenClaim({
```
wadealexc, line 257: Can we pull this out of the internal method? It's only 4 LoC, no need to bury it.

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/6f3d65e2dc454dcc0a3a57451e003455a69fa0fb
Good catch

src/contracts/core/PaymentCoordinatorStorage.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/6f3d65e2dc454dcc0a3a57451e003455a69fa0fb

src/contracts/core/PaymentCoordinatorStorage.sol
8sunyuan: fixed

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,502 @@
...
+     * @dev Strategies must be in ascending order of addresses to check for duplicates
+     * @dev This function will revert if the `rangePayment` is malformed,
+     * e.g. if the `strategies` and `weights` arrays are of non-equal lengths
+     */
+    function payForRange(
+        RangePayment[] calldata rangePayments
+    ) external onlyWhenNotPaused(PAUSED_PAY_FOR_RANGE) nonReentrant {
+        for (uint256 i = 0; i < rangePayments.length; i++) {
+            RangePayment calldata rangePayment = rangePayments[i];
+            bytes32 rangePaymentHash = keccak256(abi.encode(msg.sender, paymentNonce, rangePayment));
```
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/6f3d65e2dc454dcc0a3a57451e003455a69fa0fb
Added paymentNonce mapping

src/contracts/core/PaymentCoordinator.sol
8sunyuan: corrected my sins

src/contracts/interfaces/IPaymentCoordinator.sol
```diff
@@ -0,0 +1,285 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
```
wadealexc, line 2: Missing pragma update here

src/test/events/IPaymentCoordinatorEvents.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
```
wadealexc, line 2: Missing pragma update

src/test/unit/PaymentCoordinatorUnit.t.sol
```diff
@@ -0,0 +1,1617 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
```
wadealexc, line 2: Missing pragma update

8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/f947cbc4661436e1b90a55531b900ed5f0a6e16e
Got it, should have searched whole repo to start

src/contracts/core/PaymentCoordinator.sol
stevennevins: I second the modifier comment

src/contracts/core/PaymentCoordinator.sol
wadealexc: personal preference is that it's not `public` -- but i can live with it!

src/contracts/core/PaymentCoordinator.sol
wadealexc: [6f3d65e](https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/6f3d65e2dc454dcc0a3a57451e003455a69fa0fb)

src/contracts/core/PaymentCoordinator.sol
wadealexc: [6f3d65e](https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/6f3d65e2dc454dcc0a3a57451e003455a69fa0fb)

src/contracts/core/PaymentCoordinator.sol
wadealexc: [6f3d65e](https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/6f3d65e2dc454dcc0a3a57451e003455a69fa0fb)

src/test/unit/PaymentCoordinatorUnit.t.sol
wadealexc: [f947cbc](https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/f947cbc4661436e1b90a55531b900ed5f0a6e16e)

src/test/events/IPaymentCoordinatorEvents.sol
wadealexc: [f947cbc](https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/f947cbc4661436e1b90a55531b900ed5f0a6e16e)

src/contracts/interfaces/IPaymentCoordinator.sol
wadealexc: [f947cbc](https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/f947cbc4661436e1b90a55531b900ed5f0a6e16e)

src/contracts/interfaces/IPaymentCoordinator.sol
ypatil12: Just realized, we need to add the earner in this event too!

I'm fine with indexed or not indexed, but definitely need earner so the UI knows how much has been claimed already

src/contracts/core/PaymentCoordinatorStorage.sol
```diff
@@ -0,0 +1,108 @@
...
+        uint32 _MAX_RETROACTIVE_LENGTH,
+        uint32 _MAX_FUTURE_LENGTH,
+        uint32 _GENESIS_PAYMENT_TIMESTAMP
+    ) {
+        delegationManager = _delegationManager;
+        strategyManager = _strategyManager;
+        MAX_PAYMENT_DURATION = _MAX_PAYMENT_DURATION;
+        MAX_RETROACTIVE_LENGTH = _MAX_RETROACTIVE_LENGTH;
+        MAX_FUTURE_LENGTH = _MAX_FUTURE_LENGTH;
+        GENESIS_PAYMENT_TIMESTAMP = _GENESIS_PAYMENT_TIMESTAMP;
```
ypatil12, line 99: We need to make sure that `_GENESIS_PAYMENT_TIMESTAMP` lands on a calculationIntervalSeconds

ypatil12: How do we enforce it? Check in deploy script or codified

8sunyuan: `calculationIntervalSeconds` is being initialized so it probably has to be double checked in deploy script

src/contracts/interfaces/IPaymentCoordinator.sol
ypatil12: Based on standup today:

We need to emit (earner, claimer, claimedAmount, token, root). Leaf doesn't really matter anymore

8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/f13e1c21c5c4a90f1550fb80c132428bbacd731c
Updated and tested

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,474 @@
...
+                currAddress < address(strategy),
+                "PaymentCoordinator._payForRange: strategies must be in ascending order to handle duplicates"
+            );
+            currAddress = address(strategy);
+        }
+
+        rangePayment.token.safeTransferFrom(msg.sender, address(this), rangePayment.amount);
+    }
+
+    function _checkClaim(PaymentMerkleClaim calldata claim, DistributionRoot memory root) internal view returns (bool) {
```
wadealexc, line 263: The `bool` return for this method is kinda useless, because if anything fails verification, the method will revert.

Can we remove this?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,474 @@
...
+
+    /// @notice return the hash of the earner's leaf
+    function calculateEarnerLeafHash(EarnerTreeMerkleLeaf calldata leaf) public pure returns (bytes32) {
+        return keccak256(abi.encodePacked(leaf.earner, leaf.earnerTokenRoot));
+    }
+
+    /// @notice returns the hash of the earner's token leaf
+    function calculateTokenLeafHash(TokenTreeMerkleLeaf calldata leaf) public pure returns (bytes32) {
+        return keccak256(abi.encodePacked(leaf.token, leaf.cumulativeEarnings));
+    }
```
wadealexc, line 369: We should consider adding some specific delimiters to each of these hash calculations.

i.e:

```
uint8 EARNER_LEAF = 1;
uint8 TOKEN_LEAF = 2;

earnerLeafHash = keccak(EARNER_LEAF, earner, tokenRoot);

tokenLeafHash = keccak(TOKEN_LEAF, token, cumulativeEarnings);
```

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,474 @@
...
+        if (claimer == address(0)) {
+            claimer = earner;
+        }
+        require(msg.sender == claimer, "PaymentCoordinator.processClaim: caller is not valid claimer");
+        for (uint256 i = 0; i < claim.tokenIndices.length; i++) {
+            TokenTreeMerkleLeaf calldata tokenLeaf = claim.tokenLeaves[i];
+
+            // Calculate amount to claim and update cumulativeClaimed
+            // Will revert if new leaf cumulativeEarnings is less than cumulative claimed
+            uint256 claimAmount = tokenLeaf.cumulativeEarnings - cumulativeClaimed[earner][tokenLeaf.token];
```
wadealexc, line 181: Doing an explicit check/require for the underflow case here will make it easier to write tests!

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,474 @@
...
+
+    /**
+     * @notice verify inclusion of the token claim proof in the earner token root hash (earnerTokenRoot).
+     * The token leaf comprises of the IERC20 token and cumulativeAmount of earnings.
+     * @param earnerTokenRoot root hash of the earner token subtree
+     * @param tokenLeafIndex index of the token leaf
+     * @param tokenProof proof of the token leaf in the earner token subtree
+     * @param tokenLeaf token leaf to be verified
+     */
+    function _verifyTokenClaimProof(
```
wadealexc, line 302: Because we don't validate the size of `tokenLeafIndex` relative to `tokenProof`, there are multiple indices that are valid for any single given proof.

We're not using the indices to check whether a leaf has been claimed or not, but we should probably adhere to best practices anyway!

IIRC, indices should meet this criterion: `index < 2**(proof.length/32)`

(this applies to the other merkle proof method below)

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,474 @@
...
+        }
+    }
+
+    /**
+     * @notice Creates a new distribution root. activatedAt is set to block.timestamp + activationDelay
+     * @param root The merkle root of the distribution
+     * @param paymentCalculationEndTimestamp The timestamp until which payments have been calculated
+     * @dev Only callable by the paymentUpdater
+     */
+    function submitRoot(
```
wadealexc, line 195: Should this be pausable?

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,474 @@
...
+        bytes32 root,
+        uint32 paymentCalculationEndTimestamp
+    ) external onlyPaymentUpdater {
+        require(
+            paymentCalculationEndTimestamp > currPaymentCalculationEndTimestamp,
+            "PaymentCoordinator.submitRoot: new root must be for newer calculated period"
+        );
```
wadealexc, line 202: What is this used for? It doesn't seem to be a particularly useful check - there are no rules enforcing this field's value except that it's larger than the last time `submitRoot` was called.

If (for whatever reason) the payment updater calls this with `type(uint32).max`, `submitRoot` is bricked until we upgrade and fix it.

Is there some upper bound on how much this can increase by, or some relation it should have to `block.timestamp`?

8sunyuan: This check is just a sanity check for the paymentUpdater to submit roots sequentially and `currPaymentCalculationEndTimestamp` is read from the backend to know where to start calculating payments from.
We have MAX_RETROACTIVE_LENGTH and MAX_FUTURE_LENGTH for valid startTimestamps for RangePayments.
I think we can just reuse MAX_FUTURE_LENGTH to check @gpsanant 
```
        require(
            paymentCalculationEndTimestamp < block.timestamp + MAX_FUTURE_LENGTH,
            "PaymentCoordinator.submitRoot: paymentCalculationEndTimestamp too far in the future"
        );
```

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/24fb70726a39a125e61cc19a361d7cc20aa7b930
Added pause flag

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/24fb70726a39a125e61cc19a361d7cc20aa7b930
Added checks and tests for both verifyToken and verifyEarner

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/24fb70726a39a125e61cc19a361d7cc20aa7b930
Added

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/24fb70726a39a125e61cc19a361d7cc20aa7b930
Added and updated backend claimgen as well

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/24fb70726a39a125e61cc19a361d7cc20aa7b930

src/contracts/core/PaymentCoordinator.sol
8sunyuan: Used bit shifting for cheaper gas

src/contracts/core/PaymentCoordinator.sol
8sunyuan: Will go with this since it doesn't make sense for payments to be calculated in the future
```
        require(
            paymentCalculationEndTimestamp < block.timestamp,
            "PaymentCoordinator.submitRoot: paymentCalculationEndTimestamp can't be in the future"
        );
```

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,501 @@
...
+        bytes32 root,
+        uint32 paymentCalculationEndTimestamp
+    ) external onlyWhenNotPaused(PAUSED_SUBMIT_ROOTS) onlyPaymentUpdater {
+        require(
+            paymentCalculationEndTimestamp > currPaymentCalculationEndTimestamp,
+            "PaymentCoordinator.submitRoot: new root must be for newer calculated period"
+        );
+        require(
+            paymentCalculationEndTimestamp < block.timestamp,
+            "PaymentCoordinator.submitRoot: paymentCalculationEndTimestamp can't be in the future"
```
wadealexc, line 217: Requesting removal of the `'` apostrophe in `can't`, because it's making github's syntax highlighting wonky as fuck

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,501 @@
...
+     * @notice Sets the address of the entity that can claim payments on behalf of the earner
+     * @param claimer The address of the entity that can claim payments on behalf of the earner
+     * @dev Only callable by the `earner`
+     */
+    function setClaimerFor(address claimer) external {
+        address earner = msg.sender;
+        address prevClaimer = claimerFor[earner];
+        claimerFor[earner] = claimer;
+        emit ClaimerForSet(earner, prevClaimer, claimer);
+    }
```
wadealexc, line 385: This and other state changing `external` methods should be above our `internal` methods, just to keep with the convention used in other files.

`pure` and `view` methods can stay down here

8sunyuan: separated by external, internal, and then view/pure
https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/02a4fcacbb2016a91bfe09150ddc758721d26540#diff-2a28246efa54c09dff5f0afef55f490d130b1a13c5b7e30efb51d5647c52a1a2

src/contracts/core/PaymentCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/3866a4a20ad488378275e6fdb7c5a8ac4c6c2b09

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,525 @@
...
+            require(
+                tokenLeaf.cumulativeEarnings >= currCumulativeClaimed,
+                "PaymentCoordinator.processClaim: cumulativeEarnings must be gte than cumulativeClaimed"
+            );
+
+            // Calculate amount to claim and update cumulativeClaimed
+            uint256 claimAmount = tokenLeaf.cumulativeEarnings - currCumulativeClaimed;
+            cumulativeClaimed[earner][tokenLeaf.token] = tokenLeaf.cumulativeEarnings;
+
+            tokenLeaf.token.safeTransfer(claimer, claimAmount);
```
seongyun-ko, line 200: I think we can save gas by reducing the number of calls to `safeTransfer` since these all towards the same claimer.
The value can be aggregated to a local variable.

```
uint256 totalClaimAmount = 0;
for (uint256 i = 0; i < claim.tokenIndices.length; i++) {
            TokenTreeMerkleLeaf calldata tokenLeaf = claim.tokenLeaves[i];

            // Calculate amount to claim and update cumulativeClaimed
            uint256 claimAmount = tokenLeaf.cumulativeEarnings - currCumulativeClaimed;
            cumulativeClaimed[earner][tokenLeaf.token] = tokenLeaf.cumulativeEarnings;
            totalClaimAmount += claimAmount;
        }
        tokenLeaf.token.safeTransfer(claimer, totalClaimAmount);
```

seongyun-ko: Plus, a simple batch function for the multiple earners with the same claimer would help too :)

```
function processClaim(PaymentMerkleClaim[] calldata claims)
```

8sunyuan: Interesting, so batch claiming for different earners, but assuming they all have the same claimer

src/contracts/core/PaymentCoordinator.sol
```diff
@@ -0,0 +1,529 @@
...
+    /// @dev Index for flag that pauses submitRoots
+    uint8 internal constant PAUSED_SUBMIT_ROOTS = 3;
+
+    /// @dev Salt for the earner leaf, meant to distinguish from tokenLeaf since they have the same sized data
+    uint8 internal constant EARNER_LEAF_SALT = 0;
+    /// @dev Salt for the token leaf, meant to distinguish from earnerLeaf since they have the same sized data
+    uint8 internal constant TOKEN_LEAF_SALT = 1;
+
+    /// @dev Chain ID at the time of contract deployment
+    uint256 internal immutable ORIGINAL_CHAIN_ID;
```
wadealexc, line 46: We have some constants/immutables here, and some in `PaymentCoordinatorStorage`. Is there a reason we have these here? Otherwise I'd say we should put them all in one place for consistency.

wadealexc (approved):
This looks awesome! Great work - super clean code.

I'm going to request one additional test: for the index bounds check added to `_verifyEarnerClaimProof` and `_verifyTokenClaimProof`, I'd just like us to verify that a claim using a leaf at exactly the maximum size for the tree in question is in fact claimable. Just to sanity check that the bound check is correct!

Approving the PR as everything else looks good :)

src/contracts/core/PaymentCoordinator.sol
8sunyuan: Personally prefer the immutables being in `PaymentCoordinatorStorage` as it has the constructor defining them there (although we could just have the immutables in `PaymentCoordinator` itself).
Cleaned them up with constants into `PaymentCoordinator` and immutables in `PaymentCoordinatorStorage`
https://github.com/Layr-Labs/eigenlayer-contracts/pull/514/commits/90e157c318910a501ca4e4db9c2b6ebe0240d80c

8sunyuan:
@wadealexc 
https://github.com/Layr-Labs/eigenlayer-contracts/blob/b0499ebc78d311fae1cd5d6e43f3f8cd1719c36b/src/test/unit/PaymentCoordinatorUnit.t.sol#L1735
Unit test for the max earnerIndex and max tokenIndex for full binary tree of 8 leaves for both earners and token leaves.

https://github.com/Layr-Labs/eigenlayer-contracts/blob/b0499ebc78d311fae1cd5d6e43f3f8cd1719c36b/src/test/unit/PaymentCoordinatorUnit.t.sol#L1688
Unit test for reverting on larger indexes that could pass the proofs before the added require

src/contracts/core/PaymentCoordinator.sol
seongyun-ko: yes. @8sunyuan we would set the claimer for the +22k eigenpods to the same address and let it claim on behalf.

OR if the operator fee can be set to 100% I think we can play with it.

src/contracts/core/PaymentCoordinator.sol
wadealexc: have tried implementing this here: https://github.com/Layr-Labs/eigenlayer-contracts/pull/530/files

8sunyuan: Thanks for this 👍

src/test/EigenPod.t.sol
```diff
@@ -1256,29 +1256,6 @@ contract EigenPodTests is ProofParsing, EigenPodPausingConstants {
...
-        bytes32 leaf,
-        uint256 index,
-        bytes memory proof
-    ) public {
-        cheats.assume(proof.length % 32 != 0);
-        cheats.expectRevert(
-            bytes("Merkle.processInclusionProofKeccak: proof length should be a non-zero multiple of 32")
-        );
-        Merkle.verifyInclusionKeccak(proof, root, leaf, index);
-    }
```
wadealexc, line 1280: Hmm... this should be fine to remove. I checked `BeaconChainProofs`, and we're already requiring specific lengths for each of the proofs, so we shouldn't need extra checks/tests here.

wadealexc: oh yeah and we're using sha256 there, anyway!

8sunyuan: Yup, I checked that the Merkle keccak256 functions weren't being used anywhere

8sunyuan:
PaymentCoordinator looking good to merge right now as part of the MVP release. Looking to go ahead with audits shortly after

PR #515, opened 2024-04-16
feat: partial withdrawal batching

wadealexc:
Mostly uses the design described [here](https://hackmd.io/U36dE9lnQha3tbf7D0GtKw), but have made some tweaks.

Notes:
* Beacon chain timestamps: moving off of Succinct means we actually use timestamps differently, and we need to make sure this doesn't cause any issues.
  * `startCheckpoint` sets a checkpoint timestamp of `block.timestamp` and uses the parent beacon block root for checkpoint proofs. This means the checkpoint timestamp technically corresponds to the slot AFTER the beacon block root.
  * OTOH, Succinct's oracle timestamps use the slot that the beacon block root comes from.


src/contracts/pods/EigenPod.sol
```diff
@@ -14,42 +14,33 @@ import "../libraries/Endian.sol";
...
  * @author Layr Labs, Inc.
  * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
- * @notice The main functionalities are:
- * - creating new ETH validators with their withdrawal credentials pointed to this contract
- * - proving from beacon chain state roots that withdrawal credentials are pointed to this contract
- * - proving from beacon chain state roots the balances of ETH validators with their withdrawal credentials
- *   pointed to this contract
- * - updating aggregate balances in the EigenPodManager
- * - withdrawing eth when withdrawals are initiated
  * @notice This EigenPod Beacon Proxy implementation adheres to the current Capella consensus specs
```
gpsanant, line 27: deneb lol

src/contracts/pods/EigenPod.sol
```diff
@@ -169,133 +135,144 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-        // Verify passed-in beaconStateRoot against oracle-provided block root:
-        BeaconChainProofs.verifyStateRootAgainstLatestBlockRoot({
-            latestBlockRoot: eigenPodManager.getBlockRootAtTimestamp(oracleTimestamp),
-            beaconStateRoot: stateRootProof.beaconStateRoot,
-            stateRootProof: stateRootProof.proof
+        // Snapshot pod balance at the start of the checkpoint. Once the checkpoint is finalized,
+        // this amount will be added to the total validator balance delta and credited as shares.
+        uint256 podBalanceWei = 
+            address(this).balance
+                - nonBeaconChainETHBalanceWei
```
gpsanant, line 172: should this just be autorestaked?

src/contracts/pods/EigenPod.sol
```diff
@@ -169,133 +135,144 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+     * (see `_updateCheckpoint` for more details)
+     * @dev This method can only be called when there is a currently-active checkpoint.
+     * @param stateRootProof proves a beacon state root against the checkpoint's `beaconBlockRoot`
+     * @dev Note that if the `beaconStateRoot` has already been verified for this checkpoint, it does
+     * not need to be verfied again.
+     * @param proofs Proofs for one or more validator current balances against the checkpoint's `beaconStateRoot`
      */
-    function verifyAndProcessWithdrawals(
-        uint64 oracleTimestamp,
+    function verifyCheckpointProof(
```
gpsanant, line 200: Proofs

src/contracts/pods/EigenPod.sol
wadealexc: `nonBeaconChainETHBalanceWei`? Yeah I'm all for it.

Also worth noting that the pod balance calculated here truncates any wei amount so that it sits in storage as gwei. we should figure out if that's okay or not! (the amount isn't fully lost; you can always top up and prove another checkpoint)

src/contracts/pods/EigenPodStorage.sol
```diff
@@ -0,0 +1,57 @@
...
+    uint256 public nonBeaconChainETHBalanceWei;
+
+    /// @notice This variable tracks the total amount of partial withdrawals claimed via merkle proofs prior to a switch to ZK proofs for claiming partial withdrawals
+    uint64 __deprecated_sumOfPartialWithdrawalsClaimedGwei;
+
+    /// @notice Number of validators with proven withdrawal credentials, who do not have proven full withdrawals
+    uint256 public activeValidatorCount;
+
+    /// @notice The timestamp of the last checkpoint finalized
+    uint64 public lastFinalizedCheckpoint;
```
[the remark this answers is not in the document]
wadealexc: Yeah, good call out... I didn't use this because it's a really long name. Any other suggestions?

src/contracts/pods/EigenPod.sol
```diff
@@ -454,24 +508,25 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
         require(
             validatorInfo.status == VALIDATOR_STATUS.INACTIVE,
-            "EigenPod.verifyCorrectWithdrawalCredentials: Validator must be inactive to prove withdrawal credentials"
+            "EigenPod._verifyWithdrawalCredentials: validator must be inactive to prove withdrawal credentials"
         );
 
-        // Ensure the `validatorFields` we're proving have the correct withdrawal credentials
+        // Validator should not already be in the process of exiting
+        require(
+            validatorFields.getExitEpoch() == FAR_FUTURE_EPOCH,
```
gpsanant, line 516: do we need this check?

wadealexc: i had originally included it because `verifyWithdrawalCredentials` was requiring that `beaconTimestamp >= currentCheckpointTimestamp`. For the `==` case, if your validator is already exited, you may end up proving withdrawal credentials but your validator's ETH is already in the pod, meaning you'd be able to double your shares.

I removed the `==` condition though and made it strictly greater than. MAYBE we don't need this check... but IMO it's probably just safer to keep it? If your validator is about to exit, maybe don't verify its WC?

src/contracts/pods/EigenPod.sol
```diff
@@ -454,24 +508,25 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+            "EigenPod._verifyWithdrawalCredentials: validator must not be exiting"
+        );
+
+        // Ensure the validator's withdrawal credentials are pointed at this pod
         require(
             validatorFields.getWithdrawalCredentials() == bytes32(_podWithdrawalCredentials()),
```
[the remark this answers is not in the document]
wadealexc: oh, that's a good idea! yeah, we could do that. will add it to a list of features to discuss in the PR comment above.

src/contracts/pods/EigenPod.sol
wadealexc: [9359ccd](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/9359ccd9d94b3e93f1ec74c63137326627527c3a)

src/contracts/pods/EigenPod.sol
wadealexc: [9359ccd](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/9359ccd9d94b3e93f1ec74c63137326627527c3a)

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -144,6 +140,37 @@ library BeaconChainProofs {
...
+            Merkle.verifyInclusionSha256({
+                proof: proof.proof,
+                root: beaconStateRoot,
+                leaf: proof.balanceRoot,
+                index: balanceIndex
+            }),
+            "BeaconChainProofs.verifyValidatorBalance: Invalid merkle proof"
+        );
+
+        return getBalanceAtIndex(proof.balanceRoot, validatorIndex);
```
gpsanant, line 171: TODO?

src/contracts/pods/EigenPod.sol
```diff
@@ -485,256 +541,108 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+     * @dev Finish progress on a checkpoint and store it in state.
+     * @dev If the checkpoint has no proofs remaining, it is finalized:
+     * - a share delta is calculated and sent to the `EigenPodManager`
+     * - the checkpointed `podBalance` is added to `withdrawableRestakedExecutionLayerGwei`
+     * - `lastFinalizedCheckpoint` is updated
+     * - `currentCheckpoint` and `currentCheckpointTimestamp` are deleted
+     */
+    function _updateCheckpoint(Checkpoint memory checkpoint) internal {
+        if (checkpoint.proofsRemaining == 0) {
+            int256 totalShareDelta =
```
gpsanant, line 628: this isn't a delta?

gpsanant: in addition, this needs to be a multiple of gwei (required by EPM), which may be invalidated by selfdestruct. we should probably get rid of that requirement

wadealexc: yes it is

wadealexc: https://hackmd.io/U36dE9lnQha3tbf7D0GtKw#What-is-a-Checkpoint

wadealexc: the question of what to do with any < 1 gwei amount left over is a good question though. i'm not 100% sure what to do with it.

src/contracts/pods/EigenPod.sol
wadealexc: Actually I thought of a reason we still need this check -- if these things happen *within the same epoch, in this order*:
1. validator V exits from beacon chain; ETH hits pod
2. pod owner starts checkpoint
3. pod owner verifies V's withdrawal credentials using a timestamp after the checkpoint starts, but still in the same epoch

... then V's effective balance will not reflect the exit and they'll get double counted (ETH in pod counted during checkpoint + effective balance counted in verifyWC).

To prevent this, we require that `verifyWC` both:
* uses a timestamp after the current/last finalized checkpoints
* is not callable if the validator has initiated an exit

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -376,11 +265,18 @@ library BeaconChainProofs {
     }
 
     /**
-     * @dev Retrieves a validator's withdrawable epoch
+     * @dev Returns true IFF the validator is marked slashed
+     */
+    function getSlashStatus(bytes32[] memory validatorFields) internal pure returns (bool) {
+        return validatorFields[VALIDATOR_SLASHED_INDEX] != 0;
+    }
```
ChaoticWalrus, line 272: might want to just call this `isValidatorSlashed` or something? this kind of naming makes it even more unambiguous that returning 'true' means the validator is slashed

src/contracts/pods/EigenPod.sol
```diff
@@ -59,43 +55,21 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-
-    /// @notice an indicator of whether or not the podOwner has ever "fully restaked" by successfully calling `verifyCorrectWithdrawalCredentials`.
-    bool public hasRestaked;
-
-    /// @notice This is a mapping of validatorPubkeyHash to timestamp to whether or not they have proven a withdrawal for that timestamp
-    mapping(bytes32 => mapping(uint64 => bool)) public provenWithdrawal;
+    /// @notice If a validator is slashed on the beacon chain and their balance has not been checkpointed
+    /// within `TIME_TILL_STALE_BALANCE` of the current block, they are eligible to be marked "stale"
+    /// via `verifyStaleBalance`.
+    uint256 internal constant TIME_TILL_STALE_BALANCE = 2 weeks;
```
ChaoticWalrus, line 64: so this means people would have 2 weeks to update the balance of a validator that gets slashed on the beacon chain?
we should discuss the right value for this, but am fine leaving that as a detail until later.

src/contracts/pods/EigenPod.sol
```diff
@@ -169,133 +140,114 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-     * @param validatorFields are the fields of the "Validator Container", refer to consensus specs
-     * @dev For more details on the Beacon Chain spec, see: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#validator
+     * @dev Create a checkpoint used to prove this pod's active validator set. Checkpoints are completed 
+     * by submitting one checkpoint proof per ACTIVE validator. During the checkpoint process, the total 
+     * change in ACTIVE validator balance is tracked, and any validators with 0 balance are marked `WITHDRAWN`.
+     * @dev Once finalized, the pod owner is awarded shares corresponding to:
+     * - the total change in their ACTIVE validator balances
+     * - any ETH in the pod not already awarded shares.
+     * @dev A checkpoint cannot be created if the pod already has an outstanding checkpoint. If
+     * this is the case, the pod owner MUST complete the existing checkpoint before starting a new one.
```
ChaoticWalrus, line 161: what would be the side effects of allowing cancelling a checkpoint and starting a new one?
is there intermediate storage that would get messed up, or is it more about making sure stale balances can be proved by an external party without the pod owner griefing this external party or something?

src/contracts/pods/EigenPod.sol
```diff
@@ -169,133 +140,114 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-                validatorFieldsProofs[i], // Use validator fields proof because contains the effective balance
-                validatorFields[i]
-            );
-        }
-        eigenPodManager.recordBeaconChainETHBalanceUpdate(podOwner, sharesDeltaGwei * int256(GWEI_TO_WEI));
+    function startCheckpoint()
+        external
+        onlyEigenPodOwner() 
+        onlyWhenNotPaused(PAUSED_START_CHECKPOINT) 
+        afterRestaking(uint64(block.timestamp)) /// TODO - this is the wrong condition
```
ChaoticWalrus, line 167: is this incorrect because of the difference between our existing oracle and the direct usage of EIP-4788 here?

src/contracts/pods/EigenPod.sol
```diff
@@ -169,133 +140,114 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        // Process each checkpoint proof submitted
+        for (uint256 i = 0; i < proofs.length; i++) {
+
+            // Process a checkpoint proof for a validator. 
+            // - the validator MUST be in the ACTIVE state
+            // - the validator MUST NOT have already been proven for this checkpoint
+            //
+            // If the proof shows the validator has a balance of 0, they are marked `WITHDRAWN`.
+            // The assumption is that if this is the case, any withdrawn ETH was already in
+            // the pod when `startCheckpoint` was originally called.
```
ChaoticWalrus, line 220: is this an OK assumption? seems like the ETH could ultimately be recovered/proven in a subsequent checkpoint?
mostly making sure there is no case where ETH could get stuck.

src/contracts/pods/EigenPodStorage.sol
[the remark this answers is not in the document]
gpsanant: im the world renowned namer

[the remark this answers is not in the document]
wadealexc: ser i asked if you had suggestions not to see your credentials

src/contracts/libraries/BeaconChainProofs.sol
wadealexc: fair, the naming was supposed to be in line with the other validatorFields query methods, but you're right that `isValidatorSlashed` is much clearer. Down to change this.

src/contracts/pods/EigenPod.sol
wadealexc: If their last proof for this validator was 2 weeks ago and the validator gets slashed, they only have 6 hours.

OTOH if they proved the validator like last block and then they got slashed, they have 2 weeks!

src/contracts/pods/EigenPod.sol
```diff
@@ -59,43 +55,21 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+    /// @notice If a validator is slashed on the beacon chain and their balance has not been checkpointed
+    /// within `TIME_TILL_STALE_BALANCE` of the current block, they are eligible to be marked "stale"
+    /// via `verifyStaleBalance`.
+    uint256 internal constant TIME_TILL_STALE_BALANCE = 2 weeks;
 
-    /// @notice This is a mapping that tracks a validator's information by their pubkey hash
-    mapping(bytes32 => ValidatorInfo) internal _validatorPubkeyHashToInfo;
+    /// @notice A `verifyStaleBalance` proof MUST be older than `STALENESS_GRACE_PERIOD`, to give time
+    /// to a pod owner to prove the slashed validator's balance
+    uint256 internal constant STALENESS_GRACE_PERIOD = 6 hours;
```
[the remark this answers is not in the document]
wadealexc: We may not need the grace period though, because the only repercussion is "start a checkpoint". In that case, the `TIME_TILL_STALE_BALANCE` might be sufficient.

After all, in order to do a proof, you'd need to do a checkpoint anyway :D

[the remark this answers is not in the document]
wadealexc: (the repercussion used to be "worse", hence the grace period)

src/contracts/pods/EigenPod.sol
wadealexc: indeed, the intermediate storage is the issue. we do multiple state updates in the `_validatorPubkeyHashToInfo` mapping. it may be possible to remove all of those and make a fully self-contained checkpoint system, but it adds a lot of complexity.

src/contracts/pods/EigenPod.sol
wadealexc: i'm not 100% sure it's wrong, just smells funny. I need to think about what happens if you call `activateRestaking` and then try to verify WC. 

Although with the new PR to change the way `activateRestaking` works, this may be less of a problem.

src/contracts/pods/EigenPod.sol
wadealexc: Should be a good assumption! It's kinda the main thing we need to be true, so if there are problems with it, speak up!

I describe how the assumption works here: https://hackmd.io/@-HV50kYcRqOjl_7du8m1AA/SkJPfqBeC#What-Gets-Snapshotted

src/contracts/libraries/BeaconChainProofs.sol
wadealexc: [ff7fd32](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/ff7fd32a609654127efda187f23d3a48bd4b6c1e)

src/contracts/pods/EigenPodStorage.sol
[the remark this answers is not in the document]
wadealexc: [a4615b7](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/a4615b7ad2111b82946061bc2500dda1b74cabc3)

src/contracts/pods/EigenPod.sol
```diff
@@ -169,133 +140,114 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-                stateRootProof.beaconStateRoot,
-                validatorFieldsProofs[i], // Use validator fields proof because contains the effective balance
-                validatorFields[i]
-            );
-        }
-        eigenPodManager.recordBeaconChainETHBalanceUpdate(podOwner, sharesDeltaGwei * int256(GWEI_TO_WEI));
+    function startCheckpoint()
+        external
+        onlyEigenPodOwner() 
+        onlyWhenNotPaused(PAUSED_START_CHECKPOINT)
```
wadealexc, line 166: Note: this pause condition needs to be enforced for `verifyStaleBalance` as well, since that method can start a checkpoint.

wadealexc:
> @wadealexc can we add an event when a checkpoint is created with the checkpoint timestamp and the beacon block root
> 
> also an event with the index of the validator and the checkpoint timestamp [whenever a validator is proven for a checkpoint](https://github.com/Layr-Labs/eigenlayer-contracts/blob/e34c7eab69b1114cf31995fd0bd58e49a23523de/src/contracts/pods/EigenPod.sol#L551)

Added [705b81d](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/705b81d7b38fa5962cd8d2cd58ba695245d72adb)

src/contracts/pods/EigenPod.sol
```diff
@@ -485,256 +505,126 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+            validatorInfo.status == VALIDATOR_STATUS.ACTIVE,
+            "EigenPod._verifyCheckpointProof: validator must be ACTIVE"
         );
 
-        // 2. Balance updates should only be performed on "ACTIVE" validators
+        // Ensure we aren't proving a validator twice for the same checkpoint. This will fail if:
+        // - validator submitted twice during this checkpoint
+        // - validator withdrawal credentials verified after checkpoint starts, then submitted
+        //   as a checkpoint proof
+        // TODO - we might want to "skip" and emit an event, rather than revert?
```
ChaoticWalrus, line 535: I think we should skip here instead of reverting. emitting an event in the case of skipping seems...reasonable, at least.
my thoughts: if we revert, then there is a potential griefing vector, whereby an attacker can cause a large transaction (proving e.g. 50 balances) to revert, by submitting their own transaction that just proves a single balance.
I don't think it's a particularly realistic griefing concern, but I'd rather just not have the attack vector present at all.
Skipping would in general also resolve race conditions more elegantly, IMO.

src/contracts/pods/EigenPod.sol
```diff
@@ -485,256 +505,126 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-        // Update validator balance and timestamp, and save to state:
-        validatorInfo.restakedBalanceGwei = newRestakedBalanceGwei;
-        validatorInfo.mostRecentBalanceUpdateTimestamp = oracleTimestamp;
-        _validatorPubkeyHashToInfo[validatorPubkeyHash] = validatorInfo;
+        // Update validator state. If their new balance is 0, they are marked `WITHDRAWN`
+        validatorInfo.restakedBalanceGwei = newBalanceGwei;
+        validatorInfo.mostRecentBalanceUpdateTimestamp = beaconTimestamp;
+        if (newBalanceGwei == 0) {
+            activeValidatorCount--;
+            validatorInfo.status = VALIDATOR_STATUS.WITHDRAWN;
```
ChaoticWalrus, line 554: probably want to emit an event when this happens?

src/contracts/pods/EigenPod.sol
```diff
@@ -485,256 +505,126 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+                (int256(checkpoint.podBalanceGwei) + checkpoint.balanceDeltasGwei) * int256(GWEI_TO_WEI);
+
+            // Add any native ETH in the pod to `withdrawableRestakedExecutionLayerGwei`
+            // ... this amount can be withdrawn via the `DelegationManager` withdrawal queue
+            withdrawableRestakedExecutionLayerGwei += uint64(checkpoint.podBalanceGwei);
+
+            // Finalize the checkpoint
+            lastCheckpointTimestamp = currentCheckpointTimestamp;
+            delete currentCheckpointTimestamp;
+            delete currentCheckpoint;
```
ChaoticWalrus, line 621: probably want some kind of event here too?

src/contracts/pods/EigenPod.sol
```diff
@@ -761,24 +671,12 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
         return sha256(abi.encodePacked(validatorPubkey, bytes16(0)));
     }
 
-    /**
-     * Calculates delta between two share amounts and returns as an int256
-     */
-    function _calculateSharesDelta(uint64 newAmountGwei, uint64 previousAmountGwei) internal pure returns (int256) {
+    /// @dev Calculates the delta between two Gwei amounts, converts to Wei, and returns as an int256
```
ChaoticWalrus, line 674: nit: this comment says it converts to wei, but the function does not. (amounts actually remain in gwei / at whatever magnitude they were previously)

src/contracts/pods/EigenPodStorage.sol
```diff
@@ -0,0 +1,54 @@
...
+    uint256 public activeValidatorCount;
+
+    /// @notice The timestamp of the last checkpoint finalized
+    uint64 public lastCheckpointTimestamp;
+
+    /// @notice The timestamp of the currently-active checkpoint. Will be 0 if there is not active checkpoint
+    uint64 public currentCheckpointTimestamp;
+
+    /// @notice The current checkpoint, if there is one active
+    Checkpoint public currentCheckpoint;
```
ChaoticWalrus, line 46: definitely a nit, but we've generally followed the pattern elsewhere of having structs themselves be internal, with public getter functions, so that the function can 'return a struct' instead of returning a tuple

ChaoticWalrus (dismissed):
This is looking 🔥 
Seems like there's some other work to review + potentially merge into this first -- e.g. https://github.com/Layr-Labs/eigenlayer-contracts/pull/524 -- but I'm not seeing any other big reason to block merging this right now.

The biggest open question for me is if there are any bad side effects of changing the oracle semantics from returning (beacon block root of timestamp) to returning (_parent_ beacon block root of timestamp), but it sounds like worst case we could pause proofs for ~24 hours preceding an upgrade?

I'm going to recommend a bit more discussion + review in general before merging, but if we want to merge and do some of that work later I could be convinced here.

Thanks for the work that went into this; really appreciating all the explanations in HackMD, the PR itself, code comments, and in discussion ❤️

ChaoticWalrus:
> Decide if startCheckpoint should revert if there's no pod balance to prove. Chris says this would be helpful for the frontend, but we may want to allow 0-balance checkpoints sometimes? Maybe only for stale balances?

On this front, its a little unclear to me quite what the desired behavior should be, but I _think_ just auto-completing the checkpoint is OK, and would be nice and simple from a code clarity perspective.
If you short-circuit in this case, it seems you would have to send the user their funds (instead of reverting)? Otherwise can't ETH get stuck in an EigenPod if someone sends the pod ETH but never points any validators to it?

src/contracts/pods/EigenPod.sol
gpsanant: yespls!

src/contracts/pods/EigenPod.sol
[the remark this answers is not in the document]
wadealexc: yep, i'm leaning towards keeping this as-is (check for 0x01 creds), because i think we're likely to do a small patch before Pectra anyway to prep for that upgrade.

we can add this at that time.

src/contracts/pods/EigenPod.sol
wadealexc: i hadn't actually considered this griefing vector, that's a great point in favor of skipping!

was more thinking this might accidentally occur, but you make a good point that this could be intentionally abused to make proving a checkpoint awful.

wadealexc: Fleshing out the griefing idea - let's say you have 1000 validators to prove for your checkpoint, and block gas limit says you can prove them in batches of 25 per tx.

Someone can frontrun any one of these batches with a single proof to cause your whole batch to fail, and they can continue to do this until your checkpoint is complete.

wadealexc: Sounds like a nightmare. Let's skip :D 

@gpsanant any input? should we skip + event? or just skip?

src/contracts/pods/EigenPod.sol
wadealexc: [7aa93fb](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/7aa93fb4216af788d48eed2cbc81bfbcc7b42d6e)

src/contracts/pods/EigenPodStorage.sol
wadealexc: I can make this change, np. So weird that the public struct variables don't just return the struct lol

wadealexc:
> This is looking 🔥 Seems like there's some other work to review + potentially merge into this first -- e.g. #524 -- but I'm not seeing any other big reason to block merging this right now.

i think we should keep this as a feature branch for the time being, and merge to dev when we want to start deploying to a testnet/talking about releases?

i gather the vibe was to keep this and slashing on separate branches for the time being to make sure we could launch this without slashing if we wanted to.

ChaoticWalrus:
> I don't think we should auto-complete a checkpoint, because that implies all our validator balances are up to date. We may be talking past each other -- I'm seeing 2 concepts here:
> 
>     1. Short-circuiting and auto-completing a checkpoint when `activeValidatorCount == 0`. This should be fine and is already implemented in this PR. It just means the pod owner has no verified withdrawal creds, so nothing to prove.
> 
>     2. What to do when there is no ETH in the pod that hasn't already been backed by shares (i.e. `podBalanceGwei` is calculated to be 0 on [this line](https://github.com/layr-labs/eigenlayer-contracts/blob/ebb1201e3622953e073b7491de1891ed23343c2f/src/contracts/pods/EigenPod.sol#L574)).
> 
> 
> My idea for this second case is that we could add a `bool` flag to `startCheckpoint` that allows you to tell the contract to revert in case you don't want to prove a "useless" checkpoint. IMO we should keep the ability to prove a "useless" checkpoint for something like `verifyStaleBalance`, for example, which would always want to start a checkpoint, even if it's "useless".

Yes, I was totally misunderstanding, my bad. It seems to like we basically _need_ to keep the ability to prove a checkpoint that would not award any additional shares (the non-reverting behavior), but I think a flag that the caller can set to revert in such a scenario seems like a reasonable feature addition.

src/contracts/pods/EigenPod.sol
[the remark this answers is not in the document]
wadealexc: [eec67ae](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/eec67aeadf696f2058ff85e578531ee0dd0e20d5)

src/contracts/pods/EigenPodStorage.sol
wadealexc: [e7bd4bf](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/e7bd4bf768422c5005567e75a01b7f82d061225c)

src/contracts/pods/EigenPod.sol
wadealexc: [a644d48](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/a644d481384069337c5582a3a779c69e90393eac)

src/contracts/pods/EigenPod.sol
gpsanant: skip is a good idea.

event is also good, but not needed for backend.

wadealexc:
Note from a call I'm having w/ Chris/Scott/others -- just had an idea to remove the `activateRestaking` method entirely in favor of just calling `startCheckpoint`.

Would reduce code flows needed on the frontend.

src/contracts/pods/EigenPod.sol
```diff
@@ -347,18 +288,69 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+     * @param proof the fields of the beacon chain "Validator" container, along with a merkle proof against
+     * the beacon state root. See the consensus specs for more details:
+     * https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#validator
+     *
+     * @dev Staleness conditions:
+     * - Validator's last balance update is older than `beaconTimestamp` by `TIME_TILL_STALE_BALANCE`
+     * - Validator MUST be in `ACTIVE` status in the pod
+     * - Validator MUST be slashed on the beacon chain
+     */
+    function verifyStaleBalance(
```
ChaoticWalrus, line 308: I'm trying to figure out if this is enough conditions. should the caller also prove that the current balance is lower than the stored one?
I am trying to consider the inconvenience to a user when they have many validators pointed to a pod but only one validator is slashed -- it feels like the current logic would essentially force such a user to either exit the slashed validator or do a checkpoint every 2 weeks (as long as someone calls this function)? perhaps that's OK UX tho

ChaoticWalrus (commented):
As a general comment, if we can remove all "M1" functionality + deprecate M1-specific storage (like `lastWithdrawalTimestamp`), that would be huge. Would still require validation of the upgrade path though.

src/contracts/pods/EigenPod.sol
wadealexc: > should the caller also prove that the current balance is lower than the stored one?

well, one difficult thing here is that the method currently has access to effective balance, not current balance. we could add more to the method, but i'm not sure if the complexity is worth it

youfoundron (commented):
addresses the concerns in #556

src/contracts/pods/EigenPod.sol
```diff
@@ -159,143 +113,141 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+                continue;
+            }
+            
+            // Ensure we aren't proving a validator twice for the same checkpoint. This will fail if:
+            // - validator submitted twice during this checkpoint
+            // - validator withdrawal credentials verified after checkpoint starts, then submitted
+            //   as a checkpoint proof
+            if (validatorInfo.mostRecentBalanceUpdateTimestamp >= beaconTimestamp) {
+                continue;
+            }
```
wadealexc, line 209: Bug: this condition prevents the following flow -
1. verifyWC + startCheckpoint in the same block
3. completeCheckpoint

wadealexc: Fixed in [6f9da7f](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/6f9da7f9e7b6358eb2cf5307cf88c2df802aad88)

wadealexc:
Adding a note about the bugfix here: [6f9da7f](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/6f9da7f9e7b6358eb2cf5307cf88c2df802aad88)

I also made an additional change to `startCheckpoint` that prevents multiple checkpoints being started in the same block. Two reasons:

1. It makes no sense to checkpoint twice in a block
2. The second checkpoint wouldn't actually be completeable because all checkpointed validators would have a `lastCheckpointedAt` time equal to the current block. This means `verifyCheckpointProofs` would skip these validators because it would assume they had already been proven

src/contracts/pods/EigenPod.sol
```diff
@@ -441,44 +402,46 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-         * Deserialize the balance field from the Validator struct.  Note that this is the "effective" balance of the validator
-         * rather than the current balance.  Effective balance is generated via a hystersis function such that an effective
-         * balance, always a multiple of 1 ETH, will only lower to the next multiple of 1 ETH if the current balance is less
-         * than 0.25 ETH below their current effective balance.  For example, if the effective balance is 31ETH, it only falls to
-         * 30ETH when the true balance falls below 30.75ETH.  Thus in the worst case, the effective balance is overestimating the
-         * actual validator balance by 0.25 ETH. 
-         */
-        uint64 validatorEffectiveBalanceGwei = validatorFields.getEffectiveBalanceGwei();
+        // Get the validator's effective balance. Note that this method uses effective balance, while
+        // `verifyBalanceUpdates` uses current balance. Effective balance is updated per-epoch - so it's
```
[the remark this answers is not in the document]
wadealexc: [6e1e4bf](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/6e1e4bfb833fb5efcd98bd5927b3621582384941)

src/contracts/pods/EigenPod.sol
```diff
@@ -488,274 +451,182 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+    function _getParentBlockRoot(uint64 timestamp) internal view returns (bytes32) {
+        require(
+            block.timestamp - timestamp < BEACON_ROOTS_HISTORY_BUFFER_LENGTH * 12,
+            "EigenPod._getParentBlockRoot: timestamp out of range"
+        );
+
+        (bool success, bytes memory result) =
+            BEACON_ROOTS_ADDRESS.staticcall(abi.encode(timestamp));
+
+        if (success && result.length > 0) {
```
gpsanant, line 625: require?

src/contracts/pods/EigenPod.sol
```diff
@@ -488,274 +451,182 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        // Note: On finalization, `podBalanceGwei` is added to `withdrawableRestakedExecutionLayerGwei`
+        // to denote that it has been credited with shares. Because this value is denominated in gwei, 
+        // `podBalanceGwei` is also converted to a gwei amount here. This means that any sub-gwei amounts 
+        // sent to the pod are not credited with shares and are therefore not withdrawable. 
+        // This can be addressed by topping up a pod's balance to a value divisible by 1 gwei.
+        uint64 podBalanceGwei = 
+            uint64(address(this).balance / GWEI_TO_WEI) - withdrawableRestakedExecutionLayerGwei;
+
+        // If the caller doesn't want a "0 balance" checkpoint, revert
+        if (revertIfNoBalance && podBalanceGwei == 0) {
```
gpsanant, line 556: require

src/contracts/pods/EigenPod.sol
```diff
@@ -488,274 +451,182 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-        });
-
-        /**
-         * Finally, the validator is fully withdrawn. Update their status and place in state:
-         */
-
-        if (validatorInfo.status != VALIDATOR_STATUS.WITHDRAWN) {
-            activeValidatorCount--;
-            validatorInfo.status = VALIDATOR_STATUS.WITHDRAWN;
+            _currentCheckpoint = checkpoint;
```
gpsanant, line 604: imo take out the conditional and move this case up to `_startCheckpoint`

gpsanant: oops misunderstood

src/contracts/pods/EigenPod.sol
```diff
@@ -488,274 +451,182 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        validatorInfo.lastCheckpointedAt = checkpointTimestamp;
 
-        // Done with proofs! Now update the validator's balance and send to the EigenPodManager if needed
+        // If the validator's new balance is 0, mark them withdrawn
+        if (newBalanceGwei == 0) {
+            activeValidatorCount--;
+            validatorInfo.status = VALIDATOR_STATUS.WITHDRAWN;
+            // If we reach this point, `balanceDeltaGwei` should always be negative,
+            // so this should be a safe conversion
+            exitedBalanceGwei = uint64(uint128(-balanceDeltaGwei));
```
gpsanant, line 508: noting that this may slashed and not actually in the pod

src/contracts/pods/EigenPod.sol
```diff
@@ -488,274 +451,182 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-            validatorFields: validatorFields,
-            validatorFieldsProof: validatorFieldsProof,
-            validatorIndex: validatorIndex
-        });
+        validatorInfo.restakedBalanceGwei = newBalanceGwei;
+        validatorInfo.lastCheckpointedAt = checkpointTimestamp;
 
-        // Done with proofs! Now update the validator's balance and send to the EigenPodManager if needed
+        // If the validator's new balance is 0, mark them withdrawn
+        if (newBalanceGwei == 0) {
```
gpsanant, line 503: they may be slashed to 0, in which case if they make subsequent deposits, those deposits are wrecked since the validator is marked as withdrawn

if validators are getting slashed, there are bigger issues to deal with, so seems fine

noting the edge case though

src/contracts/pods/EigenPod.sol
```diff
@@ -159,143 +113,141 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-                stateRootProof.beaconStateRoot,
-                withdrawalProofs[i],
-                validatorFieldsProofs[i],
-                validatorFields[i],
-                withdrawalFields[i]
-            );
+        // Process each checkpoint proof submitted
+        uint64 exitedBalancesGwei;
+        for (uint256 i = 0; i < proofs.length; i++) {
+            ValidatorInfo memory validatorInfo = _validatorPubkeyHashToInfo[proofs[i].pubkeyHash];
```
gpsanant, line 194: stuffing all of the validator specific checks into `_verifyCheckpointProof` cleaner imo since we write to it there

src/contracts/pods/EigenPod.sol
```diff
@@ -488,274 +451,182 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-            "EigenPod._verifyAndProcessWithdrawal: Validator never proven to have withdrawal credentials pointed to this contract"
+            currentCheckpointTimestamp == 0, 
+            "EigenPod._startCheckpoint: must finish previous checkpoint before starting another"
         );
 
-        // Ensure we don't process the same withdrawal twice
         require(
-            !provenWithdrawal[validatorPubkeyHash][withdrawalTimestamp],
-            "EigenPod._verifyAndProcessWithdrawal: withdrawal has already been proven for this timestamp"
+            lastCheckpointTimestamp != uint64(block.timestamp),
```
gpsanant, line 539: could we add a comment about the case this addresses?

src/contracts/pods/EigenPod.sol
```diff
@@ -488,274 +451,182 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+            int256 totalShareDeltaWei =
+                (int128(uint128(checkpoint.podBalanceGwei)) + checkpoint.balanceDeltasGwei) * int256(GWEI_TO_WEI);
+
+            // Add any native ETH in the pod to `withdrawableRestakedExecutionLayerGwei`
+            // ... this amount can be withdrawn via the `DelegationManager` withdrawal queue
+            withdrawableRestakedExecutionLayerGwei += checkpoint.podBalanceGwei;
+
+            // Finalize the checkpoint
+            lastCheckpointTimestamp = currentCheckpointTimestamp;
+            delete currentCheckpointTimestamp;
```
gpsanant, line 597: `currentCheckpointTimestamp = 0` u freak

src/contracts/pods/EigenPod.sol
```diff
@@ -354,18 +291,68 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+
+        // Validator must be slashed on the beacon chain
+        require(
+            proof.validatorFields.isValidatorSlashed(),
+            "EigenPod.verifyStaleBalance: validator must be slashed to be marked stale"
+        );
+
+        // Verify passed-in `beaconStateRoot` against the beacon block root
+        BeaconChainProofs.verifyStateRoot({
+            beaconBlockRoot: _getParentBlockRoot(beaconTimestamp),
```
gpsanant, line 342: ew so `_getParentBlockRoot` is called twice?

gpsanant: stale path maybe ok

src/contracts/pods/EigenPod.sol
```diff
@@ -354,18 +291,68 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-        emit NonBeaconChainETHWithdrawn(recipient, amountToWithdraw);
-        _sendETH_AsDelayedWithdrawal(recipient, amountToWithdraw);
+
+        // Validator must be checkpoint-able
+        require(
+            validatorInfo.status == VALIDATOR_STATUS.ACTIVE,
+            "EigenPod.verifyStaleBalance: validator is not active"
+        );
+
+        // Validator must be slashed on the beacon chain
```
gpsanant, line 334: in the case where there is a huge liveness failure for no slashing this doesnt work...

src/contracts/pods/EigenPod.sol
wadealexc: WDYM, how is it called twice?

wadealexc: this method only calls it once

src/contracts/pods/EigenPod.sol
wadealexc: yes but it's harder to get the `continue` behavior used in the loop here. any suggestions?

we could return a "skip" boolean but that's kinda messy

src/contracts/pods/EigenPod.sol
wadealexc: > those deposits are wrecked since the validator is marked as withdrawn

nope, those deposits would get picked up in the next checkpoint as `podBalanceGwei` and you'd be able to exit them the same as consensus rewards or other unaccounted-for native ETH

src/contracts/pods/EigenPod.sol
wadealexc: i have a giant disclaimer comment above the `checkpointBalanceExitedGwei` state var - have a look through that and lmk if you think it needs more.

src/contracts/pods/EigenPod.sol
wadealexc: `require(!revertIfNoBalance || podBalanceGwei != 0)`? or something?

idk i think the if statement is wayyy easier to understand

src/contracts/pods/EigenPod.sol
wadealexc: PEPE is not resilient against the beacon chain itself failing, and I think that's a fair compromise. If the beacon chain itself is failing I'm not really even sure how the execution layer copes with that, so... not sure what I'm supposed to do about that.

Ultimately the goal is to do better than this stale balance function post-Pectra, but for now I think this is good enough.

wadealexc: I am open to other ideas for `verifyStaleBalance` if you have them, but I don't want to complicate things too much.

src/contracts/pods/EigenPod.sol
wadealexc: [4a923d0](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/4a923d09fe9dbbf4c175502d410feef5d07e5b8d)

src/contracts/pods/EigenPod.sol
wadealexc: [4a923d0](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/4a923d09fe9dbbf4c175502d410feef5d07e5b8d)

src/contracts/pods/EigenPod.sol
wadealexc: [49769d4](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/49769d436747913271e4f41d78d1f50ce9e398be)

I made this a little better by moving the actual state write into the parent method, making it more clear that there's a state update happening. I played with moving some of the checks/updates around and anything else felt messy.

Sidenote, for whatever reason the change in this commit saves 131 gas per validator :shrug:

src/contracts/pods/EigenPod.sol
gpsanant: inside of startCheckpoint, but i realized they're for different timestamps

src/contracts/pods/EigenPod.sol
gpsanant: lgtm

src/contracts/pods/EigenPod.sol
gpsanant: ah yep u right

wadealexc:
Created a test to specifically check skipped slot behavior here: https://gist.github.com/wadealexc/5bdaf0eb1bac3373b9d408992bc91fd7

Wasn't sure where to include it in our tests since it requires compiling with a different evm version than we do by default. (speaking of, should we change that?)

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -119,6 +104,93 @@ interface IEigenPod {
...
+    /**
+     * @dev Verify one or more validators have their withdrawal credentials pointed at this EigenPod, and award
+     * shares based on their effective balance. Proven validators are marked `ACTIVE` within the EigenPod, and
+     * future checkpoint proofs will need to include them.
+     * @dev Withdrawal credential proofs MUST NOT be older than the `lastCheckpointTimestamp` OR `currentCheckpointTimestamp`.
+     * @dev Validators proven via this method MUST NOT have an exit epoch set already.
+     * @param beaconTimestamp the beacon chain timestamp sent to the 4788 oracle contract. Corresponds
+     * to the parent beacon block root against which the proof is verified.
+     * @param stateRootProof proves a beacon state root against a beacon block root
+     * @param validatorIndices a list of validator indices being proven stale
```
ChaoticWalrus, line 146: nit: `proven stale` => just `proven`

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -119,6 +104,93 @@ interface IEigenPod {
...
+    function startCheckpoint(bool revertIfNoBalance) external;
+
+    /**
+     * @dev Progress the current checkpoint towards completion by submitting one or more validator
+     * checkpoint proofs. Anyone can call this method to submit proofs towards the current checkpoint.
+     * For each validator proven, the current checkpoint's `proofsRemaining` decreases.
+     * @dev If the checkpoint's `proofsRemaining` reaches 0, the checkpoint is finalized.
+     * (see `_updateCheckpoint` for more details)
+     * @dev This method can only be called when there is a currently-active checkpoint.
+     * @param balanceContainerProof proves the beacon's current balance container root against a checkpoint's `beaconBlockRoot`
```
ChaoticWalrus, line 128: `beacon's current balance container` => `beacon _chain_'s balance container`

ChaoticWalrus: actually I think this is fine

src/contracts/pods/EigenPod.sol
```diff
@@ -159,143 +113,144 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        onlyWhenNotPaused(PAUSED_START_CHECKPOINT) 
+    {
+        _startCheckpoint(revertIfNoBalance);
+
+        /// Legacy support for pods deployed pre-M2 that never activated restaking. `startCheckpoint`
+        /// can activate restaking, allowing them to continue using their pods via the checkpoint
+        /// system.
+        if (!hasRestaked) {
+            hasRestaked = true;
+            emit RestakingActivated(podOwner);
```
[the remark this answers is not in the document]
wadealexc: good q. i think no, because `verifyStaleBalance` requires you to have a validator in the `ACTIVE` state -- which should be impossible unless `hasRestaked == true`.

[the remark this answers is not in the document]
wadealexc: (In both M2 and PEPE, `verifyWithdrawalCredentials` requires `hasRestaked == true`)

src/contracts/pods/EigenPod.sol
wadealexc: resolving because i have since implemented this

src/contracts/pods/EigenPod.sol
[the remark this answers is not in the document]
ChaoticWalrus: ahh good point! it wouldn't even be possible to trigger `verifyStaleBalance` in this state, so the behavior doesn't matter. that's great!

src/contracts/pods/EigenPod.sol
```diff
@@ -354,18 +291,68 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+        bytes32 validatorPubkey = proof.validatorFields.getPubkeyHash();
+        ValidatorInfo memory validatorInfo = _validatorPubkeyHashToInfo[validatorPubkey];
+
+        // Validator must be eligible for a staleness proof
         require(
-            amountToWithdraw <= nonBeaconChainETHBalanceWei,
-            "EigenPod.withdrawnonBeaconChainETHBalanceWei: amountToWithdraw is greater than nonBeaconChainETHBalanceWei"
+            beaconTimestamp > validatorInfo.lastCheckpointedAt + TIME_TILL_STALE_BALANCE,
+            "EigenPod.verifyStaleBalance: validator balance is not stale yet"
         );
```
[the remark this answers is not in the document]
wadealexc: have revisited, going to remove the timing window entirely

src/contracts/interfaces/IEigenPod.sol
wadealexc: [b707336](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/b7073366427da7b01315a3aeefe54f55fe3ff8ea)

wadealexc:
Feature request - change `EPM.PodSharesUpdated` event to emit a pod owner's total shares

wadealexc:
Feature request - allow the pod owner to approve a secondary address to do certain things, like call `verifyWithdrawalCredentials` and `startCheckpoint`. This would allow pod owners to create a hot wallet they use for proofs, without requiring their cold wallet to sign transactions as frequently.

IMO this is important because the `podOwner` address is address that "owns" all your funds! It shouldn't be used for day-to-day management.

src/contracts/pods/EigenPod.sol
```diff
@@ -383,32 +392,18 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
-        emit RestakingActivated(podOwner);
-    }
-
-    /// @notice Called by the pod owner to withdraw the balance of the pod when `hasRestaked` is set to false
-    function withdrawBeforeRestaking() external onlyEigenPodOwner hasNeverRestaked {
-        _processWithdrawalBeforeRestaking(podOwner);
+    /// @notice Allows the owner of a pod to update the proof submitter, a permissioned
+    /// address that can call `startCheckpoint` and `verifyWithdrawalCredentials`.
+    /// @dev Note that EITHER the podOwner OR proofSubmitter can access these methods,
+    /// so it's fine to set your proofSubmitter to 0 if you want the podOwner to be the
```
bxmmm1, line 398:
```suggestion
    /// so it's fine to set your proofSubmitter to address(0) if you want the podOwner to be the
```

src/contracts/pods/EigenPod.sol
```diff
@@ -383,32 +392,18 @@ contract EigenPod is IEigenPod, Initializable, ReentrancyGuardUpgradeable, Eigen
...
+    /// only address that can call these methods.
+    /// @param newProofSubmitter The new proof submitter address. If set to 0, only the
```
bxmmm1, line 400:
```suggestion
    /// @param newProofSubmitter The new proof submitter address. If set to address(0), only the
```

PR #522, opened 2024-04-18
feat: StrategyFactory contract

ChaoticWalrus:
Proposal Doc: https://docs.google.com/document/d/1C1nN6BkCMS0Z9EnizU0pNYNNOiRd_FH6E2Vd-ZPULro/edit?usp=sharing

## High-Level Notes on Design
This avoid making any changes to existing "core" protocol contracts, and is instead purely additive.
In particular, this change is effectively turning the `strategyWhitelister` role in the StrategyManager into a smart contract -- the ability to whitelist particular strategies is maintained with a pass-through function, while the ability to whitelist factory-generated strategy contracts is made permissionless.

## Summary
The StrategyFactory allows any caller to deploy a [StrategyBase contract](https://github.com/Layr-Labs/eigenlayer-contracts/blob/dev/src/contracts/strategies/StrategyBase.sol) for any ERC20 token
- Technically, a BeaconProxy contract is deployed, pointing to a reference implementation
- Upgrades are controlled by a new BeaconProxy, with the same control of upgrades as our existing contracts
- Pausing power is also given to the same PauserRegistry as we use now
- The token cannot already have a Strategy deployed for it (only one deployment from the factory per token)
- We store a mapping(IERC20 token => IStrategy) public tokenStrategies and check against previous writes
- Owner can pass-through this contract to whitelist new strategy contracts
- The StrategyFactory contract is itself upgradeable, to potentially allow other feature additions in the future

Still needs a few things, including documentation that provides warnings against using reentrant tokens (e.g. [ERC777](https://eips.ethereum.org/EIPS/eip-777)), or other non-standard tokens, and more extensive testing

## Deployment
- Deploy Upgradeable Beacon for Beacon Proxies, with ownership given to the executor multisig (so upgrade rights work like our existing contracts)
- Deploy StrategyFactory implementation, pointing to Strategy Manager (immutable variable)
- Deploy StrategyFactory (proxy), giving ownership to Ops Multisig (so it maintains similar manual-whitelisting powers), pausing powers to the existing protocol-wide PausingRegistry, and proxy upgrade powers to the existing protocol-wide ProxyAdmin (+ pointing to the Upgradeable Beacon & StrategyFactory implementation deployed above)
- Transfer `strategyWhitelister` role in the StrategyManager to the StrategyFactory (proxy) -- **this will require an executor multisig action and thus be subject to a 10-day timelock delay if accomplished via the ops multisig**


src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,108 @@
...
+    ) external onlyOwner {
+        strategyManager.addStrategiesToDepositWhitelist(strategiesToWhitelist, thirdPartyTransfersForbiddenValues);
+        for (uint256 i = 0; i < strategiesToWhitelist.length; ++i) {
+            IERC20 underlyingToken = strategiesToWhitelist[i].underlyingToken();
+            _setStrategyForToken(underlyingToken, strategiesToWhitelist[i]);
+        }
+    }
+
+    // @notice Owner-only function to add (existing) Strategy contracts to the `tokenStrategies` mapping
+    function editTokenStrategiesMapping(
```
gpsanant, line 86: tbh just put in initializer?

src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,108 @@
...
+            )
+        );
+        _setStrategyForToken(token, strategy);
+    }
+
+    /** 
+     * @notice Owner-only function to pass through a call to `StrategyManager.addStrategiesToDepositWhitelist`
+     * @dev Also adds the `strategiesToWhitelist` to the `tokenStrategies` mapping
+     */
+    function whitelistStrategies(
```
gpsanant, line 74: shouldn't we just accept an array of tokens here instead of strategies and the strategy is looked up in the `tokenStrategies` mapping?

ChaoticWalrus: what? super confused on what you think the intended usage is

src/contracts/strategies/StrategyFactory.sol
ChaoticWalrus: it could also be in the initializer, but I think this is desirable regardless

src/contracts/strategies/StrategyFactory.sol
ChaoticWalrus: looks like I had missed perhaps the most important function call to have in the `deployNewStrategy` function. added here; please resolve this thread if this makes sense now.
https://github.com/Layr-Labs/eigenlayer-contracts/pull/522/commits/0e572603d34df67a65b32c8a31f4d59f6774fe5b

src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,113 @@
...
+ * @notice TODO: write this
+ */
+contract StrategyFactory is OwnableUpgradeable, Pausable {
+
+    uint8 internal constant PAUSED_NEW_STRATEGIES = 0;
+
+    /// @notice EigenLayer's StrategyManager contract
+    IStrategyManager public immutable strategyManager;
+
+    StrategyBase public immutable strategyImplementation;
```
ChaoticWalrus, line 22: should perhaps make this modifiable? in case an upgrade is initiated for all existing Strategy contracts

src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,113 @@
...
+        tokenStrategies[token] = strategy;
+        emit StrategySetForToken(token, strategy);
+    }
+
+    /**
+     * @dev This empty reserved space is put in place to allow future versions to add new
+     * variables without shifting down storage in the inheritance chain.
+     * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+     */
+    uint256[49] private __gap;
```
ChaoticWalrus, line 112: TODO: set the correct size for this

src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,140 @@
...
+
+    /// @notice EigenLayer's StrategyManager contract
+    IStrategyManager public immutable strategyManager;
+
+    StrategyBase public strategyImplementation;
+
+    ProxyAdmin public eigenLayerProxyAdmin;
+
+    // @notice Mapping token => strategy contract for the token
+    mapping(IERC20 => IStrategy) public tokenStrategies;
```
wadealexc, line 28: Do we want to do a storage contract like we do for everything else?

wadealexc (changes requested):
Overall looks fine, but IMO we want the BeaconProxy pattern!

ChaoticWalrus:
> Overall looks fine, but IMO we want the BeaconProxy pattern!

switch to beacon proxies implemented in https://github.com/Layr-Labs/eigenlayer-contracts/pull/522/commits/5d19597b8dc0e80ccf58319c7fa9728c192dc57b

src/contracts/strategies/StrategyFactory.sol
ChaoticWalrus: seems like a reasonable plan; I'll get that done shortly

src/contracts/interfaces/IStrategyFactory.sol
```diff
@@ -0,0 +1,53 @@
...
+    ) external;
+
+    // @notice Owner-only function to add (existing) Strategy contracts to the `tokenStrategy` mapping
+    function editTokenStrategiesMapping(
+        IERC20[] calldata tokens,
+        IStrategy[] calldata strategies
+    ) external;
+
+    // @notice Owner-only function to modify the `strategyBeacon`
+    function setStrategyBeacon(IBeacon _strategyBeacon) external;
```
ChaoticWalrus, line 46: open question: do we need this function (and associated event) at all?

src/contracts/strategies/StrategyFactory.sol
ChaoticWalrus: completed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/522/commits/2fa5492c0be328630bd9815226c734e092ee8d5e which also created an interface file

src/contracts/interfaces/IStrategyFactory.sol
```diff
@@ -0,0 +1,53 @@
...
+    function deployNewStrategy(IERC20 token) external returns (IStrategy newStrategy);
+
+    /** 
+     * @notice Owner-only function to pass through a call to `StrategyManager.addStrategiesToDepositWhitelist`
+     * @dev Also adds the `strategiesToWhitelist` to the `tokenStrategy` mapping
+     */
+    function whitelistStrategies(
+        IStrategy[] calldata strategiesToWhitelist,
+        bool[] calldata thirdPartyTransfersForbiddenValues
+    ) external;
```
scotthconner, line 37: Can you help me understand the use case for exposing this?

src/contracts/interfaces/IStrategyFactory.sol
scotthconner: setStrategyBeacon seems fine, the editTokenStrategiesMapping() is gunna blow up any audit because we can't guarantee the state at any given point. Even in the best case this seems like a huge footgun.

src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,108 @@
...
+                address(strategyBeacon),
+                abi.encodeWithSelector(StrategyBase.initialize.selector, token, pauserRegistry)
+            )
+        ));
+        _setStrategyForToken(token, strategy);
+        IStrategy[] memory strategiesToWhitelist = new IStrategy[](1);
+        bool[] memory thirdPartyTransfersForbiddenValues = new bool[](1);
+        strategiesToWhitelist[0] = strategy;
+        thirdPartyTransfersForbiddenValues[0] = false;
+        strategyManager.addStrategiesToDepositWhitelist(strategiesToWhitelist, thirdPartyTransfersForbiddenValues);
```
scotthconner, line 63: I feel like if the strategyManager takes a dependency on this contract and reads the mapping, we can avoid all of this state copying.

src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,108 @@
...
+        return strategy;
+    }
+
+    /** 
+     * @notice Owner-only function to pass through a call to `StrategyManager.addStrategiesToDepositWhitelist`
+     * @dev Also adds the `strategiesToWhitelist` to the `tokenStrategy` mapping
+     */
+    function whitelistStrategies(
```
scotthconner, line 71: Seems unnecessary.

src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,108 @@
...
+    ) external onlyOwner {
+        strategyManager.addStrategiesToDepositWhitelist(strategiesToWhitelist, thirdPartyTransfersForbiddenValues);
+        for (uint256 i = 0; i < strategiesToWhitelist.length; ++i) {
+            IERC20 underlyingToken = strategiesToWhitelist[i].underlyingToken();
+            _setStrategyForToken(underlyingToken, strategiesToWhitelist[i]);
+        }
+    }
+
+    // @notice Owner-only function to add (existing) Strategy contracts to the `tokenStrategy` mapping
+    function editTokenStrategiesMapping(
```
scotthconner, line 83: This seems danger - we want to be able to guarantee what strategies are added to that mapping, their implementations, their proper beacon proxies, etc.

scotthconner (changes requested):
These comments here mostly reflect what I put in the technical proposal. Nothing super critical,  but understanding the core contract interactions and where state is managed is one piece that sticks out to me.

src/contracts/interfaces/IStrategyFactory.sol
ChaoticWalrus: using `setStrategyBeacon` could/would result in a partition of the factory-deployed contracts into different sets that get upgraded via calling different beacons. I asked this because I'm just really not seeing a sensible use for it; it made more sense with the TUP pattern I had before.

I'm totally fine with removing `editTokenStrategiesMapping` -- I agree the contract would have easier-to-understand, better formally-defined behavior without it. It's essentially a governance-controlled "make an irregular state change" button, with limited practical use (but we can discuss more what I had in mind?), so I'm fine just getting rid of it

src/contracts/interfaces/IStrategyFactory.sol
ChaoticWalrus: given how the current whitelister role works (unique power in StrategyManager), the simplest way to make this idea work was to plan to transfer the role to the Factory contract but leave this method as a passthrough.

this would be useful in case there is ever any desire to (manually) add any "non-standard" strategies -- e.g. like the existing EIGEN strategy.

src/contracts/strategies/StrategyFactory.sol
ChaoticWalrus: it's one-time copying vs. always querying the factory, which amortizes quite quickly to be more efficient.
I don't care too strongly, just felt that not making any changes to the existing core contracts would be simpler to review + faster to execute

src/contracts/strategies/StrategyFactory.sol
ChaoticWalrus: that's fair; like many owner-only functions, this could be misused.
the severity of the misuse depends on what value you place on the storage -- I was viewing it primarily as a deployment-duplication-prevention mechanism, but it could potentially serve more of a purpose

ChaoticWalrus: function removed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/522/commits/b88cdb5617d8ba51b15b6472f655d8043d1929ee

src/contracts/interfaces/IStrategyFactory.sol
ChaoticWalrus: I decided to remove this function in https://github.com/Layr-Labs/eigenlayer-contracts/pull/522/commits/b88cdb5617d8ba51b15b6472f655d8043d1929ee
I'm not seeing a practical use for it

src/contracts/strategies/StrategyFactory.sol
ChaoticWalrus: function removed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/522/commits/b88cdb5617d8ba51b15b6472f655d8043d1929ee so this discussion is no longer relevant

src/contracts/strategies/StrategyFactory.sol
ChaoticWalrus: agreed offline on the approach taken here being OK; this minimizes core contract changes (no changes to StrategyManager), instead opting to turn an existing role (the `strategyWhitelister`) into a smart contract.

src/contracts/strategies/StrategyFactory.sol
ChaoticWalrus: agreed offline that some passthrough seems prudent; this function is now a pure passthrough (doesn't edit any storage in the StrategyFactory) as of https://github.com/Layr-Labs/eigenlayer-contracts/pull/522/commits/b88cdb5617d8ba51b15b6472f655d8043d1929ee

src/contracts/interfaces/IStrategyFactory.sol
ChaoticWalrus: agreed offline on this passthrough being prudent.

src/contracts/strategies/StrategyFactoryStorage.sol
```diff
@@ -0,0 +1,24 @@
...
+import "../interfaces/IStrategyFactory.sol";
+
+/**
+ * @title Storage for the StrategyFactory contract.
+ * @author Layr Labs, Inc.
+ * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
+ */
+abstract contract StrategyFactoryStorage is IStrategyFactory {
+    // @notice Upgradeable beacon which new Strategies deployed by this contract point to
+    IBeacon public strategyBeacon;
```
wadealexc, line 13: Should add `immutable`?

src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,92 @@
...
+ * @dev This may not be compatible with non-standard ERC20 tokens. Caution is warranted.
+ */
+contract StrategyFactory is StrategyFactoryStorage, OwnableUpgradeable, Pausable {
+    uint8 internal constant PAUSED_NEW_STRATEGIES = 0;
+
+    /// @notice EigenLayer's StrategyManager contract
+    IStrategyManager public immutable strategyManager;
+
+    /// @notice Since this contract is designed to be initializable, the constructor simply sets the immutable variables.
+    constructor(IStrategyManager _strategyManager) {
```
ypatil12, line 24: We should have a way to prevent strategy deployments of LST tokens that already exist. Recommend adding this in initializer

ypatil12 (changes requested):

src/contracts/strategies/StrategyFactory.sol
ChaoticWalrus: I had something like this before and ended up deleting it, after Scott pointed out that -- at least with the functions as I had laid them out -- admin actions would have made it possible to violate the "no duplicate strategies" property.

IMO the storage in this contract is currently primarily serving preventing duplicates as well as providing a nice little `token => strategy` connection, which we currently lack elsewhere (although I agree this excludes existing Strategies).

Existing strategies may vary (like the EIGEN) strategy and require a different upgrade process (due to being TUPs instead of BeaconProxies), so it's debatable whether they should be in the same mapping; we could always maintain two mappings though.

I've also thought about maintaining a _list_ of all deployed strategies from this contract, similar to how the Uniswap Factory (V2, at least) keeps a list of all deployed pairs, but we should always be able to index this list based on events, so it's of debatable value.

I think this is the area where we have the least precision on what we want + could make changes though! Perhaps @0xClandestine can lead figuring out the best path forwards here.

src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,105 @@
...
+        thirdPartyTransfersForbiddenValues[0] = false;
+        strategyManager.addStrategiesToDepositWhitelist(strategiesToWhitelist, thirdPartyTransfersForbiddenValues);
+        return strategy;
+    }
+
+    /**
+     * @notice Owner-only function to prevent strategies from being created for given tokens.
+     * @param tokens An array of token addresses to blacklist.
+     */
+    function blacklistTokens(IERC20[] calldata tokens) external onlyOwner {
```
ypatil12, line 77: Is the plan to blacklist already deployed tokens via the initializer? We don't want to get frontrun on a permissionless deployment

ypatil12 (changes requested):
One comment

src/contracts/strategies/StrategyFactory.sol
ypatil12: Resolved offline -> the deployment requires us to wait to transfer the `strategyWhitelister` role so we won't need to add this functionality in the initializer

src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,105 @@
...
+            "StrategyFactory.deployNewStrategy: Strategy already exists for token"
+        );
+        IStrategy strategy = IStrategy(
+            address(
+                new BeaconProxy(
+                    address(strategyBeacon),
+                    abi.encodeWithSelector(StrategyBase.initialize.selector, token, pauserRegistry)
+                )
+            )
+        );
```
wadealexc, line 63: Should we be using Create2 like we do in the EigenPodManager?

tlip: As an outsider, I definitely +1 this

src/contracts/strategies/StrategyFactory.sol
```diff
@@ -0,0 +1,119 @@
...
+        _setStrategyBeacon(_strategyBeacon);
+    }
+
+    /**
+     * @notice Deploy a new strategyBeacon contract for the ERC20 token.
+     * @dev A strategy contract must not yet exist for the token.
+     * $dev Immense caution is warranted for non-standard ERC20 tokens, particularly "reentrant" tokens
+     * like those that conform to ERC777.
+     */
+    function deployNewStrategy(IERC20 token)
```
mpjunior92, line 46: I could not find an event emitted when a new strategy gets deployed. EigenPodManager, for example, emits a `PodDeployed` event whenever an EigenPod is created, which can be used in [subgraph templates](https://thegraph.com/docs/en/developing/creating-a-subgraph/#data-source-templates) to index instantiated contracts. Is it possible to do the same here?

ypatil12: We can use the `StrategyAddedToDepositWhitelist` event in the StrategyManager instead? Or do you want to track the count of contracts deployed from this contract?

mpjunior92: `StrategyAddedToDepositWhitelist` tracks only EigenLayer approved strategies that are shown on the UI, right?

ypatil12: When we deploy a strategy through this method, it is added to the deposit whitelist. So it would be both EigenLayer approved AND permissionless strategies

wadealexc (approved):

wadealexc:
All the activity you're seeing here is:
- squashing the PR into a single commit to make rebasing easier
- rebasing on top of dev to make merging clean

PR #528, opened 2024-04-26
POC: slashing accounting

ChaoticWalrus:
Complete WIP. Created for illustrative purposes; there are likely bugs and this is quite incomplete and the time of opening this PR.

Adding my comments from elsewhere:
I believe that these are the necessary kinds of calculations that need to happen, but they can be redistributed / flipped in various ways.
In writing this POC, I kept feeling like there’s a simpler way to do this but have convinced myself several times that there is not.

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -86,14 +86,14 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
 
     function operatorShares(address operator, IStrategy strategy) public view returns (uint256) {
         uint256 scalingFactor = currentShareScalingFactor(operator, strategy);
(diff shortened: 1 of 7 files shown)


ChaoticWalrus:
Potentially fraught interaction with deposits:
if deposits are instantaneous (+ instantaneously credited) then users may be subject to slashing for Slashing Requests that occurred prior to their deposit

ChaoticWalrus:
Update: **this model is currently super broken**. I need to move some of the storage into a staker-specific struct (presumably within the StrategyManager) in order to ensure that new stakers are not penalized for past slashing events / don't end up socializing past losses.

ChaoticWalrus:
Closing in favor of a refactoring in https://github.com/Layr-Labs/eigenlayer-contracts/pull/543

PR #533, opened 2024-05-05
feat: add payments off-chain assumptions

ypatil12:
Add off-chain assumptions from payments calculation

- `SNAPSHOT_CADENCE`: Requirement for `CALCULATION_INTERVAL_SECONDS ` to be a multiple of this value
- `amount <= MAX_PAYMENT_AMOUNT`. This is 1e38-1, the max possible value in iceberg to do math without a loss of precision

### DIFF
--- a/docs/core/PaymentCoordinator.md
+++ b/docs/core/PaymentCoordinator.md
@@ -28,7 +28,7 @@ This document is organized according to the following themes (click each to be t
 * [Distributing and Claiming Payments](#distributing-and-claiming-payments)
 * [System Configuration](#system-configuration)
 * [Payments Merkle Tree Structure](#payments-merkle-tree-structure)
-* [Off-Chain Calculation](#offchain-calculation)
+* [Off Chain Calculation](#off-chain-calculation)
 
 #### Important state variables
 
@@ -356,7 +356,7 @@ The payment merkle tree is structured in the diagram below:
 
 ![.](../images/PaymentCoordinator_Merkle_Tree.png)
 
-### Off-Chain Integrations
+### Off Chain Calcalculation
 
 Payments are calculated via an off-chain data pipeline. The pipeline takes snapshots of core contract state at the `SNAPSHOT_CADENCE`, currently said to once per day. It then combines these snapshots with any active payments to calculate what the single day payout of an earner is. Every `CALCULATION_INTERVAL_SECONDS` payouts are accumulated up to `lastPaymentTimestamp + CALCULATION_INTERVAL_SECONDS` and posted on-chain by the entity with the `paymentUpdater` role. 
 
--- a/docs/core/PaymentCoordinator.md
+++ b/docs/core/PaymentCoordinator.md
@@ -356,7 +356,9 @@ The payment merkle tree is structured in the diagram below:
 
 ![.](../images/PaymentCoordinator_Merkle_Tree.png)
 
-### Off Chain Calcalculation
+---
+
+### Off Chain Calculation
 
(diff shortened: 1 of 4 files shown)


docs/core/PaymentCoordinator.md
```diff
@@ -28,6 +28,7 @@ This document is organized according to the following themes (click each to be t
 * [Distributing and Claiming Payments](#distributing-and-claiming-payments)
 * [System Configuration](#system-configuration)
 * [Payments Merkle Tree Structure](#payments-merkle-tree-structure)
+* [Off-Chain Calculation](#offchain-calculation)
```
8sunyuan, line 31: link to the header doesn't work, try doing `#off-chain-integrations`

src/contracts/core/PaymentCoordinatorStorage.sol
```diff
@@ -30,6 +30,10 @@ abstract contract PaymentCoordinatorStorage is IPaymentCoordinator {
     uint32 public immutable MAX_FUTURE_LENGTH;
     /// @notice absolute min timestamp (seconds) that a payment can start at
     uint32 public immutable GENESIS_PAYMENT_TIMESTAMP;
+    /// @notice The cadence at which a snapshot is taken offchain for calculating payment distributions
+    uint32 public constant SNAPSHOT_CADENCE = 1 days;
```
8sunyuan, line 34: Do we need the cadence to be stored in the contract? It's not being used in the contract and I'm not sure if we ever need to RPC call this offchain.

ypatil12: We don't use RPC to call this off chain, but it should be the case that `CALCULATION_INTERVAL_SECONDS` is a multiple of this, which is why I added it

src/contracts/core/PaymentCoordinatorStorage.sol
```diff
@@ -30,6 +30,10 @@ abstract contract PaymentCoordinatorStorage is IPaymentCoordinator {
     uint32 public immutable MAX_FUTURE_LENGTH;
     /// @notice absolute min timestamp (seconds) that a payment can start at
     uint32 public immutable GENESIS_PAYMENT_TIMESTAMP;
+    /// @notice The cadence at which a snapshot is taken offchain for calculating payment distributions
+    uint32 public constant SNAPSHOT_CADENCE = 1 days;   
+    /// @notice The maximum payment token amount for a single range payment, constrained by off-chain calculation
+    uint256 public constant MAX_PAYMENT_AMOUNT = 1e38 - 1;
```
8sunyuan, line 36: For clarity and consistency with other constants, should be added to the interface as well

docs/core/PaymentCoordinator.md
```diff
@@ -352,4 +354,12 @@ Claimers can selectively choose which token leaves to prove against and claim ac
...
 
-![.](../images/PaymentCoordinator_Merkle_Tree.png)
\ No newline at end of file
+![.](../images/PaymentCoordinator_Merkle_Tree.png)
+
+---
+
+### Off Chain Calculation
+
+Payments are calculated via an off-chain data pipeline. The pipeline takes snapshots of core contract state at the `SNAPSHOT_CADENCE`, currently said to once per day. It then combines these snapshots with any active payments to calculate what the single day payout of an earner is. Every `CALCULATION_INTERVAL_SECONDS` payouts are accumulated up to `lastPaymentTimestamp + CALCULATION_INTERVAL_SECONDS` and posted on-chain by the entity with the `paymentUpdater` role.
```
wadealexc, line 363: currently "set" to once per day

docs/core/PaymentCoordinator.md
```diff
@@ -105,6 +106,7 @@ The payment distribution amongst the AVS's Operators and delegated Stakers is de
     * Requirements from calling internal function `_payForRange()`
         * `rangePayment.strategiesAndMultipliers.length > 0`
         * `rangePayment.amount > 0`
+        * `rangePayment.amount < MAX_PAYMENT_AMOUNT`
```
wadealexc, line 109: the code uses `<=`

wadealexc (approved):
LGTM

PR #534, opened 2024-05-06
chore: cleanup `make deps` to be OS aware

seanmcgary:
Add support for setting things up on MacOS and Linux (ubuntu/debian).


wadealexc (commented):
One thing I'm noticing - we have both a `/script` and `/scripts` folder. We should probably rename one, or collapse the contents into a single folder; 2 is confusing.

wadealexc:
Another thing -- my precommit hook gave me an error:

```
./scripts/compile-bindings.sh
Fatal: Failed to generate ABI binding: 70:22: mixed named and unnamed parameters (and 10 more errors)
```

... and it spat out a ton of generated bindings. Can we quiet that down, or is it necessary?

wadealexc (changes requested):
See comments above

PR #537, opened 2024-05-06
test: tests for single earner leaf or single token leaf

8sunyuan:
added unit tests for when there is a single earner leaf or a single token leaf in an earner's subtree. This results in the proof length being 0 and the hash of the leaf being the root itself.


wadealexc (approved):
awesome <3

PR #543, opened 2024-05-07
POC: slashing accounting 2 - electric boogaloo

ChaoticWalrus:
This is initially a refactor of https://github.com/Layr-Labs/eigenlayer-contracts/pull/528 (closed in favor of this) and definitely a POC right now.
I plan on adding more features, cleaning it up a bit, adding more documentation and explanation, etc.

Edit:
The most core notion in this PR is a 'rebasing-like' mechanism for enabling slashing accounting. 
A new mapping` operator => strategy => share scalingFactor` provides a factor for converting between "non-normalized" Strategy shares and our existing concept of Strategy shares.
Existing storage is converted to non-normalized shares, and (normalized) shares are queried via view functions, providing backwards-compatibility of the interface (although it may break implicit assumptions made by integrations about share amounts changing).
A new library (https://github.com/Layr-Labs/eigenlayer-contracts/blob/poc-slashing-accounting-2-electric-boogaloo/src/contracts/libraries/SlashingAccountingUtils.sol) provides much of the associated functionality which is used across the Slasher, DelegationManager, StrategyManager, and EigenPodManager contracts.
Slashing is accomplished via a `request => window in which cancellation can take place => execution` lifecycle. Slashing must be executed in epoch order (for a given Strategy, operator tuple), and can only be executed once per (operator, Strategy, epoch) tuple.

A major TODO is figuring out how to deal with “100% slashing” and “maximum cumulative slashing” in an elegant way. The current approach feels a little "ham-fisted" -- slashing is capped at 99.99% per epoch, and once a max cumulative slashing factor is reached, share amounts are effectively zeroed out. If we want to keep something similar, we may wish to limit further delegations of shares when an operator is close to the max factor, or similar, to ensure that what would otherwise be 30% slashing events don't wipe out all user shares by crossing this threshold.

We should discuss how to transition existing events + how best to structure new events. One particular callout here is that at the time of writing (May 17, 2024), many existing events now use non-normalized share amounts instead (normalized) share amounts.

A big risk in this model is likely various types of rounding errors.
This should be a point of interrogation, with time dedicated to theoretical reasoning + testing of this in particular. I _suspect_ that  the property that `sumOfStakerShares == delegatedShares` is preserved, but this may currently be violated due to rounding.
Note as well that this model already uses "scaling up" to non-normalized share amounts, which makes it so precision is gained with slashing. Reversing the sense of this scaling would cause precision to be lost with slashing, which could have serious adverse consequences.
In general, scaling is implemented as:

- scale up from "normalized" shares to non-normalized amounts when passing from Strategies or EigenPods to StrategyManager or EigenPodManager
- scale down from non-normalized shares to normalized amounts when passing from StrategyManager or EigenPodManager to Strategies or EigenPods
- pass non-normalized amounts around in "internal" calls between other EigenLayer protocol contracts

There is one notable weird case with scaling when completing a withdrawal "as shares" of beacon chain ETH, in which we scale down, scale up, call DM=>EPM, and ultimately scale down once more. This can presumably be improved.

This PR takes inspiration from brief notes https://docs.google.com/document/d/1lQBAxBAoQ8Ri29rQqOZQDYvwIeWcii-PU-AGMsSvpmE/edit?usp=sharing and a doc on proposed interfaces https://docs.google.com/document/d/1cgV8G4fn82VtqgdnKPmSu6P6hsCWGdW1okTP5M_7GHI/edit?usp=sharing
I have chosen to leave off some of the "minutiae" for the moment, for the sake of brevity and clarity.

Functionality for editing slashable amounts / rates and optimized data structures are notably missing from this PR, despite being planned.

This PR also ignores the concept of Operator Sets for the moment; this may be added later once greater agreement on this concept has been reached.

~~This PR currently lacks historical 'shareFactors' which are important for not "over-slashing" queued withdrawals.~~ 
edit: these have been added

~~I am also considering adding view functions for share amounts that include *pending* slashing amounts, but these may rely on other as-of-yet undecided functionality.~~
edit: POCs for these have been added as well

Potentially fraught interaction with deposits:
If deposits are instantaneous (+ instantaneously credited) then users may be subject to slashing for Slashing Requests that occurred prior to their deposit. This downside is notably present in this preliminary work.

Adding a note:
I think there might be some weird edge case interactions between this accounting model and share debt in EigenPods. At some point I convinced myself there wouldn't be any bug here, but it's at least an odd spot worthy of some investigation + testing.


ChaoticWalrus:
update: I've realized the current implementation -- where "rebased" or "scaled up" shares are stored in the withdrawal struct itself -- is broken when completing a withdrawal as shares and the staker has switched delegated operators in the interim. will likely need to add converting back-and-forth to 'true share amounts' to resolve this issue.

src/contracts/core/StrategyManager.sol
```diff
@@ -439,7 +461,7 @@ contract StrategyManager is
         uint256[] memory shares = new uint256[](strategiesLength);
 
         for (uint256 i = 0; i < strategiesLength; ) {
-            shares[i] = stakerStrategyShares[staker][stakerStrategyList[staker][i]];
+            shares[i] = nonNormalizedStakerStrategyShares[staker][stakerStrategyList[staker][i]];
```
gpsanant, line 464: i believe this is only used by the UI and should be normalized

src/contracts/libraries/SlashingAccountingUtils.sol
```diff
@@ -0,0 +1,66 @@
...
+
+    function getEpochFromTimestamp(uint256 timestamp) internal pure returns (int256) {
+        return (int256(timestamp) - int256(EPOCH_GENESIS_TIMESTAMP)) / int256(EPOCH_LENGTH_SECONDS);
+    }
+
+    function currentEpoch() internal view returns (int256) {
+        return getEpochFromTimestamp(block.timestamp);
+    }
+
+    function scaleUp(uint256 shares, uint256 scalingFactor) internal pure returns (uint256) {
```
gpsanant, line 35: perhaps `normalize` and `denormalize` would be clearer?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -111,11 +125,15 @@ contract EigenPodManager is
...
-            "EigenPodManager.recordBeaconChainETHBalanceUpdate: sharesDelta must be a whole Gwei amount");
-        int256 currentPodOwnerShares = podOwnerShares[podOwner];
-        int256 updatedPodOwnerShares = currentPodOwnerShares + sharesDelta;
-        podOwnerShares[podOwner] = updatedPodOwnerShares;
+        int256 currentPodOwnerShares = nonNormalizedPodOwnerShares[podOwner];
+
+        // find the nonNormalizedShares amount
+        address operator = delegationManager.delegatedTo(podOwner);
+        uint256 scalingFactor = slasher.shareScalingFactor(operator, beaconChainETHStrategy);
+        int256 nonNormalizedSharesDelta = SlashingAccountingUtils.scaleUp(sharesDelta, scalingFactor);
```
gpsanant, line 133: should be `scaleDown` here?

src/contracts/libraries/SlashingAccountingUtils.sol
```diff
@@ -0,0 +1,66 @@
...
+        return (shares * int256(scalingFactor)) / int256(SHARE_CONVERSION_SCALE);
+    }
+
+    // @notice Overloaded version of `scaleDown` that accepts a signed integer shares amount
+    function scaleDown(int256 shares, uint256 scalingFactor) internal pure returns (int256) {
+        return (shares * int256(SHARE_CONVERSION_SCALE)) / int256(scalingFactor);
+    }
+
+    function findNewScalingFactor(uint256 scalingFactorBefore, uint256 bipsToSlash) internal pure returns (uint256) {
+        require(bipsToSlash != 0, "cannot slash for 0%");
```
gpsanant, line 54: add pretty revert msg

src/contracts/libraries/SlashingAccountingUtils.sol
```diff
@@ -0,0 +1,66 @@
...
+        require(bipsToSlash < BIPS_FACTOR, "cannot slash more than 99.99% at once");
+        uint256 scalingFactorAfter;
+        // deal with edge case of operator being slashed repeatedly, inflating scalingFactor to max uint size
+        // TODO: figure out more nuanced / appropriate way to handle this 'edge case', e.g. deciding if deposits should be blocked when close to limit
+        if (MAX_SCALING_FACTOR / scalingFactorBefore >= bipsToSlash) {
+            scalingFactorAfter = type(uint256).max;
```
gpsanant, line 60: TODO: make sure that `int256(type(uint256).max) = 2^255 - 1`

src/contracts/libraries/SlashingAccountingUtils.sol
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * @title Library of utilities for calculations related to slashing in EigenLayer
+ * @author Layr Labs, Inc.
+ * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
+ */
+library SlashingAccountingUtils {
+
+    // TODO: set this based on payments epochs
+    uint256 internal constant EPOCH_GENESIS_TIMESTAMP = 1606824000;
```
gpsanant, line 12: perfer another library for epoch logic

src/contracts/libraries/SlashingAccountingUtils.sol
gpsanant: ah realizing this is more POC than i thot, can ignore

src/contracts/core/Slasher.sol
```diff
@@ -27,7 +28,218 @@ import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
...
+            int256 changeInPendingSlashedBips = _changeInPendingSlashedBips({
+                bipsAllowedToSlash: bipsAllowedToSlash,
+                requestedSlashedBipsBefore: requestedSlashedBipsBefore,
+                requestedSlashedBipsAfter: requestedSlashedBipsAfter
+            });
+
+            // TODO: events
+            requestedSlashedBips[operator][strategy][epoch][avs] = requestedSlashedBipsAfter;
+            pendingSlashedBips[operator][strategy][epoch] = uint256(
+                int256(pendingSlashedBips[operator][strategy][epoch]) + int256(changeInPendingSlashedBips)
```
gpsanant, line 178: if we're converting to int256 here, might as well make bipsToSlash an int256 everywhere to make sure we don't hit overflow

src/contracts/core/Slasher.sol
```diff
@@ -27,7 +28,218 @@ import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
...
+            );
+        }
+    }
+
+    function reduceRequestedSlashedBips(
+        address operator,
+        IStrategy[] memory strategies,
+        // TODO: naming
+        uint256 bipsToReduce
```
gpsanant, line 187: maybe we just have a single function with an int256 here

src/contracts/libraries/SlashingAccountingUtils.sol
```diff
@@ -0,0 +1,66 @@
...
+
+    // TODO: define appropriately
+    uint256 internal constant EPOCH_LENGTH_SECONDS = 2 weeks;
+
+    uint256 internal constant SHARE_CONVERSION_SCALE = 1e18;
+
+    // an amount of shares over this will cause overflow when multiplying by `SHARE_CONVERSION_SCALE`
+    uint256 internal constant MAX_VALID_SHARES = type(uint96).max;
```
gpsanant, line 20: TODO: Enforce in strategy contracts or somewhere

src/contracts/pods/EigenPodManager.sol
gpsanant: oops confusing

src/contracts/core/DelegationManager.sol
```diff
@@ -19,6 +21,35 @@ import "./DelegationManagerStorage.sol";
...
+        return SlashingAccountingUtils.scaleDown(nonNormalizedOperatorShares[operator][strategy], scalingFactor);
+    }
+
+    // TODO: decide if this function needs to go in the interface at all
+    function pendingWithdrawalData(bytes32 withdrawalRoot) public view returns (PendingWithdrawalData memory) {
+        return _pendingWithdrawalData[withdrawalRoot];        
+    }
+
+    // @notice Returns 'true' if the `withdrawalRoot` corresponds to a queued-but-not-completed withdrawal, and 'false' otherwise
+    function pendingWithdrawals(bytes32 withdrawalRoot) public view returns (bool) {
```
gpsanant, line 42: `isWithdrawalPending`

src/contracts/core/DelegationManager.sol
```diff
@@ -730,13 +810,16 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
             strategies: strategies,
-            shares: shares
+            shares: nonNormalizedShares
         });
 
         bytes32 withdrawalRoot = calculateWithdrawalRoot(withdrawal);
 
         // Place withdrawal in queue
-        pendingWithdrawals[withdrawalRoot] = true;
+        _pendingWithdrawalData[withdrawalRoot] = PendingWithdrawalData({
```
gpsanant, line 819: 🧑‍🍳

gpsanant: are u sure this storage matches

src/contracts/core/DelegationManager.sol
```diff
@@ -657,63 +737,63 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+     * @dev If the `operator` is indeed an operator, then the operator's delegated nonNormalizedShares in the `strategies` are also decreased appropriately.
      * @dev If `withdrawer` is not the same address as `staker`, then thirdPartyTransfersForbidden[strategy] must be set to false in the StrategyManager.
      */
     function _removeSharesAndQueueWithdrawal(
         address staker, 
         address operator,
         address withdrawer,
         IStrategy[] memory strategies, 
-        uint256[] memory shares
+        uint256[] memory nonNormalizedShares
```
gpsanant, line 763: potentially might want to think about how to make UX better here so external entities don't think about normalization.

we could denormalize inside of this function, and return the normalized withdrawn shares during completion for example

src/contracts/core/DelegationManager.sol
```diff
@@ -583,8 +617,12 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
             );
         }
 
+        // TODO: currently, this _inclusive_, meaning the completed withdrawal will include slashing effects that were enacted in the `epochForEndOfSlashability`
+        // TODO: note that we again probably want the first real epoch to be epoch 1, so that existing queued withdrawals are safe from slashing
+        int256 epochForEndOfSlashability = withdrawalCreationEpoch(withdrawalRoot) + 1;
```
gpsanant, line 622: + 2

src/contracts/core/DelegationManager.sol
```diff
@@ -613,40 +668,65 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+                    );
                     /**
                     * Update shares amount depending upon the returned value.
                     * The return value will be lower than the input value in the case where the staker has an existing share deficit
                     */
                     uint256 increaseInDelegateableShares = eigenPodManager.addShares({
-                        podOwner: staker,
-                        shares: withdrawal.shares[i]
+                        podOwner: withdrawal.staker,
+                        shares: nonNormalizedShares
```
gpsanant, line 705: param name should be `nonNormalizedShares`?

src/contracts/libraries/SlashingAccountingUtils.sol
gpsanant, line 35: perhaps

`normalize(ISlasher slasher, uint256 shares, uint32 epoch)`

and such makes the most sense? reduces repetitive slasher calls everywhere

gpsanant: not sure how much gas this wastes when we reuse factor?

src/contracts/core/DelegationManager.sol
```diff
@@ -358,8 +389,11 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
                 // create the new storage
                 bytes32 newRoot = calculateWithdrawalRoot(migratedWithdrawal);
                 // safety check to ensure that root doesn't exist already -- this should *never* be hit
-                require(!pendingWithdrawals[newRoot], "DelegationManager.migrateQueuedWithdrawals: withdrawal already exists");
-                pendingWithdrawals[newRoot] = true;
+                require(!pendingWithdrawals(newRoot), "DelegationManager.migrateQueuedWithdrawals: withdrawal already exists");
```
gpsanant, line 392: there are no unmigrated withdrawals (pending review), so ideally we can delete this after review

src/contracts/core/Slasher.sol
gpsanant: to combine reduce and increase

gpsanant: check my interface in gitbook

src/contracts/core/DelegationManager.sol
ChaoticWalrus: this is backwards compatible 🤷 
I agree the name you suggest would be better though

src/contracts/core/DelegationManager.sol
ChaoticWalrus: no, there's a TODO. I'm like 75% on it

src/contracts/core/DelegationManager.sol
ChaoticWalrus: if you do that then you have to do more lookups when the withdrawal is completed.
Specifically, you would need to look up the conversion ratio both when the withdrawal was queued and at the end of slashability, instead of just the latter.
what you're suggesting sounds more like what I had in the first version https://github.com/Layr-Labs/eigenlayer-contracts/pull/528 if you want to take a look at that.
there could also be a way to format things to avoid this? or the 2-lookup tradeoff could be worth it.

src/contracts/core/DelegationManager.sol
ChaoticWalrus: yes

src/contracts/core/DelegationManager.sol
ChaoticWalrus: yes, I'd like to delete but there's a bunch of tests and interface stuff so I didn't bother.
we should get rid of it though; that'll help with contract code size for sure.

src/contracts/libraries/SlashingAccountingUtils.sol
ChaoticWalrus: this is a reasonable suggestion. I would like good naming for these.
perhaps `renormalize` when returning to just 'shares'
`denormalizeShares` and `renormalizeToShares`, or something?

src/contracts/libraries/SlashingAccountingUtils.sol
ChaoticWalrus: I think int256(type(uint256).max) is actually -2^255

why the TODO?

src/contracts/libraries/SlashingAccountingUtils.sol
8sunyuan: Yes and have PaymentCoordinator read GENESIS TIMESTAMP from the same library

src/contracts/libraries/SlashingAccountingUtils.sol
```diff
@@ -0,0 +1,66 @@
...
+
+    uint256 internal constant BIPS_FACTOR = 10000;
+
+    // TODO: explain this better. basically seems like we may need to set some max factor beyond which shares are just zeroed out
+    uint256 internal constant MAX_SCALING_FACTOR = type(uint256).max / (MAX_VALID_SHARES * SHARE_CONVERSION_SCALE);
+
+    function getEpochFromTimestamp(uint256 timestamp) internal pure returns (int256) {
```
8sunyuan, line 27: Does it make sense to have negative epochs if a timestamp before GENESIS is specified? I guess it's cleaner than having the view function revert

ChaoticWalrus: that was my line of reasoning as well.

src/contracts/core/DelegationManager.sol
gpsanant: ah rip nice

src/contracts/libraries/SlashingAccountingUtils.sol
gpsanant: works for me

src/contracts/libraries/SlashingAccountingUtils.sol
gpsanant: If that's true, then the int256 version of scaleUp and scaleDown are probably broken? they need to large and small negative numbers

src/contracts/libraries/SlashingAccountingUtils.sol
gpsanant: eh, i prefer revert. signed integers scare me. maybe skill issue tho

src/contracts/core/Slasher.sol
```diff
@@ -27,7 +28,218 @@ import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
...
+        address avs, 
+        IStrategy strategy, 
+        int256 epoch
+    ) public pure returns (uint256) {
+        // TODO: this is a stub. it needs implementation
+        return 10;
+    }
+
+    // @notice fetches the maximum slashing rate -- expressed in bips -- by an AVS for the shares of a certain strategy for a certain epoch
+    function getMaxSlashingRate(
```
8sunyuan, line 144: Noting this will need to be an interface added to middleware ServiceManagerBase as well as `createSlashingRequest`,  `reduceRequestedSlashedBips`, `executeSlashing`

src/contracts/libraries/SlashingAccountingUtils.sol
8sunyuan: +1 `normalize` and `denormalize` something like

```
function normalize(nonNormalizedShares, uint256 scalingFactor) returns (uint256)

function denormalize(normalizedShares, uint256 scalingFactor) returns (uint256)
```

src/contracts/core/DelegationManager.sol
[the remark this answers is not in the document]
ChaoticWalrus: ahh, yes, this check is totally missing right now

src/contracts/core/Slasher.sol
```diff
@@ -27,7 +28,218 @@ import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
...
+                requestedSlashedBipsAfter = requestedSlashedBipsBefore - bipsToReduce;
+            }
+            // TODO: consider loss of precision here. this is undesirable; we should fix it
+            uint256 bipsAllowedToSlash =
+                getSlashableBips(operator, avs, strategy, epoch) * getMaxSlashingRate(avs, strategy, epoch) / SlashingAccountingUtils.BIPS_FACTOR;
+            int256 changeInPendingSlashedBips = _changeInPendingSlashedBips({
+                bipsAllowedToSlash: bipsAllowedToSlash,
+                requestedSlashedBipsBefore: requestedSlashedBipsBefore,
+                requestedSlashedBipsAfter: requestedSlashedBipsAfter
+            });
```
ChaoticWalrus, line 209: this implementation multiplies slashable bips by max slashing rate and then compares that to the requested slashed bips, but what it should be doing is comparing the requested slashed bips to the max slashing rate and then multiplying by the slashable bips

ChaoticWalrus: h/t to @8sunyuan for pointing this out

src/contracts/core/StrategyManager.sol
ChaoticWalrus: ah, good point. perhaps we want a new `getNonNormalizedDeposits` function, and we keep the behavior of `getDeposits` (returning normalized shares) the same as it was before.

src/contracts/core/Slasher.sol
ChaoticWalrus: how are you feeling on this?
we could name it `modifyRequestedSlashedBips` or something?
I know we spoke about this some in person; my current feeling is that the model we're picturing for AVSs would have different access controls for increasing the slashed bips and decreasing them, so...I'm unsure there's much benefit to unifying.

src/contracts/libraries/SlashingAccountingUtils.sol
ChaoticWalrus: I think it's significant savings when we reuse the factor (on the order of several thousand gas for every reuse)

gpsanant: How often are we reusing?

src/contracts/core/StrategyManager.sol
ChaoticWalrus: addressed in this commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/543/commits/a48492a849dea0c48eacd46f353f36168a7fd213
we need to decide which functions to keep and which to scrap, or else we'll end up with both non- and normalized versions of everything

src/contracts/libraries/SlashingAccountingUtils.sol
ChaoticWalrus: renaming done here https://github.com/Layr-Labs/eigenlayer-contracts/pull/543/commits/94a529babf81fe8a108d1354be8cf7155c4d9498, resolving thread

src/contracts/libraries/SlashingAccountingUtils.sol
ChaoticWalrus: epoch logic now has its own library https://github.com/Layr-Labs/eigenlayer-contracts/pull/543/commits/5074afb01bdc7ae00a47866b2699424b5b780234

src/contracts/core/Slasher.sol
ChaoticWalrus: currently we are avoiding overflow by constraining inputs to `createSlashingRequest` be <= 10000 bips, so it's impossible to send enough transactions to overflow.
I think this is a reasonable suggestion though!

src/contracts/core/DelegationManager.sol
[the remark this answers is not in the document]
ChaoticWalrus: added here https://github.com/Layr-Labs/eigenlayer-contracts/pull/543/commits/84277d54f012d76cf211e4683ca748cfad922776
along with a TODO for another check that seems apt

src/contracts/core/Slasher.sol
ChaoticWalrus: I believe I've addressed this in https://github.com/Layr-Labs/eigenlayer-contracts/pull/543/commits/8a9447a260f59fe2d3a94c210b68067592bcdced

src/contracts/core/Slasher.sol
```diff
@@ -27,7 +28,199 @@ import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
...
+        }
+    }
+
+    function _changeInPendingSlashedBips(
+        uint256 slashableBips,
+        uint256 requestedSlashedBipsBefore,
+        uint256 requestedSlashedBipsAfter
+    ) internal pure returns (int256) {
+        return int256(slashableBips) * (int256(requestedSlashedBipsAfter) - int256(requestedSlashedBipsAfter));
+    }
```
ChaoticWalrus, line 222: @gpsanant I'm sorry but you butchered this function in commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/543/commits/9cfcdc17c9728684063d975e08f8fad75c1127a6
at present, this is bugged in two ways:

1. if implemented as it appears you intended, then an AVS could slash for any amount -- including well in excess of 10000 bips
2. this is definitely not implemented as intended, since you are subtracting `requestedSlashedBipsAfter` from itself and thus it will always return `0`

There are other options but I personally recommend reverting to something with equivalent behavior to what was present prior to your commit

gpsanant:
closed in favor of #582

PR #549, opened 2024-05-15
feat: modify withdrawal storage model

ChaoticWalrus:
This is one option for how to migrate from existing storage to new, epoch-focused behavior.
The work is certainly incomplete; I'm opening this draft PR as a demonstration of the work involved + potential first step towards completing it.

PR #550, opened 2024-05-15
feat: add storage to track the epoch in which withdrawals were queued

ChaoticWalrus:
Alternative approach to that in https://github.com/Layr-Labs/eigenlayer-contracts/pull/549
This is a much simpler, "lightweight" change, but involves adding additional storage + notably writing to one more storage slot per newly queued withdrawal.
This PR was opened for illustrative purposes; I think we can fairly easily solve the issue here by simply packing the new storage into the existing `pendingWithdrawals` mapping.


ChaoticWalrus:
This PR is clearly superior to its competition in https://github.com/Layr-Labs/eigenlayer-contracts/pull/549. Since It's being merged into a 'POC' draft PR anyway, I'm just going to go ahead and merge this now.

PR #554, opened 2024-05-21
fix: track exit gwei in checkpoint

pooleja:
This PR shows an example of how we can track the amount of exited validator balance in a checkpoint.

The problem for LRTs is that they need to track ETH exiting the EigenPod and treat rewards generated from staking vs exited validators differently from the protocol contracts.  For example, many LRTs take a protocol fee (e.g. 10%) from the rewards generated from staking and should NOT take this same fee from ETH leaving the EigenPod due to a validator exit.

This is just a quick example of how we could track it:
* Do not delete the last checkpoint when it is finished.  This allows the EigenPod owner to ensure that they have captured the expected amount of ETH from the previous checkpoint before starting a new one.
* In `_verifyCheckpointProof()`, if the validator is marked as exited, return the balance delta
* In the `CheckPoint` struct, track the exited balance delta in addition to the total

### DIFF
--- a/script/deploy/devnet/M2_Deploy_From_Scratch.s.sol
+++ b/script/deploy/devnet/M2_Deploy_From_Scratch.s.sol
@@ -187,7 +187,6 @@ contract Deployer_M2 is Script, Test {
             ethPOSDeposit,
             delayedWithdrawalRouter,
             eigenPodManager,
-            MAX_RESTAKED_BALANCE_GWEI_PER_VALIDATOR,
             GOERLI_GENESIS_TIME
         );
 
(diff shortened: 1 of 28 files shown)


wadealexc:
> I'm not 100% sure how you're using this, but I'm wondering if there's an issue where you go to query `currentCheckpoint` and someone has already overwritten it with a new checkpoint.

ah i guess the pod owner would have to call `startCheckpoint`, but i don't know how your pod owner is set up - maybe this is fine?

pooleja:
Yeah @wadealexc - I initially thought we could just track it when we submit the proofs, but since the proof submission call is unguarded, we can't guarantee we are the ones to finalized the checkpoint.  Since the start checkpoint call is guarded, then we can ensure we have saved off the amount prior to starting a new one.

I was also wondering if we needed to keep a historical list, but if we can be guaranteed to get the data before starting a new checkpoint, that should be enough.

wadealexc:
> I was also wondering if we needed to keep a historical list, but if we can be guaranteed to get the data before starting a new checkpoint, that should be enough.

Makes sense, but I wonder if other LRTs or integrators have the same guarantees. I'm tempted to opt for something like the mapping I suggested above - would something like that work?

script/deploy/devnet/M2_Deploy_From_Scratch.s.sol
emagici, line 1: better to remove the unused code

pooleja:
> > I was also wondering if we needed to keep a historical list, but if we can be guaranteed to get the data before starting a new checkpoint, that should be enough.
> 
> Makes sense, but I wonder if other LRTs or integrators have the same guarantees. I'm tempted to opt for something like the mapping I suggested above - would something like that work?

Yeah, keeping a historical list would work as well and provide better on-chain debugging if anything goes wrong.

> > I was also wondering if we needed to keep a historical list, but if we can be guaranteed to get the data before starting a new checkpoint, that should be enough.
> 
> Makes sense, but I wonder if other LRTs or integrators have the same guarantees. I'm tempted to opt for something like the mapping I suggested above - would something like that work?

Yeah, I think keeping a historical list of checkpoints would work as well and also provide better on-chain debugging and recovering if anything were to go awry.  My initial approach was just the lowest impact way I could think of.

Also I would probably suggest having `startCheckpoint()` return the checkpoint ID of the newly created  checkpoint, as a caller that is tracking these would need to immediately query it anyways to save it off (assuming the mapping is indexed by timestamp).

wadealexc:
Thanks again for the PR - closing this as it has been implemented in the partial withdrawal PR: https://github.com/Layr-Labs/eigenlayer-contracts/pull/515

PR #556, opened 2024-05-21
fix: defend against checkpoint proof griefing

youfoundron:
Currently `EigenPod::verifyCheckpointProofs` is a function used to batch checkpoint proof submissions.

It is important to note that anyone can submit proofs directly to EigenPod for proof verification. If a validator balance update has already been verified in a started checkpoint, any subsequent call for re-verification will revert.

Because of this an attacker might construct a griefing attack that effectively DOS'es an LRT protocol's attempt to verify proofs for a batch of validators.

All the attacker has to do is to keep watching the mempool & simply front-run the batch verification transaction by submitting a proof for just one single validator in the batch.

When the actual transaction comes for execution, entire batch fails as one of the  proofs is already verified.

To prevent this, I am proposing modifying `EigenPod::verifyCheckpointProofs` so that only the pod owner can make checkpoint proof submissions.

### DIFF
--- a/src/contracts/pods/EigenPod.sol
+++ b/src/contracts/pods/EigenPod.sol
@@ -175,7 +175,7 @@ contract EigenPod is
 
     /**
      * @dev Progress the current checkpoint towards completion by submitting one or more validator
-     * checkpoint proofs. Anyone can call this method to submit proofs towards the current checkpoint.
+     * checkpoint proofs. Only the pod owner can call this method to submit proofs towards the current checkpoint.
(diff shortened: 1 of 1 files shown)


wadealexc:
Should be addressed by [dcfce19](https://github.com/Layr-Labs/eigenlayer-contracts/pull/515/commits/dcfce19c2995a5858036c431cb03bdc0fc491e24) "skipping" duplicate submissions rather than reverting

Feel free to reopen or flag if that doesn't fix the issue! Good callout though :)

PR #557, opened 2024-05-22
fix: disallow null stake delegations

RonTuretzky:
This check will prevent stakers from delegating if they don't have any stake, in unity with the Eigenlayer frontend. 
This will prevent unintended behavior of users and downstream edge case handling for payments (eg: not giving rewards to stakers who have 0 stake from the operator's earnings reciever address.) 

@samlaf


ypatil12:
Closing due to staleness, happy to discuss offline if you want @RonTuretzky

PR #559, opened 2024-05-22
fix: psigp audit fix

8sunyuan:
Attached the SigP audit report for the `PaymentCoordinator` Contract
The fixes are just removing unnecessary require checks since we are using an incrementing nonce as part of the `RangePayment` hash.
Also pushed updated contract bindings


wadealexc (approved):
LGTM!

PR #562, opened 2024-05-23
chore: deprecate earningsReceiver

8sunyuan:
**Description**
`IDelegationManager.OperatorDetails.earningsReceiver` is not being used as part of payments so to avoid confusion with the upcoming release of Payments we are deprecating this address slot in the `OperatorDetails` struct.

Since a non-zero address of `earningsReceiver` is being used to check whether an address is an operator or not, the `isOperator(operator)` function is now changed to simply check `delegatedTo[operator] == operator`. That is a operator is an operator iff they are delegated to themselves.

**Changes**
- renamed `earningsReceiver` to `__deprecated_earningsReceiver`
- `isOperator` checks an address is self delegated now
- `_delegate()` no longer checks the `staker` is not delegated and that `operator` is an operator. It assumes these checks are made prior to the internal call. This was needed to fix operator registration because of how `isOperator` was changed. 
- bunch of unit test fixes and docs
- deleted commented out tests in `Delegation.t.sol`


src/contracts/core/DelegationManager.sol
```diff
@@ -91,17 +91,14 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
      * @dev Note that the `metadataURI` is *never stored * and is only emitted in the `OperatorMetadataURIUpdated` event
      */
     function registerAsOperator(
         OperatorDetails calldata registeringOperatorDetails,
         string calldata metadataURI
     ) external {
-        require(
-            _operatorDetails[msg.sender].earningsReceiver == address(0),
-            "DelegationManager.registerAsOperator: operator has already registered"
-        );
```
wadealexc, line 104: I think we should default to setting the value of this deprecated field to zero.

We dont need to keep this require here, but if we plan to repurpose this field later, we need to make sure people aren't already setting it to a nonzero value.

wadealexc (changes requested):
See comment above - if we want to leave the option to use this field, we need to ensure it stays zeroed out.

src/contracts/core/DelegationManager.sol
ChaoticWalrus: it has already been set to a nonzero value by all existing users 😞 
one option I suggested was setting it to the operator's own address, but I think setting it to zero could also be fine.

ChaoticWalrus:
Looks like this PR breaks a Prover rule
https://prover.certora.com/output/83341/72a015ef94bd4d09b5b8422ed3c36059?anonymousKey=f746a2604541cb957aa0857fc48b48d920cf0af8
`operatorCannotUnregister` is breaking on a call to because `delegateTo`, ultimately because `EigenPodManager.podOwnerShares` is resulting in a HAVOC summary; I'm pretty sure this can be fixed by using a NONDET summary for this call in the spec file

ChaoticWalrus (approved):
I'm personally OK with this, but I would like to see the Prover error fixed prior to merging, as it seems like merging this as-is would cause our CI to fail consistently

src/contracts/core/DelegationManager.sol
8sunyuan: Since all existing registered operators have nonzero values I thought its best to not make assumptions on whatever value this field is and let it be whatever people have it set to.

If we use this field later, we need to allow for a period for operators to update it and then assume afterwards that it's configured.

wadealexc: ah right i missed that we currently require this to be nonzero. yikes!

if we use this field later... we could do that or we might want to migrate and set it to 0. or set it to a sane default.

wadealexc (approved):
LGTM

PR #569, opened 2024-05-28
get rid of max slashing rate

gpsanant:
the max slashing rate should simply be the max bips opted in by the operator

an AVS will slash the maximum in a worst case scenario. further slashing conditions and slashing parameterization is in scope for a future release


gpsanant:
@8sunyuan approved of the concept, not impl offline

PR #572, opened 2024-05-28
Opset manager poc

gpsanant:
todo:
- opsetmanager
- stake locking


src/contracts/core/OperatorSetManager.sol
```diff
@@ -0,0 +1,164 @@
...
+
+    function _isAllocatorFor(address sender, address operator) internal view returns (bool) {
+        // Implement allocator check logic
+        return true; // Stubbed for now
+    }
+
+    function _hashOperatorSet(
+        OperatorSet memory operatorSet
+    ) internal pure returns (bytes32) {
+        OperatorSet[] memory operatorSetArray = new OperatorSet[](1);
```
8sunyuan, line 160: why can't we just `abi.encode(operatorSet)`?

PR #574, opened 2024-05-29
chore: remove migrateWithdrawals

8sunyuan:
All StrategyManager withdrawals have been verified to have been completed so this backwards compatible legacy function is no longer needed. This PR removes associated logic from `StrategyManager` and `DelegationManager`.

Also for consistency prepended deprecated variables with `__deprecated_`


wadealexc (approved):
LGTM

PR #579, opened 2024-06-03
Feat/operator sets

ypatil12:
POC of Operator Set Release


src/contracts/core/AVSDirectory.sol
```diff
@@ -77,7 +184,11 @@ contract AVSDirectory is
         );
         require(
             delegation.isOperator(operator),
-            "AVSDirectory.registerOperatorToAVS: operator not registered to EigenLayer yet");
+            "AVSDirectory.registerOperatorToAVS: operator not registered to EigenLayer yet"
+        );
+        require(
+            !isOperatorSetAVS[msg.sender]
```
gpsanant, line 190: `operatorAVSOperatorSetCount != 0`?

src/contracts/core/AVSDirectory.sol
```diff
@@ -52,6 +55,110 @@ contract AVSDirectory is
...
+     * @dev msg.sender is used as the AVS
+     * @dev operator must be registered for msg.sender AVS and the given 
+     * operator set
+     */
+    function deregisterOperatorFromOperatorSet(
+        address operator,
+        uint32 operatorSetID
+    ) external onlyWhenNotPaused(PAUSED_OPERATOR_REGISTER_DEREGISTER_TO_AVS) {
+        require(
+            operatorSetRegistrations[msg.sender][operator][operatorSetID] == true,
```
gpsanant, line 146: u can get rid `== true`

src/contracts/core/AVSDirectory.sol
```diff
@@ -52,6 +55,110 @@ contract AVSDirectory is
...
+        // Mark the salt as spent
+        operatorSaltIsSpent[operator][operatorSignature.salt] = true;
+
+        // Set the operator as registered if not already
+        if (avsOperatorStatus[msg.sender][operator] != OperatorAVSRegistrationStatus.REGISTERED) {
+            avsOperatorStatus[msg.sender][operator] = OperatorAVSRegistrationStatus.REGISTERED;
+            emit OperatorAVSRegistrationStatusUpdated(operator, msg.sender, OperatorAVSRegistrationStatus.REGISTERED);
+        }
+
+        emit OperatorAddedToOperatorSet(operator, msg.sender, operatorSetID);
```
gpsanant, line 127: emit before if

src/contracts/core/AVSDirectory.sol
scotthconner: we are still not allowing the operator to call this?

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -30,17 +35,30 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
...
-    /// @dev Salt is used in the `registerOperatorToAVS` function.
+    /// @dev Salt is used in the `registerOperatorToAVS` and `registerOperatorToOperatorSet` function.
     mapping(address => mapping(bytes32 => bool)) public operatorSaltIsSpent;
 
-    constructor(IDelegationManager _delegation) {
+    /// @notice Mapping: AVS => whether or not the AVS uses operator set
+    mapping(address => bool) public isOperatorSetAVS;
+
+    /// @notice Mapping: AVS => operatorSetID => strategy => whether or not the strategy is in the operator set
+    mapping(address => mapping(uint32 => mapping(IStrategy => bool))) public operatorSetStrategies;
```
scotthconner, line 45: what does "in the operator set" mean? That they can slash it? or that they are paying for it?

scotthconner (commented):
just a few questions.

src/contracts/interfaces/IAVSDirectory.sol
```diff
@@ -19,6 +26,22 @@ interface IAVSDirectory is ISignatureUtils {
...
+     */
+    event OperatorAddedToOperatorSet(address operator, OperatorSet operatorSet);
+    
+    /**
+     * @notice Emitted when an operator is removed from an operator set
+     */
+	event OperatorRemovedFromOperatorSet(address operator, OperatorSet operatorSet);
+
+    /// @notice Emitted when a strategy is added to an operator set
+	event OperatorSetStrategyAdded(OperatorSet operatorSet, IStrategy strategy);
```
ypatil12, line 40: TODO: index these

src/contracts/core/AVSDirectory.sol
```diff
@@ -127,6 +238,53 @@ contract AVSDirectory is
...
+     * @param operatorSetID The ID of the operator set
+     * @param strategies The strategies to add to the operator set
+     * TODO: align with team on if we want to keep the two require statements
+     */
+    function addStrategiesToOperatorSet(
+        uint32 operatorSetID,
+        IStrategy[] calldata strategies
+    ) external {
+        uint256 strategiesToAdd = strategies.length;
+        for (uint256 i = 0; i < strategiesToAdd; i++) {
```
ypatil12, line 252: Change to `.length` in iterator since using calldata

src/contracts/core/AVSDirectory.sol
ypatil12: Are you suggesting to only check `operatorAVSOperatorSetCount != 0`? The end result of doing so is that the AVS is not forced into becoming an operator set AVS, which we want to do here

src/contracts/core/AVSDirectory.sol
ypatil12: Correct @scotthconner . Operators still have to register via the AVS. We are considering adding functionality for operators to list the operator sets they want in this contract - "standby mode"

src/contracts/core/AVSDirectory.sol
ypatil12: I prefer this way since the log indices are ordered avs registration then operator set registration

src/contracts/core/AVSDirectoryStorage.sol
ypatil12: They can eventually slash it if the operator has also set its strategy in the operator set to slashable. Mainly used for the UI thought to properly account for strategies in each operator set.

src/contracts/interfaces/IAVSDirectory.sol
8sunyuan, line 1: file spacing is a little wacky here and on the other files too. I had the same thing with copying interfaces from gitbook

src/contracts/interfaces/IAVSDirectory.sol
```diff
@@ -35,6 +58,59 @@ interface IAVSDirectory is ISignatureUtils {
...
+	 * @param operator the address of the operator to be added to the operator set
+	 * @param operatorSetID the ID of the operator set
+	 * @param operatorSignature the signature of the operator on their intent to register
+	 * @dev msg.sender is used as the AVS
+	 * @dev operator must not have a deregistration from the operator set
+	 * @dev if this is the first operator set in the AVS that the operator is 
+	 * registering for, a OperatorAVSRegistrationStatusUpdated event is emitted with 
+	 * a REGISTERED status
+	 */
+	function registerOperatorToOperatorSet(address operator, uint32 operatorSetID, ISignatureUtils.SignatureWithSaltAndExpiry memory operatorSignature) external;
```
8sunyuan, line 73: Can we make this an array of OperatorSetIds, update to `uint32[] calldata operatorSetIds`? Otherwise the Operator needs to provide a signature for each operatorSetId they are registering for.
Similarly for deregister as well, take an array of OperatorSetIds

0xClandestine:
I have quite a bit of stashed changes.

src/contracts/interfaces/IAVSDirectory.sol
ypatil12: cc @0xClandestine

src/test/unit/AVSDirectoryUnit.t.sol
```diff
@@ -170,14 +172,137 @@ contract AVSDirectoryUnitTests is EigenLayerUnitTestSetup, IAVSDirectoryEvents {
     }
 }
 
+contract AVSDirectoryUnitTests_registerOperatorToOperatorSet is AVSDirectoryUnitTests {
+    event OperatorAddedToOperatorSet(address operator, IAVSDirectory.OperatorSet operatorSet);
```
ypatil12, line 176: Nit: import events from IAVSDirectoryEvents

src/test/unit/AVSDirectoryUnit.t.sol
```diff
@@ -170,14 +172,137 @@ contract AVSDirectoryUnitTests is EigenLayerUnitTestSetup, IAVSDirectoryEvents {
...
+
+    function testFuzz_revert_SignatureIsExpired(
+        address operator,
+        uint32 oid,
+        bytes32 salt,
+        uint256 expiry
+    ) public virtual {
```
ypatil12, line 183: might want to add `filterFuzzedAddressInputs(operator)`

ypatil12 (commented):
Nit testing comments

8sunyuan:
Another nit: lets stick to `OperatorSetId` to be consistent with how we call it `operatorId` in the middleware repo

src/test/unit/AVSDirectoryUnit.t.sol
8sunyuan: +1 for all fuzzed address inputs, we get a couple flakey fuzz tests in our CI occasionally when we forget this modifier

src/contracts/interfaces/IAVSDirectory.sol
```diff
@@ -17,7 +25,25 @@ interface IAVSDirectory is ISignatureUtils {
     event AVSMetadataURIUpdated(address indexed avs, string metadataURI);
```
8sunyuan, line 25: We should have `updateAVSMetadataURI` and events updated with operatorSets so AVSs can update accordingly to their operatorSets. Especially needed for the "meta" AVSs who need operator sets for UI purposes

src/contracts/interfaces/IAVSDirectory.sol
0xClandestine: done :)

src/contracts/interfaces/IAVSDirectory.sol
ypatil12: @stevennevins is leading design for operator set metadata, assuming it will be in separate PR

src/contracts/interfaces/IAVSDirectory.sol
```diff
@@ -2,58 +2,171 @@
...
+     *  @notice Emitted when an operator's registration status with an AVS is updated.
+     *  Specifically, when an operator enters its first operator set for an AVS, or
+     *  when it is removed from the last operator set.
      */
+    event OperatorAVSRegistrationStatusUpdated(
+        address indexed operator, address indexed avs, OperatorAVSRegistrationStatus status
+    );
+
+    /// @notice Emitted when an operator is added to an operator set.
+    event OperatorAddedToOperatorSet(address operator, OperatorSet operatorSet);
```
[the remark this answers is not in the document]
0xClandestine: I was thinking the same.

Will include.

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -16,31 +14,58 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
...
+    mapping(address => mapping(uint32 => mapping(IStrategy => bool))) public isOperatorSetStrategy;
+
+    /// @notice Mapping: avs => operator => operatorSetID => whether the operator is registered for the operator set
+    mapping(address => mapping(address => mapping(uint32 => bool))) public isOperatorInOperatorSet;
+
+    /// @notice Mapping: avs => operator => number of operator sets the operator is registered for the AVS
+    mapping(address => mapping(address => uint256)) public operatorAVSOperatorSetCount;
+
+    /// @notice Mapping: avs = operator => Whether the given operator set in standby mode or not
+    mapping(address => mapping(address => bool)) public onStandby;
```
stevennevins, line 58: Should this be by operatorSetId?

0xClandestine: My understanding is that when a user puts an AVS in standby it allows the AVS to void the need for a signature when registering the operator to various operator sets. 

With this in mind `operatorSetId` didn't seam needed for storage.

0xClandestine: Yeah I think my interpretation of the spec is a bit off, will fix.

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -16,31 +14,58 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
...
-    /// @dev Salt is used in the `registerOperatorToAVS` function.
+    /// @dev Salt is used in the `registerOperatorToAVS` and `registerOperatorToOperatorSet` function.
     mapping(address => mapping(bytes32 => bool)) public operatorSaltIsSpent;
 
-    constructor(IDelegationManager _delegation) {
+    /// @notice Mapping: AVS => whether or not the AVS uses operator set
+    mapping(address => bool) public isOperatorSetAVS;
+
+    /// @notice Mapping: AVS => operatorSetID => strategy => whether or not the strategy is in the operator set
+    mapping(address => mapping(uint32 => mapping(IStrategy => bool))) public isOperatorSetStrategy;
```
stevennevins, line 49: Do you think it would be worth enumerating the strategies so we can get all registered strategies?

0xClandestine:
``` 
    function cancelSalt(bytes32 salt) external {
        require(!operatorSaltIsSpent[msg.sender][salt], "AVSDirectory.cancelSalt: cannot cancel spent salt");
        operatorSaltIsSpent[msg.sender][salt] = true;
    }
```

This check only saves gas during revert cases, at the expense of normal use. Imo this check should be removed.

src/contracts/core/AVSDirectoryStorage.sol
0xClandestine: In a perfect world everything would be enumerated and mappings like this could be bitmaps. The issue with enumerating here is that we would have to store a constant list of strategies which is possible but a bit unconventional.

src/contracts/core/AVSDirectory.sol
```diff
@@ -48,21 +56,201 @@ contract AVSDirectory is
...
+            }
+
+            // Assert `operator` has not already been registered to `operatorSetIds[i]`.
+            require(
+                !isOperatorInOperatorSet[msg.sender][operator][operatorSetIds[i]],
+                "AVSDirectory.registerOperatorToOperatorSets: operator already registered to operator set"
+            );
+
+            // Mutate calling AVS to operator set AVS status, preventing further legacy registrations.
+            if (!isOperatorSetAVS[msg.sender]) isOperatorSetAVS[msg.sender] = true;
```
ypatil12, line 184: Nit: This doesn't need to be set in every iteration? Just at the end of the function?

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -16,31 +14,55 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
     bytes32 public constant OPERATOR_AVS_REGISTRATION_TYPEHASH =
         keccak256("OperatorAVSRegistration(address operator,address avs,bytes32 salt,uint256 expiry)");
 
+    /// @notice The EIP-712 typehash for the `OperatorSetRegistration` struct used by the contract
+    bytes32 public constant OPERATOR_SET_REGISTRATION_TYPEHASH =
+        keccak256("OperatorSetRegistration(address avs,uint32[] operatorSetIds,bytes32 salt,uint256 expiry)");
+
+    /// @notice The EIP-712 typehash for the `StandbyParams` struct used by the contract
+    bytes32 public constant OPERATOR_STANDBY_UPDATE =
+        keccak256("OperatorStandbyUpdate(address avs,uint32 id,bool onStandby,bytes32 salt,uint256 expiry)");
```
ypatil12, line 23: This typehash should be over a list of StandbyUpdates, not a single one?

ypatil12 (commented):
Added 2 minor comments. Also, please resolve the comments on the test file if not already done :)

src/contracts/core/AVSDirectory.sol
```diff
@@ -48,21 +56,201 @@ contract AVSDirectory is
...
+                operator,
+                calculateUpdateStandbyDigestHash(standbyParams, operatorSignature.salt, operatorSignature.expiry),
+                operatorSignature.signature
+            );
+            // Spend salt.
+            operatorSaltIsSpent[operator][operatorSignature.salt] = true;
+        }
 
+        for (uint256 i; i < standbyParams.length; ++i) {
+            onStandby[standbyParams[i].operatorSet.avs][operator][standbyParams[i].operatorSet.id] =
```
gpsanant, line 103: i feel like this mapping should be `onStandby[operator][avs][id]`

src/contracts/core/AVSDirectory.sol
```diff
@@ -48,21 +56,201 @@ contract AVSDirectory is
...
+                "AVSDirectory.deregisterOperatorFromOperatorSet: operator not registered for operator set"
+            );
+
+            // Mutate `isOperatorInOperatorSet` to `false` for `operatorSetIds[i]`.
+            isOperatorInOperatorSet[msg.sender][operator][operatorSetIds[i]] = false;
+
+            // Decrement `operatorAVSOperatorSetCount` by 1.
+            // The above assertion makes underflow logically impossible here.
+            unchecked {
+                --operatorAVSOperatorSetCount[msg.sender][operator];
```
gpsanant, line 228: why dont u subtract the length outside?

0xClandestine: oops

PR #581, opened 2024-06-05
chore: dev into testnet-holesky

8sunyuan:
Merging `dev` for the testnet Holesky Rewards release
Release tag after merging will be `v0.3.0-holesky-rewards`


src/contracts/core/StrategyManager.sol
```diff
@@ -195,19 +195,6 @@ contract StrategyManager is
         strategy.withdraw(recipient, token, shares);
     }
 
-    /// @notice Function called by the DelegationManager as part of the process of transferring existing queued withdrawals from this contract to that contract.
-    /// @dev This function is expected to be removed in the next upgrade, after all queued withdrawals have been migrated.
-    function migrateQueuedWithdrawal(DeprecatedStruct_QueuedWithdrawal memory queuedWithdrawal) external onlyDelegationManager returns(bool, bytes32) {
```
gpsanant, line 200: if this isn't on testnet are we ok making the release?

8sunyuan: This function doesn't get called regardless because we redeployed onto Holesky last minute for M2. We never had version of M1 contracts. 
Can queue multisig txs after to upgrade the SM and DM to be accurate with this branch

wadealexc (approved):

PR #582, opened 2024-06-08
feat: slashing

gpsanant:
This PR is the initial implementation of slashing for EigenLayer.

It has 2 large changes which are
- the operator set manager. which newly handles
  - setting of magnitudes
  - stake locking
- the slasher and corresponding DM/SM/EPM changes
  - rebasing like accounting mechanism to handle slashing
  - the request/cancel/execute flow for slashing requests

big todos are:
- how to handle 100% slashing
- eigenpod handling
- stake locking accounting
- events


src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+     * @param strategies the list of strategies slashing requested is being modified for
+     * @param bipsToIncrease the basis points slashing to modify for given strategies
+     */
+    function increaseRequestedBipsToSlash(
+        address operator,
+        bytes4 operatorSetID,
+        IStrategy[] memory strategies,
+        uint32 bipsToIncrease
+    ) external {
+        require(bipsToIncrease > 0, "Slasher.increaseRequestedBipsToSlash: bipsToIncrease must be positive");
```
8sunyuan, line 34: Looks like this was changed to uint32 for the bips param. Don't need this require since `_modifyRequestedBipsToSlash` already requires non-zero value

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+        IStrategy[] memory strategies,
+        uint32 epoch,
+        uint32 bipsToReduce
+    ) external {
+        uint32 currentEpoch = EpochUtils.currentEpoch();
+        require(
+            epoch == currentEpoch || epoch + 1 == currentEpoch,
+            "Slasher.reduceRequestedBipsToSlash: can only reduce for current or previous epoch"
+        );
+        require(bipsToReduce > 0, "Slasher.reduceRequestedBipsToSlash: bipsToReduce must be negative");
```
8sunyuan, line 66: Similar to above, also not needed

src/contracts/core/OperatorSetManager.sol
```diff
@@ -0,0 +1,208 @@
...
+                epoch = _lockedMagnitudeUpdates[operator][strategy][i - 1].startEpoch;
+                break;
+            }
+        }
+
+        MagnitudeUpdate[] storage operatorSetMagnitudeUpdates = _operatorSetMagnitudeUpdates[operator][strategy][_hashOperatorSet(operatorSet)];
+        uint192 operatorSetMagnitude = 0;
+
+        // iterate from the latest operator set magnitude update to the oldest
+        for (uint256 i = operatorSetMagnitudeUpdates.length; i > 0; i--) {
```
8sunyuan, line 177: I feel like its possible to combine these two for loops since 
`operatorSetMagnitudeUpdates.length <= totalMagnitudeUpdates[operator][strategy]`

src/contracts/core/OperatorSetManager.sol
```diff
@@ -0,0 +1,208 @@
...
+        uint192 totalMagnitude = 0;
+        // iterate through total magnitude updates to get the total magnitude
+        for (uint256 i = _totalMagnitudeUpdates[operator][strategy].length; i > 0; i--) {
+            if (_totalMagnitudeUpdates[operator][strategy][i - 1].epoch <= epoch) {
+                totalMagnitude = _totalMagnitudeUpdates[operator][strategy][i - 1].totalMagnitude;
+                break;
+            }
+        }
+
+        return uint16(operatorSetMagnitude * SlashingAccountingUtils.BIPS_FACTOR / totalMagnitude);
```
8sunyuan, line 193: should be noted that if there are totalMagnitude updates then this view function will revert 
since `totalMagnitude = 0`

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+        address operator,
+        bytes4 operatorSetID,
+        IStrategy[] memory strategies,
+        uint32 bipsToIncrease
+    ) external {
+        require(bipsToIncrease > 0, "Slasher.increaseRequestedBipsToSlash: bipsToIncrease must be positive");
+        require(
+            bipsToIncrease < SlashingAccountingUtils.BIPS_FACTOR,
+            "Slasher.increaseRequestedBipsToSlash: bipsToIncrease must be less than BIPS_FACTOR"
+        );
```
ChaoticWalrus, line 38: should we allow equality? seems like probably?

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+     *
+     * @param operator the operator to slash
+     * @param strategies the list of strategies to execute slashing for
+     * @param epoch the epoch in which the slashing requests to execute were made
+     */
+    function executeSlashing(address operator, IStrategy[] memory strategies, uint32 epoch) external {
+        require(
+            EpochUtils.currentEpoch() > EpochUtils.getMinExecutionEpochFromRequestEpoch(epoch),
+            "Slasher.executeSlashing: current epoch must be greater than the minimum execution epoch"
+        );
```
ChaoticWalrus, line 85: strict inequality seems incorrect here? shouldn't equality be OK?

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+            slashingRequestIds[operator][strategy].lastExecutedSlashingRequestId = slashingRequest.id;
 
-    function resetFrozenStatus(address[] calldata) external {}
+            // truncate to BIPS_FACTOR_SQUARED if it exceeds it
+            if (slashingRequest.slashingRate > SlashingAccountingUtils.BIPS_FACTOR_SQUARED) {
+                slashingRequest.slashingRate = SlashingAccountingUtils.BIPS_FACTOR_SQUARED;
+            } else if (slashingRequest.slashingRate == 0) {
+                // if the slashing rate is 0 due to cancellation, skip the slashing
+                continue;
+            }
```
ChaoticWalrus, line 102: I think we should revert instead of continuing in this case?
there is no race condition; is there a reason to prefer this behavior?

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+        }
+    }
 
-    function latestUpdateBlock(address, address) external view returns (uint32) {}
+    // includes pending but unexecuted slashings
+    function pendingShareScalingFactor(address operator, IStrategy strategy) public view returns (uint64) {
+        uint64 pendingScalingFactor = SlashingAccountingUtils.findNewScalingFactor({
+            scalingFactorBefore: _shareScalingFactor[operator][strategy],
+            // TODO: this particular lookup could potentially get more complex? e.g. also including the trailing epoch's pending amounts
+            rateToSlash: slashingRequests[operator][strategy][EpochUtils.currentEpoch()].slashingRate
```
ChaoticWalrus, line 257: seems this should indeed be more complex -- we should presumably look up at least the previous epoch's value as well?

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+        require(bipsToModify != 0, "Slasher._modifyRequestedBipsToSlash: cannot modify slashing by 0");
 
-    function middlewareTimesLength(address) external view returns (uint256) {}
+        IOperatorSetManager.OperatorSet memory operatorSet =
+            IOperatorSetManager.OperatorSet({avs: msg.sender, id: operatorSetID});
+        bytes32 operatorSetHash = _hashOperatorSet(operatorSet);
+        for (uint256 i = 0; i < strategies.length; ++i) {
+            IStrategy strategy = strategies[i];
+            uint32 requestedSlashedBipsBefore = requestedSlashedBips[operator][strategy][epoch][operatorSetHash];
+            int32 requestedSlashedBipsAfter = int32(requestedSlashedBipsBefore) + bipsToModify;
```
ChaoticWalrus, line 309: BUG: this is *not* dealing correctly with the case of `requestedSlashedBipsAfter` exceeding 10,000!
see similar comment on the closed PR here https://github.com/Layr-Labs/eigenlayer-contracts/pull/543#discussion_r1651198852

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+        emit RequestedBipsToSlashModified(epoch, operator, operatorSet, strategies, bipsToModify);
+    }
 
-    function operatorWhitelistedContractsLinkedListEntry(
-        address,
-        address
-    ) external view returns (bool, uint256, uint256) {}
+    /// @notice Returns the epoch for lookup in the slashing request history and whether
+    /// there were any slashing requests that happened before or at the given epoch
+    function _getLookupEpoch(address operator, IStrategy strategy, uint32 epoch) internal view returns (uint32, bool) {
```
ChaoticWalrus, line 341: IMO this is abysmal naming. in particular this is looking up the most recent epoch where slashing occurred for the (operator, strategy) tuple, which is at or prior to the `epoch` -- *and* returning false if no such value is found

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+    /// @dev slashingRequests[operator][strategy][epochForLookup] must be set, use _getLookupEpoch
+    function _canWithdraw(address operator, IStrategy strategy, uint32 epoch) internal view returns (bool) {
+        return slashingRequests[operator][strategy][epoch].id
+            <= slashingRequestIds[operator][strategy].lastExecutedSlashingRequestId;
+    }
 
+    function _hashOperatorSet(IOperatorSetManager.OperatorSet memory operatorSet) internal pure returns (bytes32) {
+        IOperatorSetManager.OperatorSet[] memory operatorSetArray = new IOperatorSetManager.OperatorSet[](1);
+        operatorSetArray[0] = operatorSet;
+        return keccak256(abi.encode(operatorSetArray));
```
ChaoticWalrus, line 366: is there a reason to hash a length-1 array of the struct instead of just hashing the struct itself?

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+     */
+    function canWithdraw(address operator, IStrategy strategy, uint32 epoch) public view returns (bool) {
+        bool __canWithdraw = true;
+        (uint32 lookupEpoch, bool found) = _getLookupEpoch(operator, strategy, epoch);
+        if (found) {
+            // if the slashing request is not executed, the operator cannot withdraw
+            __canWithdraw = _canWithdraw(operator, strategy, lookupEpoch);
+        }
+        return __canWithdraw;
+    }
```
ChaoticWalrus, line 235: do you maybe just want to make this function call `getWithdrawabilityAndScalingFactorAtEpoch` and then only return the boolean value?
then we can also eliminate the abstraction of `_canWithdraw`...

src/contracts/libraries/SlashingAccountingUtils.sol
```diff
@@ -0,0 +1,58 @@
...
+    function normalize(int256 nonNormalizedShares, uint64 scalingFactor) internal pure returns (int256) {
+        return (nonNormalizedShares * int256(uint256(SHARE_CONVERSION_SCALE))) / int256(uint256(scalingFactor));
+    }
+
+    // TODO: consider possible loss of precision and its consequences
+    // @dev note that rateToSlash is in parts per BIPS_FACTOR_SQUARED, i.e. in parts per 1e8
+    function findNewScalingFactor(uint64 scalingFactorBefore, uint64 rateToSlash) internal pure returns (uint64) {
+        require(rateToSlash != 0, "cannot slash for 0%");
+        require(rateToSlash <= BIPS_FACTOR_SQUARED, "cannot slash more than 100% at once");
+        uint64 scalingFactorAfter;
```
ChaoticWalrus, line 48: wait, since when are scaling factors only uint64's ? 
this change makes `MAX_SCALING_FACTOR` incorrect!
this also makes overflow of scaling factors possible, and drastically reduces the repeated slashing events to get to the (appropriate) `MAX_SCALING_FACTOR`.
@gpsanant I am assuming you made this change -- we should discuss!

src/contracts/core/SlasherStorage.sol
```diff
@@ -0,0 +1,63 @@
...
+
+    /// @notice Mapping: Operator => Strategy => epoch => SlashingRequest
+    mapping(address => mapping(IStrategy => mapping(uint32 => SlashingRequest))) public slashingRequests;
+
+    /**
+     * @notice Mapping: operator => strategy => share scalingFactor,
+     * stored in the "SHARE_CONVERSION_SCALE", i.e. scalingFactor = 2 * SHARE_CONVERSION_SCALE indicates a scalingFactor of "2".
+     * Note that a value of zero is treated as one, since this means that the operator has never been slashed
+     */
+    mapping(address => mapping(IStrategy => uint64)) internal _shareScalingFactor;
```
ChaoticWalrus, line 40: making these uint64's is a huge mistake. we should correct this.

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+                slashingRequestIds[operator][strategy].lastCreatedSlashingRequestId = requestId;
+                slashingRequest.id = requestId;
+            }
 
-    function getMiddlewareTimesIndexServeUntilBlock(address, uint32) external view returns (uint32) {}
+            // pending slashing rate is modified by the bips to modify * the slashable bips
+            // when divided by BIPS_FACTOR**2, this will give the actual proportion of the
+            // stake is being modified. this is done for accuracy when slashable bips are small
+            slashingRequest.slashingRate = uint64(
+                int64(slashingRequest.slashingRate)
```
8sunyuan, line 329: order of operations looks incorrect here. Need to add bipsToModify to slashingRate first(add parenthesis) before multiplying with the operator's slashable bips/magnitude

8sunyuan: Nvm the slashingRate is parts per hundred million (BIPS_FACTOR_SQUARED)
The slashingRate is adding the product of the operator's configured magnitude and the requested bips to slash by the AVS and adding this to the existing slashingRate. Makes sense now

src/contracts/core/Slasher.sol
8sunyuan: Should be >= 
`EpochUtils.currentEpoch() >= EpochUtils.getMinExecutionEpochFromRequestEpoch(epoch)`
the right term returns epoch + 2 so the current epoch has to be at least epoch + 3 which is incorrect.

wadealexc (changes requested):
Request - this needs to branch off of whatever operator sets PR we currently have, so it's clear what we're changing from there to here

src/contracts/core/SlasherStorage.sol
```diff
@@ -0,0 +1,63 @@
...
+    struct SlashingRequest {
+        uint32 id; // an incrementing ID for each slashing request
+        uint64 slashingRate; // This is parts per (BIPS_FACTOR**2), i.e. parts per 1e8, pphm = parts per hundred million, to slash upon execution
+        uint64 scalingFactor; // the scaling factor to apply to the operator's shares. this is only set upon execution
+    }
+
+    struct SlashingRequestIds {
+        uint32 lastCreatedSlashingRequestId; // the last slashing request ID that was created
+        uint32 lastExecutedSlashingRequestId; // the last slashing request ID that was executed
+    }
```
wadealexc, line 27: Names are long as heck. How about:
- `lastCreatedId`
- `lastExecutedId`

src/contracts/libraries/EpochUtils.sol
```diff
@@ -0,0 +1,37 @@
...
+        return getEpochFromTimestamp(block.timestamp);
+    }
+
+    function getNextSlashingParameterEffectEpoch() internal view returns (uint32) {
+        return currentEpoch() + 3;
+    }
+
+    function getMinExecutionEpochFromRequestEpoch(uint32 requestEpoch) internal pure returns (uint32) {
+        return requestEpoch + 2;
+    }
```
wadealexc, line 32: name is way too long

`getMinExecutionEpoch`

src/contracts/core/DelegationManager.sol
```diff
@@ -535,137 +572,151 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
     }
 
-    // @notice Increases `operator`s delegated shares in `strategy` by `shares` and emits an `OperatorSharesIncreased` event
-    function _increaseOperatorShares(address operator, address staker, IStrategy strategy, uint256 shares) internal {
-        operatorShares[operator][strategy] += shares;
-        emit OperatorSharesIncreased(operator, staker, strategy, shares);
+    // @notice Increases `operator`s delegated nonNormalizedShares in `strategy` by `nonNormalizedShares` and emits an `OperatorSharesIncreased` event
+    function _increaseOperatorShares(address operator, address staker, IStrategy strategy, uint256 nonNormalizedShares) internal {
+        nonNormalizedOperatorShares[operator][strategy] += nonNormalizedShares;
+        emit OperatorSharesIncreased(operator, staker, strategy, nonNormalizedShares);
```
wadealexc, line 666: Generally, I am not a fan of changing the existing "shares" naming to `nonNormalizedShares`.

IMO we keep the existing "shares" concept as-is, and use `normalizedShares` after we do whatever conversion normalization entails. Changing these interfaces and variable names though both:
* breaks more things
* requires more thinking

src/contracts/core/DelegationManager.sol
```diff
@@ -535,137 +572,151 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
                 /**
-                 * This call will revert if it would reduce the Staker's virtual beacon chain ETH shares below zero.
+                 * This call will revert if it would reduce the Staker's virtual beacon chain ETH nonNormalizedShares below zero.
                  * This behavior prevents a Staker from queuing a withdrawal which improperly removes excessive
-                 * shares from the operator to whom the staker is delegated.
+                 * nonNormalizedShares from the operator to whom the staker is delegated.
                  * It will also revert if the share amount being withdrawn is not a whole Gwei amount.
                  */
-                eigenPodManager.removeShares(staker, shares[i]);
+                eigenPodManager.removeShares(staker, nonNormalizedShares[i]);
```
wadealexc, line 712: Continuing my prev comment, this naming thing is also gonna be inconsistent. See here -- why is this not `epm.removeNonNormalizedShares`?

How do we know `removeShares` wants a non-normalized parameter?

yuck. let's please keep the existing `shares` concept as-is.

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+    /**
+     * @notice Called by an AVS to increase its own slashing request for a given
+     * operator set and operator in the current epoch
+     *
+     * @param operator the operator that the calling AVS is to increase the bips they want to slash
+     * @param operatorSetID the id of the operator set the AVS is increasing their slashing for
+     * @param strategies the list of strategies slashing requested is being modified for
+     * @param bipsToIncrease the basis points slashing to modify for given strategies
+     */
+    function increaseRequestedBipsToSlash(
```
wadealexc, line 28: Not convinced we need an "increase" method. IMO we should have:
* `requestSlash`
* `reduceSlash`
* `executeSlash`

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+     * @return whether the withdrawal can be executed
+     */
+    function canWithdraw(address operator, IStrategy strategy, uint32 epoch) public view returns (bool) {
+        bool __canWithdraw = true;
+        (uint32 lookupEpoch, bool found) = _getLookupEpoch(operator, strategy, epoch);
+        if (found) {
+            // if the slashing request is not executed, the operator cannot withdraw
+            __canWithdraw = _canWithdraw(operator, strategy, lookupEpoch);
+        }
+        return __canWithdraw;
```
wadealexc, line 234: This can be refactored to just use `canWithdraw` logic directly

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+     * @param strategy the strategy the withdrawal is from
+     * @param epoch the last epoch the withdrawal was slashable until
+     * @return whether the withdrawal can be executed
+     * @return whether there was a slashing request for the given operator and strategy at the given epoch
+     */
+    function getWithdrawabilityAndScalingFactorAtEpoch(
+        address operator,
+        IStrategy strategy,
+        uint32 epoch
+    ) external view returns (bool, uint64) {
```
wadealexc, line 209: Mega wordy. Suggestions:
- `getWithdrawability`
- `getScalingFactor`

Don't need to include the full depth and breadth of function in the name. Good documentation will make these pop out -- currently you're forcing this to be called as follows:

```solidity
(bool canWithdraw, uint64 scalingFactor) =
  slasher.getWithdrawabilityAndScalingFactorAtEpoch(operator, strategy, epoch);
```

Naming the return variables is already going to make the usage apparent. It could be super simple instead:

```solidity
(bool canWithdraw, uint64 scalingFactor) = 
  slasher.getWithdrawability(operator, strategy, epoch);
```

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+        address operator,
+        IStrategy strategy,
+        uint32 epoch
+    ) external view returns (uint32) {
+        uint64 totalSlashingRate = slashingRequests[operator][strategy][epoch].slashingRate;
+        if (totalSlashingRate >= SlashingAccountingUtils.BIPS_FACTOR_SQUARED) {
+            totalSlashingRate = SlashingAccountingUtils.BIPS_FACTOR_SQUARED;
+        }
+        return uint32(totalSlashingRate);
+    }
```
wadealexc, line 195: Can we simplify this to like a oneliner that uses `Math.max`?

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+        IStrategy strategy,
+        IOperatorSetManager.OperatorSet calldata operatorSet,
+        uint32 epoch
+    ) external view returns (uint32) {
+        uint32 requestedSlashingRate = getRequestedSlashingRate(operator, strategy, operatorSet, epoch);
+        if (requestedSlashingRate >= SlashingAccountingUtils.BIPS_FACTOR_SQUARED) {
+            requestedSlashingRate = uint32(SlashingAccountingUtils.BIPS_FACTOR_SQUARED);
+        }
+        return requestedSlashingRate;
+    }
```
wadealexc, line 170: Can we simplify this to like a oneliner that uses `Math.max`?

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+        bytes32 operatorSetHash = _hashOperatorSet(operatorSet);
+        for (uint256 i = 0; i < strategies.length; ++i) {
+            IStrategy strategy = strategies[i];
+            uint32 requestedSlashedBipsBefore = requestedSlashedBips[operator][strategy][epoch][operatorSetHash];
+            int32 requestedSlashedBipsAfter = int32(requestedSlashedBipsBefore) + bipsToModify;
+            // don't allow underflow
+            if (requestedSlashedBipsAfter < 0) {
+                bipsToModify = -int32(requestedSlashedBipsBefore);
+                requestedSlashedBipsAfter = 0;
+            }
```
wadealexc, line 314: Why are we doing this rather than reverting?

src/contracts/core/DelegationManager.sol
```diff
@@ -535,137 +572,151 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+                        operator: currentOperator,
+                        // the 'staker' here is the address receiving new shares
+                        staker: withdrawal.staker,
+                        strategy: withdrawal.strategies[i],
+                        nonNormalizedShares: nonNormalizedShares
+                    });
                 }
-                unchecked { ++i; }
             }
+            unchecked { ++i; }
```
wadealexc, line 657: Let's put this into the `for` loop itself rather than incrementing way down here.

We've been bitten by this before - and this method is getting massive now...

src/contracts/core/DelegationManager.sol
```diff
@@ -535,137 +572,151 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
             );
         }
 
+        uint32 epochForEndOfSlashability = EpochUtils.getEndOfSlashabilityEpoch(withdrawalCreationEpoch(withdrawalRoot));
+        require(EpochUtils.currentEpoch() > epochForEndOfSlashability,
+            "DelegationManager._completeQueuedWithdrawal: withdrawal is still slashable");
```
wadealexc, line 577: I feel like this shouldn't apply for stakers that weren't delegated to an operator when they started withdrawing.

src/contracts/core/DelegationManager.sol
```diff
@@ -535,137 +572,151 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
             );
         }
 
+        uint32 epochForEndOfSlashability = EpochUtils.getEndOfSlashabilityEpoch(withdrawalCreationEpoch(withdrawalRoot));
```
wadealexc, line 575: Why do we need to store the creation epoch? We have `withdrawal.startBlock`; can't we calculate it from that?

src/contracts/core/DelegationManager.sol
```diff
@@ -535,137 +572,151 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
-                    withdrawer: msg.sender,
-                    strategy: withdrawal.strategies[i],
-                    shares: withdrawal.shares[i],
-                    token: tokens[i]
+                require(canWithdraw, "DelegationManager._completeQueuedWithdrawal: withdrawal is still slashable");
+
+                shares = SlashingAccountingUtils.normalize({
+                    nonNormalizedShares: withdrawal.shares[i], 
+                    scalingFactor: scalingFactor
                 });
```
wadealexc, line 604: This is fucking awful

After looking around, it looks like `normalize` is literally only called AFTER calls to the Slasher.

How about we have the Slashed do this normalization instead, and have it return already-normalized data?

src/contracts/core/Slasher.sol
wadealexc: See #609

src/contracts/core/DelegationManager.sol
wadealexc: See #609

src/contracts/core/DelegationManager.sol
wadealexc: See #609

src/contracts/core/DelegationManager.sol
wadealexc: See #609

src/contracts/core/StrategyManager.sol
```diff
@@ -28,6 +30,25 @@ contract StrategyManager is
...
+    // includes the effect of all unexecuted but pending slashings
+    function stakerStrategySharesIncludingPendingSlashings(address staker, IStrategy strategy) external view returns (uint256) {
+        address operator = delegation.delegatedTo(staker);
+        uint64 scalingFactor = slasher.pendingShareScalingFactor(operator, strategy);
+        return SlashingAccountingUtils.normalize({
+            nonNormalizedShares: nonNormalizedStakerStrategyShares[staker][strategy],
+            scalingFactor: scalingFactor
+        });
+    }
+
```
wadealexc, line 51: Awful naming. Suggestions:
* `normalizedStakerShares`
* `pendingNormalizedStakerShares`

Other suggestion is to just have these methods live in the slasher.

src/contracts/core/StrategyManager.sol
```diff
@@ -269,33 +290,33 @@ contract StrategyManager is
...
+    function _addShares(address staker, IERC20 token, IStrategy strategy, uint256 nonNormalizedShares) internal {
         // sanity checks on inputs
         require(staker != address(0), "StrategyManager._addShares: staker cannot be zero address");
-        require(shares != 0, "StrategyManager._addShares: shares should not be zero!");
+        require(nonNormalizedShares != 0, "StrategyManager._addShares: nonNormalizedShares should not be zero!");
 
-        // if they dont have existing shares of this strategy, add it to their strats
-        if (stakerStrategyShares[staker][strategy] == 0) {
+        // if they dont have existing nonNormalizedShares of this strategy, add it to their strats
+        if (nonNormalizedStakerStrategyShares[staker][strategy] == 0) {
```
wadealexc, line 308: There is so much diff here that doesn't contribute to the implementation in any way.

Let's keep the original variable names here.

src/contracts/core/Slasher.sol
```diff
@@ -1,102 +1,368 @@
...
+            uint64 scalingFactorBefore = shareScalingFactor(operator, strategy);
+            uint64 scalingFactorAfter = SlashingAccountingUtils.findNewScalingFactor({
+                scalingFactorBefore: scalingFactorBefore,
+                rateToSlash: slashingRequest.slashingRate
+            });
+            // update storage to reflect the slashing
+            slashedEpochHistory[operator][strategy].push(epoch);
+            _shareScalingFactor[operator][strategy] = scalingFactorAfter;
+            // TODO DONE?: note that this behavior risks off-by-one errors. it needs to be clearly defined precisely how the historical storage is supposed to work
+            slashingRequest.scalingFactor = scalingFactorAfter;
```
wadealexc, line 113: This doesn't update anything, storage-wise.

But - do we even want to do an update here? It seems like we're done with this `slashingRequest`

PR #587, opened 2024-06-17
Feat: make distributionRoots internal

ypatil12:
Make `distributionRoots` internal so we can expose public getters. 

Prev PR had duplicated bytecode. h/t @8sunyuan for the suggestion :)


seanmcgary (approved):

8sunyuan (approved):
TY for this, just need updated bindings though and lgtm

PR #597, opened 2024-06-24
test: certora spec for RewardsCoordinator

8sunyuan:
Included spec as a github workflow with `certora-prover-conf.yml`


certora/specs/core/RewardsCoordinator.spec
```diff
@@ -0,0 +1,113 @@
...
+
+// Given a processClaim passes, second processClaim reverts
+// Assumptions
+// - checkClaim(claim) == true and does not revert
+// - claim has non-empty tokenLeaves
+/// status: pass
+rule claimWithduplicateTokenLeafs(env e, IRewardsCoordinator.RewardsMerkleClaim claim, address recipient) {
+    require claim.tokenLeaves.length == 1;
+    
+    checkClaim(e, claim);
```
ChaoticWalrus, line 36: what's the point of this call?

certora/conf/rewardsCoordinator.conf
```diff
@@ -0,0 +1,28 @@
...
+    "packages": [
+        "@openzeppelin=lib/openzeppelin-contracts",
+        "@openzeppelin-upgrades=lib/openzeppelin-contracts-upgradeable"
+    ],
+    "link": [],
+    "optimistic_fallback": true,
+    "optimistic_loop": true,
+    "optimistic_hashing": true,
+    "rule_sanity": "basic",
+    "exclude_rule": ["checkClaimNeverFalse"],
```
ChaoticWalrus, line 19: so is this rule just failing to pass right now then?

certora/specs/core/RewardsCoordinator.spec
```diff
@@ -0,0 +1,113 @@
...
+        e.msg.sender == claim.earnerLeaf.earner;
+
+    processClaim@withrevert(e, claim, recipient);
+
+    assert !lastReverted => canWork;
+}
+
+/// checkClaim either returns True or reverts otherwise
+invariant checkClaimNeverFalse(env e, IRewardsCoordinator.RewardsMerkleClaim claim)
+    !checkClaim(e, claim);
```
ChaoticWalrus, line 113: is this the spec we landed on? seems odd that a `checkX` function would ever revert instead of returning false, unless the input was severely malformed :/

certora/specs/core/RewardsCoordinator.spec
```diff
@@ -0,0 +1,113 @@
...
+    // Call checkClaim to ensure it does not revert in rule since it never returns false
+    checkClaim(e, claim);
+
+    bool canWork = claimerFor(e, claim.earnerLeaf.earner) != 0 ? 
+        e.msg.sender == claimerFor(e, claim.earnerLeaf.earner) :
+        e.msg.sender == claim.earnerLeaf.earner;
+
+    processClaim@withrevert(e, claim, recipient);
+
+    assert !lastReverted => canWork;
```
ChaoticWalrus, line 108: can you assert equality here? that would be stronger if possible; I'm not sure it would work with the summarization of `verifyInclusionKeccak` though...

ChaoticWalrus (approved):
LGTM. I have a few minor questions but nothing that I think is particularly worth blocking on.

certora/harnesses/RewardsCoordinatorHarness.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.12;
+
+import "../../src/contracts/core/RewardsCoordinator.sol";
+
+contract RewardsCoordinatorHarness is RewardsCoordinator {
```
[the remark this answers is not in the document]
8sunyuan: This was made in case there were any internal functions needed to be made. Looks like we can just remove though

certora/specs/core/RewardsCoordinator.spec
8sunyuan: `checkClaim` calls the internal functions `_checkClaim` which will revert on any invalid inputs and invalid merkle proofs. We decided on leaving that internal function as is rather than trying to refactor the merkle library and changing all the requires to if checks that return false. 

I think the `checkClaim` view function could probably be refactored though to either return True or False by doing a STATICCALL to `_checkClaim`

certora/specs/core/RewardsCoordinator.spec
8sunyuan: Its to ensure that the claim is valid.
since checkClaim doesn't have @withrevert any successful paths of this rule will include checkClaim not reverting. Probably more clear by just having `require checkClaim`

ChaoticWalrus: makes sense, but adding 'require' syntax would definitely be clearer

certora/conf/rewardsCoordinator.conf
teryanarmen: No the rule is called in paymentCoordinatorWithoutSanity.conf. The reason for the separate conf is that this rule fails the sanity check since checkClaim reverts instead of returning false.

certora/specs/core/RewardsCoordinator.spec
teryanarmen: The summarization may be an issue but all other revert reasons will also be problematic. First its things like the contract being paused, in reentrant mode, and >0 msg.value being sent, then compiler checks which are harder to understand.

certora/specs/core/RewardsCoordinator.spec
```diff
@@ -110,4 +111,116 @@ rule claimCorrectness(env e, IRewardsCoordinator.RewardsMerkleClaim claim, addre
...
 invariant checkClaimNeverFalse(env e, IRewardsCoordinator.RewardsMerkleClaim claim)
-    !checkClaim(e, claim);
\ No newline at end of file
+    !checkClaim(e, claim);
+
+rule rewardsSubmissionIntegrity(env e) 
+    {
+    IRewardsCoordinator.RewardsSubmission[] rewardsSubmission1;
+    IRewardsCoordinator.RewardsSubmission[] rewardsSubmission2;
+    IRewardsCoordinator.RewardsSubmission[] rewardsSubmissions;
```
ChaoticWalrus, line 120: this rule is checking batch equivalency, right?

gpsanant:
can we close or merge this? @8sunyuan

PR #599, opened 2024-06-25
feat: disable pending roots

8sunyuan:
Add the ability for the `RewardsUpdater` to disable their own posted roots when they are pending and not yet activated(7 day delay from when a root is submitted).

TODO
- [X] Add and fix tests
- [X] Update /docs


src/contracts/core/RewardsCoordinator.sol
```diff
@@ -488,8 +505,13 @@ contract RewardsCoordinator is
...
     }
 
+    /// @notice loop through the distribution roots from reverse and get latest root that is not disabled
     function getCurrentDistributionRoot() external view returns (DistributionRoot memory) {
-        return _distributionRoots[_distributionRoots.length - 1];
+        for (uint256 i = _distributionRoots.length; i > 0; i--) {
+            if (!_distributionRoots[i - 1].disabled) {
+                return _distributionRoots[i - 1];
+            }
+        }
```
ChaoticWalrus, line 514: Fairly unrelated to this PR, but is the intention here that this is the latest *active* root, or the latest root in general (which may be waiting to become active)?
Seems like this is written to be the latter; I think that's an OK choice but we should be explicit about the intention.

ChaoticWalrus (commented):
LGTM overall. Not giving explicit approval because I want to make sure we align offline first.

src/contracts/core/RewardsCoordinator.sol
8sunyuan: Yeah wasn't actually sure on the desired behavior from the view funcs here. Perhaps multiple interfaces could be added including one that finds the latest root that is both active AND not disabled

ChaoticWalrus: I'd be fine with both, personally. Could be good to ask a question or two (on Slack) + consider naming for these function(s) to maximize clarity. I don't think this PR needs to be the place to make this decision though.

ChaoticWalrus (approved):
Approving now. AFAICT there's no opposition to this slight tweak.

wadealexc (approved):
LGTM

PR #601, opened 2024-06-26
Stake locking

gpsanant:
giga poc of stake locking

Essentially, this adds a list of the ranges of stake locking with elements of the form:
```solidity
struct StakeLock {
        uint32 startEpoch; // inclusive
        uint32 endEpoch; // inclusive
}
```

Then, any slashing requests lookup what lock they're in or create a new one and update the `lockedUntilEpoch` to the next epoch and `lockedMagnitude` according to the rules described in the doc to the expected magnitude for the next epoch.

```solidity
struct TotalMagnitudeUpdate {
        uint32 epoch;
        uint64 nonSlashableMagnitude;
        uint64 totalAllocatedMagnitude;
        uint32 lockedUntilEpoch;
        uint64 lockedTotalAllocatedMagnitude;
}

struct MagnitudeUpdate {
        uint32 epoch; // the epoch the magnitude update is for
        uint64 magnitude; // the magnitude for the epoch
        uint32 lockedUntilEpoch; // the latest epoch at which this update's magnitude should be read
        uint64 lockedMagnitude; // the magnitude the operator is locked to
}
```
There exists the gross accounting issue of how `lockedMagnitude`s become the `magnitude`s when the next epoch comes around. The conditions on which this should happen is when the current epoch is the `lockedUntilEpoch`. If this condition is met, we copy over the `lockedMagnitude` into the `magnitude` field. Otherwise, we edit `getSlashableBips` to make this check and calculate the bips using the correct magnitude.

notes: 
- not hooked up to slasher yet
- increases not implemented
- no increases on cancellation yet


src/contracts/core/AVSDirectory.sol
```diff
@@ -336,12 +338,322 @@ contract AVSDirectory is
...
+        require(
+            msg.sender == address(slasher), "OperatorSetManager.decreaseAndLockMagnitude: Caller is not the slasher"
+        );
+        uint32 currentEpoch = EpochUtils.currentEpoch();
+
+        for (uint256 i = 0; i < strategies.length; i++) {
+            uint32 startEpoch = currentEpoch;
+            // lock magnitude updates for the operator, inplace if possible
+            {
+                uint256 lockedMagnitudeUpdatesLength = _lockedMagnitudeUpdates[operator][strategies[i]].length;
```
8sunyuan, line 492: updated endEpochs need to be currentEpoch + 1

gpsanant:
good night sweet child

PR #602, opened 2024-06-26
chore: cleanup and readability

8sunyuan:
Mostly some cleanup to  `_modifyRequestedBipsToSlash` to remove the use of casting and using signed ints


src/contracts/core/Slasher.sol
```diff
@@ -352,6 +362,15 @@ contract Slasher is SlasherStorage {
         return (epochForLookup, found);
     }
 
+    /// Consider adding to SlashingAccountingUtils
+    function _addBips(uint32 a, uint32 b) private pure returns (uint32) {
+        return (a + b > SlashingAccountingUtils.BIPS_FACTOR) ? uint32(SlashingAccountingUtils.BIPS_FACTOR) : a + b;
```
gpsanant, line 367: we shouldn't cap here, we cap at the execution phase

8sunyuan: But when you take this value and multiply it with the operators magnitude it could exceed the configured magnitude for the operator allowing the AVS to slash more than should be actually allowed.  @ChaoticWalrus had a comment on this on the slashing PR.

PR #607, opened 2024-07-01
feat: operator sets

ypatil12:
Pending Performance Based Rewards


src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -16,31 +14,55 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
...
-    /// @dev Salt is used in the `registerOperatorToAVS` function.
+    /// @dev Salt is used in the `registerOperatorToAVS` and `registerOperatorToOperatorSet` function.
     mapping(address => mapping(bytes32 => bool)) public operatorSaltIsSpent;
 
-    constructor(IDelegationManager _delegation) {
+    /// @notice Mapping: AVS => whether or not the AVS uses operator set
+    mapping(address => bool) public isOperatorSetAVS;
+
+    /// @notice Mapping: avs => operator => operatorSetID => whether the operator is registered for the operator set
+    mapping(address => mapping(address => mapping(uint32 => bool))) public isOperatorInOperatorSet;
```
wadealexc, line 49: Repetitive/wordy. Suggestions:
* `isInOperatorSet`
* `isMemberOfOperatorSet`

src/contracts/core/AVSDirectory.sol
```diff
@@ -48,21 +56,200 @@ contract AVSDirectory is
...
+     *  @param operatorSignature The signature of the operator on their intent to register.
+     *
+     *  @dev msg.sender is used as the AVS.
+     *  @dev The operator must not have a pending deregistration from the operator set.
+     *  @dev If this is the first operator set in the AVS that the operator is
+     *  registering for, a OperatorAVSRegistrationStatusUpdated event is emitted with
+     *  a REGISTERED status.
      */
-    function registerOperatorToAVS(
+    function registerOperatorToOperatorSets(
```
wadealexc, line 123: Wordy. I know this fits with `registerOperatorToAVS` below, but I'm not a fan of the word "Operator" twice in the function name - makes parsing harder.

Suggestions: 
* `registerToOperatorSets`
* `registerOperatorToSets`
* `addOperatorToSets`
* `addToOperatorSets`

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -16,31 +14,55 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
...
     /// @notice Mapping: operator => 32-byte salt => whether or not the salt has already been used by the operator.
-    /// @dev Salt is used in the `registerOperatorToAVS` function.
+    /// @dev Salt is used in the `registerOperatorToAVS` and `registerOperatorToOperatorSet` function.
     mapping(address => mapping(bytes32 => bool)) public operatorSaltIsSpent;
 
-    constructor(IDelegationManager _delegation) {
+    /// @notice Mapping: AVS => whether or not the AVS uses operator set
+    mapping(address => bool) public isOperatorSetAVS;
+
+    /// @notice Mapping: avs => operator => operatorSetID => whether the operator is registered for the operator set
```
wadealexc, line 48: Mapping ordering here feels incorrect.

The relationship is that an AVS has operator sets, and operator sets within an AVS have operators.

I would expect the mapping to be `avs -> operator set -> operator`

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -16,31 +14,55 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
...
+    mapping(address => mapping(address => mapping(uint32 => bool))) public isOperatorInOperatorSet;
+
+    /// @notice Mapping: avs => operator => number of operator sets the operator is registered for the AVS
+    mapping(address => mapping(address => uint256)) public operatorAVSOperatorSetCount;
```
wadealexc, line 52: Wordy. Suggestions:
* `operatorSetCountByAVS`
* `numOperatorSetsByAVS`

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -16,31 +14,55 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
...
+
+    /// @notice Mapping: avs = operator => operatorSetId => Whether the given operator set in standby mode or not
```
wadealexc, line 54: Mapping order feels incorrect; see prev comment

src/contracts/core/AVSDirectory.sol
```diff
@@ -48,21 +56,200 @@ contract AVSDirectory is
...
+    /**
+     * @notice Updates the standby parameters for an operator across multiple operator sets.
+     * Allows the AVS to add the operator to a given operator set if they are not already registered.
+     *
+     * @param operator The address of the operator for which the standby parameters are being updated.
+     * @param standbyParams The new standby parameters for the operator.
+     * @param operatorSignature If non-empty, the signature of the operator authorizing the modification.
+     *                  If empty, the `msg.sender` must be the operator.
+     */
+    function updateStandbyParams(
```
wadealexc, line 74: This needs to be deeper in the file -- functions at the top are registration/deregistration. "update" methods should be below those for the easiest reading experience

src/contracts/interfaces/IAVSDirectory.sol
```diff
@@ -2,58 +2,143 @@
...
-        REGISTERED          // Operator registered to AVS
+        UNREGISTERED, // Operator is not registered with the AVS.
+        REGISTERED // Operator is registered with the AVS.
+
+    }
+
+    struct OperatorSet {
+        address avs;
+        uint32 id;
+    }
```
wadealexc, line 18: This struct really isn't being utilized well.

In 99% of places where the actual code deals with operator sets, we aren't using this struct. We only use it in events, and as an additional field in `StandbyParam`

However, even in this latter case, the use of a nested struct in `StandbyParam` makes `updateStandbyParams` really messy.

IMO we should just change `StandbyParam` to have all 3 fields rather than a nested struct.

src/contracts/core/AVSDirectory.sol
```diff
@@ -48,21 +56,200 @@ contract AVSDirectory is
...
+        operatorSaltIsSpent[operator][operatorSignature.salt] = true;
+
+        // Register `operator` if not already registered.
+        if (avsOperatorStatus[msg.sender][operator] != OperatorAVSRegistrationStatus.REGISTERED) {
+            avsOperatorStatus[msg.sender][operator] = OperatorAVSRegistrationStatus.REGISTERED;
+            emit OperatorAVSRegistrationStatusUpdated(operator, msg.sender, OperatorAVSRegistrationStatus.REGISTERED);
+        }
+
+        // Mutate calling AVS to operator set AVS status, preventing further legacy registrations.
+        if (!isOperatorSetAVS[msg.sender]) isOperatorSetAVS[msg.sender] = true;
```
wadealexc, line 168: I'm not a fan of this - I think this opens the door to footgun accidentally making your AVS operator sets only. 

If we're committed to supporting non-operator set AVSs, we need an explicit method to upgrade like `becomeOperatorSetAVS`

wadealexc: I'm not sure how we'd ensure that new AVSs are operator set AVSs by default though....

wadealexc (commented):
One of the usecases is:

> As an AVS, I would like to distinguish between the strategies in each of my operator sets on the UI

I don't see strategies represented super well with these operator sets.

1. There's no clear "create operator set" method (yes, it can be in the middleware contracts, but I'm questioning why we don't have an explicit creation method here)
2. There's no clear binding between an operator set and the strategies it represents

src/contracts/core/AVSDirectory.sol
```diff
@@ -48,21 +56,200 @@ contract AVSDirectory is
...
+        // Assert operator's signature has not expired.
+        require(
+            operatorSignature.expiry >= block.timestamp,
+            "AVSDirectory.registerOperatorToOperatorSets: operator signature expired"
+        );
+        // Assert operator's signature `salt` has not already been spent.
+        require(
+            !operatorSaltIsSpent[operator][operatorSignature.salt],
+            "AVSDirectory.registerOperatorToOperatorSets: salt already spent"
+        );
```
wadealexc, line 142: These checks can be skipped if the operator is on standby, right?

src/contracts/core/AVSDirectoryStorage.sol
stevennevins: `ifMemberOf`

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -16,31 +14,55 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
...
+
+    /// @notice The EIP-712 typehash for the `StandbyParams` struct used by the contract
+    bytes32 public constant OPERATOR_STANDBY_UPDATE =
+        keccak256("OperatorStandbyUpdate(StandbyParam[] standbyParams,bytes32 salt,uint256 expiry)");
+    
     /// @notice The DelegationManager contract for EigenLayer
     IDelegationManager public immutable delegation;
 
+    /// @notice The StrategyManager contract for EigenLayer
+    IStrategyManager public immutable strategyManager;
```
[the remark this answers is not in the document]
0xClandestine: Will remove, I think it was original code from AVSDirectory...

src/contracts/core/AVSDirectoryStorage.sol
0xClandestine: I like `isMemberOf` short and concise, thoughts @wadealexc ?

src/contracts/core/AVSDirectory.sol
0xClandestine: Will think on this one a bit more...

src/contracts/core/AVSDirectoryStorage.sol
0xClandestine: This was also noted by @ypatil12 will fix...

src/contracts/core/AVSDirectoryStorage.sol
0xClandestine: WIll fix...

src/contracts/core/AVSDirectoryStorage.sol
wadealexc: IMO `isMemberOf == isMember` - the "of" adds pretty much nothing.

If we want to do `isMember`, that's fine - but if that's too short, we might want to consider `isMemberOfSet`?

wadealexc: I think I prefer `isMemberOfSet` over `isMember` actually. The latter would work well in library format (`operator.isMember()`), but reading it as a mapping, it's harder to tell what's going on

0xClandestine: I personally like isMemberOf(operator, avs, set) because it suggests an `<operator> is member of <avs> <operator_set>`.

src/contracts/core/AVSDirectoryStorage.sol
0xClandestine: Rethinking on this I feel like opeartor should always be first if the state variable is returning data about an operator.

0xClandestine:
Was able to condense 4 mappings into 2 with the recent storage overhaul, it allows for the original interface if we want to add some getters. 

Saves several SLOADs/SSTOREs throughout `AVSDirectory`.

src/contracts/core/AVSDirectory.sol
```diff
@@ -127,20 +355,59 @@ contract AVSDirectory is
...
-    function updateAVSMetadataURI(string calldata metadataURI) external {
-        emit AVSMetadataURIUpdated(msg.sender, metadataURI);
+
+    /**
+     * @notice Helper function used by migration & registration functions to register an operator to operator sets.
+     * @param avs The AVS that the operator is registering to.
+     * @param operator The operator to register.
+     * @param operatorSetIds The IDs of the operator sets.
+     */
+    function _registerToOperatorSets(address avs, address operator, uint32[] calldata operatorSetIds) internal {
```
8sunyuan, line 369: Nit: could we order params as operator, avs, opSetIds? More consistent with interfaces and I prefer having AVS and operatorSetIds next to eachother as the operatorSet.

src/contracts/core/AVSDirectory.sol
```diff
@@ -52,67 +52,295 @@ contract AVSDirectory is
...
-     * @param operatorSignature The signature, salt, and expiry of the operator's signature.
+     * @notice Called by an AVS to create a list of new operatorSets.
+     *
+     * @param operatorSetIds The IDs of the operator set to initialize.
+     *
+     * @dev msg.sender must be the AVS.
+     * @dev The AVS may create operator sets before it becomes an operator set AVS.
+     */
+    function createOperatorSets(uint32[] calldata operatorSetIds) external {
+        for (uint256 i = 0; i < operatorSetIds.length; ++i) {
```
8sunyuan, line 68: Can we have another pause flag for this function?

src/contracts/core/AVSDirectory.sol
```diff
@@ -52,67 +52,295 @@ contract AVSDirectory is
...
+     *  @param operatorSignature The signature of the operator on their intent to register.
+     *
+     *  @dev msg.sender is used as the AVS.
+     *  @dev The operator must not have a pending deregistration from the operator set.
+     */
+    function registerOperatorToOperatorSets(
+        address operator,
+        uint32[] calldata operatorSetIds,
+        ISignatureUtils.SignatureWithSaltAndExpiry memory operatorSignature
+    ) external override onlyWhenNotPaused(PAUSED_OPERATOR_REGISTER_DEREGISTER_TO_AVS) {
```
8sunyuan, line 143: Should perhaps use a different pause flag for legacy registrations/dereg vs operator set registration/dereg

src/contracts/core/AVSDirectory.sol
```diff
@@ -52,67 +52,295 @@ contract AVSDirectory is
...
+                avsOperatorStatus[msg.sender][operators[i]] == OperatorAVSRegistrationStatus.REGISTERED,
+                "AVSDirectory.migrateOperatorsToOperatorSets: operator already migrated or not a legacy registered operator"
+            );
+
+            // Migrate operator to operator sets.
+            _registerToOperatorSets(msg.sender, operators[i], operatorSetIds[i]);
+
+            // Deregister operator from AVS - this prevents the operator from being migrated again since
+            // the AVS can no longer use the legacy M2 registration path
+            avsOperatorStatus[msg.sender][operators[i]] = OperatorAVSRegistrationStatus.UNREGISTERED;
```
8sunyuan, line 121: Maybe should note this storage slot is to be deprecated in a future release once AVSs have migrated

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -42,5 +55,5 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[47] private __gap;
+    uint256[43] private __gap;
```
8sunyuan, line 58: I'm seeing 3 new mappings but we decremented gap by 4??

src/contracts/core/AVSDirectory.sol
ypatil12: Updated

src/contracts/core/AVSDirectory.sol
ypatil12: Not sure why we would ever pause this?

src/contracts/core/AVSDirectory.sol
ypatil12: Added comment

src/contracts/core/AVSDirectoryStorage.sol
ypatil12: Fixed

ypatil12:
Combined with slashing: https://github.com/Layr-Labs/eigenlayer-contracts/pull/679

PR #608, opened 2024-07-01
[Draft]: Operator Set Initialization

ypatil12:
Adds storage for existence of operator sets within an AVS. Useful for rewards where we check rewards are being made to operator sets that exist

Storage is getting quite verbose. Will pare down in separate PR


stevennevins:
What do you think about also adding an entry point for flipping the boolean for `isOperatorSetAVS` to true

PR #610, opened 2024-07-02
feat: clean up slashing withdrawal completion

wadealexc:
* addresses a handful of my comments on the main slashing PR
* cleans up withdrawal completion logic to use fewer external calls and be less confusing
* TODO: EpochUtils.slashingEndEpoch calculation

### DIFF
--- a/src/contracts/core/Slasher.sol
+++ b/src/contracts/core/Slasher.sol
@@ -10,6 +10,8 @@ import "../libraries/SlashingAccountingUtils.sol";
  * @notice Slasher
  */
 contract Slasher is SlasherStorage {
+    using EpochUtils for *;
+
     constructor(
         IStrategyManager _strategyManager,
         IDelegationManager _delegationManager,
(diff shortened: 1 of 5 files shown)


src/contracts/core/Slasher.sol
```diff
@@ -194,27 +196,39 @@ contract Slasher is SlasherStorage {
...
+            uint256 amount = withdrawal.shares[i];
+
+            (uint32 lookupEpoch, bool found) = _getLookupEpoch(prevOperator, strategy, slashingEndEpoch);
+
+            // Can't withdraw or redeposit here
+            // TODO - may want to add amounts[i].canWithdraw == false?
+            // not sure if we want to treat that differently to just amount == 0
+            if (!found || !_canWithdraw(prevOperator, strategy, lookupEpoch)) {
+                continue;
+            }
```
ChaoticWalrus, line 218: AFAICT we _need_ to treat the 'false' case differently that 'amount = 0' -- otherwise (I think, at least), we risk losing user funds by "zeroing out" part/all of their withdrawal, instead of reverting and forcing them to wait longer

src/contracts/interfaces/ISlasher.sol
```diff
@@ -148,19 +148,15 @@ interface ISlasher {
...
-     */
-    function getWithdrawabilityAndScalingFactorAtEpoch(
-        address operator,
-        IStrategy strategy,
-        uint32 epoch
-    ) external view returns (bool, uint64);
+	struct WithdrawableAmounts {
+		uint256 withdrawAmount;
+		uint256 redepositAmount;
+	}
```
ChaoticWalrus, line 154: could maybe make this an array that depends on the context of being fully withdrawn vs. redeposited?
but now that I suggest that... context-dependent storage of any kind tends to be a nightmare.
I guess I'm just seeing some inefficiency here, but I suppose that can be dealt with later; for now I agree this seems clearer.

ChaoticWalrus (changes requested):
Definitely looks like an improvement in readability.
I think the fact that `getWithdrawableAmounts` doesn't differentiate between the zero amount case and the "can't withdraw" case is a bug (but acknowledge I could be wrong); otherwise LGTM

src/contracts/core/Slasher.sol
wadealexc: okay, good point. i can adjust the struct to have a field specifying the staker isn't eligible for withdrawal yet.

src/contracts/interfaces/ISlasher.sol
wadealexc: Since we aren't actually storing this anywhere, I don't think what you're suggesting is crazy

src/contracts/libraries/EpochUtils.sol
```diff
@@ -14,6 +14,11 @@ library EpochUtils {
     // TODO: define appropriately
     uint256 internal constant EPOCH_LENGTH_SECONDS = 1 weeks;
 
+    function slashingEndEpoch(uint64 startBlock) internal pure returns (uint32) {
+        // TODO idk convert to timestamp, then epoch, then add slashability
+        return 0;
+    }
```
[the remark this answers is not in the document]
ChaoticWalrus: it's looking like we will nix epochs all-together, which should hopefully help resolve this

8sunyuan:
Closing, https://github.com/Layr-Labs/eigenlayer-contracts/pull/679 is most up-to-date impl

PR #611, opened 2024-07-02
chore: upgrade preprod eigenpods

wadealexc:


8sunyuan (approved):
LGTM, not sure if you want to also change the eigenlayer_addresses json config as well with the new EP implementation address but we can always do that later

PR #613, opened 2024-07-02
feat: epoch operator set reg/dereg

0xClandestine:


src/contracts/interfaces/IAVSDirectory.sol
```diff
@@ -12,6 +12,17 @@ interface IAVSDirectory is ISignatureUtils {
...
+    enum EpochStates {
+        NULL,
+        REGISTERED,
+        DEREGISTERED
+    }
+
+    struct OperatorInfo {
+        uint248 lastEpoch;
+        bool onStandby;
+    }
```
ChaoticWalrus, line 24: we already have a mapping called OperatorInfo in the middleware contracts repo.
I'd recommend something more specific/description, but don't have the best suggestion right now. perhaps when/if I review in depth.

Issue #592, opened 2024-06-21
Feature: Add `forge fmt` to CI


0xClandestine:
**User story**

As a developer, I want to enforce forge fmt in our CI GitHub actions so that all code merged into the repository is consistently formatted.

**Actions**
- [x] Update the GitHub Actions workflow to run forge fmt with check.
- [x] Update contributing guidelines to include the new requirement.

**Acceptance Criteria**

* CI runs forge fmt on every pull request.
* The pull request fails if forge fmt reports issues.
* Contributing guidelines are updated.

PR #614, opened 2024-07-03
feat: add `forge fmt src/contracts` to CI

0xClandestine:
https://github.com/Layr-Labs/eigenlayer-contracts/issues/592


0xClandestine:
Rather than enforcing the use of `forge fmt` across the entire repo, I used `forge fmt src/contracts`.

0xClandestine:
> Rather than enforcing the use of `forge fmt` across the entire repo, I used `forge fmt src/contracts`.

Also not sure if I properly added the CI script to workflows...

src/contracts/core/AVSDirectory.sol
```diff
@@ -89,9 +92,7 @@ contract AVSDirectory is
 
         // Check that the signature is valid
         EIP1271SignatureUtils.checkSignature_EIP1271(
-            operator,
-            operatorRegistrationDigestHash,
-            operatorSignature.signature
+            operator, operatorRegistrationDigestHash, operatorSignature.signature
         );
```
wadealexc, line 96: I'm fine with what `forge fmt` has done here but for the record, the actual way we do this in our codebases is that typically methods with 3+ parameters are called using the explicit parameter/field name syntax, e.g:

```solidity
EIP1271SignatureUtils.checkSignature_EIP1271({
  signer: operator,
  digestHash: operatorRegistrationDigestHash,
  signature: operatorSignature.signature
});
```

src/contracts/core/DelegationManager.sol
```diff
@@ -373,12 +392,7 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
             address operator = delegatedTo[staker];
 
             // subtract strategy shares from delegate's shares
-            _decreaseOperatorShares({
-                operator: operator,
-                staker: staker,
-                strategy: strategy,
-                shares: shares
-            });
+            _decreaseOperatorShares({operator: operator, staker: staker, strategy: strategy, shares: shares});
```
wadealexc, line 395: Oof. This one is pretty bad.

wadealexc: I see it's trying to make this a single line because it fits in the line length, but when we use explicit parameter-named syntax like this, it's almost universally clearer to have each param on a single line

src/contracts/core/AVSDirectory.sol
wadealexc: ok i see we are not calling `checkSignature` like this super consistently, so maybe it's fine for this method. i dont feel super strongly about it

src/contracts/core/DelegationManager.sol
```diff
@@ -467,37 +483,27 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        (IStrategy[] memory strategies, uint256[] memory shares) = getDelegatableShares(staker);
 
         for (uint256 i = 0; i < strategies.length;) {
-            _increaseOperatorShares({
-                operator: operator,
-                staker: staker,
-                strategy: strategies[i],
-                shares: shares[i]
-            });
+            _increaseOperatorShares({operator: operator, staker: staker, strategy: strategies[i], shares: shares[i]});
```
wadealexc, line 502: Yep, same thing again - not a fan of collapsing this to one line. :(

src/contracts/core/DelegationManager.sol
```diff
@@ -554,30 +559,31 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
                     shares: withdrawal.shares[i],
                     token: tokens[i]
                 });
-                unchecked { ++i; }
+                unchecked {
+                    ++i;
+                }
             }
-        // Award shares back in StrategyManager/EigenPodManager. If withdrawer is delegated, increase the shares delegated to the operator
+            // Award shares back in StrategyManager/EigenPodManager. If withdrawer is delegated, increase the shares delegated to the operator
```
wadealexc, line 566: Don't like this but maybe the comment just needs to be in the else branch entirely

src/contracts/core/DelegationManager.sol
```diff
@@ -628,26 +636,23 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
         for (uint256 i = 0; i < strategies.length;) {
             // Similar to `isDelegated` logic
             if (operator != address(0)) {
-                _decreaseOperatorShares({
-                    operator: operator,
-                    staker: staker,
-                    strategy: strategies[i],
-                    shares: shares[i]
-                });
+                _decreaseOperatorShares({operator: operator, staker: staker, strategy: strategies[i], shares: shares[i]});
```
wadealexc, line 655: :(

wadealexc (changes requested):
Generally I don't hate what I see

But see my comments above -- it has a habit of moving explicit-parameter function calls into a single line, which I really do not like. Is there a way to fix that specifically?

src/contracts/core/AVSDirectory.sol
8sunyuan: Also personally prefer multi-line for paramaters when it exceeds line length

src/contracts/core/DelegationManager.sol
0xClandestine: Yeah not a huge fan either...

I'll see if there's a way to avoid it via `foundry.toml`

#### Possible solutions
1) Use Foundry's comment flags like this:
```
// forgefmt: disable-next-item 
```
2) Maybe reduce max line length a bit?

0xClandestine:
Fix all of the above issues using:

```
// forgefmt: disable-next-item
```

Or by moving comments around...

src/contracts/core/DelegationManager.sol
```diff
@@ -486,18 +505,14 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
...
+        (IStrategy[] memory strategies, uint256[] memory shares) = getDelegatableShares(staker);
 
         for (uint256 i = 0; i < strategies.length;) {
-            _increaseOperatorShares({
-                operator: operator,
-                staker: staker,
-                strategy: strategies[i],
-                shares: shares[i]
-            });
+            _increaseOperatorShares({operator: operator, staker: staker, strategy: strategies[i], shares: shares[i]});
```
wadealexc, line 511: Needs forgefmt disable

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -445,12 +458,7 @@ contract RewardsCoordinator is
         // Verify inclusion of earner leaf
         bytes32 earnerLeafHash = calculateEarnerLeafHash(earnerLeaf);
         require(
-            Merkle.verifyInclusionKeccak({
-                root: root,
-                index: earnerLeafIndex,
-                proof: earnerProof,
-                leaf: earnerLeafHash
-            }),
+            Merkle.verifyInclusionKeccak({root: root, index: earnerLeafIndex, proof: earnerProof, leaf: earnerLeafHash}),
```
wadealexc, line 461: Needs forgefmt disable

src/contracts/libraries/Endian.sol
```diff
@@ -12,14 +12,8 @@ library Endian {
...
-        return
-            (n >> 56) |
-            ((0x00FF000000000000 & n) >> 40) |
-            ((0x0000FF0000000000 & n) >> 24) |
-            ((0x000000FF00000000 & n) >> 8) |
-            ((0x00000000FF000000 & n) << 8) |
-            ((0x0000000000FF0000 & n) << 24) |
-            ((0x000000000000FF00 & n) << 40) |
-            ((0x00000000000000FF & n) << 56);
+        return (n >> 56) | ((0x00FF000000000000 & n) >> 40) | ((0x0000FF0000000000 & n) >> 24)
```
wadealexc, line 15: Needs forgefmt disable

src/contracts/pods/EigenPodManager.sol
```diff
@@ -272,21 +281,24 @@ contract EigenPodManager is
...
                 return sharesAfter;
             }
         } else {
             // if the shares started positive and became negative, then the decrease in delegateable shares is the starting share amount
             if (sharesAfter <= 0) {
                 return (-sharesBefore);
-            // if the shares started positive and stayed positive, then the change in delegateable shares
-            // is the difference between starting and ending amounts
+                // if the shares started positive and stayed positive, then the change in delegateable shares
+                // is the difference between starting and ending amounts
```
wadealexc, line 301: Needs cleanup to move comments into correct blocks

wadealexc (changes requested):
See comments + please generate bindings

wadealexc (approved):
LGTM

PR #615, opened 2024-07-03
feat: interface to create rewards for avs

8sunyuan:
Use case: as a `rewardsForAllSubmitter` I wish to create a rewardsSubmission on behalf of an AVS. This will distribute out rewards to all of the AVSs operators and their delegated stakers


src/contracts/core/RewardsCoordinator.sol
```diff
@@ -143,6 +143,43 @@ contract RewardsCoordinator is
...
+            RewardsSubmission calldata rewardsSubmission = rewardsSubmissions[i];
+            uint256 nonce = submissionNonce[avs];
+            bytes32 rewardsSubmissionHash = keccak256(abi.encode(avs, nonce, rewardsSubmission));
+
+            _validateRewardsSubmission(rewardsSubmission);
+
+            isAVSRewardsSubmissionHash[avs][rewardsSubmissionHash] = true;
+            submissionNonce[avs] = nonce + 1;
+
+            emit AVSRewardsSubmissionCreated(avs, nonce, rewardsSubmissionHash, rewardsSubmission);
```
ypatil12, line 178: Note that we cannot distinguish, via events off-chain, between rewards submitted _by_ AVSs and _on behalf_ of AVSs. Is this by design? If we do care, then the bronze/silver tables need to be updated.

PR #619, opened 2024-07-09
refactor: remove cancelling from slashing

8sunyuan:
Removing cancelling from Slashing request lifecycle but big TODO is to implement this at the AVS middleware contracts level


src/contracts/core/DelegationManager.sol
```diff
@@ -572,10 +563,6 @@ contract DelegationManager is Initializable, OwnableUpgradeable, Pausable, Deleg
             );
         }
 
-        uint32 epochForEndOfSlashability = EpochUtils.getEndOfSlashabilityEpoch(withdrawalCreationEpoch(withdrawalRoot));
```
gpsanant, line 575: we still need this to only factor in slashing from up until the epoch after they queued the withdrawal

8sunyuan: Are you not still subject to slashing while queued withdrawing?

gpsanant: a withdrawal queued in epoch `n` should be subject to slashing until the end of epoch `n+1` and no longer

8sunyuan: Right because the withdrawal can be completed whenever and not necessarily in epoch n + 2

8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/619/commits/0e3dd0843b36d4e3847e0ddbbd8ff8cdb633439a

gpsanant:
@8sunyuan can this be closed?

PR #629, opened 2024-07-16
feat: add proof submitter address

wadealexc:
From Slack:

I have a small feature request for PEPE -- allowing a pod owner to approve a second address to start checkpoints and verify validator credentials. This change would be a new state variable plus a single, small method in EigenPod.sol that allows the pod owner to change the `proofSubmitter` address.

I think this would be super helpful for things like our CLI, which allows you to pass in a private key to submit transactions. However, Stakers will not want to submit their pod owner's private key to the CLI, because that key has access to literally all of their funds. We need a way for them to set up a hot wallet to accomplish their daily proving tasks.


gpsanant (approved):

src/contracts/pods/EigenPodStorage.sol
```diff
@@ -71,10 +71,16 @@ abstract contract EigenPodStorage is IEigenPod {
     /// @notice The current checkpoint, if there is one active
     Checkpoint internal _currentCheckpoint;
 
+    /// @notice An address with permissions to call `startCheckpoint` and `verifyCheckpointProofs`, set
+    /// by the podOwner. This role exists to allow a podOwner to designate a hot wallet that can call
+    /// these methods, allowing the podOwner to remain a cold wallet that is only used to manage funds.
+    /// @dev If this address is NOT set, only the podOwner can call `startCheckpoint` and `verifyCheckpointProofs`
```
8sunyuan, line 77: typo here, should be `verifyWithdrawalCredentials`

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -186,10 +189,25 @@ interface IEigenPod {
...
+
     /*******************************************************************************
                                    VIEW METHODS
     *******************************************************************************/
 
+    /// @notice An address with permissions to call `startCheckpoint` and `verifyCheckpointProofs`, set
+    /// by the podOwner. This role exists to allow a podOwner to designate a hot wallet that can call
+    /// these methods, allowing the podOwner to remain a cold wallet that is only used to manage funds.
+    /// @dev If this address is NOT set, only the podOwner can call `startCheckpoint` and `verifyCheckpointProofs`
+    function proofSubmitter() external view returns (address);
```
8sunyuan, line 209: same thing here too with comments

wadealexc: [b23f172](https://github.com/Layr-Labs/eigenlayer-contracts/pull/629/commits/b23f1720212c84c44874c5247baa59bdf3152770)

src/contracts/pods/EigenPodStorage.sol
wadealexc: [b23f172](https://github.com/Layr-Labs/eigenlayer-contracts/pull/629/commits/b23f1720212c84c44874c5247baa59bdf3152770)

PR #633, opened 2024-07-17
Remove allocator opset reg

gpsanant:


8sunyuan (approved):
approved and side note: The fact that no interfaces changes to OperatorSets are needed in Slashing is a good indicator on designs

PR #634, opened 2024-07-17
LST/Token and Deposit Flow Clarification

wesfloyd:


wesfloyd:
Alternatively, stylistically if you feel the "Depositing Natively Restaked beacon chain ETH includes the following high level steps:" is redundant - should we add a step between steps 1 and 2 to explicitly indicate the user should "Generate Proof via (eigenpod-proofs-generation)[https://github.com/Layr-Labs/eigenpod-proofs-generation] library?

ypatil12:
Closing in favor of new docs/flows

PR #635, opened 2024-07-17
fix: delegateTo address(0)

8sunyuan:
Noticed a flakey test here where `delegateTo(address(0))` doesn't actually revert because the `isOperator` definition was changed to be `delegatedTo[operator] == operator` and its default value in storage is ofc the 0 address, meaning the 0 address is technically an operator and calling delegateTo does not revert. The unit test below doesn't revert as expected from this scenario.
https://github.com/Layr-Labs/eigenlayer-contracts/actions/runs/9976519451/job/27568908582#step:6:5327

However our `isDelegated` helper actually checks `delegatedTo[staker] != address(0)` so you could delegate to the 0 address but it would never show a staker as delegated. 

PR fixes this by simply changing `isOperator` check to be `operator != address(0) && delegatedTo[operator] == operator` so calling `delegateTo(address(0))` will revert 

Note: Also fixed coverage in CI failing


ChaoticWalrus (approved):
LGTM.

The key thing (discussed offline):
 the `isOperator` definition/behavior changed when we deprecated the earnings receiver, so we had a bug where the zero address was being defined as an operator. This fixes the bug.

PR #631, opened 2024-07-16
docs: add research docs

gpsanant:

PR #636, opened 2024-07-18
feat: operator set migration; completely separate legacy and opset reg

ypatil12:
Bring back the OperatorAVSRegistration storage that was previously overwritten.

Segregates the registration/deregistration on M2 Legacy & Operator Sets for consistency in offchain code paths. As a result, we remove the numAVSRegistered struct. 

The storage may be able to be condensed still, but haven't put further thought into this.   

Pending: Bring Back OpSet struct in a separate PR

### DIFF
--- a/src/contracts/core/AVSDirectory.sol
+++ b/src/contracts/core/AVSDirectory.sol
@@ -86,7 +86,7 @@ contract AVSDirectory is
     }
 
     /**
-     *  @notice Called by AVSs to add an operator to an operator set.
+     *  @notice Called by AVSs to add an operator to a list of operatorSets.
      *
      *  @param operator The address of the operator to be added to the operator set.
      *  @param operatorSetIds The IDs of the operator sets.
@@ -94,9 +94,6 @@ contract AVSDirectory is
      *
      *  @dev msg.sender is used as the AVS.
      *  @dev The operator must not have a pending deregistration from the operator set.
-     *  @dev If this is the first operator set in the AVS that the operator is
-     *  registering for, a OperatorAVSRegistrationStatusUpdated event is emitted with
-     *  a REGISTERED status.
      */
     function registerOperatorToOperatorSets(
         address operator,
--- a/src/contracts/core/AVSDirectory.sol
+++ b/src/contracts/core/AVSDirectory.sol
@@ -62,13 +62,25 @@ contract AVSDirectory is
      * @param operatorSetId The ID of the operator set to initialize.
      *
      * @dev msg.sender must be the AVS.
+     * @dev The AVS may create operator sets before it becomes an operator set AVS.
      */
     function createOperatorSet(uint32 operatorSetId) external {
         require(
-            !isOperatorSet[msg.sender][operatorSetId], "AVSDirectory._createOperatorSet: operator set already exists"
+            !isOperatorSet[msg.sender][operatorSetId], "AVSDirectory.createOperatorSet: operator set already exists"
         );
         isOperatorSet[msg.sender][operatorSetId] = true;
         emit OperatorSetCreated(msg.sender, operatorSetId);
+    }
+
+    /**
+     * @notice Sets the AVS as an operator set AVS, preventing legacy M2 operator registrations.
+     * 
+     * @dev msg.sender must be the AVS. 
+     */
+    function becomeOperatorSetAVS() external {
+        require(!isOperatorSetAVS[msg.sender], "AVSDirectory.becomeOperatorSetAVS: already an operator set AVS");
+        isOperatorSetAVS[msg.sender] = true;
+        emit AVSMigratedToOperatorSets(msg.sender);
     }
 
     /**
@@ -99,6 +111,11 @@ contract AVSDirectory is
             !operatorSaltIsSpent[operator][operatorSignature.salt],
             "AVSDirectory.registerOperatorToOperatorSets: salt already spent"
         );
+        // Assert that the AVS is an operator set AVS.
+        require(
+            isOperatorSetAVS[msg.sender],
+            "AVSDirectory.registerOperatorToOperatorSets: AVS is not an operator set AVS"
+        );
         // Assert `operator` is actually an operator.
         require(
             delegation.isOperator(operator),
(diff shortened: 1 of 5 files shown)


src/contracts/core/AVSDirectory.sol
```diff
@@ -52,37 +52,89 @@ contract AVSDirectory is
...
+            emit OperatorSetCreated(msg.sender, operatorSetIds[i]);
+        }
+    }
+
+    /**
+     * @notice Sets the AVS as an operator set AVS, preventing legacy M2 operator registrations.
+     *
+     * @dev msg.sender must be the AVS.
+     */
+    function becomeOperatorSetAVS() external {
```
gpsanant, line 83: do we need this? it should just be upon first `createOperatorSets`

src/contracts/core/AVSDirectory.sol
```diff
@@ -52,37 +52,89 @@ contract AVSDirectory is
...
+     * @dev The AVS can no longer register operators via the legacy M2 registration path once it begins migration
+     * @dev The operator is deregistered from the M2 legacy AVS once migrated
+     */
+    function migrateOperatorsToOperatorSets(
+        address[] calldata operators,
+        uint32[][] calldata operatorSetIds
+    ) external override onlyWhenNotPaused(PAUSED_OPERATOR_REGISTER_DEREGISTER_TO_AVS) {
+        // Assert that the AVS is an operator set AVS.
         require(
-            !isOperatorSet[msg.sender][operatorSetId], "AVSDirectory._createOperatorSet: operator set already exists"
```
gpsanant, line 68: we should keep this check

src/contracts/core/AVSDirectory.sol
```diff
@@ -94,16 +146,20 @@ contract AVSDirectory is
...
-            "AVSDirectory.registerOperatorToOperatorSets: salt already spent"
-        );
         // Assert `operator` is actually an operator.
         require(
             delegation.isOperator(operator),
             "AVSDirectory.registerOperatorToOperatorSets: operator not registered to EigenLayer yet"
         );
+        // Assert that the AVS is an operator set AVS.
+        require(
+            isOperatorSetAVS[msg.sender], "AVSDirectory.registerOperatorToOperatorSets: AVS is not an operator set AVS"
```
gpsanant, line 156: instead, check that the operator set exists here

src/contracts/core/AVSDirectory.sol
```diff
@@ -52,37 +52,89 @@ contract AVSDirectory is
...
+     * @dev The operator is deregistered from the M2 legacy AVS once migrated
+     */
+    function migrateOperatorsToOperatorSets(
+        address[] calldata operators,
+        uint32[][] calldata operatorSetIds
+    ) external override onlyWhenNotPaused(PAUSED_OPERATOR_REGISTER_DEREGISTER_TO_AVS) {
+        // Assert that the AVS is an operator set AVS.
         require(
-            !isOperatorSet[msg.sender][operatorSetId], "AVSDirectory._createOperatorSet: operator set already exists"
+            isOperatorSetAVS[msg.sender], "AVSDirectory.migrateOperatorsToOperatorSets: AVS is not an operator set AVS"
```
gpsanant, line 106: no need to make this check, register to operator sets will fail

src/contracts/core/AVSDirectory.sol
gpsanant: or actually, check in the internal function

src/contracts/core/AVSDirectory.sol
```diff
@@ -232,15 +289,13 @@ contract AVSDirectory is
             "AVSDirectory.registerOperatorToAVS: operator signature expired"
         );
 
-        MemberInfo storage member = memberInfo[msg.sender][operator];
+        // Assert that the AVS is not an operator set AVS.
+        require(!isOperatorSetAVS[msg.sender], "AVSDirectory.registerOperatorToAVS: AVS is an operator set AVS");
```
gpsanant, line 293: we should put this in the docs. registration is bricked once migration starts

src/contracts/core/AVSDirectory.sol
```diff
@@ -288,36 +343,71 @@ contract AVSDirectory is
...
+            );
+
+            // Assert `operator` has not already been registered to `operatorSetIds[i]`.
+            require(
+                !isMember[avs][operator][operatorSetIds[i]],
+                "AVSDirectory._registerOperatorToOperatorSets: operator already registered to operator set"
+            );
+
+            // Mutate `isMember` to `true`.
+            isMember[avs][operator][operatorSetIds[i]] = true;
```
gpsanant, line 384: prefer `isOperatorInOperatorSet`

gpsanant: idc tho, change the view function in the spec if u prefer this tho

src/contracts/core/AVSDirectory.sol
```diff
@@ -52,37 +52,89 @@ contract AVSDirectory is
...
+     * @dev The AVS may create operator sets before it becomes an operator set AVS.
      */
-    function createOperatorSet(uint32 operatorSetId) external {
+    function createOperatorSets(uint32[] calldata operatorSetIds) external {
+        for (uint256 i = 0; i < operatorSetIds.length; ++i) {
+            require(
+                !isOperatorSet[msg.sender][operatorSetIds[i]],
+                "AVSDirectory.createOperatorSet: operator set already exists"
+            );
+            isOperatorSet[msg.sender][operatorSetIds[i]] = true;
```
gpsanant, line 73: can you update the view functions in the spec accordingly?

src/contracts/core/AVSDirectory.sol
ypatil12: This prevents footguns for AVSs. Suggestion by @stevennevins and @wadealexc

src/contracts/core/AVSDirectory.sol
ypatil12: Operator set existence is checked in the internal function so we don't duplicate code in migration + registration. We check whether AVS is an operator set AVS in the parent function since we don't need to check it for every operatorSetID of the AVS - gas savings, even though its a warm sload

src/contracts/core/AVSDirectory.sol
ypatil12: Assume you mean for `createOperatorSets` function?

src/contracts/core/AVSDirectory.sol
ypatil12: cc @0xClandestine on naming finalization - remember you were playing around with some ideas here

src/contracts/core/AVSDirectory.sol
ypatil12: Added here: https://github.com/Layr-Labs/eigenlayer-contracts/pull/631/commits/c3b723b42895361f2e1b6aaec2e135944161ca2d

src/contracts/core/AVSDirectory.sol
stevennevins: I was thinking of it as we don't want to let createOperatorSets be called unintentionally. Requiring that this is called first prevents that

PR #637, opened 2024-07-18
feat: operator set migration

ypatil12:
This PR adds the migration feature for operator sets. 

In lieu of adding additional storage, we use the previous `OperatorAVSStatus` mapping to gate migrations. If the operator is registered, we allow migration. Once the migration has completed, we deregister the operator from M2, which prevents future migrations.

### DIFF
--- a/src/contracts/core/AVSDirectory.sol
+++ b/src/contracts/core/AVSDirectory.sol
@@ -98,7 +98,7 @@ contract AVSDirectory is
      * @dev The operator is deregistered from the M2 legacy AVS once migrated
      */
     function migrateOperatorsToOperatorSets(
-        address[] operators,
+        address[] calldata operators,
         uint32[][] calldata operatorSetIds
     ) external override onlyWhenNotPaused(PAUSED_OPERATOR_REGISTER_DEREGISTER_TO_AVS) {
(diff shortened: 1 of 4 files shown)


0xClandestine (approved):

src/contracts/core/AVSDirectory.sol
```diff
@@ -67,24 +67,63 @@ contract AVSDirectory is
...
+     * @dev The operator is deregistered from the M2 legacy AVS once migrated
+     */
+    function migrateOperatorsToOperatorSets(
+        address[] calldata operators,
+        uint32[][] calldata operatorSetIds
+    ) external override onlyWhenNotPaused(PAUSED_OPERATOR_REGISTER_DEREGISTER_TO_AVS) {
+        // Assert that the AVS is an operator set AVS.
+        require(isOperatorSetAVS[msg.sender], "AVSDirectory.migrateOperatorsToOperatorSets: AVS is not an operator set AVS");
+
+        for (uint256 i = 0; i < operators.length; i++) {
```
stevennevins, line 107: Don't we need another for loop to register for each operator set for each operator?

stevennevins (approved):

stevennevins:
lgtm

PR #640, opened 2024-07-22
SDA - two contract solution

gpsanant:
The approach in this PR is to have an AllocatorManager contract which bridges to the previous DelegationManager to
1. make sure that stakers delegated to pre-SDA operators cant delegate to allocators
2. complete handoffs from pre-SDA operators to allocators 
3. migrate state for stakers that were handed off by pre-SDA operators

In addition, the DM `getOperatorShares` and `operatorShares` are reimplemented to return 0 for operators pending migration and read from XYZ new contract for the allocated shares to an operator.

Big TODOs:
- [ ] Hook up with total magnitude stuff and making withdrawal logic sound with scaled shares
- [ ] Implementing `getOperatorShares` and `operatorShares` to account for migrations and SDA
- [ ] Address EigenPodManager/StrategyManager interactions
- [x] State migration for handed off stakers
- [ ] TODOs :)

Choices of note:
1. Operator addresses must be either pre-SDA or post-SDA
2. Stakers must either be pre-SDA or post-SDA, but can migrate on the same address from pre to post

Notes on edge cases:

1. stakers cannot queue DM withdrawals after shares are migratable leading to them possibly having withdrawals in both contracts
3. what happens when handed off stakers or pre-SDA operators interact with contracts before migration is completed

### DIFF
--- a/docs/releases/slashing-release/interfaces/allocator.md
+++ b/docs/releases/slashing-release/interfaces/allocator.md
@@ -4,26 +4,108 @@ The `OperatorDetails` is modified to replace the `earningsReciever` field with t
 
 ```solidity
 interface IDelegationManager {
+	/// STRUCTS
+
+	/**
+     * Struct type used to specify an existing queued withdrawal. Rather than storing the entire struct, only a hash is stored.
+     * In functions that operate on existing queued withdrawals -- e.g. completeQueuedWithdrawal`, the data is resubmitted and the hash of the submitted
+     * data is computed by `calculateWithdrawalRoot` and checked against the stored hash in order to confirm the integrity of the submitted data.
+     */
+    struct Withdrawal {
+        // The address that originated the Withdrawal
+        address staker;
+        // The staker's allocator at the time of thhe queing of the withdrawal
+        address allocator;
+        // The address that can complete the Withdrawal + will receive funds when completing the withdrawal
+        address withdrawer;
+        // Nonce used to guarantee that otherwise identical withdrawals have unique hashes
+        uint256 nonce;
+        // Block number when the Withdrawal was created
+        uint32 startBlock;
+        // Array of strategies that the Withdrawal contains
+        IStrategy[] strategies;
+        // Array containing the amount of shares in each Strategy in the `strategies` array
+        uint256[] shares;
+    }
+
 	/// EVENTS
 
-	/// @notice Emitted when an operator updates their OperatorDetails to @param newOperatorDetails
+	// @notice Emitted when a new allocator is registered in EigenLayer
+    event AllocatorRegistered(address indexed allocator, address selectionApprover);
+
+    /// @notice Emitted when an allocator updates their selection approver
+    event SelectionApproverUpdated(address indexed allocator, address selectionApprover);
+
+	/**
+     * @notice Emitted when @param allocator indicates that they are updating their MetadataURI string
+     * @dev Note that these strings are *never stored in storage* and are instead purely emitted in events for off-chain indexing
+     */
+    event AllocatorMetadataURIUpdated(address indexed allocator, string metadataURI);
+
+	/// @notice Emitted when an operator queues a handoff of their stake to a target allocator
 	event AllocatorHandoffQueued(address operator, address allocator, uint32 effectTimestamp);
 
-	
+	/// @notice Emitted when a handoff from an operator to an allocator is completed for a set of strategies
 	event AllocatorHandoffCompleted(address operator, address allocator, IStrategy[] strategies);
+
+	/// @notice Emitted whenever an allocator's shares are increased for a given strategy. Note that shares is the delta in the allocator's shares.
+    event AllocatorSharesIncreased(address indexed allocator, address staker, IStrategy strategy, uint256 shares);
+
+    /// @notice Emitted whenever an allocator's shares are decreased for a given strategy. Note that shares is the delta in the allocator's shares.
+    event AllocatorSharesIncreased(address indexed allocator, address staker, IStrategy strategy, uint256 shares);
+
+	/// @notice Emitted when @param staker selects @param allocator as their allocator
+    event StakerSelected(address indexed staker, address indexed allocator);
+
+    /// @notice Emitted when @param staker unselects their current allocator
+    event StakerUnselected(address indexed staker, address indexed allocator);
+
+    /// @notice Emitted when @param staker unselects an allocator via a call not originating from the staker themself
+    event StakerForceUnselected(address indexed staker, address indexed allocator);
 	
 	/// EXTERNAL - STATE MODIFYING
+
+	/**
+     * @notice Registers the caller as an allocator in EigenLayer.
+     * @param selectionApprover is the address that must approve of the selection of a staker's stake to the allocator. If set to 0, no approval is 
+	 * required.
+     * @param metadataURI is a URI for the allocator's metadata, i.e. a link providing more details on the allocator.
+     *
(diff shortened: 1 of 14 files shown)


docs/releases/slashing-release/interfaces/allocator.md
```diff
@@ -1,61 +1,277 @@
...
+        // NEW: The staker's allocator at the time of thhe queing of the withdrawal
+        address allocator;
+        // The address that can complete the Withdrawal + will receive funds when completing the withdrawal
+        address withdrawer;
+        // Nonce used to guarantee that otherwise identical withdrawals have unique hashes
+        uint256 nonce;
+        // Block number when the Withdrawal was created
+        uint32 startBlock;
+        // Array of strategies that the Withdrawal contains
+        IStrategy[] strategies;
```
8sunyuan, line 26: Is it time to just reduce this to a single strategy? We already do this in `queueWithdrawal` where we have multiple withdrawal roots created and we know its going to be more expensive to complete withdrawals in the case of slashing events.

gpsanant: it fucks up completion of existing withdrawals :(

unless we want to do a far more invasive upgrade where we upgrade to single strat withdrawals, wait a while, autocomplete all multistrategy withdrawals, and then remove autocomplete functions...

8sunyuan: Yup, basically the same thing we did with M2 withdrawals migration. I guess this will just remain as tech debt for the foreseeable future but it would be nice to address this at some point.

src/contracts/core/AllocatorManager.sol
```diff
@@ -0,0 +1,965 @@
...
+	 * @param strategies the strategies to be handed off
+	 * 
+	 * @dev must be called 14 days after the handoff was queued
+	 * @dev the allocator's shares are incremented by the operator's shares for each strategy and the operator's shares are decremented by the operator's
+	 * shares for each strategy.
+	 * @dev if all strategies are not handed off, this function can be called by anyone else to 
+	 * complete the handoff for different strategies
+	 */
+	function completeAllocatorHandoff(address operator, address allocator, IStrategy[] calldata strategies) external {
+        uint256[] memory shares = delegationManager.getMigratableOperatorShares(operator, allocator, strategies);
```
ChaoticWalrus, line 219: does this function also zero out the operator's shares? otherwise I'd be worried this could be called multiple times

src/contracts/core/AllocatorManagerStorage.sol
```diff
@@ -0,0 +1,120 @@
...
+    /// @notice Mapping: staker => cumulative number of queued withdrawals they have ever initiated.
+    /// @dev This only increments (doesn't decrement), and is used to help ensure that otherwise identical withdrawals have unique hashes.
+    mapping(address => uint256) public cumulativeWithdrawalsQueued;
+
+
+    /**
+     * @notice Minimum delay enforced by this contract per Strategy for completing queued withdrawals. Measured in blocks, and adjustable by this contract's owner,
+     * up to a maximum of `MAX_WITHDRAWAL_DELAY_BLOCKS`. Minimum value is 0 (i.e. no delay enforced).
+     */
+    mapping(IStrategy => uint256) public strategyWithdrawalDelayBlocks;
```
ChaoticWalrus, line 105: feels like a possible opportunity to switch this from blocks to seconds/time, as the rest of the protocol has moved in that direction already

src/contracts/core/AllocatorManagerStorage.sol
```diff
@@ -0,0 +1,120 @@
...
+
+    struct AllocatorDetails {
+        address delegationApprover;
+        bool isAllocator; // if true, then the address is an allocator. Needed beacause selectionApprover can be 0.
+    }
+
+    mapping(address => AllocatorDetails) internal _allocatorDetails;
+
+    /// @notice Mapping: staker => whether the staker is a post SDA staker (i.e. they have no left over state unmigrated from the DelegationManager).
+    mapping(address => bool) public isPostSDAStaker;
```
ChaoticWalrus, line 65: can presumably be grouped with `_delegatedTo` for storage packing optimization

src/contracts/core/DelegationManager.sol
```diff
@@ -563,6 +600,11 @@ contract DelegationManager is
         // Remove `withdrawalRoot` from pending roots
         delete pendingWithdrawals[withdrawalRoot];
 
+        require(
+            !isPendingHandoff(withdrawal.delegatedTo) || receiveAsTokens,
+            "DelegationManager._completeQueuedWithdrawal: a staker pending handoff cannot withdraw as shares"
+        );
```
ChaoticWalrus, line 606: feels like a potential major UX headache? Very confusing UX for stakers that get hit by this; users could be forced to withdraw from system when e.g. trying to change delegation.
I'm presuming this behavior is to simplify accounting? what's the alternative solution?

src/contracts/core/DelegationManager.sol
```diff
@@ -664,6 +706,7 @@ contract DelegationManager is
         IStrategy[] memory strategies,
         uint256[] memory shares
     ) internal returns (bytes32) {
+        require(!isCompletableHandoff(operator), "DelegationManager._removeSharesAndQueueWithdrawal: staker's operator is in the middle of a handoff");
```
ChaoticWalrus, line 709: so stakers cannot queue new withdrawals once the handoff is completable? 
this is enforcing that shares must be moved to the AllocatorManager and a new withdrawal queued there instead, right?

src/contracts/core/DelegationManager.sol
```diff
@@ -821,6 +864,18 @@ contract DelegationManager is
         return delegatedTo[operator] == operator;
     }
 
+    function isPendingHandoff(address operator) public view returns (bool) {
+        return _handoffs[operator].completableTimestamp != 0;
+    }
```
ChaoticWalrus, line 869: this isn't 'pending' it's just 'initiated ever'?

src/contracts/core/DelegationManager.sol
```diff
@@ -842,14 +897,49 @@ contract DelegationManager is
...
+     * @dev if any of the shares are 0, it will revert
+     */
+    function getMigratableOperatorShares(address operator, address allocator, IStrategy[] calldata strategies) external returns (uint256[] memory) {
+        // TODO: onlyAllocatorRegistry
+        require(_handoffs[operator].allocator == allocator, "DelegationManager.getMigratableOperatorShares: allocator mismatch");
+        require(isCompletableHandoff(operator), "DelegationManager.getMigratableOperatorShares: handoff not completable");
+        uint256[] memory shares = _getOperatorShares(operator, strategies);
+        for (uint256 i = 0; i < strategies.length; ++i) {
+            // overwrite to 0 because they're migrated
+            require(shares[i] != 0, "DelegationManager.getMigratableOperatorShares: operator has active shares");
```
ChaoticWalrus, line 914: this require seems unnecessary? I think this just introduces a race condition in multiple calls to migrate operator shares with no clear benefit?

src/contracts/core/StrategyManagerStorage.sol
```diff
@@ -25,6 +26,7 @@ abstract contract StrategyManagerStorage is IStrategyManager {
 
     // system contracts
     IDelegationManager public immutable delegation;
+    IAllocatorManager public immutable allocatorManager = IAllocatorManager(address(0));
```
ChaoticWalrus, line 29: TODO: set this in constructor

src/contracts/core/AllocatorManager.sol
ChaoticWalrus: it does!

PR #641, opened 2024-07-23
feat: poc for slashing allocators

8sunyuan:


src/contracts/core/AVSDirectory.sol
```diff
@@ -490,4 +491,381 @@ contract AVSDirectory is
...
+    ) external returns (QueuedReallocation[] memory queuedReallocations) {
+        AllocatorOperatorDetails storage allocatorDetails = allocatorOperatorDetails[allocator][operator];
+        require(
+            allocatorDetails.isAllocatorFor,
+            "AVSDirectory.queueReallocation: allocator is not an allocator for operator"
+        );
+        if (msg.sender != allocator) {
+            _verifyAllocatorSignature(allocator, operator, allocatorSignature);
+        }
+        uint32 reallocationNonce = allocatorDetails.reallocationNonce;
```
gpsanant, line 627: this isn't used?

src/contracts/core/AVSDirectory.sol
```diff
@@ -490,4 +491,381 @@ contract AVSDirectory is
...
+
+        QueuedReallocation[] memory queuedAllocations = new QueuedReallocation[](magnitudeAdjustments.length);
+
+        for (uint256 i = 0; i < magnitudeAdjustments.length; ++i) {
+            uint64 scalingFactor = _getCurrentScalingFactor(
+                operator,
+                _hashStrategyAndOperatorSet(strategy, magnitudeAdjustments[i].operatorSet)
+            );
+            uint64 normalizedMagnitudeIncrement = SlashingAccountingUtils.normalizeMagnitude({
+                nonNormalizedMagnitude: magnitudeAdjustments[i].magnitudeDiff,
```
gpsanant, line 713: i feel like they should send the magnitude rather than the nonnormalized magnitude and we should convert on the fly

src/contracts/core/AVSDirectory.sol
```diff
@@ -490,4 +491,381 @@ contract AVSDirectory is
...
+                timestamp: uint32(block.timestamp)
+            });
+
+            // set queued allocation hash struct in mapping as pending
+            bytes32 reallocationHash = calculateQueuedReallocationHash(queuedAllocations[i]);
+            queuedReallocationHashes[reallocationHash] = true;
+
+            // decrement from non slashable magnitude
+            nonslashableMagnitude -= normalizedMagnitudeIncrement;
+            ++nonce;
```
gpsanant, line 739: nit: use length of array

src/contracts/core/AVSDirectory.sol
```diff
@@ -490,4 +491,381 @@ contract AVSDirectory is
...
+                queuedScalingFactor: scalingFactor,
+                timestamp: uint32(block.timestamp)
+            });
+
+            // set queued allocation hash struct in mapping as pending
+            bytes32 reallocationHash = calculateQueuedReallocationHash(queuedAllocations[i]);
+            queuedReallocationHashes[reallocationHash] = true;
+
+            // decrement from non slashable magnitude
+            nonslashableMagnitude -= normalizedMagnitudeIncrement;
```
gpsanant, line 738: need to overwrite storage

src/contracts/core/AVSDirectory.sol
```diff
@@ -490,4 +491,381 @@ contract AVSDirectory is
...
+        uint32 nonce = allocatorDetails.reallocationNonce;
+        QueuedReallocation[] memory queuedDeallocations = new QueuedReallocation[](magnitudeAdjustments.length);
+
+        for (uint256 i = 0; i < magnitudeAdjustments.length; ++i) {
+            bytes32 opSetKey = _hashStrategyAndOperatorSet(strategy, magnitudeAdjustments[i].operatorSet);
+            uint64 scalingFactor = _getCurrentScalingFactor(
+                operator,
+                _hashStrategyAndOperatorSet(strategy, magnitudeAdjustments[i].operatorSet)
+            );
+            uint64 normalizedMagnitudeDecrement = SlashingAccountingUtils.normalizeMagnitude({
```
gpsanant, line 762: do we want an `expectedScalingFactor` argument?

gpsanant: to revert upon slashing race conditions

src/contracts/core/AVSDirectory.sol
```diff
@@ -490,4 +491,381 @@ contract AVSDirectory is
...
+                nonNormalizedMagnitude: magnitudeAdjustments[i].magnitudeDiff,
+                scalingFactor: scalingFactor
+            });
+            require(
+                normalizedMagnitudeDecrement <= _nonNormalizedMagnitudes[operator][opSetKey][allocator],
```
gpsanant, line 767: need to decrement `_nonNormalizedMagnitudes`

src/contracts/core/AVSDirectory.sol
8sunyuan: Was thinking the exact same thing looking at it. Allocators shouldn't have to think about the normalized magnitudes

src/contracts/core/AVSDirectory.sol
8sunyuan: are we doing unbacked increases? Or backed increases.
If the latter then we need the same running queued allocation sum as before

gpsanant: yep we should do the latter imo.

however, it kinda sucks that the allocations take two steps, since the UI for stakers is "possibly more risky" wonder if we could get rid of that

src/contracts/core/AVSDirectory.sol
```diff
@@ -490,4 +491,381 @@ contract AVSDirectory is
...
+        IStrategy strategy,
+        OperatorSet opSet,
+        address allocator,
+        uint32 bipsToSlash
+    );
+
+    uint256 constant MAX_ALLOCATORS_PER_OPERATOR = 10; 
+    uint256 constant QUEUED_REALLOCATION_DELAY = 21 days; // placeholder val for now
+
+    uint64 constant DEFAULT_TOTAL_MAGNITUDE = 1e18; // 1e18 = 1
```
ChaoticWalrus, line 552: we almost certainly want this as large as possible. we can discuss, but even making magnitudes uint64s (instead of a larger type) has some (theoretical, at least) negative consequences

PR #644, opened 2024-07-24
Feat: add `sharesToUnderlying` event

0xClandestine:


ypatil12 (dismissed):
LGTM. Let's just add a sanity test?

src/contracts/strategies/StrategyBase.sol
```diff
@@ -223,8 +223,8 @@ contract StrategyBase is Initializable, Pausable, IStrategy {
      * @return The amount of underlying tokens corresponding to the input `amountShares`
      * @dev Implementation for these functions in particular may vary significantly for different strategies
      */
-    function sharesToUnderlying(uint256 amountShares) public view virtual override returns (uint256) {
-        return sharesToUnderlyingView(amountShares);
+    function sharesToUnderlying(uint256 amountShares) public virtual override returns (uint256 underlying) {
+        emit SharesToUnderlying(underlying = sharesToUnderlyingView(amountShares));
```
bowenli86, line 227: it should be the ratio, not total underlying token. The ratio can be by setting amountShares = 10^8

besides, felt like emitting total shares is also super useful and emitting both would actually be better? @ypatil12 wdyt?

ypatil12: Ahh gotcha. Okay let's emit the ratio

src/contracts/strategies/StrategyBase.sol
```diff
@@ -220,11 +220,11 @@ contract StrategyBase is Initializable, Pausable, IStrategy {
...
      * @notice In contrast to `sharesToUnderlyingView`, this function **may** make state modifications
      * @param amountShares is the amount of shares to calculate its conversion into the underlying token
-     * @return The amount of underlying tokens corresponding to the input `amountShares`
+     * @return underlying The amount of underlying tokens corresponding to the input `amountShares`
      * @dev Implementation for these functions in particular may vary significantly for different strategies
      */
-    function sharesToUnderlying(uint256 amountShares) public view virtual override returns (uint256) {
-        return sharesToUnderlyingView(amountShares);
+    function sharesToUnderlying(uint256 amountShares) public virtual override returns (uint256 underlying) {
+        emit SharesToUnderlying(1e8 * amountShares / (underlying = sharesToUnderlyingView(amountShares)));
```
bowenli86, line 227: rename to SharesToUnderlyingTokenRatio?

also this need to have the strategy address in the event

PR #648, opened 2024-07-26
feat: add pass thoughts for all `strategyWhitelister` gated functions

ypatil12:
Add pass throughs for `removeStrategiesFromDepositWhitelist` and `setThirdPartyTransfersForbidden`


wadealexc (approved):
looks great, thank you!

PR #654, opened 2024-08-01
Task: require solhint as part of continuous integration

jac18281828:
1. require solhint in GitHub format action
2. include npm in dev container
3. add vscode tasks.json for local build
4. format action is now lint and check format action


ypatil12 (approved):
LGTM

PR #656, opened 2024-08-05
feat: updated slashing allocations

8sunyuan:
Updated to original interface. TL;DR
- single allocator per operator
- 1 tx for allocations where all allocations need to come from nonslashableMagnitude (we are calling this `freeMagnitude` in storage
- queue, complete tx steps for performing deallocations since deallocations could still be slashed while queued.
- `slashOperator`, slashes current magnitude as well as queued deallocations. 

### TODOs:
Several TODOs commented out in the code as well

#### Deposits/Withdraws of Shares [NOT STARTED]
- [ ] Update deposits/withdraws based on slashed allocations
#### Allocator Configuration [WIP]
- [ ] Figure out a tolerable rounding scheme for slashing magnitude allocations
- [ ] Add signature verification to relevant functions
- [ ] Decide on where operator<>allocator relationship stored. `OperatorDetails.earningsReceiver`?
- [x] Create internal function to loop from current magnitude, and decrement current and all future values. Likely requires modifying CheckpointsUpgradeable library to acommodate this. Used for `queueDeallocation()`, `slashOperator()`
- [ ] Events


src/contracts/pods/EigenPodManager.sol
```diff
@@ -289,6 +289,7 @@ contract EigenPodManager is
             if (sharesAfter <= 0) {
                 // if the shares started negative and stayed negative, then there cannot have been an increase in delegateable shares
                 return 0;
+                // if the shares started negative and became positive, then the increase in delegateable shares is the ending share amount
```
gpsanant, line 292: ?

src/contracts/core/AVSDirectory.sol
```diff
@@ -52,137 +61,662 @@ contract AVSDirectory is
...
+     * @notice For a single strategy, allocate magnitude to a list of operator sets and then update freeMagnitude
+     * for the strategy after all allocations
+     * @param operator address to allocate magnitude for
+     * @param allocation magnitude adjustment for a single strategy and multiple operator sets
+     * @param effectTimestamp timestamp when the allocation will take effect
+     */
+    function _allocate(address operator, MagnitudeAdjustment calldata allocation, uint32 effectTimestamp) internal {
+        IStrategy strategy = allocation.strategy;
+        OperatorSet[] calldata operatorSets = allocation.operatorSets;
+        uint64 freeAllocatableMagnitude = freeMagnitude[operator][strategy];
```
gpsanant, line 502: just `freeMangnitude` should be fine?

src/contracts/core/AVSDirectory.sol
```diff
@@ -52,137 +61,662 @@ contract AVSDirectory is
...
+
+        for (uint256 i = 0; i < operatorSets.length; ++i) {
+            // 1. check freeMagnitude available, that is the allocation of stake is backed and not slashable
+            require(
+                freeAllocatableMagnitude >= allocation.magnitudeDiffs[i],
+                "AVSDirectory.queueAllocations: insufficient available free magnitude to allocate"
+            );
+            // 2. rate limiting of 1 pending allocation at a time
+            (bool exists, uint32 timestamp, uint224 value) = _magnitudeUpdate[operator][strategy][operatorSets[i].avs][operatorSets[i]
```
gpsanant, line 511: we need to make this agnostic to the number of pending changes

src/contracts/core/AVSDirectory.sol
```diff
@@ -52,137 +61,662 @@ contract AVSDirectory is
...
+     * @param strategy the strategy to complete deallocations for
+     * @param operatorSet the operator set to complete deallocations for the given strategy
+     */
+    function _completeDeallocation(
+        address operator,
+        IStrategy strategy,
+        OperatorSet calldata operatorSet
+    ) internal returns (uint64 freeMagnitudeToAdd) {
+        QueuedDeallocation[] memory queuedDeallocation =
+            _queuedDeallocations[operator][strategy][operatorSet.avs][operatorSet.operatorSetId];
```
gpsanant, line 595: how does this work if you set this to memory?

src/contracts/libraries/MagnitudeCheckpoints.sol
```diff
@@ -0,0 +1,266 @@
...
+
+        uint256 pos = _upperBinaryLookup(self._magnitudes, timestamp, low, high);
+
+        return pos == 0 ? (0, pos) : (_unsafeAccess(self._magnitudes, pos - 1)._value, pos - 1);
+    }
+
+    /// @notice WARNING: this function is only used because of the invariant property
+    /// that from the current timestamp, all future checkpointed magnitude values are strictly > current value.
+    /// Use function with extreme care for other situations.
+    function decrementCurrentAndFutureCheckpoints(
```
gpsanant, line 244: Not current, maybe `decrementAtAndAfter`

src/contracts/libraries/MagnitudeCheckpoints.sol
```diff
@@ -0,0 +1,266 @@
...
+        History storage self,
+        uint32 timestamp,
+        uint224 decrementValue
+    ) internal {
+        (uint224 value, uint256 pos) = upperLookupRecentWithPos(self, timestamp);
+
+        // if there is no checkpoint, return
+        pos == 0 ? type(uint256).max : pos;
```
gpsanant, line 252: why does this mean there is no checkpoint?

src/contracts/libraries/MagnitudeCheckpoints.sol
```diff
@@ -0,0 +1,266 @@
...
+                low = mid + 1;
+            }
+        }
+
+        uint256 pos = _upperBinaryLookup(self._magnitudes, timestamp, low, high);
+
+        return pos == 0 ? (0, pos) : (_unsafeAccess(self._magnitudes, pos - 1)._value, pos - 1);
+    }
+
+    /// @notice WARNING: this function is only used because of the invariant property
```
gpsanant, line 241: can a library inherit another library, would prefer that...

src/contracts/core/AVSDirectory.sol
8sunyuan: we simply read the freeMagnitude to add back and add to running sum. We just update the `_nextDeallocationIndex` for the next time completeDeallocations is called

src/contracts/libraries/MagnitudeCheckpoints.sol
8sunyuan: `upperLookupRecent` returns value of 0 if there is no checkpoint, this should probably be changed to check value == 0 instead because there could be a checkpoint value at pos 0.

src/contracts/libraries/MagnitudeCheckpoints.sol
8sunyuan: The checkpoints upgradeable library has additional checkpoint structs and functions for different size keys, values pairs like below. In an attempt to save some additional bytecode bloat and to make the naming more clear I felt this was more convenient. Also wanted to add functions like `decrementCurrentAndFutureCheckpoints`.
Could be worth discussing

```
struct Trace160 {
      Checkpoint160[] _checkpoints;
  }

  struct Checkpoint160 {
      uint96 _key;
      uint160 _value;
  }
 ```

File: src/contracts/core/AVSDirectory.sol (hunk 3 above)
Comment (8sunyuan) [reply src/contracts/core/AVSDirectory.sol:511]: Something like below?

```
// read number of checkpoints after current timestamp
(uint224 value, uint256 pos) = _magnitudeUpdate[operator][strategy][operatorSets[i].avs][operatorSets[i]
    .operatorSetId].upperLookupRecentWithPos(uint32(block.timestamp));
// value is 0 if no checkpoint exists at current timestamp
if (value != 0) {
    require(
        pos == _magnitudeUpdate[operator][strategy][operatorSets[i].avs][operatorSets[i].operatorSetId].length() - 1,
        "AVSDirectory.queueAllocations: only one pending allocation allowed for op, opSet, strategy"
    );
}
```

PR #659, opened 2024-08-05
feat: limit total shares on `StrategyBase`

ypatil12:
In order to not brick offchain ZB infra for rewards, we limit the total number of shares to `1e38-1` in `StrategyBase`. All of our current strategies have total supply's FAR less than this number (max value uint120).


gpsanant (approved):

wadealexc (changes requested):
Just needs a docs update, then it's good.

PR #662, opened 2024-08-07
Feat: enumerating sets operators are in

0xClandestine:


0xClandestine:
Need to get coverage back up before merging...

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -43,22 +47,46 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
...
+    function isMember(address operator, OperatorSet memory operatorSet) public view returns (bool) {
+        return _operatorSetsMemberOf[operator].contains(_encodeOperatorSet(operatorSet));
+    }
+
+    function _encodeOperatorSet(OperatorSet memory operatorSet) internal view returns (bytes32) {
+        return bytes32(abi.encodePacked(operatorSet.avs, uint96(operatorSet.operatorSetId)));
+    }
+
+    function _decodeOperatorSet(bytes32 encoded) internal view returns (OperatorSet memory) {
+        return OperatorSet({
```
stevennevins, line 84: wdyt about using a custom type (OperatorSet) for bytes32

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -43,22 +47,46 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
...
+    mapping(address => EnumerableSet.Bytes32Set) internal _operatorSetsMemberOf;
+
     /// @notice Total amount of operators sets created.
     uint256 public operatorSetCount;
 
     constructor(IDelegationManager _delegation) {
         delegation = _delegation;
     }
 
+    function memberFor(address operator, uint256 index) public view returns (OperatorSet memory) {
```
ypatil12, line 67: This functionality should be in the implementation contract (as opposed to storage)

ypatil12 (commented):
1 organization comment. Also curious about the `memberFor` use case

src/contracts/core/AVSDirectoryStorage.sol
0xClandestine: Don't think it's needed here, will add better documentation though.

src/contracts/core/AVSDirectory.sol
```diff
@@ -149,40 +425,132 @@ contract AVSDirectory is
      *
      */
 
+    /// @notice Returns operator sets an operator is registered to in the order they were registered.
+    /// @param operator The operator address to query.
+    /// @param index The index of the enumerated list of operator sets.
+    function operatorSetsMemberOf(address operator, uint256 index) public view returns (OperatorSet memory) {
```
stevennevins, line 431: It feels like the CLI would want to just return the list of operator sets that the operator is in vs having to call to get the total and then calling with each index or paging through them

0xClandestine: There's a pagination fn that was just added for this `operatorSetsMemberOf(operator, start, length)`, pagination is needed to avoid gas limit related issues (if wanted).

stevennevins: Even paging through them doesn't feel ideal.  I think it's okay to assume there won't be 1000s of operator sets for an operator

PR #663, opened 2024-08-07
Feat/operator set rewards

ypatil12:
Pending Tests


src/contracts/core/RewardsCoordinator.sol
```diff
@@ -185,6 +207,55 @@ contract RewardsCoordinator is
...
+                rewardsSubmission.token,
+                rewardsSubmission.amount,
+                rewardsSubmission.startTimestamp,
+                rewardsSubmission.duration,
+                OPERATOR_SET_MAX_RETROACTIVE_LENGTH,
+                OPERATOR_SET_GENESIS_REWARDS_TIMESTAMP
+            );
+
+            isAVSRewardsSubmissionHash[msg.sender][rewardsSubmissionHash] = true;
+            submissionNonce[msg.sender] = nonce + 1;
```
8sunyuan, line 248: do we want submissionNonce mapping to be [msg.sender][operatorSetId]? Maybe overkill and not needed. But could be easy way to check number of rewards submissions per opSet

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -185,6 +207,55 @@ contract RewardsCoordinator is
...
+     *
+     * @dev msg.sender is the AVS in the operatorSet the rewards submission is being made to
+     * @dev AVSs set their reward type depending on what metric they want rewards
+     * distributed proportional to
+     * @dev The tokens in the rewards submissions are sent to the `RewardsCoordinator` contract
+     * @dev Strategies of each rewards submission must be in ascending order of addresses to check for duplicates
+     */
+    function rewardOperatorSetForRange(OperatorSetRewardsSubmission[] calldata rewardsSubmissions)
+        external
+        onlyWhenNotPaused(PAUSED_AVS_REWARDS_SUBMISSION)
```
8sunyuan, line 225: Lets use another pause flag to distinguish vs legacy interface

src/contracts/core/RewardsCoordinatorStorage.sol
```diff
@@ -77,24 +90,34 @@ abstract contract RewardsCoordinatorStorage is IRewardsCoordinator {
...
     mapping(address => uint256) public submissionNonce;
+
     /// @notice Mapping: avs => avsRewardsSubmissionHash => bool to check if rewards submission hash has been submitted
     mapping(address => mapping(bytes32 => bool)) public isAVSRewardsSubmissionHash;
+
     /// @notice Mapping: avs => rewardsSubmissionForALlHash => bool to check if rewards submission hash for all has been submitted
     mapping(address => mapping(bytes32 => bool)) public isRewardsSubmissionForAllHash;
-    /// @notice Mapping: address => bool to check if the address is permissioned to call createRewardsForAllSubmission
+
+    /// @notice Mapping: address => bool to check if the address is permissi oned to call createRewardsForAllSubmission
```
8sunyuan, line 100: nit: typo

src/contracts/core/RewardsCoordinator.sol
ypatil12: Interesting... I think there is a use case off chain but not necessarily on-chain for being able to get this #

ypatil12: Going to not implement for now unless we have a concrete use case

PR #664, opened 2024-08-07
fix: compile warnings

0xClandestine:


wadealexc:
image

Still getting some compiler warnings -- must be new since the rebase!

Looks good otherwise, thanks for the cleanup.

wadealexc (changes requested):
See above ^ should be good once these are fixed

ypatil12 (approved):

wadealexc (approved):

PR #676, opened 2024-08-10
Support third party withdrawals with queueWithdrawalsWithSignature

peitalin:
**Description**
Support third party withdrawals by adding a `queueWithdrawalWithSignature` function in DelegationManager so that a contract can queue withdrawals on behalf of a user with their signature.

**Use Cases**
At the moment a contract (e.g. a contract facilitating L2 <-> L1 restaking and withdrawing) can conduct deposits on behalf of a staker with `depositIntoStrategyWithSignature` in the StrategyManager, but can't queue withdrawals on behalf of a staker. Adding this feature would enable withdrawing via L2.

The token which we are using for Dual Staking has most of it's liquidity and users on L2, and so we wanted to allow our users to both stake and withdraw from Eigenlayer from L2 instead of: requiring our users to bridge ETH back to L1, queueWithdrawal from Eigenlayer, waiting for the withdrawal period, completeWithdrawal, then bridging back to L2.

**Update:**
The Eigenlayer team mentioned this feature existed in the past but was removed due to phishing concerns. 
I'm wondering if adding a `whitelistThirdPartyWithdrawer(contractAddr, strategy)` with checks in the `queueWithdrawalWithSignature` function addresses these concerns sufficiently.

At the moment, queueWithdrawals already checks if [thirdPartyTransfersForbidden = false](https://github.com/Layr-Labs/eigenlayer-contracts/blob/00fc4b95e9c1a5c4f370e41f56d01052d186da07/src/contracts/core/DelegationManager.sol#L696) to allow third party withdrawals, however `thirdPartyTransfersForbidden` defaults to `false` in [StrategyFactory.deployNewStrategy()](https://github.com/Layr-Labs/eigenlayer-contracts/blob/00fc4b95e9c1a5c4f370e41f56d01052d186da07/src/contracts/strategies/StrategyFactory.sol#L68), so I'm unsure if the team wants to add a separate `thirdPartyWithdrawalsForbidden = false` requirement to allow for different defaults for third party deposits, and third party withdrawals. 

Happy to add a separate `thirdPartyWithdrawalsForbidden` feature in the PR as well if the team prefers the latter.

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -316,47 +316,49 @@ contract DelegationManager is
      * All withdrawn shares/strategies are placed in a queue and can be fully withdrawn after a delay.
      */
     function queueWithdrawalsWithSignature(
-        QueuedWithdrawalParams[] calldata queuedWithdrawalParams,
-        address staker,
-        bytes memory signature
+        QueuedWithdrawalWithSignatureParams[] calldata queuedWithdrawalWithSigParams
     )
         external
         onlyWhenNotPaused(PAUSED_ENTER_WITHDRAWAL_QUEUE)
         returns (bytes32[] memory)
     {
-        bytes32[] memory withdrawalRoots = new bytes32[](queuedWithdrawalParams.length);
+        bytes32[] memory withdrawalRoots = new bytes32[](queuedWithdrawalWithSigParams.length);
         address operator = delegatedTo[msg.sender];
 
-        for (uint256 i = 0; i < queuedWithdrawalParams.length; i++) {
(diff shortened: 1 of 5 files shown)


peitalin:
Edit: Seems like this feature puts the DelegationManager contract size just above the size limit:
```
| Contract                    | Size (B) | Margin (B) |
|-----------------------------|----------|------------|
| DelegationManager           |   25,019 |       -443 |
```

However if we shorten some of the require error messages we see:
```
| Contract                    | Size (B) | Margin (B) |
|-----------------------------|----------|------------|
| DelegationManager           |   24,388 |        188 |
```

PR #680, opened 2024-08-13
feat: refactor to free all opsets on allo/deallo

8sunyuan:
Deallocations store the magnitudeDiff and completable timestamp in a (op, Strategy) mapping called `_pendingFreeMagnitude` which gets read from whenever allocate/deallocate is called to update freeMagnitude when possible. However, slashing would require brute force reading all these deallocations and additionally storing the respective operatorSet, so we also push the array index of the deallocation to an array in a (op, opSet, Strategy) mapping. This gets read upon each `slashOperator` call and provides the pointer lookup in `_pendingFreeMagnitude` to slash deallocations if needed.


src/contracts/core/AVSDirectory.sol
```diff
@@ -581,42 +586,40 @@ contract AVSDirectory is
...
-     * @notice Completes queued deallocations for a single strategy and multiple operator sets
-     * @param operator address to complete deallocations for
-     * @param strategy the strategy to complete deallocations for
-     * @param operatorSet the operator set to complete deallocations for the given strategy
-     */
-    function _completeDeallocation(
+    /// @dev read through pending free magnitudes and add to freeMagnitude if completableTimestamp is >= block timestamp
+    /// In additiona to updating freeMagnitude, updates next starting index to read from for pending free magnitudes after completing
+    function _updateFreeMagnitude(
         address operator,
```
gpsanant, line 604: probably want to set an `endBeforeIndex` here that is ignored if set to `0`

src/contracts/core/AVSDirectory.sol
```diff
@@ -581,42 +586,40 @@ contract AVSDirectory is
...
+            freeMagnitude[operator][strategy] += pendingFreeMagnitude.magnitudeDiff;
+            ++nextIndex;
         }
-        _nextDeallocationIndex[operator][strategy][operatorSet.avs][operatorSet.operatorSetId] = i;
-        return freeMagnitudeToAdd;
+        // update next pending free magnitude index to start from after adding all completable magnitudes
+        _nextPendingFreeMagnitudeIndex[operator][strategy] = nextIndex;
     }
 
     /// @dev Check for max number of pending deallocations, ensuring <= MAX_PENDING_UPDATES
```
gpsanant, line 625: rename from pending dealllocations

src/contracts/core/AVSDirectory.sol
```diff
@@ -625,13 +628,15 @@ contract AVSDirectory is
...
-        QueuedDeallocation[] memory queuedDeallocations =
-            _queuedDeallocations[operator][strategy][operatorSet.avs][operatorSet.operatorSetId];
-        uint256 length = queuedDeallocations.length;
-
+        uint256 length = _queuedDeallocationIndices[operator][strategy][operatorSet.avs][operatorSet.operatorSetId].length;
+                
         for (uint256 i = length; i > 0; --i) {
+            // index of pendingFreeMagnitude/deallocation to check for slashing
+            uint256 index = _queuedDeallocationIndices[operator][strategy][operatorSet.avs][operatorSet.operatorSetId][i - 1];
+            PendingFreeMagnitude memory pendingFreeMagnitude = _pendingFreeMagnitude[operator][strategy][index];
```
gpsanant, line 636: this will underflow if length = 0

src/contracts/core/AVSDirectory.sol
```diff
@@ -625,13 +628,15 @@ contract AVSDirectory is
         IStrategy strategy,
         OperatorSet calldata operatorSet
```
gpsanant, line 629: shouldn't this function be edited to not be operatorSet specific

src/contracts/core/AVSDirectory.sol
```diff
@@ -625,13 +628,15 @@ contract AVSDirectory is
...
     ) internal view returns (uint64 freeMagnitudeToAdd) {
```
gpsanant, line 630: doesn't even touch this return var?

src/contracts/core/AVSDirectory.sol
```diff
@@ -372,22 +369,30 @@ contract AVSDirectory is
...
+                    // therefore older deallocations will also be completable. Since this is ordered by completableTimestamps break loop now
+                    if (pendingFreeMagnitude.completableTimestamp >= uint32(block.timestamp)) {
                         break;
                     }
+
+                    // pending deallocation is still within slashable window, slash magnitudeDiff and add to slashedMagnitude
+                    uint64 slashedAmount =
+                            uint64(uint256(bipsToSlash) * uint256(pendingFreeMagnitude.magnitudeDiff) / BIPS_FACTOR);
+                    pendingFreeMagnitude.magnitudeDiff -= slashedAmount;
+                        slashedMagnitude += slashedAmount;
```
gpsanant, line 395: weird indent prob need tabs/spaces

src/contracts/core/AVSDirectory.sol
```diff
@@ -372,22 +369,30 @@ contract AVSDirectory is
...
+            // if NOT completable, then slash
             {
-                uint256 queuedDeallocationsLen =
-                    _queuedDeallocations[operator][strategies[i]][msg.sender][operatorSetId].length;
-                for (uint256 j = queuedDeallocationsLen; j > 0; --j) {
-                    QueuedDeallocation storage queuedDeallocation =
-                        _queuedDeallocations[operator][strategies[i]][msg.sender][operatorSetId][j - 1];
-
-                    // if queued deallocation is still within slashable window, then slash and keep track of sum to decrement from totalMagnitude
-                    if (uint32(block.timestamp) + ALLOCATION_DELAY > queuedDeallocation.completableTimestamp) {
```
gpsanant, line 383: man what was this check :)

src/contracts/core/AVSDirectory.sol
```diff
@@ -581,42 +586,40 @@ contract AVSDirectory is
...
-                QueuedDeallocation({
+            // 4. push PendingFreeMagnitude and respective array index into (op,opSet,Strategy) queued deallocations
+            uint256 index = _pendingFreeMagnitude[operator][strategy].length;
+            _pendingFreeMagnitude[operator][strategy].push(
+                PendingFreeMagnitude({
                     magnitudeDiff: deallocation.magnitudeDiffs[i],
                     completableTimestamp: completableTimestamp
                 })
             );
+            _queuedDeallocationIndices[operator][strategy][operatorSets[i].avs][operatorSets[i].operatorSetId].push(index);
```
gpsanant, line 597: imo its time combine `_magnitudeUpdates` and `_queuedDeallocationIndices`

src/contracts/core/AVSDirectory.sol
8sunyuan: too restrictive? allocations are limited to be 1 pending allocation per (op, opSet, Strategy).
Would be a bit inconsistent if deallos are (op, opSet) but its definitely a uglier check here vs just reading from pendingFreeMagnitude array

src/contracts/core/AVSDirectory.sol
8sunyuan: brain melt

src/contracts/core/AVSDirectory.sol
8sunyuan: yeaaah forge fmt

src/contracts/core/AVSDirectory.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/680/commits/d33bceec29869d0ee995a73cefc5c98e08d72c27

src/contracts/core/AVSDirectory.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/680/commits/d33bceec29869d0ee995a73cefc5c98e08d72c27
Changed to `_checkQueuedDeallocations`, since we have pendingFreeMagnitude now

src/contracts/core/AVSDirectory.sol
8sunyuan: Think we are okay with just completing all. This will be called in both allocate/deallocate functions to bound the "numToComplete" from increasing past 1.

src/contracts/core/AVSDirectory.sol
gpsanant: didn't understand usage, my bad

src/contracts/core/AVSDirectory.sol
8sunyuan: Synced offline but there is a bound on (op, strategy, opSet) queuedDeallos but not on pendingFreeMagnitude(op,strategy) because multiple opSets can be queued and not completed

src/contracts/core/AVSDirectory.sol
```diff
@@ -276,9 +276,18 @@ contract AVSDirectory is
         uint32 effectTimestamp = uint32(block.timestamp) + ALLOCATION_DELAY;
         for (uint256 i = 0; i < allocations.length; ++i) {
             // 1. For the given (operator,strategy) clear all pending free magnitude for the strategy and update freeMagnitude
-            _updateFreeMagnitude(operator, allocations[i].strategy);
+            // numToComplete = 0 defaults to completing all pending deallocations up to uint8 max (256)
+            _updateFreeMagnitude({
```
gpsanant, line 280: shouldn't we accept `numToComplete` to this interface?

src/contracts/core/AVSDirectory.sol
```diff
@@ -302,9 +311,18 @@ contract AVSDirectory is
         // deallocate for each strategy
         for (uint256 i = 0; i < deallocations.length; ++i) {
             // 1. For the given (operator,strategy) clear all pending free magnitude for the strategy and update freeMagnitude
-            _updateFreeMagnitude(operator, deallocations[i].strategy);
+            // numToComplete = 0 defaults to completing all pending deallocations up to uint8 max (256)
+            _updateFreeMagnitude({
+                operator: operator, 
+                strategy: deallocations[i].strategy,
+                numToComplete: 0
```
gpsanant, line 318: same here

gpsanant (approved):

src/contracts/core/AVSDirectory.sol
8sunyuan: opting to keep separate mappings otherwise lookup for indices in slashOperator would require looking through future magnitudeUpdates as well

src/contracts/core/AVSDirectory.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/680/commits/c3d36fdef27069dc8b643ecba7b638963756a7a6

src/contracts/core/AVSDirectory.sol
8sunyuan: leaving out of interface for simplicity, if gas costs are an issue then can call `updateFreeMagnitude` separately

PR #683, opened 2024-08-13
feat: performance based rewards

0xClandestine:


src/contracts/core/RewardsCoordinator.sol
```diff
@@ -82,7 +84,9 @@ contract RewardsCoordinator is
         uint32 _MAX_FUTURE_LENGTH,
         uint32 _GENESIS_REWARDS_TIMESTAMP,
         uint32 _OPERATOR_SET_GENESIS_REWARDS_TIMESTAMP,
-        uint32 _OPERATOR_SET_MAX_RETROACTIVE_LENGTH
+        uint32 _OPERATOR_SET_MAX_RETROACTIVE_LENGTH,
+        uint32 _GENESIS_PERFORMANCE_REWARDS_TIMESTAMP,
```
ypatil12, line 88: Expecting this to be the same, since performance based rewards are also operator set based. Don't make the change yet since still deciding.

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -229,35 +238,65 @@ contract RewardsCoordinator is
...
-                ++i;
-            }
+            // Emit event to track reward submission.
+            emit OperatorSetRewardCreated(msg.sender, currentNonce, rewardsSubmissionHash, rewardsSubmission);
         }
+        // Increase avs submission nonce by rewards submissions submitted.
+        submissionNonce[msg.sender] = startingNonce + rewardsSubmissions.length;
+    }
+
+    function rewardOperatorsForPerformance(PerformanceRewardsSubmission calldata rewardsSubmission)
```
ypatil12, line 274: Any reason why this isn't an array? Assume call data will just get too big? That's fine with me.

ypatil12: Do you think we should make the other reward submission functions just structs instead of arrays of structs? cc @8sunyuan

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -229,35 +238,65 @@ contract RewardsCoordinator is
...
+        external
+        onlyWhenNotPaused(PAUSED_REWARD_PERFORMANCE)
+        nonReentrant
+    {
+        // Cache nonce for later use.
```
ypatil12, line 279: Nit: the `startTimestamp + duration` should be < block.timestamp as we want the reward to be completely retroactive.

src/contracts/interfaces/IRewardsCoordinator.sol
```diff
@@ -234,6 +247,13 @@ interface IRewardsCoordinator {
         uint256 claimedAmount
     );
 
+    event PerformanceBasedRewardCreated(
+        address indexed avs,
+        uint256 indexed submissionNonce,
+        bytes32 indexed rewardsSubmissionHash,
+        PerformanceRewardsSubmission rewardsSubmission
```
ypatil12, line 254: Can we emit the RewardType here too? Just because commissions are set based on the RewardType

src/contracts/interfaces/IRewardsCoordinator.sol
```diff
@@ -93,6 +94,18 @@ interface IRewardsCoordinator {
         uint32 duration;
     }
 
+    /// @notice Used for performance based reward submissions.
+    struct PerformanceRewardsSubmission {
+        uint32 operatorSetId;
+        address[] operators;
+        uint32[] scores;
+        IStrategy[] strategies;
```
ypatil12, line 102: We may have to update this to `StrategiesAndMultipliers`. Differs from spec. Will keep you updated

ypatil12 (commented):
Minor comments. Appreciate the optimizations : )

src/contracts/interfaces/IRewardsCoordinator.sol
```diff
@@ -93,6 +94,18 @@ interface IRewardsCoordinator {
         uint32 duration;
     }
 
+    /// @notice Used for performance based reward submissions.
+    struct PerformanceRewardsSubmission {
```
ypatil12, line 98: Can we also add natspec for each of these. Need to describe what scores & operators are.

src/contracts/core/RewardsCoordinator.sol
ypatil12, line 274: `scores.length` must equal `operators.length`

src/contracts/interfaces/IRewardsCoordinator.sol
ypatil12: uint32[] scores specify how the AVS subjectively values each operator's performance. 

strategies[] specify which strategies to reward given stakers of an operator. Note: we likely have to change this to `StrategiesAndMultipliers` to be able to calculate the `stakeWeight` of a staker for performance based rewards.

src/contracts/interfaces/IRewardsCoordinator.sol
```diff
@@ -349,6 +369,8 @@ interface IRewardsCoordinator {
      */
     function rewardOperatorSetForRange(OperatorSetRewardsSubmission[] calldata rewardsSubmissions) external;
 
+    function rewardOperatorsForPerformance(PerformanceRewardsSubmission calldata rewardsSubmission) external;
```
ypatil12, line 372: Nit: natspec

src/contracts/interfaces/IRewardsCoordinator.sol
```diff
@@ -19,7 +19,8 @@ interface IRewardsCoordinator {
 
     /// @notice Reward type
     enum RewardType {
-        DELEGATED_STAKE
+        DELEGATED_STAKE,
+        PERFORMANCE
```
ypatil12, line 23: Shouldn't we only allow commissions to be set for valid reward types? cc @8sunyuan ?

src/contracts/interfaces/IRewardsCoordinator.sol
ypatil12, line 254: Can we emit the reward type here too?

ypatil12 (commented):
Additional comments

src/contracts/core/RewardsCoordinator.sol
0xClandestine: Okay wasn't clear on that, will fix.

src/contracts/core/RewardsCoordinator.sol
0xClandestine: It wasn't defined as an array in the notion doc, was curious about that though.

src/contracts/core/RewardsCoordinator.sol
0xClandestine: Sounds like we need another check in _validateRewardsSubmission()?

src/contracts/interfaces/IRewardsCoordinator.sol
0xClandestine: I tried adding natspec, I'm just not sure its correct. Might be a good idea for you to add docs since you're more familiar with off-chain.

src/contracts/core/RewardsCoordinator.sol
0xClandestine: Nice catch. Will fix.

src/contracts/interfaces/IRewardsCoordinator.sol
0xClandestine: Read above^^^

src/contracts/interfaces/IRewardsCoordinator.sol
0xClandestine: Yessir.

0xClandestine: Should we also be doing this for the other events, and while we're on the topic I feel like event names for reward submissions should be more consistent since they all serve the same purpose, generally.

ypatil12: Let's chat about this one tmrw?

src/contracts/core/RewardsCoordinator.sol
ypatil12: Nah, because only this reward cannot go in the future. The rest of the rewards can. Doesn't make sense to give a reward for performance in the future if there is no performance yet :)

src/contracts/core/RewardsCoordinator.sol
ypatil12: Assume we are EigenDA and have 200 operators & 200 scores. What would be the max size of the array?

ypatil12: My guess is definitely < 5 elements? But also doing an array is consistent with how we are have the rest of the functions, just unsure if its practical..

gpsanant:
can we close this @0xClandestine @ypatil12

PR #687, opened 2024-08-15
init

Sidu28:


Sidu28:
still runs out of gas for me
Screenshot 2024-08-15 at 10 21 31 PM

PR #691, opened 2024-08-16
feat: set allocations

8sunyuan:
Draft for changing allocate/deallocate to just `setAllocations` which can specify absolute magnitude values vs previously we had `magnitudeDiffs`. Depending on the current magnitude value assigned for the (op, Strategy, opSet) tuple, it will either allocate/deallocate accordingly.

NOTE: in `setAllocations` any magnitude values that are equal to the current value are ignored as it is neither a allocation or deallocation. 

Additional changes included:
- expectedTotalMagnitude check
- initializeable allocationDelays that could be set one time by an operator's allocator
- Checkpoints library optimization: reuse length check in `upperLookupRecentWithPos`

PR #694, opened 2024-08-17
stakeRoot

gpsanant:
builds off of #691 

the StakeRoot protocol


gpsanant:
crazy git things happened, see #715

PR #695, opened 2024-08-18
start on scaledShares

gpsanant:
starting yet another time on share scaling

DM changes started. lots of work left on SM and EPM and tests

TODO:

- [ ] Need to fix pre-upgrade withdrawals and converting between `startBlock` to `startTimestamp` as part of the Withdrawal struct handling. 
- [X] ~initializeAllocationDelay in AVSD/Slasher? Or in DM upon operator registration~ Moved it to DelegationManager
- [ ] Effects of lagging state of beacon chain slashing and a simultaneous slashing event on core contracts


src/contracts/core/DelegationManager.sol
```diff
@@ -40,6 +41,12 @@ contract DelegationManager is
     // @dev Maximum Value for `stakerOptOutWindowBlocks`. Approximately equivalent to 6 months in blocks.
     uint256 public constant MAX_STAKER_OPT_OUT_WINDOW_BLOCKS = (180 days) / 12;
 
+    // the number of 12-second blocks in 30 days (60 * 60 * 24 * 30 / 12 = 216,000)
+    uint256 public constant MAX_WITHDRAWAL_DELAY_BLOCKS = 216_000;
+
+    // The max configurable withdrawal delay per strategy. Set to 30 days in seconds
+    uint256 public constant MAX_WITHDRAWAL_DELAY = 2_592_000;
```
gpsanant, line 48: just use `30 days`

src/contracts/core/DelegationManager.sol
```diff
@@ -40,6 +41,12 @@ contract DelegationManager is
     // @dev Maximum Value for `stakerOptOutWindowBlocks`. Approximately equivalent to 6 months in blocks.
     uint256 public constant MAX_STAKER_OPT_OUT_WINDOW_BLOCKS = (180 days) / 12;
 
+    // the number of 12-second blocks in 30 days (60 * 60 * 24 * 30 / 12 = 216,000)
+    uint256 public constant MAX_WITHDRAWAL_DELAY_BLOCKS = 216_000;
```
gpsanant, line 45: just use `30 days / 12`

src/contracts/core/DelegationManager.sol
```diff
@@ -311,93 +327,87 @@ contract DelegationManager is
...
      * will simply be transferred to the caller directly.
-     * @dev middlewareTimesIndex is unused, but will be used in the Slasher eventually
      * @dev beaconChainETHStrategy shares are non-transferrable, so if `receiveAsTokens = false` and `withdrawal.withdrawer != withdrawal.staker`, note that
      * any beaconChainETHStrategy shares in the `withdrawal` will be _returned to the staker_, rather than transferred to the withdrawer, unlike shares in
      * any other strategies, which will be transferred to the withdrawer.
      */
     function completeQueuedWithdrawal(
         Withdrawal calldata withdrawal,
         IERC20[] calldata tokens,
-        uint256 middlewareTimesIndex,
```
gpsanant, line 327: we should have never had this, but we can't break this interface

src/contracts/core/DelegationManager.sol
8sunyuan: Not sure why I did this

src/contracts/core/DelegationManager.sol
8sunyuan: We absolutely can't break this? Is the assumption for our TUPS contracts is that all interfaces are final?
I would deem that for a smart contract system that is not fully implemented (missing rewards/slashing) its pretty reasonable to expect interfaces to change. But I guess it's not much benefit for a lot of downstream impact.

ypatil12: Slashing is a breaking change, feel like this is fine? Worst case, add the new version and say we will deprecate the `middlewareTimesIndex` post slashing.

src/contracts/core/DelegationManager.sol
```diff
@@ -231,9 +238,9 @@ contract DelegationManager is
             "DelegationManager.undelegate: caller cannot undelegate staker"
```
gpsanant, line 238: require allocation delay to register

src/contracts/core/DelegationManager.sol
```diff
@@ -253,14 +260,14 @@ contract DelegationManager is
                 IStrategy[] memory singleStrategy = new IStrategy[](1);
                 uint256[] memory singleShare = new uint256[](1);
                 singleStrategy[0] = strategies[i];
-                singleShare[0] = shares[i];
+                singleShare[0] = scaledShares[i];
```
gpsanant, line 263: singleScaledShare?

src/contracts/core/DelegationManager.sol
gpsanant: slashing is not a breaking change to deposits/withdrawals if we don't want it to be

src/contracts/core/DelegationManager.sol
```diff
@@ -528,132 +560,174 @@ contract DelegationManager is
...
-    // @notice Decreases `operator`s delegated shares in `strategy` by `shares` and emits an `OperatorSharesDecreased` event
-    function _decreaseOperatorShares(address operator, address staker, IStrategy strategy, uint256 shares) internal {
+    // @notice Decreases `operator`s delegated scaled shares in `strategy` by `scaledShares` and emits an `OperatorSharesDecreased` event
+    function _decreaseOperatorScaledShares(address operator, address staker, IStrategy strategy, uint256 scaledShares) internal {
         // This will revert on underflow, so no check needed
-        operatorShares[operator][strategy] -= shares;
-        emit OperatorSharesDecreased(operator, staker, strategy, shares);
+        operatorScaledShares[operator][strategy] -= scaledShares;
+        // TODO: What to do about event wrt scaling?
+        emit OperatorSharesDecreased(operator, staker, strategy, scaledShares);
```
gpsanant, line 726: @ypatil12 @bowenli86 what do you think about pipeline/event concerns with slashing?

src/contracts/core/DelegationManager.sol
```diff
@@ -528,132 +560,174 @@ contract DelegationManager is
...
+
+        // read delegated operator for scaling and adding shares back if needed
+        address currentOperator = delegatedTo[msg.sender];
+        // descale shares to get the "real" strategy share values of the withdrawal
+        uint256[] memory descaledShares = ShareScalingLib.descaleSharesAtTimestamp(
+            avsDirectory,
+            withdrawal.delegatedTo,
+            withdrawal.strategies,
+            withdrawal.scaledShares,
+            withdrawal.startTimestamp + avsDirectory.DEFAULT_ALLOCATION_DELAY()
```
gpsanant, line 599: need the deallocation delay here?

src/contracts/core/DelegationManager.sol
```diff
@@ -510,15 +534,23 @@ contract DelegationManager is
         delegatedTo[staker] = operator;
         emit StakerDelegated(staker, operator);
 
-        (IStrategy[] memory strategies, uint256[] memory shares) = getDelegatableShares(staker);
+        (IStrategy[] memory strategies, uint256[] memory shares) = getDelegatableScaledShares(staker);
+
+        // NOTE: technically the function returns scaled shares, so it needs to be descaled before scaling to the
+        // the new operator according to their totalMagnitude. However, since the staker is not delegated,
+        // scaledShares = shares so we can skip descaling and read these as descaled shares.
+        uint256[] memory scaledShares = ShareScalingLib.scaleShares(avsDirectory, operator, strategies, shares);
```
gpsanant, line 542: we don't need to scale here

src/contracts/core/DelegationManager.sol
```diff
@@ -510,15 +534,23 @@ contract DelegationManager is
         delegatedTo[staker] = operator;
         emit StakerDelegated(staker, operator);
 
-        (IStrategy[] memory strategies, uint256[] memory shares) = getDelegatableShares(staker);
+        (IStrategy[] memory strategies, uint256[] memory shares) = getDelegatableScaledShares(staker);
```
gpsanant, line 537: call this `scaledShares`

src/contracts/core/DelegationManager.sol
```diff
@@ -528,132 +560,174 @@ contract DelegationManager is
...
         );
+        require(
+            tokens.length == withdrawal.strategies.length,
+            "DelegationManager._completeQueuedWithdrawal: input length mismatch"
+        );
+
+        // read delegated operator for scaling and adding shares back if needed
+        address currentOperator = delegatedTo[msg.sender];
+        // descale shares to get the "real" strategy share values of the withdrawal
+        uint256[] memory descaledShares = ShareScalingLib.descaleSharesAtTimestamp(
```
gpsanant, line 594: i feel like we should descale only at the SM/EPM level?

gpsanant: unless completing as shares

gpsanant: actually, descaling here is fine, reduces complexity in EPM and SM

src/contracts/core/DelegationManager.sol
```diff
@@ -528,132 +560,174 @@ contract DelegationManager is
...
+        uint256[] memory scaledShares = ShareScalingLib.scaleShares(
+            avsDirectory,
+            currentOperator,
+            withdrawal.strategies,
+            descaledShares
+        );
 
+        for (uint256 i = 0; i < withdrawal.strategies.length;) {
+            require(
+                withdrawal.startTimestamp + strategyWithdrawalDelays[withdrawal.strategies[i]] <= block.number,
```
gpsanant, line 668: This doesn't seem to work for uncompleted withdrawals from before the upgrade?

src/contracts/core/DelegationManager.sol
8sunyuan: > i feel like we should descale only at the SM/EPM level?

tried to keep as much scaling/descaling done at DM level to reduce complexity. Since deposits aren't done here however there will have to be some helper view calls to scale share prior to depositing `stakerStrategyScaleShares` or `podOwnerShares`

src/contracts/core/DelegationManager.sol
8sunyuan: Yes, you need to read the totalMagnitude at time of completion, not the current totalMagnitude in the existing helper functions

src/contracts/core/DelegationManager.sol
8sunyuan: We do, the staker is not currently delegated to anybody so their "scaledShares" aren't scaled accordingly to the new operator they are delegating to

src/contracts/core/DelegationManager.sol
8sunyuan: gp, added to pr description

src/contracts/core/StrategyManager.sol
```diff
@@ -262,90 +265,93 @@ contract StrategyManager is
...
-                "StrategyManager._addShares: deposit would exceed MAX_STAKER_STRATEGY_LIST_LENGTH"
+                "StrategyManager._addScaledShares: deposit would exceed MAX_STAKER_STRATEGY_LIST_LENGTH"
             );
             stakerStrategyList[staker].push(strategy);
         }
 
-        // add the returned shares to their existing shares for this strategy
-        stakerStrategyShares[staker][strategy] += shares;
+        // add the returned scaled shares to their existing scaled shares for this strategy
+        stakerStrategyScaledShares[staker][strategy] += scaledShares;
```
gpsanant, line 292: can we call back to the DelegationManager here to increase shares?

src/contracts/core/DelegationManager.sol
```diff
@@ -510,150 +534,205 @@ contract DelegationManager is
...
-                        _increaseOperatorShares({
-                            operator: currentOperator,
-                            // the 'staker' here is the address receiving new shares
-                            staker: msg.sender,
-                            strategy: withdrawal.strategies[i],
-                            shares: withdrawal.shares[i]
-                        });
-                    }
+                uint256 increaseInDelegateableScaledShares =
+                    eigenPodManager.addShares({podOwner: staker, shares: scaledShares[i]});
```
gpsanant, line 688: addScaledShares

src/contracts/core/DelegationManager.sol
```diff
@@ -510,150 +534,205 @@ contract DelegationManager is
...
-                shares: shares[i]
+                scaledShares: scaledShares[i]
             });
 
+            // stakerStrategyScaledShares(staker, strategies[i]) needs to be updated to be equal to newly scaled shares.
+            // we take the difference between the shares and the scaledShares to update the staker's scaledShares. The key property
+            // here is that scaledShares should be <= shares and can never be greater than shares due to totalMagnitude being
+            // a monotonically decreasing value.
+            if (shares[i] > scaledShares[i]) {
+                strategyManager.removeScaledShares(staker, strategies[i], shares[i] - scaledShares[i]);
```
gpsanant, line 558: should we remove here or "set"?

src/contracts/core/DelegationManager.sol
gpsanant: emphasis on the DEallocation

src/contracts/core/DelegationManager.sol
gpsanant: hm, i guess this is fine for now. will need to be changed upon concentration

src/contracts/core/DelegationManager.sol
```diff
@@ -510,150 +534,205 @@ contract DelegationManager is
...
                 strategy: strategies[i], 
-                shares: shares[i]
+                scaledShares: scaledShares[i]
             });
 
+            // stakerStrategyScaledShares(staker, strategies[i]) needs to be updated to be equal to newly scaled shares.
+            // we take the difference between the shares and the scaledShares to update the staker's scaledShares. The key property
+            // here is that scaledShares should be <= shares and can never be greater than shares due to totalMagnitude being
+            // a monotonically decreasing value.
+            if (shares[i] > scaledShares[i]) {
```
gpsanant, line 557: doesn't work for beaconChainETH

src/contracts/core/StrategyManager.sol
```diff
@@ -262,90 +265,93 @@ contract StrategyManager is
...
+        scaledShares = delegation.getStakerScaledShares(staker, strategy, shares);
 
-        // Increase shares delegated to operator, if needed
-        delegation.increaseDelegatedShares(staker, strategy, shares);
+        // add the returned scaled shares to the staker's existing scaled shares for this strategy
+        _addScaledShares(staker, token, strategy, scaledShares);
 
-        return shares;
+        // Increase scaled shares delegated to operator, if needed
+        delegation.increaseDelegatedScaledShares(staker, strategy, scaledShares);
```
gpsanant, line 325: wonder if a `increaseDelegatedShares` that scales and returns the scaled shares makes more sense than calling DM twice to scale and then increase?

src/contracts/pods/EigenPodManager.sol
```diff
@@ -123,16 +123,16 @@ contract EigenPodManager is
         // skip making a call to the DelegationManager if there is no change in delegateable shares
```
gpsanant, line 123: we need to scale inside of `_calculateChangeInDelegatableShares`

gpsanant: and update podOwnerShares after

src/contracts/core/DelegationManager.sol
```diff
@@ -40,6 +41,12 @@ contract DelegationManager is
     // @dev Maximum Value for `stakerOptOutWindowBlocks`. Approximately equivalent to 6 months in blocks.
     uint256 public constant MAX_STAKER_OPT_OUT_WINDOW_BLOCKS = (180 days) / 12;
 
+    // the number of 12-second blocks in 30 days (60 * 60 * 24 * 30 / 12 = 216,000)
+    uint256 public constant MAX_WITHDRAWAL_DELAY_BLOCKS = 30 days / 12;
```
wadealexc, line 45: Should be `MAX_WITHDRAWAL_DELAY / 12`, rather than reusing the "30 days" value

src/contracts/core/DelegationManager.sol
wadealexc: ew no

src/contracts/libraries/ShareScalingLib.sol
```diff
@@ -0,0 +1,111 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.12;
+
+import "../interfaces/IAVSDirectory.sol";
+
+library ShareScalingLib {
```
wadealexc, line 6: Why do we need a library for this instead of just calling a function on the AVSDirectory? It complicates the callstack and most of the work is being done in the AVSDirectory anyway..

Is this solely to save on code size?

src/contracts/core/DelegationManager.sol
```diff
@@ -311,93 +327,87 @@ contract DelegationManager is
...
-     * @dev *If the staker is actively delegated*, then increases the `staker`'s delegated shares in `strategy` by `shares`. Otherwise does nothing.
+     * @dev *If the staker is actively delegated*, then increases the `staker`'s delegated shares in `strategy` by `scaledShares`. Otherwise does nothing.
      * @dev Callable only by the StrategyManager or EigenPodManager.
      */
-    function increaseDelegatedShares(
+    function increaseDelegatedScaledShares(
         address staker,
         IStrategy strategy,
-        uint256 shares
+        uint256 scaledShares
```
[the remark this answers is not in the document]
ypatil12: 3rd parties can just convert shares into nominal amounts via helper functions

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -42,17 +46,16 @@ abstract contract DelegationManagerStorage is IDelegationManager {
...
+     * @notice Mapping: operator => strategy => total number of scaled shares in the strategy delegated to the operator.
      * @dev By design, the following invariant should hold for each Strategy:
-     * (operator's shares in delegation manager) = sum (shares above zero of all stakers delegated to operator)
-     * = sum (delegateable shares of all stakers delegated to the operator)
+     * (operator's scaled shares in delegation manager) = sum (scaled shares above zero of all stakers delegated to operator)
+     * = sum (delegateable scaled shares of all stakers delegated to the operator)
+     * @dev FKA `operatorShares`
      */
-    mapping(address => mapping(IStrategy => uint256)) public operatorShares;
+    mapping(address => mapping(IStrategy => uint256)) public operatorScaledShares;
```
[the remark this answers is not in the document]
ypatil12: Alternatively: https://github.com/Layr-Labs/eigenlayer-contracts/pull/695#discussion_r1731617824. Just trying to think of ways to simplify the naming or even accounting here

src/contracts/core/DelegationManager.sol
wadealexc: Let's break the interface. Fuck this param, it's time

wadealexc: > slashing is not a breaking change to deposits/withdrawals if we don't want it to be

i'd like it to be so we can kill this param

src/contracts/core/DelegationManager.sol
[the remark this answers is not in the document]
ypatil12: Actually, this is a better commen: https://github.com/Layr-Labs/eigenlayer-contracts/pull/695#discussion_r1731745687

src/contracts/core/DelegationManager.sol
wadealexc: But they're not scaled at this point, because the staker isn't delegated? If i understand the comment below correctly

Calling the "getXScaled" function here is confusing tho...

src/contracts/core/DelegationManager.sol
ypatil12: Need to sit down and look through all events to make sure we are covered

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/695/commits/287cdceb9484cf4257bfcec2da74319bba547fe1

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/695/commits/2a8d0eb2f873c733791ba1e3f5b3ac8eda3e79b7

src/contracts/core/StrategyManager.sol
8sunyuan: Now I'm wondering if we should just store `stakerStrategyScaledShares` and `podOwnerScaledShares` in the DelegationManager if all deposits are going to require calling the DelegationManager to scale regardless if the staker is delegated or not hmmm

src/contracts/pods/EigenPodManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/695/commits/2a8d0eb2f873c733791ba1e3f5b3ac8eda3e79b7
https://github.com/Layr-Labs/eigenlayer-contracts/pull/695/commits/c54e9c7c6e808ecdbdf34d6400dd29bbcd326b34
renamed to indicate shares

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/695/commits/49a2cd544dbf2abd0d498706621422833c8a49ab

src/contracts/libraries/ShareScalingLib.sol
8sunyuan: I think it was mainly used as a means to abstracting the scaling logic separately from the contract. These functions are internal so they don't save bytecode size. They are also not pure functions either and depend on reading state from the core contracts.
There's probably not a good enough reason to make this a library as only the DelegationManager will be calling these functions. AVSDirectory simply uses the `INITIAL_TOTAL_MAGNITUDE` constant.

src/contracts/core/DelegationManager.sol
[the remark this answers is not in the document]
8sunyuan: Probably a good idea to make the `scaledShares` mappings internal as they are not entirely useful for integrators.

src/contracts/core/AVSDirectory.sol
```diff
@@ -840,6 +838,52 @@ contract AVSDirectory is
...
+
+            unchecked {
+                ++i;
+            }
+        }
+        return totalMagnitudes;
+    }
+
+    /// @notice Returns the total magnitude of an operator for a given set of strategies at a given timestamp
+    function getTotalMagnitudesAtTimestamp(
```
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/695/commits/d9eb842932a8d3c2fff1313261ba4b631ef58afa

src/contracts/core/AVSDirectory.sol
```diff
@@ -840,6 +838,52 @@ contract AVSDirectory is
             || registrationStatus.lastDeregisteredTimestamp + DEALLOCATION_DELAY >= block.timestamp;
     }
 
+    /// @notice Returns the total magnitude of an operator for a given set of strategies
+    function getTotalMagnitudes(address operator, IStrategy[] calldata strategies) public view returns (uint64[] memory) {
```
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/695/commits/d9eb842932a8d3c2fff1313261ba4b631ef58afa

src/contracts/libraries/ShareScalingLib.sol
8sunyuan: AVSDirectory is used however to read the totalMagnitudes

src/contracts/libraries/ShareScalingLib.sol
```diff
@@ -0,0 +1,111 @@
...
+    function scaleShares(
+        IAVSDirectory avsDirectory,
+        address operator,
+        IStrategy[] memory strategies,
+        uint256[] memory shares
+    ) internal view returns (uint256[] memory) {
+        uint64[] memory totalMagnitudes = avsDirectory.getTotalMagnitudes(operator, strategies);
+        uint256[] memory scaledShares = new uint256[](shares.length);
+        for (uint256 i = 0; i < shares.length; i++) {
+            scaledShares[i] = shares[i] * INITIAL_TOTAL_MAGNITUDE / totalMagnitudes[i];
```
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/695/commits/3ce2b04f176f2c3e959ad95fb056d4a105cfe2ad

src/contracts/core/DelegationManager.sol
8sunyuan: They are technically "scaled" but because its scaled based on operator address(0) there is no difference
because the following is true `scaleShares(shares) == shares`.

So rather than doing `descaledShares = descaleShares(scaledShares)` prior to scaling shares according to the new delegated operator, we can just skip descaling the shares.

8sunyuan: I used the variable name here as `shares` to indicate it is in a sense already "descaled" but I admit it is very confusing

src/contracts/core/DelegationManager.sol
8sunyuan: brainfart https://github.com/Layr-Labs/eigenlayer-contracts/pull/695/commits/287cdceb9484cf4257bfcec2da74319bba547fe1#diff-d323417ded59a593f4ea410225a9167b24198c4d2c05cbc0fe9af47d34b88810R643

src/contracts/core/DelegationManager.sol
8sunyuan: If we don't kill middlewaretimes at this moment of time we never will.

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/695/commits/435d73d5fb2a31da76cf434dc9bc0b51b411f0d2

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -102,6 +105,22 @@
...
 
+    /**
+     * @notice Global minimum withdrawal delay for all strategy withdrawals.
+     * NOTE: This is used equivalently to minWithdrawalDelayBlocks and strategyWithdrawalDelayBlocks but
+     * using timestamps instead in the Slashing release.
+     * In addition, we now also configure withdrawal delays on a per-strategy basis.
+     * To withdraw from a strategy, max(minWithdrawalDelay, strategyWithdrawalDelay[strategy]) number of seconds must have passed.
+     * See mapping strategyWithdrawalDelay below for per-strategy withdrawal delays.
+     */
+    uint256 public minWithdrawalDelay;
```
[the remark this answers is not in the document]
8sunyuan: Definitely need to add but will address later when we account for changing the `startBlocknum` to `startTimestamp`

src/contracts/core/StrategyManager.sol
8sunyuan: Definitely a good optimization, will add this later because I think its easier to view contracts diff as currently implemented.

src/contracts/libraries/ShareScalingLib.sol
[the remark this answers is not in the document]
8sunyuan: Synced offline but we might just use magnitude amounts in the `slashOperator` interface. TBD

src/contracts/core/DelegationManagerStorage.sol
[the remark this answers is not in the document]
8sunyuan: Agreed on this. Will see other peoples opinions

src/contracts/core/DelegationManager.sol
gpsanant: ah yes, these aren't scaled. confusing definitely

src/contracts/core/DelegationManager.sol
```diff
@@ -113,18 +114,30 @@ contract DelegationManager is
...
+        uint32 allocationDelay,
         string calldata metadataURI
     ) external {
         require(!isDelegated(msg.sender), "DelegationManager.registerAsOperator: caller is already actively delegated");
         _setOperatorDetails(msg.sender, registeringOperatorDetails);
         SignatureWithExpiry memory emptySignatureAndExpiry;
         // delegate from the operator to themselves
         _delegate(msg.sender, msg.sender, emptySignatureAndExpiry, bytes32(0));
+        // set the allocation delay
+        _initializeAllocationDelay(allocationDelay);
```
gpsanant, line 126: this should call into the AVSD, so much cleaner for it to continue to live there

8sunyuan: requires a `onlyDelegationManager` check if its stored in the AVSD, seems kind of gross to me vs just doing the external call to the DM to read it when doing allocations

gpsanant: let's debate!

gpsanant (commented):
review wtf

8sunyuan (approved):
approving and moving comments to other PR

PR #697, opened 2024-08-20
feat: pepe deploy script and config cleanup

wadealexc:
* removes a lot of old configs and scripts
* cleans up mainnet config and deploy folder


wadealexc:
New mainnet config folder looks like this: 
image

wadealexc:
Rather than creating new config/address files each deploy/upgrade, we can update the existing files.

I also added a `lastUpdatedAt` field to the top of both `mainnet-addresses` and `mainnet-config` that gets printed out when you run a script that uses them. I'm setting this value to `v0.4.2-mainnet-pepe` -- the idea is to make it obvious whether config has been updated recently or you're using something outdated.

(This all needs to be documented in a `scripts/README` type file, but I'm leaving that for a followup PR.)

script/configs/mainnet/mainnet-addresses.config.json
```diff
@@ -0,0 +1,80 @@
+{
+    "lastUpdated": "v0.4.2-mainnet-pepe",
+    "addresses": {
+        "avsDirectory": "0x135dda560e946695d6f155dacafc6f1f25c1f5af",
+        "avsDirectoryImplementation": "0xdabdb3cd346b7d5f5779b0b614ede1cc9dcba5b7",
+        "beaconOracle": "0x343907185b71aDF0eBa9567538314396aa985442",
```
8sunyuan, line 6: noting beaconOracle and DWR addresses not needed after this upgrade

8sunyuan (approved):
LGTM

PR #679, opened 2024-08-12
feat: slashing release

8sunyuan:
Updated 1/03/2025
### AVSDirectory
- Currently on mainnet, all operator<>AVS registration takes place on the AVSDirectory contract which AVS contracts call into along with an operator signature to confirm operator registration in the EigenLayer protocol. This registration/deregistration flow will change in the Slashing release with the introduction of operatorSets. More on this below...
- The AVSDirectory contract will be unchanged and allow for backwards compatibility. It will be deprecated (all data indexing ignored) at some future date when AVSs have migrated their systems over to the newly defined contract, the AllocationManager.

### New Core contract: AllocationManager! (basically the Slasher contract)

- Operator<>AVS registration/deregistration should be migrated here for existing AVSs if they wish to utilize Slashing functionality within EigenLayer. In addition to a different address entrypoint(previously AVSDirectory), the flow has actually changed to an operator registering themselves with the AllocationManager, and the AllocationManager will then call the corresponding callback setup by the AVS themselves. This is contrary to what exists today on the AVSDirectory which expects to be called by an AVS with an operator signature providing "registration consent".
- AVSs now have the ability to define subsets of operators called operatorSets. This allows for more granularity around defining slashing and rewards conditions for their different operators and their tasks. An operator can be exclusively in one operatorSet or many all at once. It is up to the AVS to determine their own registration/deregistration conditions and the Strategies that are in each of their operatorSets.
- The AllocationManager is the Slasher contract, pending decision on rename or just changing back to Slasher. 
- This implements "unique security", stake is not reused across AVSs/operatorSets and operators allocate slashable proportions to each operatorSet. Imagine this as a pie chart where each slice gets given to a different operatorSet.
- `modifyAllocations` can be called by an operator to configure for a given operatorSet, the slashable proportions for each Strategies in the operatorSet. Ex. If an AVS operatorSet has stETH and rETH Strategies in their operatorSet, as an operator I can allocate 50% of my delegated stETH shares and 25% of my delegated rETH shares to be slashable by this operatorSet. An allocation can be thought of as the slashable proportion defined over the tuple (operator, operatorSet, Strategy).
[...]

PR #714, opened 2024-08-29
Feat: opSet -> operators enumerable Set

ypatil12:
An added benefit is getting rid of the `operatorSetMemberCount` mapping


src/contracts/core/AVSDirectory.sol
```diff
@@ -431,39 +433,86 @@ contract AVSDirectory is
...
         }
     }
+    
+    /**
+     * @notice Returns an array of operators registered to the operatorSet. 
+     * @param operatorSet The operatorSet to query.
+     * @param start The starting index of the array to query.
+     * @param length The amount of items of the array to return.
+     */
+    function getOperatorsInOperatorSet(
```
gpsanant, line 480: great this is exactly what i need!!

src/contracts/core/AVSDirectory.sol
```diff
@@ -417,9 +419,9 @@ contract AVSDirectory is
                 "AVSDirectory._deregisterOperatorFromOperatorSet: operator not registered for operator set"
             );
 
-            --operatorSetMemberCount[avs][operatorSetIds[i]];
-
-            _operatorSetsMemberOf[operator].remove(_encodeOperatorSet(operatorSet));
+            bytes32 encodedOperatorSet = _encodeOperatorSet(operatorSet);
+            _operatorSetsMemberOf[operator].remove(encodedOperatorSet);
```
gpsanant, line 423: hm. instead of the is member check you could acually just

```solidity
require(_operatorSetsMemberOf[operator].remove(encodedOperatorSet))
```

i think

gpsanant:
Can you change base to #679

src/contracts/core/AVSDirectory.sol
ypatil12: Yup, good suggestion. Updated reg and dereg methods

PR #716, opened 2024-08-29
chore: some changes for making deployments

shrimalmadhur:
- Adding deployment scripts
- Adding `getSlashableMagnitudes` method to get slashable magnitude for given operator and set of strategies 
- Update all scripts to take avsDirectory as input

### DIFF
--- a/src/contracts/core/AVSDirectory.sol
+++ b/src/contracts/core/AVSDirectory.sol
@@ -914,6 +914,29 @@ contract AVSDirectory is
     /**
      * @param operator the operator to get the slashable magnitude for
      * @param strategies the strategies to get the slashable magnitude for
+     * 
+     * @return operatorSets the operator sets the operator is a member of and the current slashable magnitudes for each strategy
+     */
+    function getCurrentSlashableMagnitudes(
+        address operator,
+        IStrategy[] calldata strategies,
+    ) external view returns (OperatorSet[] memory, uint64[][] memory) {
+        OperatorSet[] memory operatorSets = getOperatorSetsOfOperator(operator, 0, _operatorSetsMemberOf[operator].length());
+        uint64[][] memory slashableMagnitudes = new uint64[][](strategies.length);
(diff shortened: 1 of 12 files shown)


src/contracts/core/AVSDirectory.sol
```diff
@@ -911,6 +911,31 @@ contract AVSDirectory is
...
+        IStrategy[] calldata strategies,
+        uint32 timestamp
+    ) external view returns (OperatorSet[] memory, uint64[][] memory) {
+        OperatorSet[] memory operatorSets = getOperatorSetsOfOperator(operator, 0, _operatorSetsMemberOf[operator].length());
+        uint64[][] memory slashableMagnitudes = new uint64[][](strategies.length);
+        for (uint256 i = 0; i < strategies.length; ++i) {
+            slashableMagnitudes[i] = new uint64[](operatorSets.length);
+            for (uint256 j = 0; j < operatorSets.length; ++j) {
+                slashableMagnitudes[i][j] = uint64(
+                    _magnitudeUpdate[operator][strategies[i]][_encodeOperatorSet(operatorSets[j])].upperLookupLinear(timestamp)
```
8sunyuan, line 932: Does this make sense to use `upperLookupLinear` for any given arbitrary timestamp? @gpsanant what was the reasoning for this over `upperLookupRecent` for this function?

shrimalmadhur: for my use case, I will always want most recent. so I will more or less always pass time.now

8sunyuan: Don't want anyone using this view function to run out of gas while iterating backwards because the timestamp provided is really historic. The `_magnitudeUpdate` checkpoints array is unbounded in size

shrimalmadhur: ok that makes sense. added a new method

8sunyuan (approved):

PR #717, opened 2024-08-30
bEIGEN/EIGEN changes

MinisculeTarantula:
feat: add burn function and ~~one-time minter transfer~~  editable`isMinter` mapping to bEIGEN
new unit tests also included

EIGEN token also updated so that:

- `wrap` and `unwrap` now mint and burn EIGEN tokens rather than transferring them (like most standard "wrapped" tokens)
- `EIGEN.totalSupply()` is coded to return `bEIGEN.totalSupply()`

### DIFF
--- a/src/contracts/token/Eigen.sol
+++ b/src/contracts/token/Eigen.sol
@@ -123,14 +123,14 @@ contract Eigen is OwnableUpgradeable, ERC20VotesUpgradeable {
      */
     function wrap(uint256 amount) external {
         require(bEIGEN.transferFrom(msg.sender, address(this), amount), "Eigen.wrap: bEIGEN transfer failed");
-        _transfer(address(this), msg.sender, amount);
+        _mint(msg.sender, amount);
     }
 
     /**
      * @notice This function allows Eigen holders to unwrap their tokens into bEIGEN
      */
     function unwrap(uint256 amount) external {
-        _transfer(msg.sender, address(this), amount);
+        _burn(msg.sender, amount);
         require(bEIGEN.transfer(msg.sender, amount), "Eigen.unwrap: bEIGEN transfer failed");
     }
 
@@ -160,6 +160,15 @@ contract Eigen is OwnableUpgradeable, ERC20VotesUpgradeable {
             );
         }
         super._beforeTokenTransfer(from, to, amount);
+    }
+
+    /**
+     * @notice Overridden to return the total bEIGEN supply instead.
+     * @dev The issued supply of EIGEN should match the bEIGEN balance of this contract,
+     * less any bEIGEN tokens that were sent directly to the contract (rather than being wrapped)
+     */
+    function totalSupply() public view override returns (uint256) {
+        return bEIGEN.totalSupply();
     }
 
     /**
--- a/src/contracts/token/Eigen.sol
+++ b/src/contracts/token/Eigen.sol
@@ -155,7 +155,7 @@ contract Eigen is OwnableUpgradeable, ERC20VotesUpgradeable {
         if (block.timestamp <= transferRestrictionsDisabledAfter) {
             // if both from and to are not whitelisted
             require(
-                from == address(0) || from == address(this) || to == address(this) || allowedFrom[from] || allowedTo[to],
+                from == address(0) || to == address(0) || allowedFrom[from] || allowedTo[to],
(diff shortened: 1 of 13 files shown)


src/contracts/token/Eigen.sol
```diff
@@ -123,17 +123,22 @@ contract Eigen is OwnableUpgradeable, ERC20VotesUpgradeable {
...
      */
     function unwrap(uint256 amount) external {
-        _transfer(msg.sender, address(this), amount);
+        _burn(msg.sender, amount);
         require(bEIGEN.transfer(msg.sender, amount), "Eigen.unwrap: bEIGEN transfer failed");
     }
 
+    // @notice Burns EIGEN tokens held by the EIGEN token address itself
+    function burnExtraTokens() external {
+        _burn(address(this), balanceOf(address(this)));
```
gpsanant, line 139: Should this instead be
```solidity
uint256 unbackedTokens =  totalSupply() - bEIGEN.balanceOf(address(this));
require(unbackedTokens != 0, "Eigen.burnExtraTokens: No unbacked tokens to burn");
_burn(address(this), unbackedTokens);
```
essentially burning all tokens that are not backed which should only work one time for the tokens living at the Eigen token address.

My main goal with this would be so that we don't fuck up invariants when someone stupidly sends Eigen tokens directly to the Eigen token contract.

MinisculeTarantula: discussed offline; we decided to remove this function and go another route for fixing total supply, see https://github.com/Layr-Labs/eigenlayer-contracts/pull/717/commits/07c1fa6c7f9c4a631d9fa4f89a80f0526e0838c9

src/contracts/token/Eigen.sol
```diff
@@ -134,11 +134,6 @@ contract Eigen is OwnableUpgradeable, ERC20VotesUpgradeable {
         require(bEIGEN.transfer(msg.sender, amount), "Eigen.unwrap: bEIGEN transfer failed");
     }
 
-    // @notice Burns EIGEN tokens held by the EIGEN token address itself
-    function burnExtraTokens() external {
```
gpsanant: pls update pr comment

src/contracts/token/Eigen.sol
```diff
@@ -134,11 +134,6 @@ contract Eigen is OwnableUpgradeable, ERC20VotesUpgradeable {
         require(bEIGEN.transfer(msg.sender, amount), "Eigen.unwrap: bEIGEN transfer failed");
```
gpsanant, line 134: can we delete the mint function

MinisculeTarantula: I have been aiming for minimal code changes. I think this might complicate existing tests as well :/

src/contracts/token/Eigen.sol
MinisculeTarantula: done, ty

gpsanant (approved):

src/contracts/token/Eigen.sol
```diff
@@ -123,14 +123,14 @@ contract Eigen is OwnableUpgradeable, ERC20VotesUpgradeable {
...
-        _transfer(address(this), msg.sender, amount);
+        _mint(msg.sender, amount);
     }
 
     /**
      * @notice This function allows Eigen holders to unwrap their tokens into bEIGEN
      */
     function unwrap(uint256 amount) external {
-        _transfer(msg.sender, address(this), amount);
+        _burn(msg.sender, amount);
```
wadealexc, line 133: I think the `_transfer` version of these methods caused us to add `from == this || to == this` to `_beforeTokenTransfer`

Now that we're getting rid of it in favor of `_mint` and `_burn`, that method needs to account for `_beforeTokenTransfer` being called with `from == address(0)` for mint, and `to == address(0)` for burn.

I think only the former of those two is supported - does this mean I won't be able to `unwrap`?

MinisculeTarantula: you are absolutely correct. fixed in https://github.com/Layr-Labs/eigenlayer-contracts/pull/717/commits/c53aa74b7b2860d30dbf1a9f3f86eca22aa305a2 good catch and thank you for this!

src/contracts/token/Eigen.sol
MinisculeTarantula: resolved this offline, keeping code changes minimal here

PR #720, opened 2024-08-30
Feat: Allow operators to set allocation delay multiple times

ypatil12:
- Operators can only set a new delay if a subsequent allocation would be AFTER the latest allocation timestamp

TBD:
- Enforcing a global delay for this allocation being active


src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -85,6 +85,13 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
     /// @notice Mapping: operator => strategy => operatorSet (encoded) => list of queuedDeallocation indices
     mapping(address => mapping(IStrategy => mapping(bytes32 => uint256[]))) internal _queuedDeallocationIndices;
 
+    /// @notice Mapping: operator => allocation delay (in seconds) for the operator.
+    /// This determines how long it takes for allocations to take effect in the future.
+    mapping(address => uint32) internal _operatorAllocationDelay;
```
gpsanant, line 90: just `allocationDelay` is cleaner imo

src/contracts/core/AVSDirectory.sol
```diff
@@ -261,6 +264,27 @@ contract AVSDirectory is
...
+
+    /**
+     * @notice Called by operators to set their allocation delay. 
+     * @param delay the allocation delay in seconds
+     * @dev msg.sender is assumed to be the operator
+     * @dev Fails if setting a delay would result in the ability to set an
+     *      an allocation that is active BEFORE the latest set allocation. 
+     */
+    function setAllocationDelay(uint32 delay) external {
+        require(
```
gpsanant, line 276: we need to require that the delay is nonzero otherwise it will be set to 21 days

src/contracts/core/DelegationManager.sol
```diff
@@ -115,30 +114,18 @@ contract DelegationManager is
      */
     function registerAsOperator(
         OperatorDetails calldata registeringOperatorDetails,
-        uint32 allocationDelay,
```
gpsanant, line 118: we should still call into the AVSD imo

src/contracts/core/AVSDirectory.sol
```diff
@@ -261,6 +264,27 @@ contract AVSDirectory is
...
+            delegation.isOperator(msg.sender),
+            "avsDirectory.initializeAllocationDelay: operator not registered to EigenLayer yet"
+        );
+        require(
```
gpsanant, line 280: first time set is not delayed?

src/contracts/core/AVSDirectory.sol
```diff
@@ -261,6 +264,27 @@ contract AVSDirectory is
...
+            delay + block.timestamp >= latestAllocationEffectTimestamp[msg.sender],
```
gpsanant, line 281: what does this condition get us?

8sunyuan: We want to make sure if you decrease your allocation delay, you can't insert a pending update into the checkpoints array with the timestamps not being ordered. Although this is not feasible anyways with 1 max pending allocation?

src/contracts/core/AVSDirectory.sol
```diff
@@ -264,25 +271,38 @@ contract AVSDirectory is
...
-        _operatorAllocationDelay[msg.sender] = delay;
+
+        bool pendingInEffect = delayInfo.pendingDelay != 0 && block.timestamp >= delayInfo.pendingDelayEffectTimestamp;
+
+        if (pendingInEffect) {
+            delayInfo.delay = delayInfo.pendingDelay;
+        }
+
+        delayInfo.pendingDelay = delay;
+        delayInfo.pendingDelayEffectTimestamp = uint32(block.timestamp + ALLOCATION_DELAY_CONFIGURATION_DELAY);
```
gpsanant, line 303: `ALLOCATION_DELAY_CONFIGURATION_DELAY` should be `delayInfo.delay`

gpsanant: or rather `operatorAllocationDelay`

src/contracts/core/AVSDirectory.sol
8sunyuan: Nvm: snapshot library has a ordering check in its `_insert` function for decreasing keys

src/contracts/core/AVSDirectory.sol
```diff
@@ -261,6 +271,40 @@ contract AVSDirectory is
...
+    ) external {
+        require(
+            delegation.isOperator(msg.sender),
+            "avsDirectory.initializeAllocationDelay: operator not registered to EigenLayer yet"
+        );
+
+        AllocationDelayInfo storage delayInfo = _allocationDelayInfo[msg.sender];
+
+        require(
+            delay + block.timestamp >= delayInfo.latestAllocationEffectTimestamp,
```
8sunyuan, line 292: We don't need to check for `latestAllocationEffectTimestamp` and can remove it from the struct entirely

src/contracts/core/AVSDirectory.sol
```diff
@@ -283,13 +327,9 @@ contract AVSDirectory is
...
         );
-        IDelegationManager.AllocationDelayDetails memory details = delegation.operatorAllocationDelay(operator);
-        require(
-            details.isSet,
-            "AVSDirectory.modifyAllocations: operator must initialize allocation delay before modifying allocations"
-        );
+
         // effect timestamp for allocations to take effect. This is configurable by operators
-        uint32 effectTimestamp = uint32(block.timestamp) + details.allocationDelay;
+        uint32 effectTimestamp = uint32(block.timestamp) + operatorAllocationDelay(operator);
```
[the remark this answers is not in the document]
8sunyuan: Agreed, allocation delay would be set as 0 if never set. Should probably have a bool in the `AllocationDelayInfo` struct

src/contracts/core/AVSDirectory.sol
8sunyuan: what if its the first time being set? Is it just the default deallocation delay?

src/contracts/core/DelegationManager.sol
8sunyuan: not a fan but will delegate thoughts to @shrimalmadhur who will probably say yes

shrimalmadhur: I would like for it to be a single call when someone becomes operator - they can set it then only.
what's the reason for not having it here vs calling avs directory separately?

src/contracts/core/AVSDirectory.sol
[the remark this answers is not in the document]
8sunyuan: Or just check the timestamp is not 0

gpsanant:
already on #679

PR #721, opened 2024-08-30
chore: deploy and upgrade longtail on preprod

wadealexc:


script/utils/ExistingDeploymentParser.sol
```diff
@@ -63,6 +63,7 @@ contract ExistingDeploymentParser is Script, Test {
     StrategyFactory public strategyFactory;
     StrategyFactory public strategyFactoryImplementation;
     UpgradeableBeacon public strategyBeacon;
+    StrategyBase public strategyFactoryBeaconImplementation;
```
gpsanant, line 66: why include here if we're not parsing

gpsanant (approved):

PR #727, opened 2024-09-04
feat: custom errors

0xClandestine:


foundry.toml
```diff
@@ -16,7 +16,7 @@ optimizer_runs = 200
 # Whether or not to use the Yul intermediate representation compilation pipeline
 via_ir = false
 # Override the Solidity version (this overrides `auto_detect_solc`)
-solc_version = '0.8.12'
+solc_version = '0.8.27'
```
8sunyuan, line 19: We probably need to update the caret version in all the individual files too since custom errors in requires will not compile with a version less than 0.8.27

8sunyuan: Oh this PR is stale. https://github.com/Layr-Labs/eigenlayer-contracts/pull/734 Is up-to-date

PR #730, opened 2024-09-09
feat: split allocation logic to `AllocationManager`

0xClandestine:


src/contracts/core/AllocationManagerStorage.sol
```diff
@@ -0,0 +1,70 @@
...
+        delegation = _delegation;
+        avsDirectory = _avsDirectory;
+    }
+
+    /**
+     * @dev This empty reserved space is put in place to allow future versions to add new
+     * variables without shifting down storage in the inheritance chain.
+     * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+     */
+    uint256[35] private __gap;
```
8sunyuan, line 66: I think this should be 50-7=43 gap here.
Including `_DOMAIN_SEPARATOR` we are using 7 storage slots

src/contracts/core/AVSDirectoryStorage.sol
8sunyuan, line 1: needs to add back the gap here, these storage slots were never added to the AVSD on mainnet

8sunyuan (approved):
LGTM

PR #734, opened 2024-09-09
feat:  allocation view methods

shrimalmadhur:
- Add method for getting pending dealloc
- add method for getting pending alloc
- Fix deployment scripts


src/contracts/core/AllocationManager.sol
```diff
@@ -645,6 +645,71 @@ contract AllocationManager is
...
+                if (key > block.timestamp) {
+                    pendingMagnitude[i] = uint64(value);
+                    timestamps[i] = key;
+                } else {
+                    pendingMagnitude[i] = 0;
+                    timestamps[i] = 0;
+                }
+            }
+        }
+        return (pendingMagnitude, timestamps);
```
8sunyuan, line 676: how come you convert the timestamps from uint32 -> uint64?

src/contracts/core/AllocationManager.sol
```diff
@@ -645,6 +645,71 @@ contract AllocationManager is
...
+        uint64[] memory pendingMagnitude = new uint64[](operatorSets.length);
+        uint64[] memory timestamps = new uint64[](operatorSets.length);
+        for (uint256 i = 0; i < operatorSets.length; ++i) {
+            (bool exists, uint32 key, uint224 value) =
+                _magnitudeUpdate[operator][strategy][_encodeOperatorSet(operatorSets[i])].latestSnapshot();
+            if (exists) {
+                if (key > block.timestamp) {
+                    pendingMagnitude[i] = uint64(value);
+                    timestamps[i] = key;
+                } else {
```
8sunyuan, line 670: Would it be more useful to resize this array if you hit this else block? i.e no allocaiton for this opSet.
We can rewrite it with some assembly to resize the array.

src/contracts/core/AllocationManager.sol
```diff
@@ -645,6 +645,71 @@ contract AllocationManager is
...
+
+    /**
+     * @notice Returns the pending deallocations of an operator for a given strategy and operatorSets
+     * @param operator the operator to get the pending deallocations for
+     * @param strategy the strategy to get the pending deallocations for
+     * @param operatorSets the operatorSets to get the pending deallocations for
+     * @return pendingMagnitudeDiff the pending difference in deallocations for each operatorSet
+     * @return timestamps the timestamps for each pending deallocation
+     */
+    function getPendingDeallocations(
```
8sunyuan, line 687: Same comments for above `getPendingAllocations` apply here too

src/contracts/core/AllocationManager.sol
```diff
@@ -645,6 +645,71 @@ contract AllocationManager is
         return totalMagnitude;
     }
 
+    /**
+     * @notice Returns the pending allocations of an operator for a given strategy and operatorSets
```
8sunyuan, line 649: Might be worth noting that the assumption is that there is only MAX 1 pending allocation and deallocation. Thats why you only need to read `latestSnapshot` or use the last index in `_queuedDeallocationIndices`

script/deploy/devnet/Deploy_From_Scratch.s.sol
8sunyuan, line 1: deployment stuff being actively worked on by others, so this will likely change some time later. LGTM for now

src/contracts/core/AllocationManager.sol
shrimalmadhur: if I resize then how would the caller know what operatorSet it is for. basically keeping it as same size as operatorSet will help the caller figure out what opSet it is for. Unless we also return modified opSet array?

src/contracts/core/AllocationManager.sol
shrimalmadhur: ah yes. I will add a comment. good call

src/contracts/core/AllocationManager.sol
shrimalmadhur: oh I actually did not realize that - good catch. `forge b` did not detect it :O

8sunyuan: It's because it casted the uint and didn't perceive that as an error

shrimalmadhur: ahh auto casting. will change it to uint32 - it was not my intention. good catch

shrimalmadhur:
Not needed anymore

PR #735, opened 2024-09-09
chore: force the scratch buld on every push

jbrower95:
- Deprecate M1 & M2 build script. Each branch should keep an up-to-date version of this script.
- A github action now builds Deploy_From_Scratch on every push.
- Renamed from "deploy scratch" to "deploy local" everywhere

Follow up: I'll check that all contracts in the repo are deployed, and fail if there are any that aren't included in the scratch build (to avoid losing sync with the repo).


.github/workflows/deploy-local.yml
```diff
@@ -0,0 +1,51 @@
...
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Install lcov
+      run: |
+        sudo apt-get install lcov
+      id: lcov
+    - name: Install Foundry
+      uses: foundry-rs/foundry-toolchain@v1
+      with:
+        version: nightly
```
samlaf, line 33: do we want to pin a version?

samlaf (commented):
LGTM. Curious why the rename from devnet -> local.
Feel like devnet was more explicit, but perhaps you prefer devnet to potentially mean something distributed?

ypatil12 (commented):
Realized that `Deploy_From_Scratch` doesn't deploy the `StrategyFactory`. Would be good to add to this PR

Edit: Seems like this will be done in a follow up PR? Please confirm and I'll approve :)

.github/workflows/deploy-local.yml
jbrower95: pinned 1.2.0

jbrower95:
we'll update the script in a follow up PR (where I add logic to check that all contracts are deployed). Don't want multiple concerns in the same PR

.github/workflows/deploy-local.yml
jbrower95: ahh, this doesn't work

only nightly is supported

per their docs;
`Version to install, e.g. nightly or 1.0.0. Note: Foundry only has nightly builds for the time being.`

All other versions 404

samlaf: Hmm iirc they keep monthly versions around. But latest is fine actually it’s actually good to make sure our stuff is deployable with latest foundry

samlaf (approved):

PR #738, opened 2024-09-10
feat: legacy withdrawals and cleanup for DelegationManager

8sunyuan:
**Description**:
1. Includes support for legacy M2 withdrawals
The withdrawal struct has converted the uint32 `startBlock` field to `startTimestamp`. We simply check that this value is < `LEGACY_WITHDRAWALS_TIMESTAMP` (Jan 01 2025 17:00:00 GMT+0000) and process as an old legacy withdrawal if so. We just need to ensure this timestamped value is < whatever the current timestamp is when the upgrade takes place. It would take 600+ years for the current blocknumber on mainnet to reach this threshold. We should be good

3. Removing thirdPartyTransfersForbidden and deprecating the storage

**DM codesize after cleanup and optimizer-runs = 5**
image

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -771,7 +771,7 @@ contract DelegationManager is
     /**
      * @notice Removes `scaledShares` in `strategies` from `staker` who is currently delegated to `operator` and queues a withdrawal to the `withdrawer`.
      * @dev If the `operator` is indeed an operator, then the operator's delegated shares in the `strategies` are also decreased appropriately.
-     * @dev If `withdrawer` is not the same address as `staker`, then thirdPartyTransfersForbidden[strategy] must be set to false in the StrategyManager.
+     * @dev If `withdrawer` is not the same address as `staker`
      */
     function _removeSharesAndQueueWithdrawal(
         address staker,
@@ -783,6 +783,7 @@ contract DelegationManager is
     ) internal returns (bytes32) {
         require(staker != address(0), InputAddressZero());
         require(strategies.length != 0, InputArrayLengthZero());
+        require(staker == withdrawer, WithdrawerNotStaker());
 
         uint256[] memory scaledStakerShares = new uint256[](strategies.length);
 
@@ -835,10 +836,6 @@ contract DelegationManager is
                  */
                 eigenPodManager.removeShares(staker, sharesToDecrement);
             } else {
-                require(
-                    staker == withdrawer || !strategyManager.thirdPartyTransfersForbidden(strategies[i]),
-                    WithdrawerNotStaker()
-                );
                 // this call will revert if `scaledShares[i]` exceeds the Staker's current shares in `strategies[i]`
                 strategyManager.removeShares(staker, strategies[i], sharesToDecrement);
             }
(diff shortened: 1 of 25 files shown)


src/contracts/core/DelegationManager.sol
```diff
@@ -38,14 +38,14 @@ contract DelegationManager is
...
 
-    // The max configurable withdrawal delay per strategy. Set to 30 days in seconds
-    uint256 public constant MAX_WITHDRAWAL_DELAY = 30 days;
+    /// @dev Wed Jan 01 2025 17:00:00 GMT+0000, timestamp used to check whether a pending withdrawal
+    /// should be processed as legacy M2 or with slashing considered.
+    uint32 public constant LEGACY_WITHDRAWALS_TIMESTAMP = 1735750800;
 
-    // the number of 12-second blocks in 30 days (60 * 60 * 24 * 30 / 12 = 216,000)
-    uint256 public constant MAX_WITHDRAWAL_DELAY_BLOCKS = MAX_WITHDRAWAL_DELAY / 12;
+    uint32 public constant WITHDRAWAL_DELAY = SlashingConstants.DEALLOCATION_DELAY;
```
gpsanant, line 48: This should not be in a library, this is problematic for different envs. it should be an immutable

src/contracts/core/DelegationManager.sol
```diff
@@ -624,25 +573,36 @@ contract DelegationManager is
         require(tokens.length == withdrawal.strategies.length, InputArrayLengthMismatch());
 
         bytes32 withdrawalRoot = calculateWithdrawalRoot(withdrawal);
+        bool isLegacyWithdrawal = false;
 
         require(pendingWithdrawals[withdrawalRoot], ActionNotQueued());
-        require(withdrawal.startTimestamp + minWithdrawalDelay <= block.number, WithdrawalDelayNotElapsed());
+        require(withdrawal.startTimestamp + WITHDRAWAL_DELAY <= uint32(block.timestamp), WithdrawalDelayNotElapsed());
```
gpsanant, line 579: duplicated

src/contracts/core/DelegationManager.sol
```diff
@@ -875,10 +838,6 @@ contract DelegationManager is
                  */
                 eigenPodManager.removeShares(staker, sharesToDecrement);
             } else {
-                require(
-                    staker == withdrawer || !strategyManager.thirdPartyTransfersForbidden(strategies[i]),
```
gpsanant, line 879: we still need to make sure the staker is the withdrawer

8sunyuan: moved above, instead of checking per strategy

src/contracts/core/DelegationManager.sol
8sunyuan: Then I guess same applies for DEALLOCATION_DELAY in the AllocationManager

gpsanant (approved):

src/contracts/core/DelegationManager.sol
gpsanant: yeah. not sure the best solution is here.

I feel like it's immutables

PR #740, opened 2024-09-11
feat: allocation delay

0xClandestine:


src/contracts/core/AllocationManager.sol
```diff
@@ -403,6 +438,27 @@ contract AllocationManager is
...
     }
 
+    /**
+     * @notice Returns the allocation delay of an operator
+     * @param operator The operator to get the allocation delay for
+     * @dev Defaults to `DEFAULT_ALLOCATION_DELAY` if none is set
+     */
+    function allocationDelay(
+        address operator
+    ) public view returns (uint32 delay) {
```
8sunyuan, line 448: perhaps return a boolean as well indicating whether or not the operator has ever set an allocationDelay.
This could be checked by `pendingDelayEffectTimestamp != 0`
If it returns false, modifyAllocations should revert

src/contracts/core/AllocationManager.sol
```diff
@@ -60,6 +69,32 @@ contract AllocationManager is
         _transferOwnership(initialOwner);
     }
 
+    /**
+     * @notice Called by operators to set their allocation delay.
+     * @param delay the allocation delay in seconds
+     * @dev msg.sender is assumed to be the operator
+     * @dev Fails if setting a delay would result in the ability to set an
```
8sunyuan, line 76: remove this comment since we removed this check

src/contracts/core/AllocationManager.sol
```diff
@@ -403,6 +438,27 @@ contract AllocationManager is
...
+     */
+    function allocationDelay(
+        address operator
+    ) public view returns (uint32 delay) {
+        AllocationDelayInfo memory delayInfo = _allocationDelayInfo[operator];
+
+        bool pendingInEffect = delayInfo.pendingDelay != 0 && block.timestamp >= delayInfo.pendingDelayEffectTimestamp;
+
+        if (delayInfo.delay == 0 && !pendingInEffect) {
+            return DEFAULT_ALLOCATION_DELAY;
```
8sunyuan, line 454: Per comment above, lets get rid of `DEFAULT_ALLOCATION_DELAY` and this if block here. We also want to allow configurging 0 delay anyways without enforcing a default.

src/contracts/core/AllocationManager.sol
```diff
@@ -19,6 +19,12 @@ contract AllocationManager is
 {
     using Snapshots for Snapshots.History;
 
+    /// @dev Delay before deallocations are completable and can be added back into freeMagnitude
+    uint32 public constant DEALLOCATION_DELAY = 17.5 days;
```
8sunyuan, line 23: Remove and use the immutable DEALLOCATION_DELAY defined in AMStorage

src/contracts/core/AllocationManager.sol
```diff
@@ -333,6 +366,8 @@ contract AllocationManager is
                 info.freeMagnitude -= allocation.magnitudes[i] - uint64(currentMagnitude);
             }
         }
+
+        _allocationDelayInfo[operator].pendingDelayEffectTimestamp = allocationEffectTimestamp;
```
8sunyuan, line 370: rebase error? This doesn't make sense

src/contracts/core/DelegationManagerStorage.sol
8sunyuan, line 1: gap here needs to be incremented again

PR #741, opened 2024-09-11
Stake Root Simplifications [WIP]

shotaronowhere:
This PR represents WIP simplification and extension of the stake root contracts. It is not in a finalized state as items such as test coverage, safe type casting, natspec comments, incentive compatibility for removals, modifications needed to charging, and agreement of the simplification approach are needed.

Simplifications

- Removing checkpoint history data structures. Some of these data structures were used to track balance updates and introduced both logical and computational complexity as binary search was used in some cases. There's still a checkpoint history used for the total charge, but I think we can get rid of that with another approach (storage proof or synchronous charging).
- The async nature of charging is another point of complexity related to the blacklist window. We should discuss what the goal is here for blacklisting in general, but in the context of charging, it looks like an unneccesary complication that impacts the happy path due to an unhappy path concern. As far as I see, the original implementation delays payment of the proof submitter in case some point of the zk-proving stack is compromised, and you don't want to compensate someone for producing a proof which does not effectively compute the stake root. On the otherhand, proof charging is rate limited and should this event happen, the contract is upgradable anyways so we could effectively pause the contract. If we are designing a contract that will be immutable in the future, then that changes the analysis. I don't think the the complexity of async charging is not worth the rare edge case it protects against.
- Removal from the stake tree leaves strategies dirty

AVS' can withdraw the balances at any time and remove operator sets from the tree.


src/contracts/core/StakeRootCompendium.sol
```diff
@@ -26,170 +25,139 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
         StrategyAndMultiplier[] calldata strategiesAndMultipliers
     ) external {
-        require(
-            strategiesAndMultipliers.length > 0,
-            "StakeRootCompendium.setStrategiesAndMultipliers: no strategies and multipliers provided"
-        );
+        _addStrategiesAndMultipliers(IAVSDirectory.OperatorSet({avs: msg.sender, operatorSetId: operatorSetId}), strategiesAndMultipliers);
+    }
+
+    // natspec
```
gpsanant, line 39: delete

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -26,170 +25,139 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
+    /// @notice Initializes an operator set with a given set of strategies and multipliers
+    /// @dev This a convinience (gas saving) function for avs'
```
gpsanant, line 41: we use multiline comments stylistically
```
/**
 *
...
 */
 ```

File: src/contracts/core/StakeRootCompendium.sol [hunk 3]
```diff
@@ -26,170 +25,139 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
+    /// @param operatorSetId The ID of the operator set to initialize
+    /// @param strategiesAndMultipliers The strategies and multipliers to initialize the operator set with
+    function initOperatorSetInStakeTree(
```
gpsanant, line 44: hm.

Can we just make `addStrategiesAndMultipliers` payable and deposit if `msg.value > 0`?

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -26,170 +25,139 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
 
     /// @inheritdoc IStakeRootCompendium
     function depositForOperatorSet(IAVSDirectory.OperatorSet calldata operatorSet) external payable {
```
gpsanant, line 27: just `deposit` to be consistent with new internal funcs

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -485,9 +382,8 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
                 avsDirectory.getTotalAndAllocatedMagnitudes(operator, operatorSet, strategies);
 
             for (uint256 i = 0; i < strategies.length; i++) {
-                uint256 multipliedDelegatedShares = delegatedShares[i] * multipliers[i] / 1 ether;
-                delegatedStake += multipliedDelegatedShares * totalMagnitudes[i];
-                slashableStake += multipliedDelegatedShares * allocatedMagnitudes[i];
+                delegatedStake += delegatedShares[i] * totalMagnitudes[i] / 1 ether * multipliers[i];
```
gpsanant, line 385: i *think* the previous saves more gas in the zkVM

gpsanant: not sure though, we should benchmark

shotaronowhere: ahh okay this was inadvertent

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -234,226 +202,155 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
 
-        // keep track of the cumulative charge that would have been charged since the last change
-        // this is: 
-        // latestSnapshottedCumulativeCharge + proofs since latest snapshot * (charge / proof)
-        // latestSnapshottedCumulativeCharge + (latestCalculationTimestamp - latestSnapshotTimestamp) * (1 proof / proofInterval time) * (charge / proof)
-        // latestSnapshottedCumulativeCharge + (charge / proof) * (time since snapshot) * (1 proof / proofInterval time)
-        return latestSnapshottedCumulativeCharge + uint224(currentCharge * (latestCalculationTimestamp - latestSnapshotTimestamp) / proofInterval);
+        // update the index of the operator sets
+        operatorSetToOneBasedIndex[msg.sender][operatorSet.operatorSetId] = 0;
+        operatorSetToOneBasedIndex[msg.sender][substituteOperatorSet.operatorSetId] = operatorSetOneBasedIndex;
```
gpsanant, line 265: this should not be msg.sender, it should be the `operatorSet.avs` and `substituteOperatorSet.avs` respectively

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -234,226 +202,155 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
-    // updates total strategies and the total charge per proof whenever an operator set's number of strategies changes
-    function _updateTotals(uint256 numStrategiesBefore, uint256 numStrategiesAfter) internal {
-        totalStrategies = totalStrategies - numStrategiesBefore + numStrategiesAfter;
-        totalChargeSnapshot.push(
-            uint32(block.timestamp), 
-            uint224(totalStrategies * linearChargePerProof + constantChargePerProof)
-        );
+    // updates the deposit balance for the operator set and returns the penalty if the operator set has fallen below the minimum deposit balance
+    function _updateDepositInfo(IAVSDirectory.OperatorSet memory operatorSet) internal returns (uint256) {
+        uint256 depositBalance = getDepositBalance(operatorSet);
```
gpsanant, line 325: we should remove them from the operatorSet if they fall below the min balance here?

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -234,226 +202,155 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
-                operatorSet,
-                strategies
-            );
-        }
-
-        IAVSDirectory.OperatorSet memory substituteOperatorSet = operatorSets[operatorSets.length - 1];
-        uint224 operatorSetIndex = operatorSetToIndex[operatorSet.avs][operatorSet.operatorSetId].latest();
-        operatorSets[operatorSetIndex] = substituteOperatorSet;
-        operatorSets.pop();
+    function _updateTotalStrategies(uint256 numStrategiesBefore, IAVSDirectory.OperatorSet memory operatorSet) internal {
```
gpsanant, line 289: rename back to totalCharge because we need to call this when charges change

src/contracts/core/StakeRootCompendium.sol
shotaronowhere: yeah, definitely

src/contracts/core/StakeRootCompendium.sol
shotaronowhere: we can, i reserved that logic where it is explicitly relevant eg processing withdrawals. . .

let me ponder, can put it back in, just thinking if there was a good reason i took it out

src/contracts/core/StakeRootCompendium.sol
shotaronowhere: will do

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -234,226 +202,155 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
+        // note: operatorSetIndexPlusOne is 1 indexed
+        operatorSetToOneBasedIndex[operatorSet.avs][operatorSet.operatorSetId] = uint208(operatorSets.length);
     }
 
-    function _getPreviousCumulativeCharges(uint32 timestamp) internal view returns(uint224, uint224) {
-        return (
-            cumulativeContantChargeSnapshot.upperLookup(timestamp),
-            cumulativeLinearChargeSnapshot.upperLookup(timestamp)
+    function _removeOperatorSetFromStakeTree(IAVSDirectory.OperatorSet memory operatorSet) internal {
+        uint256 operatorSetOneBasedIndex = operatorSetToOneBasedIndex[operatorSet.avs][operatorSet.operatorSetId];
```
gpsanant, line 252: we need to update the totalCharge in this function in case this is a force removal that isn't a removal of strategies

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -26,170 +25,139 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
+            );
+        }
+
+        _updateTotalStrategies(numStrategiesBefore, operatorSet);
+    }
+
 
-        // remove the strategies and multipliers for the operator set
-        _removeStrategiesAndMultipliers(operatorSet, strategies);
+    function removeOperatorSetFromStakeTree(uint32 operatorSetId) external {
```
gpsanant, line 100: hm why do we have this?

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -26,170 +25,139 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
+        _removeOperatorSetFromStakeTree(IAVSDirectory.OperatorSet({avs: msg.sender, operatorSetId: operatorSetId}));
     }
 
     /// @inheritdoc IStakeRootCompendium
     function setOperatorSetExtraData(uint32 operatorSetId, bytes32 extraData) external {
-        (bool exists,, uint224 index) = operatorSetToIndex[msg.sender][operatorSetId].latestCheckpoint();
```
gpsanant, line 101: seems correct to revert here imo

src/contracts/core/StakeRootCompendium.sol
shotaronowhere: not necessary I guess, could remove it if it doesn't matter for AVS'

src/contracts/core/StakeRootCompendium.sol
shotaronowhere: yeah i'll put that back, it's a setter so used very rarely

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -26,170 +25,139 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
-        require(exists && index != REMOVED_INDEX, "StakeRootCompendium.setOperatorExtraData: operatorSet is not in stakeTree");
         operatorExtraDatas[msg.sender][operatorSetId][operator] = extraData;
     }
 
     /// POSTING ROOTS AND BLACKLISTING
 
     /// @inheritdoc IStakeRootCompendium
     function verifyStakeRoot(
-        uint32 calculationTimestamp,
+        uint64 calculationBlockNumber,
```
gpsanant, line 118: this should be a timestamp since everything is in timestamps

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -26,170 +25,139 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
         require(
-            !stakeRootSubmissions[submissionIndex].blacklisted,
-            "StakeRootCompendium.blacklistStakeRoot: stakeRoot already blacklisted"
+            calculationBlockNumber % proofInterval == 0,
+            "StakeRootCompendium._postStakeRoot: blockNumber must be a multiple of proofInterval"
         );
         require(
-            block.timestamp < stakeRootSubmissions[submissionIndex].blacklistableBefore,
-            "StakeRootCompendium.blacklistStakeRoot: stakeRoot cannot be blacklisted"
+            latestBlockNumberProven < calculationBlockNumber,
```
gpsanant, line 128: `calculationBlockNumber = latestBlockNumberProven + proofInterval`

And this can only be if this is not the first submission

src/contracts/core/StakeRootCompendium.sol
shotaronowhere: hmmmm this might be awkward since the zk proof is against a blockhash and blocks are sequential but calculationTimestamps can jump
BN: [ 0 ] [ 1 ] [ 2 ]
TS:   12    36   48

so there is a missing timestamp corresponding to timestamp 24. . . should a proof be skipped in this case?

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -234,226 +202,155 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
+        StakeRootSubmission memory submission = stakeRootSubmissions[calculationBlockNumber];
 
-        uint256 stakeRootSubmissionsLength = stakeRootSubmissions.length;
-        if (stakeRootSubmissionsLength != 0) {
-            require(
-                stakeRootSubmissions[stakeRootSubmissionsLength - 1].calculationTimestamp + proofInterval == calculationTimestamp,
-                "StakeRootCompendium._postStakeRoot: calculationTimestamp must be greater than the last posted calculationTimestamp"
-            );
+        if (submission.stakeRoot == bytes32(0)) {
+            submission.stakeRoot = stakeRoot;
```
gpsanant, line 339: if this is not a proof post, the submission must be proven. confirmation must happen after proofs

shotaronowhere: I was kind of overloading this functionality for the desired feature that if the provers are offline, the same data structures can be used for AVS' which may choose to use a simpler confirmation rule (trust eigen labs)

---

or what is desired as a fallback if zkprovers are all offline?

shotaronowhere: but yeah, certainly can adjust this so confirmerss strictly confirm

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -234,226 +202,155 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
+            uint224(totalStrategies * chargePerProofLinear + chargePerProofConstant)
+        );
+    }
 
-        // update the index of the operator sets
-        operatorSetToIndex[operatorSet.avs][operatorSet.operatorSetId].push(uint32(block.timestamp), REMOVED_INDEX);
-        operatorSetToIndex[operatorSet.avs][operatorSet.operatorSetId].push(uint32(block.timestamp), operatorSetIndex);
+    function _setChargePerProof(uint64 _chargePerProofLinear, uint64 _chargePerProofConstant) internal {
+        _updateCumulativeCharges();
+        chargePerProofLinear = _chargePerProofLinear;
```
gpsanant, line 299: we need to update the total Charge here

src/contracts/core/StakeRootCompendium.sol
shotaronowhere: what do you mean?

from the first require statement, we have `calculationBlockNumber % proofInterval == 0`

and from `latestBlockNumberProven < calculationBlockNumber`

we have basically that

latestBlockNumberProven = latestBlockNumberProven + n * proofInterval

---

do you mean proofs should be calculated strictly sequentially?

src/contracts/core/StakeRootCompendium.sol
shotaronowhere: yeah, i overlooked some total charging related logic, will put it in there

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -234,226 +202,155 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
-        totalStrategies = totalStrategies - numStrategiesBefore + numStrategiesAfter;
-        totalChargeSnapshot.push(
-            uint32(block.timestamp), 
-            uint224(totalStrategies * linearChargePerProof + constantChargePerProof)
-        );
+    // updates the deposit balance for the operator set and returns the penalty if the operator set has fallen below the minimum deposit balance
+    function _updateDepositInfo(IAVSDirectory.OperatorSet memory operatorSet) internal returns (uint256) {
+        uint256 depositBalance = getDepositBalance(operatorSet);
+        depositInfo[operatorSet.avs][operatorSet.operatorSetId].balance = uint96(depositBalance);
+        depositInfo[operatorSet.avs][operatorSet.operatorSetId].snapshotTimestamp = uint64(block.timestamp);
```
gpsanant, line 327: we need to update the cumulatives here

src/contracts/core/StakeRootCompendium.sol
```diff
@@ -26,170 +25,139 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
         require(
-            !stakeRootSubmissions[submissionIndex].forcePosted,
-            "StakeRootCompendium.blacklistStakeRoot: stakeRoot was force posted"
+            stakeRootSubmissions[calculationBlockNumber].proven == false, 
+            "StakeRootCompendium._postStakeRoot: calculationBlockNumber already proven"
         );
-        stakeRootSubmissions[submissionIndex].blacklisted = true;
+        // TODO: verify proof
+        _postStakeRoot(true, calculationBlockNumber, stakeRoot);
+        uint256 charge = totalChargeSnapshot.upperLookup(uint32(calculationBlockNumber));
```
gpsanant, line 137: should we just allow them to submit an index here to avoid lookup gas?

shotaronowhere: i need to put a bit more thought into the total charge, thats a component I overlooked a bit. . .

can't we avoid snapshots by having a view function for the totalCharge and doing a storage proof for that block? that's basically what the zk proof is doing anyways right? could add that to the proof? and to avoid zk tech stack failure enforce a MAX_TOTAL_CHARGE = 3 eth

alternatively to KISS we can do a little bit more accounting to have an O(1) lookup, I think there's a simple way to do it

src/contracts/core/StakeRootCompendiumStorage.sol
```diff
@@ -23,8 +23,6 @@ abstract contract StakeRootCompendiumStorage is IStakeRootCompendium, OwnableUpg
     uint32 public constant MAX_NUM_OPERATOR_SETS = 2048;
     /// @notice the maximum number of strategies that each operator set in the StakeTree can use to weight their operator stakes
     uint32 public constant MAX_NUM_STRATEGIES = 20;
-    /// @notice the placeholder index used for operator sets that are removed from the StakeTree
-    uint32 public constant REMOVED_INDEX = type(uint32).max;
 
     /// @notice the minimum balance that must be maintained for an operatorSet
     uint256 public constant MIN_DEPOSIT_BALANCE = 0.001 ether;
```
gpsanant, line 28: can we make this an immutable?

src/contracts/core/StakeRootCompendium.sol
gpsanant, line 137: set `latestBlockNumberProven`

gpsanant (approved):

PR #745, opened 2024-09-13
eigenpods slashing updates

gpsanant:
## Summary
This proposal
- this would get rid of negative shares (going forward)
- more intuitive slashing model between LSTs and EigenPods

by using a separate beaconChainScalingFactor that is decreased upon negative deltas

### WIP: EigenPod.sol

Since EigenPods are updated to record a _proportion delta_ in the EPM instead of the abosolute share delta upon deposits, an "edge" case exists where uncompleted checkpoints exist during the slashing upgrade and can no longer call back to the EPM.

I'm still looking into it but I believe our choices are the following:

1. Pause/Unpause
a. pause checkpoint creation
b. complete all existing checkpoints
c. execute the slashing upgrade
d. unpause checkpoint creation (we'd already have an unpause tx in the timelock controller)
2. Upgrade pods in UNIPRINT
a. upgrade EigenPod.sol in UNIPRINT to keep track of total beaconChain shares before and after checkpoints, but don't change any interactions (i.e. don't upgrade the EPM or DM)
b. make sure all existing checkpoints that were started before UNIPRINT are completed (via completion if necessary)
c. execute the slashing upgrade (this works because deltas are easily calculable from total beaconChain shares, which were kept track of in a)
3. Keep decreaseDelegatedShares around in the DM/EPM for the existing checkpoints at time of queueing
a. decreaseDelegatedShares would only be called for negative deltas that result from existing checkpoints from before the slashing release

(1) is less operations than (2), but more time pressure and downtime than (2).
(3) is less operations than (2), better uptime and time pressure than (1), but is the most code.

Still looking for way to calculate share deltas without changing what we're keeping track of but doesn't look promising

### Getting all shares positive

Next, we need to make sure all podOwnerShares are positive in order for checkpoints to successfully go through.

The only way that a pod's shares are negative is that the podOwner has queued a bunch of withdrawals and negative share deltas have occured afterwards. The pod's shares can always become nonnegative if all withdrawals are completed.  So in order to migrate to a nonnegative share system, we essentially break all shareDeltas from being propagated to the EPM until the podOwner completes enough withdrawals to make their shares nonnegative. 

### UX

This doesn't effect AVS UX because the podOwner's shares are not counted in the DM if they're already negative, so they don't care about further updates.

It effects podOwner UX, since they must complete their withdrawals before completing any more checkpoints. In addition, there is a small pause in checkpoint creation.

### DIFF
--- a/src/contracts/pods/EigenPod.sol
+++ b/src/contracts/pods/EigenPod.sol
@@ -61,19 +61,19 @@ contract EigenPod is Initializable, ReentrancyGuardUpgradeable, EigenPodPausingC
 
     /// @notice Callable only by the EigenPodManager
     modifier onlyEigenPodManager() {
-        require(msg.sender == address(eigenPodManager), UnauthorizedCaller());
+        require(msg.sender == address(eigenPodManager), OnlyEigenPodManager());
         _;
     }
 
     /// @notice Callable only by the pod's owner
     modifier onlyEigenPodOwner() {
-        require(msg.sender == podOwner, UnauthorizedCaller());
+        require(msg.sender == podOwner, OnlyEigenPodOwner());
         _;
     }
 
     /// @notice Callable only by the pod's owner or proof submitter
     modifier onlyOwnerOrProofSubmitter() {
-        require(msg.sender == podOwner || msg.sender == proofSubmitter, UnauthorizedCaller());
+        require(msg.sender == podOwner || msg.sender == proofSubmitter, OnlyEigenPodOwnerOrProofSubmitter());
         _;
     }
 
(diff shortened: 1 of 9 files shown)


src/contracts/pods/EigenPod.sol
```diff
@@ -635,8 +645,19 @@ contract EigenPod is Initializable, ReentrancyGuardUpgradeable, EigenPodPausingC
             delete currentCheckpointTimestamp;
             delete _currentCheckpoint;
 
+            // Calculate the slashing proportion
+            uint256 proportionPodBalanceDecrease = 0;
+            if (checkpoint.balanceDeltasGwei < 0) {
+                uint256 totalBefore = withdrawableRestakedExecutionLayerGwei + checkpoint.beaconChainBalanceBefore;
+                proportionPodBalanceDecrease = uint256(uint128(-checkpoint.balanceDeltasGwei)) * 1e18 / totalBefore;
+            }
```
wadealexc, line 653: This seems to only care about the beacon chain portion of the delta, which is incorrect.

For example, if I withdraw a validator, I'll end up with a negative delta here - but I think you're trying to capture actual asset decrease, not just beacon chain balance decrease.

You need to capture the entire share delta

gpsanant: you're right! patched

gpsanant:
merging temporarily rn

gpsanant:
will modify after further review

PR #746, opened 2024-09-13
refactor: receive as shares callback

8sunyuan:
Receive as shares is essentially duplicated logic of deposits. This includes refactoring the addShares function in SM to call increaseDelegatedShares. Changes to EPM are still pending


gpsanant (approved):

gpsanant:
pls update EPM block as well :pray:

PR #747, opened 2024-09-13
feat: bump OZ version (0.4.9)

0xClandestine:


.gitmodules
```diff
@@ -1,15 +1,9 @@
...
 [submodule "lib/ds-test"]
 	path = lib/ds-test
 	url = https://github.com/dapphub/ds-test
 [submodule "lib/forge-std"]
 	path = lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
-[submodule "lib/openzeppelin-contracts"]
-	path = lib/openzeppelin-contracts
-	url = https://github.com/OpenZeppelin/openzeppelin-contracts
 [submodule "lib/openzeppelin-contracts-v4.9.0"]
```
gpsanant, line 7: can we get rid of the `-v4.9.0` everywhere

gpsanant: oh i see you did this in the remapping, would be nice to change the folder name too

0xClandestine:
bindings likely need fixed

.gitmodules
0xClandestine: Wondering how to properly rename a submodule?

bin/storage-report.sh
```diff
@@ -0,0 +1,5 @@
+mkdir storage-report
+for file in $(find src -name '*.sol'); do
+    contract_name=$(basename "$file" .sol)
```
gpsanant, line 3: could you do it only for the `src/contracts/*.sol`?

PR #750, opened 2024-09-16
feat: ci storage reports

0xClandestine:
Adapted from @stevennevins's [implementation](https://github.com/stevennevins/ci-workflows/blob/main/.github/workflows/storage-checker.yml)


.github/workflows/storage-report.yml
```diff
@@ -0,0 +1,39 @@
+name: Check Storage Layout
+on:
+  workflow_dispatch:
+  pull_request:
```
gpsanant, line 4: can we do this on pull request into dev

.github/workflows/storage-report.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          git fetch origin $MAIN
+          git checkout $MAIN
+
+      - name: "Generate and prepare the contract artifacts for main branch"
+        run: |
+          bash bin/storage-report.sh
+
+      - name: Compare outputs
+        run: |
```
stevennevins, line 36: In my workflow I was outputting to two separate directories and the diffing the contents of the two directories

0xClandestine: Nice catch, will fix.

.github/workflows/storage-report.yml
0xClandestine: fixed

gpsanant (approved):
let's try it out

PR #754, opened 2024-09-17
fix: pending dealloc bug

shrimalmadhur:
- If there are no pending deallocations, then the length of queued dealloc is 0, so handling that case


src/contracts/core/AllocationManager.sol
```diff
@@ -750,16 +750,22 @@ contract AllocationManager is
             uint256[] memory indices =
```
gpsanant, line 750: uh can u add a todo here to not load the array into memory? not sure the exact behavior under the hood here actually

src/contracts/core/AllocationManager.sol
```diff
@@ -750,16 +750,22 @@ contract AllocationManager is
...
-                _pendingFreeMagnitude[operator][strategy][deallocationIndex];
-            if (latestPendingMagnitude.completableTimestamp > block.timestamp) {
-                pendingMagnitudeDiff[i] = latestPendingMagnitude.magnitudeDiff;
-                timestamps[i] = latestPendingMagnitude.completableTimestamp;
-            } else {
-                // There is no pending deallocation, so we set the pending magnitude and timestamp to 0
+            if (length == 0) {
+                // If there is no deallocations queued for this operator set
                 pendingMagnitudeDiff[i] = 0;
                 timestamps[i] = 0;
```
gpsanant, line 756: nit: prefer a `continue` and no indent

src/contracts/core/AllocationManager.sol
```diff
@@ -750,16 +750,22 @@ contract AllocationManager is
             uint256[] memory indices =
                 _queuedDeallocationIndices[operator][strategy][_encodeOperatorSet(operatorSets[i])];
             uint256 length = indices.length;
```
gpsanant, line 752: just use indices.length everywhere, it's clearer

src/contracts/core/AllocationManager.sol
```diff
@@ -750,16 +750,22 @@ contract AllocationManager is
...
-                // There is no pending deallocation, so we set the pending magnitude and timestamp to 0
+            if (length == 0) {
+                // If there is no deallocations queued for this operator set
                 pendingMagnitudeDiff[i] = 0;
                 timestamps[i] = 0;
+            } else {
+                uint256 deallocationIndex = indices[length - 1];
+                PendingFreeMagnitude memory latestPendingMagnitude =
+                    _pendingFreeMagnitude[operator][strategy][deallocationIndex];
+                if (latestPendingMagnitude.completableTimestamp > block.timestamp) {
```
gpsanant, line 761: condense this whole block into
```solidity
uint256[] storage deallocationIndices = _queuedDeallocationIndices[operator][strategy][_encodeOperatorSet(operatorSets[i])].length;
uint256 numDeallocations = deallocationIndices.length;
if (numDeallocations > 0 && deallocationIndices[numDeallocations - 1]) {
    PendingFreeMagnitude memory latestPendingMagnitude =
                    _pendingFreeMagnitude[operator][strategy][deallocationIndex];
    pendingMagnitudeDiff[i] = latestPendingMagnitude.magnitudeDiff;
    timestamps[i] = latestPendingMagnitude.completableTimestamp;
}
```
FYI, things default to 0, not nil so this should work.

Any reason you're not returning a list of `PendingFreeMagnitude`s?

shrimalmadhur: my noob solidity brain somehow assumed that I can't return array of objects lol.

PR #758, opened 2024-09-17
Stake root misc cleanup

gpsanant:
Addresses 
- maxTotalCharge storage var
- natspec comments
- race condition on updateBalance
- misc simplifications


src/contracts/core/StakeRootCompendium.sol
```diff
@@ -242,79 +243,35 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
 
     /// SET FUNCTIONS
 
-    function setChargePerProof(uint96 _chargePerStrategy, uint96 _chargePerOperatorSet) public onlyOwner  {
+    /// @inheritdoc IStakeRootCompendium
+    function setMaxTotalCharge(uint96 _maxTotalCharge) external onlyOwner {
```
shotaronowhere, line 247: I understand that in practice the maxTotalCharge in USD value should only go up and reflects offchain compute capacity, and due to eth price fluctuation would the charge decrease in size denominated in eth. Perhaps add a comment that if the maxTotalCharge decreases, one must also modify the chargePerProof (perStrategy and perOperatorSet) so that the totalCharge does not already exceed the new maxTotalCharge to avoid "brick"-ing operatorset removals / additions.

shotaronowhere (approved):
LGTM

Can add comments on edge cases when setting total charge and proof interval charges. I commented on the total charge in the review since it could brick functionality in an edge case, but when the proof interval decreases by more than a fraction, it's possible multiple calculation timestamps are eligible to be verified at the same time and offchain actors will need to decide what to do when that happens (especially with a round robin between multiple providers, wouldn't want a race condition) but this is a rare edge case. Worth a comment tho.

src/contracts/core/StakeRootCompendium.sol
gpsanant: ah. ill add something to not allow reducing below the current totalCharge

PR #760, opened 2024-09-17
feat: simplify rejoin

shotaronowhere:
Allows OperatorSets to rejoin without clearing previous strategiesAndMultipliers. This means OperatorSets must also be allowed to modify their strategiesAndMultipliers before rejoining.

I've also updated the comments about front running when claiming the penalty for kicking out operator sets

### DIFF
--- a/src/contracts/core/StakeRootCompendium.sol
+++ b/src/contracts/core/StakeRootCompendium.sol
@@ -73,12 +73,8 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
         StrategyAndMultiplier[] calldata strategiesAndMultipliers
     ) external {
         OperatorSet memory operatorSet = OperatorSet({avs: msg.sender, operatorSetId: operatorSetId});
-        bool isInStakeTree = _isInStakeTree(operatorSet);
-
-        if (isInStakeTree) {
-            // update the deposit balance for the operator set whenever number of strategies is changed
-            _updateDepositInfo(operatorSet);
-        }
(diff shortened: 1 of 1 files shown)


src/contracts/core/StakeRootCompendium.sol
```diff
@@ -95,24 +90,30 @@ contract StakeRootCompendium is StakeRootCompendiumStorage {
...
     /// @inheritdoc IStakeRootCompendium
     function removeStrategiesAndMultipliers(uint32 operatorSetId, IStrategy[] calldata strategies) external {
         // update the deposit balance for the operator set whenever number of strategies is changed
         OperatorSet memory operatorSet = OperatorSet({avs: msg.sender, operatorSetId: operatorSetId});
-        _updateDepositInfo(operatorSet);
+        bool isInStakeTree = _isInStakeTree(operatorSet);
 
+        if (isInStakeTree) {
+            // update the deposit balance for the operator set whenever number of strategies is changed
+            _updateDepositInfo(operatorSet);
```
gpsanant, line 115: in _updateBalanceInfo, we make sure they're in the stakeTree: https://github.com/Layr-Labs/eigenlayer-contracts/blob/0633e791cf846f363044d706139efb393b7a8c29/src/contracts/core/StakeRootCompendium.sol#L334

gpsanant: i prefer not having a pathway where they can edit when they're outside of the stakeTree, it complicates code paths

gpsanant (approved):

PR #763, opened 2024-09-18
fix: add storage report to gitignore

bowenli86:
currently building the contract will generate storage report, and people can accidentally check them in

this pr prevents people from checking in storage report into code


.gitignore
```diff
@@ -37,6 +37,7 @@ script/output/M1_deployment_data.json
 
 # autogenerated docs (you can generate these locally)
 /docs/docgen/
+/docs/storage-report/
```
0xClandestine, line 40: This was intentional so that we have a diff history for storage updates.

PR #766, opened 2024-09-18
feat: stakeroot cleanup

0xClandestine:

PR #767, opened 2024-09-18
feat: fixed accounting bug and refactoring

shotaronowhere:
There was an accounting bug when updating params this is fixed.

I'm comfortable with the refactor in #766 , so i'm building off of that and just keeping this branch separate until there's another review. If we toss out the  #766 refactor I can migrate the fix.

- simplified the naming of `totalCharge` to `chargePerProof`
- removed an "optimization" in #766 of copying storage to memory to avoid multiple sloads, but with some profiling the gas cost of memory copying doesn't make up for fewer reads from the same slot. In general KISS. Let's not over optimize here. I understand avoiding reading from new slots and storage packing, but reading the same slot is cheap. better not to sweat the small stuff
- removed the assembly and replaced with readable code


0xClandestine (approved):
lgtm

PR #775, opened 2024-09-19
feat: inheritdoc

0xClandestine:


ypatil12 (commented):
Can we validate the natspec in the interface matches what we remove? IIRC some of these do not align properly.

0xClandestine:
> Can we validate the natspec in the interface matches what we remove? IIRC some of these do not align properly.

Imo natspec needs a rework overall, lots of inconsistent format/grammar/punctuation.

ypatil12 (approved):

PR #780, opened 2024-09-24
POC: AVS Identifier

ypatil12:
POC to use identifiers instead of ServiceManager (now dispatcher) address to identify an AVS in AVSDirectory

What still needs to be done:

- [ ] Implement Dispatcher factory
- [ ] Figure out migration from old to new dispatcher
- [ ] Ensure that current ServiceManager addresses can migrate


src/contracts/core/AVSDirectory.sol
```diff
@@ -63,6 +63,22 @@ contract AVSDirectory is
      *                    EXTERNAL FUNCTIONS
      *
      */
+    
+    /// @inheritdoc IAVSDirectory
+    function claimIdentifier(bytes20 avs) external {
```
gpsanant, line 68: why not bytes32?

ypatil12: need to encode with id for enumerable set

PR #781, opened 2024-09-24
Remove Slither from CI

ypatil12:


anupsv:
removing as its currently not helpful and being more of a hindrance than productive.

anupsv (approved):

PR #784, opened 2024-09-26
Strict Custom Types

ypatil12:
Adds custom types to SM & DM, with a focus on abstracting conversions away to the SlashingLibrary. 

One item I'd like some comment on is how strict we want to be between conversions. Do we want to enforce conversion from A to B then B to C if we want to convert from A to C

We can currently do the following conversions:

1. DelegatedShares -> Shares
2. OwnedShares -> DelegatedShares
3. Shares -> DelegatedShares
4. DelegatedShares -> OwnedShares
5. Shares -> OwnedShares

Is there room for simplification here?


src/contracts/core/DelegationManager.sol
```diff
@@ -516,7 +516,7 @@ contract DelegationManager is
                 operator: operator, 
                 staker: staker, 
                 strategy: strategies[i],
-                existingShares: uint256(0).wrapShares(),
+                existingShares: Shares(0),
```
0xClandestine, line 519: Oh nice, didn't know this was valid syntax.

src/contracts/core/StrategyManager.sol
```diff
@@ -183,6 +183,7 @@ contract StrategyManager is
         IERC20 token,
         OwnedShares ownedShares
     ) external onlyDelegationManager {
+        // TODO: add Shares type to StrategyBase
```
0xClandestine, line 186: Can do this if needed.

src/contracts/libraries/SlashingLib.sol
```diff
@@ -90,15 +97,69 @@ library SlashingLib {
...
+        // Step 2: Compute currentShares + addedShares
+        uint256 ownedPlusAddedShares = currentOwnedShares.add(addedShares).unwrap();
+
+        // Step 3: Calculate newStakerScalingFactor
+        // Note: We divide by operatorMagnitude to preserve 
+
+        //TODO: figure out if we only need to do one divWad here
+        uint256 newStakerScalingFactor = 
+            ownedPlusAddedShares
+            .divWad(currentShares.unwrap() + addedOwnedShares.unwrap())
```
0xClandestine, line 159: `currentShares.add(addedOwnedShares)`

0xClandestine: Is `addedOwnedShares` defined?

0xClandestine: -> `addedShares` maybe?

0xClandestine (commented):
few comments 🤝

PR #794, opened 2024-10-03
feat: align environments

MinisculeTarantula:
feat: add checker, improve configs, cleanup

- add Current_Config_Check.s.sol that checks the active governance config

- fix configs by adding more addresses + renaming for clarity

- delete (apparently) unused files

## Context Added For Reviewers (Nov 1, 2024)

**The primary productive script here** is https://github.com/Layr-Labs/eigenlayer-contracts/blob/feat-align-environments/script/utils/CurrentConfigCheck.s.sol which was already used to align test environments to mainnet, and contains scripts for checking the existing governance config, simulating change of governance powers to include the Protocol Council, and checks for the desired end-state.

I've also made some productive changes to https://github.com/Layr-Labs/eigenlayer-contracts/blob/feat-align-environments/script/utils/ExistingDeploymentParser.sol

This is a start on a new "from scratch deployment script" but _is not functional right now_ -- I'm unclear on the plan for this, so can either delete it or put work into making it more correct, but I recognize that **it should not be left in its present state, and it is not worth reviewing right now**.

This had a bunch of conflicts with Zeus-related stuff. I just resolved them via a rebase action but we will likely need to sort things out a second time, in a more proper fashion.


MinisculeTarantula:
Update:
This PR could likely use more improvement, but is in a place where I believe it is additive + fairly well-formed, so I am marking it as ready for review, but not explicitly requesting review yet.

script/deploy/local/DeployFromScratchWithChecks.s.sol
```diff
@@ -0,0 +1,374 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.12;
+
+import "./Deploy_From_Scratch.s.sol";
```
MinisculeTarantula, line 4: note that this file has a ton of duplicated code right now. ideally most of it would be de-duplicated!

MinisculeTarantula: update: this file is still in bad shape, definitely needs more work

script/deploy/holesky/Deploy_TimelockController.s.sol
```diff
@@ -0,0 +1,77 @@
...
+        newTimelockController.grantRole(TIMELOCK_ADMIN_ROLE, executorMultisig);
+        newTimelockController.grantRole(CANCELLER_ROLE, operationsMultisig);
+        newTimelockController.grantRole(PROPOSER_ROLE, foundationMultisig);
+        newTimelockController.grantRole(EXECUTOR_ROLE, foundationMultisig);
+        newTimelockController.renounceRole(TIMELOCK_ADMIN_ROLE, msg.sender);
+
+        // STOP RECORDING TRANSACTIONS FOR DEPLOYMENT
+        vm.stopBroadcast();
+
+        require(newTimelockController.hasRole(TIMELOCK_ADMIN_ROLE, executorMultisig),
```
gpsanant, line 64: we should adda a check that the timelock itself has the `TIMELOCK_ADMIN_ROLE`

script/utils/CurrentConfigCheck.s.sol
```diff
@@ -0,0 +1,570 @@
...
+            vm.selectFork(forkId);
+        } else if (keccak256(abi.encodePacked(networkName)) == keccak256(abi.encodePacked("mainnet"))) {
+            deployedContractsConfig = "script/configs/mainnet/mainnet-addresses.config.json";
+            intialDeploymentParams = "script/configs/mainnet/mainnet-config.config.json"; 
+            forkUrl = vm.envString("RPC_MAINNET");
+            uint256 forkId = vm.createFork(forkUrl);
+            vm.selectFork(forkId);
+        }
+
+        if (keccak256(abi.encodePacked(networkName)) == keccak256(abi.encodePacked("local"))) {
```
gpsanant, line 44: `else if`

script/utils/CurrentConfigCheck.s.sol
```diff
@@ -0,0 +1,570 @@
...
+        assertEq(eigenLayerPauserReg.unpauser(), executorMultisig,
+            "eigenLayerPauserReg.unpauser() != operationsMultisig");
+        require(eigenLayerPauserReg.isPauser(operationsMultisig),
+            "operationsMultisig does not have pausing permissions");
+        require(eigenLayerPauserReg.isPauser(executorMultisig),
+            "executorMultisig does not have pausing permissions");
+        require(eigenLayerPauserReg.isPauser(pauserMultisig),
+            "pauserMultisig does not have pausing permissions");
+
+        (bool success, bytes memory returndata) = timelock.staticcall(abi.encodeWithSignature("admin()"));
```
gpsanant, line 122: why not cast to `Timelock`?

script/utils/CurrentConfigCheck.s.sol
```diff
@@ -0,0 +1,570 @@
...
+            "tokens must have different proxy admins to allow different timelock controllers");
+        require(eigenTokenTimelockController != beigenTokenTimelockController,
+            "tokens must have different timelock controllers");
+
+        // note that proxy admin owners are different but _token_ owners are the same
+        assertEq(Ownable(address(EIGEN)).owner(), address(eigenTokenTimelockController),
+            "EIGEN.owner() != eigenTokenTimelockController");
+        assertEq(Ownable(address(bEIGEN)).owner(), address(eigenTokenTimelockController),
+            "bEIGEN.owner() != eigenTokenTimelockController");
+        assertEq(eigenTokenProxyAdmin.owner(), address(eigenTokenTimelockController),
```
gpsanant, line 156: this conflicts with the `script/deploy/holesky/Eigen_Token_Deploy.s.sol` where the owner is the ops multisig

script/utils/CurrentConfigCheck.s.sol
```diff
@@ -0,0 +1,570 @@
...
+            address(beigenTokenProxyAdmin),
+            "beigenTokenProxyAdmin is not actually the admin of the bEIGEN token");
+
+        require(eigenTokenTimelockController.hasRole(eigenTokenTimelockController.PROPOSER_ROLE(), foundationMultisig),
+            "foundationMultisig does not have PROPOSER_ROLE on eigenTokenTimelockController");
+        require(eigenTokenTimelockController.hasRole(eigenTokenTimelockController.EXECUTOR_ROLE(), foundationMultisig),
+            "foundationMultisig does not have EXECUTOR_ROLE on eigenTokenTimelockController");
+        require(eigenTokenTimelockController.hasRole(eigenTokenTimelockController.CANCELLER_ROLE(), operationsMultisig),
+            "operationsMultisig does not have CANCELLER_ROLE on eigenTokenTimelockController");
+        require(eigenTokenTimelockController.hasRole(eigenTokenTimelockController.TIMELOCK_ADMIN_ROLE(), executorMultisig),
```
gpsanant, line 174: we should check that the TC is its own admin

script/utils/CurrentConfigCheck.s.sol
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    function simulateLegacyTimelockActions() public {
+        // TODO
+        // give proxy admin ownership to timelock controller
+        // eigenTokenProxyAdmin.transferOwnership(address(timelockController));
+
+        // swapOwner(address previousOwner, address oldOwner, address newOwner)
+        // TODO: figure out if this is the correct input for all chains or the communityMultisig is correct on some
+        address previousOwner = address(1);
```
gpsanant, line 460: what does this do?

script/utils/CurrentConfigCheck.s.sol
```diff
@@ -0,0 +1,570 @@
...
+        uint256[] memory values = new uint256[](tx_array_length);
+        bytes[] memory payloads = new bytes[](tx_array_length);
+        // 1. transfer upgrade rights over EIGEN token from eigenTokenProxyAdmin to eigenLayerProxyAdmin
+        targets[0] = address(eigenTokenProxyAdmin);
+        payloads[0] = abi.encodeWithSelector(ProxyAdmin.changeProxyAdmin.selector, address(EIGEN), address(eigenLayerProxyAdmin));
+        // 2. transfer ownership of EIGEN token to executorMultisig
+        targets[1] = address(EIGEN);
+        payloads[1] = abi.encodeWithSelector(Ownable.transferOwnership.selector, executorMultisig);
+        // 3. transfer ownership of bEIGEN token to executorMultisig
+        targets[2] = address(bEIGEN);
```
gpsanant, line 507: move to other function?

script/utils/CurrentConfigCheck.s.sol
MinisculeTarantula: this check is already in the `checkTimelockControllerConfig` function:
https://github.com/Layr-Labs/eigenlayer-contracts/blob/ab8f84ba5aa2fd889f50ae874e7eff4e51ee364d/script/utils/CurrentConfigCheck.s.sol#L292-L294

script/utils/CurrentConfigCheck.s.sol
MinisculeTarantula: We could potentially just delete that token deploy script -- it feels like it should become part of the "deploy from scratch" script going forwards, but I think this may not be possible right now due to not being able to compile all contracts at once (conflicting OZ versions, in particular)?

script/utils/CurrentConfigCheck.s.sol
MinisculeTarantula: thought we didn't have the interface in here 🤷 
probably should do that though

script/deploy/holesky/Deploy_TimelockController.s.sol
MinisculeTarantula: these already exist, see comment here https://github.com/Layr-Labs/eigenlayer-contracts/pull/794#discussion_r1829957661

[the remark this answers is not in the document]
MinisculeTarantula: Looks like this comment was on an old commit; this change has been made.

script/utils/CurrentConfigCheck.s.sol
```diff
@@ -0,0 +1,570 @@
...
+        (bool success, bytes memory returndata) = timelock.staticcall(abi.encodeWithSignature("admin()"));
+        require(success, "call to timelock.admin() failed");
+        address timelockAdmin = abi.decode(returndata, (address));
+        assertEq(timelockAdmin, operationsMultisig,
+            "timelockAdmin != operationsMultisig");
+
+        (success, returndata) = executorMultisig.staticcall(abi.encodeWithSignature("getOwners()"));
+        require(success, "call to executorMultisig.getOwners() failed");
+        address[] memory executorMultisigOwners = abi.decode(returndata, (address[]));
+        require(executorMultisigOwners.length == 2,
```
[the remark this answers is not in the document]
MinisculeTarantula: I literally always use require, personally.

assert still doesn't support error strings https://docs.soliditylang.org/en/v0.8.26/control-structures.html#panic-via-assert-and-error-via-require

script/utils/CurrentConfigCheck.s.sol
MinisculeTarantula: This is a required input, and the call will revert without it being correct.
(Gnosis) Safes keep a weird, internal and not-very-easy-to-query linked-list-esque(?) structure of owners which uses address(1) as one end.
Typically this will be another owner of the Safe, but in general as I understand it it's the other node which will need to be attached to the new node to complete the linked structure.

script/utils/CurrentConfigCheck.s.sol
MinisculeTarantula: no? why? this timelock is currently the owner of both EIGEN and bEIGEN, so this is how we will actually perform this action.

script/utils/CurrentConfigCheck.s.sol
MinisculeTarantula: done in https://github.com/Layr-Labs/eigenlayer-contracts/pull/794/commits/d5844681cdfcc2a90b80a80ac97d440e1bf39736

script/utils/CurrentConfigCheck.s.sol
MinisculeTarantula: switched this in https://github.com/Layr-Labs/eigenlayer-contracts/pull/794/commits/d5844681cdfcc2a90b80a80ac97d440e1bf39736

script/deploy/local/DeployFromScratchWithChecks.s.sol
MinisculeTarantula: file deleted in https://github.com/Layr-Labs/eigenlayer-contracts/pull/794/commits/e8398a1f9b559ca5ae3f5cfd60c2717c210d1e1d
I've been told Zeus development will cover this

script/utils/CurrentConfigCheck.s.sol
gpsanant: ok cool, as long as the test works

gpsanant (dismissed):

PR #796, opened 2024-10-04
fix: index out of bound error

shrimalmadhur:
- Fix index out of bound issues while `_delegate`


gpsanant (approved):

shrimalmadhur:
A lot has changed and this change is not valid anymore. so closing this.

PR #802, opened 2024-10-06
chore: add missing functions to the dm interface

stevennevins:


ypatil12:
Looks like we already added this, closing:

https://github.com/Layr-Labs/eigenlayer-contracts/blob/dc7e9293615d6d5d6489fdfb8ec74443ead539b4/src/contracts/interfaces/IDelegationManager.sol#L482-L498

PR #803, opened 2024-10-07
feat: eip712 improvements

0xClandestine:
- removes `utils/EIP1271SignatureUtils.sol`
- adds `mixins/SignatureUtils.sol`
        - New mixin uses OZ SignatureChecker,  and condenses repeat EIP-712 related logic


src/contracts/core/RewardsCoordinator.sol
```diff
@@ -527,26 +527,4 @@ contract RewardsCoordinator is
...
     }
-
-    /**
-     * @notice Getter function for the current EIP-712 domain separator for this contract.
-     *
-     * @dev The domain separator will change in the event of a fork that changes the ChainID.
-     * @dev By introducing a domain separator the DApp developers are guaranteed that there can be no signature collision.
-     * for more detailed information please read EIP-712.
-     */
-    function domainSeparator() public view returns (bytes32) {
```
gpsanant, line 538: do we just not use EIP1271?

src/contracts/mixins/SignatureUtils.sol
```diff
@@ -0,0 +1,83 @@
...
+pragma solidity ^0.8.0;
+
+import "@openzeppelin-upgrades/contracts/utils/cryptography/SignatureCheckerUpgradeable.sol";
+
+import "../interfaces/ISignatureUtils.sol";
+
+/// @title SignatureUtils
+/// @notice A mixin to provide EIP-712 signature validation utilities.
+/// @dev Domain name is hardcoded to "EigenLayer".
+abstract contract SignatureUtils is ISignatureUtils {
```
gpsanant, line 11: can we import this in all storage contracts since it has no storage itself so inheritance order doesn't matter

src/contracts/interfaces/IAVSDirectory.sol
```diff
@@ -26,6 +26,7 @@ interface IAVSDirectory is ISignatureUtils {
     error InvalidOperatorSet();
     /// @dev Thrown when `operator` is not a registered operator.
     error OperatorNotRegistered();
+
```
gpsanant, line 29: ?

gpsanant (changes requested):
few comments. can u also generate storage diff

src/contracts/core/RewardsCoordinator.sol
0xClandestine: Yeah, RC currently does not need signature utils.

src/contracts/mixins/SignatureUtils.sol
0xClandestine: We could, but it's not needed for several core contracts.

0xClandestine:
closing in favor of https://github.com/Layr-Labs/eigenlayer-contracts/pull/808

PR #804, opened 2024-10-07
slashing contract docs

gpsanant:


ypatil12 (approved):
We may want to add a separate breaking accounting changes in an additional PR, but this is good enough for write interactions related to slashing.

gpsanant:
TODO: add slashing per block req

gpsanant:
Took out oncer per block slash limit cause it was too gross. discussing offline.

PR #806, opened 2024-10-08
add events, fix bugs, abstract better

gpsanant:


src/contracts/core/AllocationManager.sol
```diff
@@ -269,28 +308,75 @@ contract AllocationManager is
...
+        IStrategy strategy,
+        bytes32 operatorSetKey,
+        MagnitudeInfo memory mInfo,
+        bool isCompletion
+    ) internal {
+        MagnitudeInfo memory mInfoStored = _operatorMagnitudeInfo[operator][strategy][operatorSetKey];
+        OperatorSet memory operatorSet = _decodeOperatorSet(operatorSetKey);
+
+        // if the current magnitude has been updated, this is a slash, emit an event
+        if (mInfoStored.currentMagnitude != mInfo.currentMagnitude && !isCompletion) {
```
ypatil12, line 349: pendingMagnitudeDelta is also changed in a slash to pending deallocation? 

Should be something like:
``` solidity
// Completions have already had magnitude info emitted from its initial deallocation
if (!isCompletion) {
       if (mInfoStored.currentMagnitude != mInfo.currentMagnitude) ....
       else if (mInfoStored.pendingMagnitudeDelta != mInfo.pendingMagnitudeDelta) ...
}
```

src/contracts/core/AllocationManager.sol
```diff
@@ -195,25 +211,48 @@ contract AllocationManager is
...
-            );
+            delegation.decreaseOperatorShares({
+                operator: operator,
+                strategy: strategies[i],
+                previousTotalMagnitude: totalMagnitudeAfterSlash + slashedMagnitude,
+                newTotalMagnitude: totalMagnitudeAfterSlash
+            });
         }
+
+        emit OperatorSlashed(operator, _decodeOperatorSet(operatorSetKey), description);
```
ypatil12, line 255: I feel like we'll have to change this to make it easier for offchain to associate the slashes  with the associated mag changes. Perhaps something like a slashID per operator??

ypatil12 (changes requested):

src/contracts/core/DelegationManager.sol
```diff
@@ -605,35 +615,25 @@ contract DelegationManager is
...
         address operator,
         address staker,
         IStrategy strategy,
         uint256 existingDepositShares,
-        uint256 addedDepositShares,
+        uint256 addedShares,
         uint64 totalMagnitude
     ) internal {
         //TODO: double check ordering here
-        operatorShares[operator][strategy] +=
```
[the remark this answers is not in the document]
gpsanant: > depositShares and are "reused" as withdrawableShares
why do you say this?

src/contracts/core/AllocationManager.sol
gpsanant: both current and pending are changed, will wrap in `if (!isCompletion)` tho

src/contracts/core/AllocationManager.sol
gpsanant: any decrease such that `block.timestamp == effectTimestamp` is a slash

any decrease in the future that overwrites a previous decrease a the same time is a slash.

I can easily add these checks in the `OperatorSetMagnitudeUpdate` and add a bool, would that work?

src/contracts/core/AllocationManager.sol
ypatil12: Ah I read this wrong since you’re also updating the mag of the pending which is still completable at the effect timestamp. This is correct

ypatil12: Wrapping conditional cleaner tho, ince

src/contracts/core/DelegationManager.sol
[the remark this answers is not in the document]
ypatil12: They are depositShares in the SM/EPM world and then the naming/meaning  changes when we call them shares in the DM world. Its correct logically but we just need to be explicit that its okay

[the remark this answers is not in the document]
gpsanant: which thing are you pointing out is misnamed?

src/contracts/core/AllocationManager.sol
gpsanant: or are you looking at connecting a reason to a slash?

perhaps we don't allow two slashes in the same tx/block for the same operatorSet?

ypatil12: How does the subgraph distinguish between two slashes for different slash reasons that occurred in the same block from a multicall?

gpsanant: right. shall we add a
```solidity
///  @notice operator => strategy => operator set (encoded) => last timestamp at which the tuple was slashed
mapping(address => mapping(IStrategy => mapping(bytes32 => uint32))) lastSlashedTimestamp;
```
and enforce its nonzero for the current timestamp

(also i hate to say it: `tstore`!)

ypatil12: That seems fine for now, not ideal

src/contracts/core/DelegationManager.sol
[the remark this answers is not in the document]
ypatil12: Not technically misnamed, just want to add a comment, I’ll make a commit. This is just me used to old naming (depositShares for sharesAdded as opposed to shares)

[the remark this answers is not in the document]
gpsanant: ah i see what you're saying

src/contracts/core/AllocationManager.sol
gpsanant: done https://github.com/Layr-Labs/eigenlayer-contracts/pull/806/commits/d94bf6f353efcdf57f00304560967ba9ca2e3220

src/contracts/core/DelegationManager.sol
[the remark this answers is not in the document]
gpsanant: addedDepositShares = addedShares for deposits

ypatil12 (approved):

PR #811, opened 2024-10-09
POC: Delegated Accounts

ypatil12:
Minimal POC of delegated accounts which includes:

- 2-step ownable for accepting agency
- agency at the granularity of func selector and target
- Ability to revoke agency so the admin OR the agent has the ability to instantly stop agency over an identifier. This is useful if the admin believes the agent is compromised or if the agent wants to service a new (identifier, target, selector) tuple

Assumes that the identifier sets the admin


src/contracts/libraries/OwnershipManagerLib.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import "../interfaces/IOwnershipManager.sol";
+
+
+library OwnershipManagerLib {
+    function getIdentifier(address agent, address ownershipManager) internal view returns (bytes20) {
+        return IOwnershipManager(ownershipManager).getIdentifier(agent, address(this), bytes4(msg.data[0:4]));
```
stevennevins, line 9: passing selector > msg.sig > slicing the calldata+casting

stevennevins: passing a storage reference to the mapping that getIdentifier uses would probably be cleaner here

PR #814, opened 2024-10-10
add Strategy <> OperatorSet mapping in storage, and APIs and events

bowenli86:


src/contracts/core/AVSDirectory.sol
```diff
@@ -451,6 +526,26 @@ contract AVSDirectory is
         }
     }
 
+    /**
+     * @notice Returns an array of strategies in the operatorSet.
+     * @param operatorSet The operatorSet to query.
+     * @param start The starting index of the array to query.
+     * @param length The amount of items of the array to return.
+     */
+    function getStrategiesInOperatorSet(
```
stevennevins, line 535: @bowenli86 would it make your life easier if we just return all the enumerated strategies in the operator set?  I don't love this pattern for paginating these values with start and length params

src/contracts/core/AVSDirectoryStorage.sol
```diff
@@ -69,6 +69,10 @@ abstract contract AVSDirectoryStorage is IAVSDirectory {
     /// @dev Each key is formatted as such: bytes32(abi.encodePacked(avs, uint96(operatorSetId)))
     mapping(bytes32 => EnumerableSet.AddressSet) internal _operatorSetMembers;
 
+    /// @notice Mapping: operatorSet => List of strategies that the operatorSet contains
+    /// @dev Each key is formatted as such: bytes32(abi.encodePacked(avs, uint96(operatorSetId)))
+    mapping(bytes32 => EnumerableSet.AddressSet) internal _operatorSetStrategies;
+
```
ypatil12, line 75: Update storage gap

ypatil12 (commented):
LGTM, just update gap

src/contracts/core/AVSDirectoryStorage.sol
gpsanant: done

ypatil12 (approved):

ypatil12 (approved):

PR #825, opened 2024-10-15
feat: two queues for allocation/deallocation

8sunyuan:
In order to preserve sortedness in the `_clearModificationQueue` helper, we need the modificationQueue array to be sorted based on effectTimestamp. However if we have allocations and deallocations stored here, ordering is not preserved because they have different delays (deallo delay is fixed while allocation delay is fixed).

Also added an additional require on allocation delays. The configured delay is bounded by `ALLOCATION_CONFIGURATION_DELAY`. This is also needed to ensure the `allocationQueue` is sorted in ascending effectTimestamps


ypatil12:
Closing since we are adding deallocations to queue instead

PR #827, opened 2024-10-15
Fix: Single Deallocation Queue

ypatil12:
See https://github.com/Layr-Labs/eigenlayer-contracts/pull/826

### DIFF
--- a/src/test/unit/AllocationManagerUnit.t.sol
+++ b/src/test/unit/AllocationManagerUnit.t.sol
@@ -833,7 +833,7 @@ contract AllocationManagerUnitTests_SlashOperator is AllocationManagerUnitTests 
 
         // Check storage after complete modification
         cheats.warp(deallocationEffectTimestamp);
-        allocationManager.clearModificationQueue(defaultOperator, _strategyMockArray(), _maxNumToClear());
+        allocationManager.clearDeallocationQueue(defaultOperator, _strategyMockArray(), _maxNumToClear());
         mInfos = allocationManager.getAllocationInfo(defaultOperator, strategyMock, allocations[0].operatorSets);
         assertEq(magnitudeAfterDeallocationSlash, mInfos[0].currentMagnitude, "currentMagnitude not updated");
         assertEq(magnitudeAfterDeallocationSlash, maxMagnitudeAfterSlash / 2, "magnitude after deallocation should be half of max magnitude, since we originally deallocated by half");
@@ -929,8 +929,8 @@ contract AllocationManagerUnitTests_SlashOperator is AllocationManagerUnitTests 
 
     /**
      * Slashes the operator after deallocation, even if the deallocation has not been cleared. Validates that:
-     * 1. Even if we do not clear modification queue, the deallocation is NOT slashed from since we're passed the deallocationEffectTimestamp
-     * 2. Validates storage post slash & post clearing modification queue
+     * 1. Even if we do not clear deallocation queue, the deallocation is NOT slashed from since we're passed the deallocationEffectTimestamp
+     * 2. Validates storage post slash & post clearing deallocation queue
      * 3. Total magnitude only decreased proportionally by the magnitude set after deallocation
      */
     function test_allocate_deallocate_slashAfterDeallocation() public {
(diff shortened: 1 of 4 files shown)


src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -1875,6 +1873,118 @@ contract AllocationManagerUnitTests_ClearModificationQueue is AllocationManagerU
...
+        assertEq(allocationEffectTimestamp, mInfos[0].effectTimestamp, "effect timestamp not correct");
+        assertLt(allocationEffectTimestamp, deallocationEffectTimestamp, "invalid test setup");
+
+        // Warp to allocation effect timestamp & clear the queue
+        cheats.warp(allocationEffectTimestamp);
+        allocationManager.clearDeallocationQueue(defaultOperator, _strategyMockArray(), _maxNumToClear());
+
+        // Validate `getAllocatableMagnitude`. Allocatable magnitude should be the difference between the total magnitude and the encumbered magnitude
+        uint64 allocatableMagnitude = allocationManager.getAllocatableMagnitude(defaultOperator, strategyMock);
+        assertEq(WAD - 33e16 - 5e17, allocatableMagnitude, "allocatableMagnitude not correct");
```
8sunyuan, line 1925: This check makes sense to me. I'm assuming the deallocation still is in the queue here at this moment of time as well right. Since the second allocation has an effectTimestamp earlier than the deallocation, and we warped to the effectTimestamp of the second allocation.

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -1875,6 +1873,118 @@ contract AllocationManagerUnitTests_ClearModificationQueue is AllocationManagerU
...
+    }
+
+    /**
+     * Allocates to opset1, allocates to opset2, deallocates from opset1. Asserts that the allocation, which has a higher
+     * effect timestamp is not blocking the deallocation.
+     * The queue looks like (note that the first allocation has been completed):
+     * 1. (allocation, opSet2, mag: 5e17, effectTimestamp: 50th day)
+     * 2. (deallocation, opSet1, mag: 0, effectTimestamp: 42.5 day)
+     */
+    function test_regression_allocationBlocksDeallocation() public {
```
8sunyuan, line 1941: nit: testname sounds like we expect the allocation to block the deallocation.
Maybe `test_regression_deallocationNotBlocked`?

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -1875,6 +1873,118 @@ contract AllocationManagerUnitTests_ClearModificationQueue is AllocationManagerU
...
+        IAllocationManagerTypes.MagnitudeAllocation[] memory thirdAllocation =
+            _generateMagnitudeAllocationCalldataForOpSet(defaultAVS, 2, 10e16, 1e18);
+        cheats.prank(defaultOperator);
+        allocationManager.modifyAllocations(thirdAllocation);
+    }
+
+    /**
+     * Allocates to opset1, allocates to opset2, deallocates from opset1. Asserts that the allocation, which has a higher
+     * effect timestamp is not blocking the deallocation.
+     * The queue looks like (note that the first allocation has been completed):
```
8sunyuan, line 1937: Maybe clarify that the actual allocation is no longer in the queue storage otherwise your description is slightly confusing. But it is helpful to know what the actual state is for the operator regarding allocations/deallocations

8sunyuan (approved):
Couple nits but looks fire otherwise

src/test/unit/AllocationManagerUnit.t.sol
ypatil12: Updated

src/test/unit/AllocationManagerUnit.t.sol
ypatil12: Good point. Clarified this

src/test/unit/AllocationManagerUnit.t.sol
ypatil12: Exactly

gpsanant (approved):

PR #830, opened 2024-10-16
fix: fixed compile

haike0513:
fixed make compile with error


ypatil12:
Thanks for this :). I'm merging this into `fix/dev-compilation` since that also fixes issues with the latest foundry version.

PR #837, opened 2024-10-16
feat: Rewards v2

0xrajath:
# Motivation

As part of the Rewards v2 release, we need to support the following:

1. Refactoring existing `commission` terminology to `split`, and `global` terminology to `default`.
2. Operator-Directed Rewards submission by AVSs.
3. Batching claiming by users to be more gas efficient.
4. Variable Operator split per AVS (Settable by the operator with a 7-day delay. Default of 10%)
5. Variable Operator split for Programmatic Incentives (Settable by the operator with a 7-day delay. Default of 10%)

# Modifications

* Refactoring existing `commission` terminology to `split`, and `global` terminology to `default` for RewardsCoordinator in `src`, `script` and `test`. 

* `RewardsCoordinator` and `IRewardsCoordinator`: 
    * `createOperatorDirectedAVSRewardsSubmission` external function and related changes that allows AVSs to submit operator-directed rewards.
    * `setOperatorAVSSplit` external function and related changes that allows operators to set their split per AVS.
    * `setOperatorPISplit` external function and related changes that allows operators to set their split for Programmatic Incentives.
    * Refactored code in `processClaim` into an internal function called `_processClaim`
    * `processClaims` external function and related changes that iterates over `_processClaim` to allow batch claiming of rewards for earners.
    * Renamed `operatorCommissionBips()` view function to `getOperatorAVSSplit()` and added logic to handle default value and operator<>avs split.
    * `getOperatorPISplit()` view function with logic to handle default value and operator PI split.
    
* `RewardsCoordinatorStorage` : 
    * Added 3 new storage variables.
    * Updated storage gap from 40 to 37 slots to accomodate new storage variables.

* Updated Test Cases
* Updated Storage Report
* Updated Bindings

# Result

Rewards v2 release.


lib/forge-std
0xrajath, line 1: Will fix this. Something happened while rebasing.

script/deploy/holesky/upgrade_preprod_rewardsCoordinator.s.sol
0xrajath, line 1: This file was touched since `dev` wasn't compiling and I manually made it `_verifyContractsInitialized(false);`. But now I need to figure out what's up with my linting. Either these existing files haven't been linted using prettier or we're currently not using forge fmt (My IDE is auto linting them on save).

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -118,11 +120,9 @@ contract RewardsCoordinator is
...
 
     /// @inheritdoc IRewardsCoordinator
-    function createAVSRewardsSubmission(RewardsSubmission[] calldata rewardsSubmissions)
-        external
-        onlyWhenNotPaused(PAUSED_AVS_REWARDS_SUBMISSION)
-        nonReentrant
-    {
+    function createAVSRewardsSubmission(
+        RewardsSubmission[] calldata rewardsSubmissions
+    ) external onlyWhenNotPaused(PAUSED_AVS_REWARDS_SUBMISSION) nonReentrant {
```
0xrajath, line 125: This was changed due to auto linting.

0xrajath: Multiple other auto-linting instances in this PR.

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -340,6 +363,83 @@ contract RewardsCoordinator is
         }
     }
 
+    /**
+     * @notice Validate a PerformanceRewardsSubmission. Called from `createAVSPerformanceRewardsSubmission`.
+     * @param performanceRewardsSubmission PerformanceRewardsSubmission to validate.
+     * @return total amount to be transferred from the avs to the contract.
+     */
+    function _validatePerformanceRewardsSubmission(
```
0xrajath, line 371: Not checking for `MAX_FUTURE_LENGTH` (Since Performance based reward submissions are retroactive) or `MAX_REWARDS_AMOUNT` (Since we no longer have the `1e38 - 1` limitation in the offchain calculation.

src/contracts/core/RewardsCoordinatorStorage.sol
```diff
@@ -89,6 +89,9 @@ abstract contract RewardsCoordinatorStorage is IRewardsCoordinator {
     /// if rewards submission hash for all stakers and operators has been submitted
     mapping(address => mapping(bytes32 => bool)) public isRewardsSubmissionForAllEarnersHash;
 
+    /// @notice Mapping: avs => avsPerformanceRewardsSubmissionHash => bool to check if performance rewards submission hash has been submitted
+    mapping(address => mapping(bytes32 => bool)) public isAVSPerformanceRewardsSubmissionHash;
```
0xrajath, line 93: Was there a reason we needed separate mappings for each reward type? I just copied the current setup but questioning the need for it here.

src/contracts/core/RewardsCoordinatorStorage.sol
```diff
@@ -120,5 +123,5 @@ abstract contract RewardsCoordinatorStorage is IRewardsCoordinator {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[40] private __gap;
+    uint256[39] private __gap;
```
0xrajath, line 126: I've updated this since we added the mapping above. Just wanted to double confirm that this looks right. I remember seeing a PR a while back changing this to 40.

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -178,41 +172,56 @@ contract RewardsCoordinator is
...
-            cumulativeClaimed[earner][tokenLeaf.token] = tokenLeaf.cumulativeEarnings;
+        _processClaim(claim, recipient);
+    }
 
-            tokenLeaf.token.safeTransfer(recipient, claimAmount);
-            emit RewardsClaimed(root.root, earner, claimer, recipient, tokenLeaf.token, claimAmount);
+    /// @inheritdoc IRewardsCoordinator
+    function processClaims(
+        RewardsMerkleClaim[] calldata claims,
+        address recipient
```
0xrajath, line 221: We could also claim for multiple recipients instead of claiming to a single recipient. Thoughts?

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -340,6 +349,83 @@ contract RewardsCoordinator is
         }
     }
 
+    /**
+     * @notice Validate a PerformanceRewardsSubmission. Called from `createAVSPerformanceRewardsSubmission`.
+     * @param performanceRewardsSubmission PerformanceRewardsSubmission to validate.
+     * @return total amount to be transferred from the avs to the contract.
+     */
+    function _validatePerformanceRewardsSubmission(
```
0xrajath, line 357: Thoughts on using `revert` with custom errors? Or just custom errors in general instead of error strings? I kept it similar to the current setup but using custom errors will be more gas efficient.

src/contracts/core/RewardsCoordinator.sol
gpsanant: single recipient seems fine

src/contracts/core/RewardsCoordinator.sol
gpsanant: let's do that with slashing like we are on #727

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -178,41 +172,56 @@ contract RewardsCoordinator is
...
+        for (uint256 i = 0; i < performanceRewardsSubmissions.length; i++) {
+            PerformanceRewardsSubmission calldata performanceRewardsSubmission = performanceRewardsSubmissions[i];
+            uint256 nonce = submissionNonce[msg.sender];
+            bytes32 performanceRewardsSubmissionHash = keccak256(
+                abi.encode(msg.sender, nonce, performanceRewardsSubmission)
+            );
+
+            uint256 totalAmount = _validatePerformanceRewardsSubmission(performanceRewardsSubmission);
+
+            isAVSPerformanceRewardsSubmissionHash[msg.sender][performanceRewardsSubmissionHash] = true;
```
gpsanant, line 197: do we need to check that it already exists

src/contracts/core/RewardsCoordinator.sol
gpsanant: could u add this in code comments?

src/contracts/core/RewardsCoordinatorStorage.sol
gpsanant: let's combine into one, agreed

src/contracts/core/RewardsCoordinator.sol
0xrajath: The nonce keeps increasing. So it can theoretically never exist already. That was the point of the nonce here.

src/contracts/core/RewardsCoordinator.sol
0xrajath: Alright holding off on custom errors for now.

src/contracts/core/RewardsCoordinatorStorage.sol
0xrajath: I'm assuming this is used by the rewards pipeline/sidecar in some form. I don't want to break backwards compatibility if that's the case.

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -340,6 +349,83 @@ contract RewardsCoordinator is
...
+        address currOperatorAddress = address(0);
+        for (uint256 i = 0; i < performanceRewardsSubmission.operatorRewards.length; ++i) {
+            OperatorReward calldata operatorReward = performanceRewardsSubmission.operatorRewards[i];
+            require(
+                operatorReward.operator != address(0),
+                "RewardsCoordinator._validatePerformanceRewardsSubmission: operator cannot be 0 address"
+            );
+            require(
+                currOperatorAddress < operatorReward.operator,
+                "RewardsCoordinator._validatePerformanceRewardsSubmission: operators must be in ascending order to handle duplicates"
```
scotthconner, line 379: Is this a more gas efficient way of not having a temporary storage slot? Seems rather restrictive on the client but not seemingly impossible I suppose.

scotthconner: Also, who cares if there are duplicates? Is it up to us to infer their intention? Are there any side effects of allowing duplicates?

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -340,6 +349,83 @@ contract RewardsCoordinator is
...
+        );
+
+        // Require performanceRewardsSubmission is for whitelisted strategy or beaconChainETHStrategy
+        address currAddress = address(0);
+        for (uint256 i = 0; i < performanceRewardsSubmission.strategiesAndMultipliers.length; ++i) {
+            IStrategy strategy = performanceRewardsSubmission.strategiesAndMultipliers[i].strategy;
+            require(
+                strategyManager.strategyIsWhitelistedForDeposit(strategy) || strategy == beaconChainETHStrategy,
+                "RewardsCoordinator._validatePerformanceRewardsSubmission: invalid strategy considered"
+            );
```
scotthconner, line 418: does this propertly support longtail assets?

src/contracts/core/RewardsCoordinator.sol
0xrajath: Is your question: why are we doing this check? If that's the case we're doing it so that the AVS can't pass in duplicate operator addresses into the reward submission, and this is the cheapest way we can check that without doing any searching of the array. 

The cli/sdk can easily sort it into this order and warn if there are duplicates before the transaction is submitted onchain.

src/contracts/core/RewardsCoordinator.sol
0xrajath: Good question. It was something I had noted down myself to check.

src/contracts/core/RewardsCoordinator.sol
0xrajath: We do this duplicate check for the Rewards v1 submission. So I'm assuming it's some constraint on the pipeline and to make some calculation easier (I'm just speculating, but I could be wrong - Need to check this assumption). 

But besides that why would an AVS need to pay the same operator 2 different amounts in the same reward submission? This seems like a nice constraint to have.

src/contracts/core/RewardsCoordinatorStorage.sol
gpsanant: @ypatil12 @8sunyuan may have opinions

src/contracts/core/RewardsCoordinator.sol
gpsanant: ah, sounds good

src/contracts/core/RewardsCoordinatorStorage.sol
ypatil12: You can have collisions, in the worst case, between `AVSRewardsSubmission`, `RewardsSubmissionForAllEarners`, and `RewardsSubmissionForAll`  if they are submitted with the same params in the same block. You should be able to combine all the AVS submitted reward types into one mapping.

ypatil12: This should be fine since we include the nonce in the rewards submission hash. cc @8sunyuan to work

ypatil12: Internally we're not using this mapping offchain

src/contracts/core/RewardsCoordinatorStorage.sol
ypatil12: Yup, this is correct

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -340,6 +349,83 @@ contract RewardsCoordinator is
...
+            require(
+                currOperatorAddress < operatorReward.operator,
+                "RewardsCoordinator._validatePerformanceRewardsSubmission: operators must be in ascending order to handle duplicates"
+            );
+            currOperatorAddress = operatorReward.operator;
+            require(
+                operatorReward.amount > 0,
+                "RewardsCoordinator._validatePerformanceRewardsSubmission: operator reward amount cannot be 0"
+            );
```
ypatil12, line 385: What are the gas costs for 200,500,1000 operators?

src/contracts/interfaces/IRewardsCoordinator.sol
```diff
@@ -60,6 +70,42 @@ interface IRewardsCoordinator {
...
+     *             <--valid range for startTimestamp--
+     *             ^
+     *         GENESIS_REWARDS_TIMESTAMP
+     *
+     *
+     * Scenario B: GENESIS_REWARDS_TIMESTAMP IS OUT OF RANGE
+     *         <--------MAX_RETROACTIVE_LENGTH--------> t (block.timestamp)
+     *         <----valid range for startTimestamp----
+     *     ^
+     * GENESIS_REWARDS_TIMESTAMP
```
ypatil12, line 87: `GENESIS_REWARDS_TIMESTAMP` is never in range?

src/contracts/core/RewardsCoordinator.sol
ypatil12: Rewards works out of the box for longtail assets without overflow since we limit the number of shares that can be minted for any strategy

src/contracts/core/RewardsCoordinatorStorage.sol
0xrajath: Oh if it's not being used offchain, what was the purpose of this mapping? I don't see it being used anywhere else besides just setting it during reward submission.

src/contracts/core/RewardsCoordinator.sol
0xrajath: Done in [179aa1f](https://github.com/Layr-Labs/eigenlayer-contracts/pull/837/commits/179aa1fb13d8d8d0aede96e87173978bfc0a5c62) as a `@dev` natspec.

src/contracts/core/RewardsCoordinatorStorage.sol
```diff
@@ -57,7 +57,7 @@ abstract contract RewardsCoordinatorStorage is IRewardsCoordinator {
      */
     DistributionRoot[] internal _distributionRoots;
 
-    /// Slot 3
+    /// Slot 2
```
0xrajath, line 60: Updated the comment here. It is in fact Slot 2 since we start from Slot 0 (rather than Slot 1)

src/contracts/interfaces/IRewardsCoordinator.sol
0xrajath: What do you mean?

src/contracts/core/RewardsCoordinator.sol
0xrajath: Is that question wrt to this check or in general how much it would cost if 200,500,1000 operators were passed in calldata?

ypatil12: The latter

src/contracts/interfaces/IRewardsCoordinator.sol
ypatil12: For the performance based rewards it's not possible to submit a reward that starts at `GENESIS_REWARDS_TIMESTAMP` (ie. scenario B)

src/contracts/core/RewardsCoordinatorStorage.sol
ypatil12: @8sunyuan do you remember? I can't recall off the top of my head

ypatil12: It appears it may actually have no use... potentially we wanted it for some sort of future pipeline integration

src/contracts/core/RewardsCoordinator.sol
0xrajath: Here's a spreadsheet based on 64 bytes/operator calldata: https://docs.google.com/spreadsheets/d/1MOFK9wJQwmcktVywyiXOqyLYPNfMNo0ko9QU0bA2ua4/edit?gid=0#gid=0

0xrajath: Do you want me to document that somewhere? Or some other form of gas report using test cases?

src/contracts/interfaces/IRewardsCoordinator.sol
0xrajath: Well you can if `MAX_RETROACTIVE_LENGTH` becomes a modifiable variable in the future. But I agree as of now it's theoretically not possible.

src/contracts/core/RewardsCoordinator.sol
0xrajath: @ypatil12 Are you saying that we don't need this check?

src/contracts/core/RewardsCoordinatorStorage.sol
0xrajath: Keeping it for uniformity's sake.

src/contracts/core/RewardsCoordinatorStorage.sol
```diff
@@ -89,6 +89,15 @@ abstract contract RewardsCoordinatorStorage is IRewardsCoordinator {
...
 
+    /// @notice Mapping: avs => avsPerformanceRewardsSubmissionHash => bool to check if performance rewards submission hash has been submitted
+    mapping(address => mapping(bytes32 => bool)) public isAVSPerformanceRewardsSubmissionHash;
+
+    /// @notice Mapping: operator => avs => OperatorCommission. The commission an operator takes for a specific AVS.
+    mapping(address => mapping(address => OperatorCommission)) internal operatorAVSCommissionBips;
+
+    /// @notice Mapping: operator => OperatorPICommission. The commission an operator takes for Programmatic Incentives.
+    mapping(address => OperatorCommission) internal operatorPICommissionBips;
+
```
0xrajath, line 100: I've marked these as `internal` since we'll have separate getters for getting the commission (with defaults included). I don't want these to be directly accessible to other external contracts.

src/contracts/libraries/TimestampCheckpoints.sol
0xrajath, line 1: Ignore this file for now. I'm still trying to think of the best way to handle overlapping delays for commission setting. Have a couple of solutions in mind.

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -271,6 +289,55 @@ contract RewardsCoordinator is
...
     }
 
+    /// @inheritdoc IRewardsCoordinator
+    function setOperatorAVSCommission(address operator, address avs, uint16 commission) external {
+        require(msg.sender == operator, "RewardsCoordinator.setOperatorAVSCommission: caller is not the operator");
+        require(
+            commission <= ONE_HUNDRED_IN_BIPS,
+            "RewardsCoordinator.setOperatorAVSCommission: commission must be <= 10000 bips"
+        );
+        uint32 activatedAt = SafeCast.toUint32(block.timestamp) + activationDelay;
```
0xrajath, line 299: I've noticed we don't safe cast in other functions. So happy to remove if we feel strongly about it. `uint32` won't overflow anyway until ~2100 I think.

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -271,6 +289,69 @@ contract RewardsCoordinator is
...
+        OperatorCommission storage operatorCommission = operatorAVSCommissionBips[operator][avs];
+
+        // If, the earlier 'new' commission is activated, we update the 'old' commission with the earlier 'new' commission.
+        // Else, the earlier 'old' commission remains the same. This is essentially resetting the activation delay window
+        // since the earlier commission setting didn't complete.
+        if (block.timestamp >= operatorCommission.activatedAt) {
+            operatorCommission.oldCommissionBips = operatorCommission.newCommissionBips;
+        }
+        operatorCommission.newCommissionBips = commission;
+        operatorCommission.activatedAt = activatedAt;
```
0xrajath, line 310: Happy to clarify if this section is confusing. It's essentially handling the case if an operator tries setting a commission again before their previous commission activates. The intended behavior here in that case is that the activation delay window is just extended starting from the current block.timestamp.

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -271,6 +293,76 @@ contract RewardsCoordinator is
...
+            operator,
+            avs,
+            activatedAt,
+            operatorCommission.oldCommissionBips,
+            commission
+        );
+    }
+
+    /// @inheritdoc IRewardsCoordinator
+    function setOperatorPICommission(
```
ypatil12, line 331: We can dedupe the setting & checking logic with a helper function `_setOperatorCommission(OperatorCommission storage operatorCommission, uint16 newCommission)`

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -477,8 +682,35 @@ contract RewardsCoordinator is
...
+            return
+                (block.timestamp >= operatorCommission.activatedAt)
+                    ? operatorCommission.newCommissionBips
+                    : operatorCommission.oldCommissionBips;
+        }
+    }
+
+    /// @inheritdoc IRewardsCoordinator
+    function getOperatorPICommission(address operator) external view returns (uint16) {
+        OperatorCommission memory operatorCommission = operatorPICommissionBips[operator];
```
ypatil12, line 702: We can dedupe the getter logic here too

ypatil12 (commented):
Minor Organizational Comments

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -187,25 +196,26 @@ contract RewardsCoordinator is
...
+            submissionNonce[avs] = nonce + 1;
 
             emit AVSPerformanceRewardsSubmissionCreated(
                 msg.sender,
+                avs,
                 nonce,
                 performanceRewardsSubmissionHash,
                 performanceRewardsSubmission
             );
+            // TODO: Figure out if avs or msg.sender should be sender here in the context of delegated accounts.
```
gpsanant, line 218: prob msg.sender makes the most sense, that way i can delegate the payer for my rewards

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -284,16 +294,73 @@ contract RewardsCoordinator is
...
+            "RewardsCoordinator.setOperatorPICommission: commission must be <= 10000 bips"
+        );
+        require(
+            commission >= MIN_PI_COMMISSION_BIPS,
+            "RewardsCoordinator.setOperatorPICommission: commission must be >= 1000 bips"
+        );
+        uint32 activatedAt = SafeCast.toUint32(block.timestamp) + activationDelay;
 
-        OperatorAVSCommission storage operatorAVSCommission = operatorAVSCommissionBips[operator][avs];
+        OperatorCommission storage operatorCommission = operatorPICommissionBips[operator];
```
gpsanant, line 346: whats the reasoning for a different map as opposed to a magic PI AVS?

src/contracts/core/RewardsCoordinator.sol
0xrajath: Yeah I feel the same. The delegated account can approve and call and submit the rewards.

0xrajath: Done in [92201fb](https://github.com/Layr-Labs/eigenlayer-contracts/pull/837/commits/92201fb1751d961015aee116b415bef3921d9f39)

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -178,41 +181,60 @@ contract RewardsCoordinator is
...
         }
     }
 
+    /// @inheritdoc IRewardsCoordinator
+    function createAVSPerformanceRewardsSubmission(
+        address avs,
+        PerformanceRewardsSubmission[] calldata performanceRewardsSubmissions
+    ) external onlyWhenNotPaused(PAUSED_AVS_PERFORMANCE_REWARDS_SUBMISSION) nonReentrant {
+        require(msg.sender == avs, "RewardsCoordinator.createAVSPerformanceRewardsSubmission: caller is not the AVS");
+        // TODO: Should we check if this is a registered AVS?
```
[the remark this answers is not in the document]
0xrajath: Sounds good. Removed `TODO` in [92201fb](https://github.com/Layr-Labs/eigenlayer-contracts/pull/837/commits/92201fb1751d961015aee116b415bef3921d9f39)

src/contracts/core/RewardsCoordinator.sol
0xrajath: Got it! Resolving this thread.

src/contracts/core/RewardsCoordinator.sol
0xrajath: Sounds good. I've noted it in the Exit criteria for technical docs.

src/contracts/interfaces/IRewardsCoordinator.sol
0xrajath: Keeping `GENESIS_REWARDS_TIMESTAMP` in the check incase `MAX_RETROACTIVE_LENGTH` becomes modifiable in the future and we get a long enough span into the past.

src/contracts/core/RewardsCoordinator.sol
0xrajath: Changed it back to casting using `uint32(*)` instead of Safecast: [4f65f54](https://github.com/Layr-Labs/eigenlayer-contracts/pull/837/commits/4f65f54414a387e5e00275c72a385f32c0cc023b)

src/contracts/core/RewardsCoordinator.sol
0xrajath: Do we foresee a third occurrence of this? If not refactoring might be overkill for the small benefit.

src/contracts/core/RewardsCoordinator.sol
0xrajath: https://github.com/Layr-Labs/eigenlayer-contracts/pull/837#discussion_r1819675600

src/contracts/core/RewardsCoordinator.sol
0xrajath: Also PI has the specific requirement of having a minimum 10% commission.

src/contracts/core/RewardsCoordinator.sol
0xrajath: Thought about it a bit more and decided to refactor and dedupe it: [9f18c44](https://github.com/Layr-Labs/eigenlayer-contracts/pull/837/commits/9f18c445e548736086ffccd8df7ac57e25433ce5)

src/contracts/core/RewardsCoordinator.sol
0xrajath: Thought about it a bit more and decided to refactor and dedupe it: [9f18c44](https://github.com/Layr-Labs/eigenlayer-contracts/pull/837/commits/9f18c445e548736086ffccd8df7ac57e25433ce5)

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -430,30 +564,104 @@ contract RewardsCoordinator is
...
+     * @param activatedAt The timestamp when the split is activated.
+     */
+    function _setOperatorSplit(OperatorSplit storage operatorSplit, uint16 split, uint32 activatedAt) internal {
+        // If, the earlier 'new' split is activated, we update the 'old' split with the earlier 'new' split.
+        // Else, the earlier 'old' split remains the same. This is essentially resetting the activation delay window
+        // since the earlier split setting didn't complete.
+        if (block.timestamp >= operatorSplit.activatedAt) {
+            if (operatorSplit.activatedAt == 0) {
+                // If the operator split has not been initialized yet, set the old split to the default split.
+                operatorSplit.oldSplitBips = defaultOperatorSplitBips;
```
0xrajath, line 637: Added the following condition for the edge case of getting correct operator split before the first time the split is activated. Discovered it while writing test cases.

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -267,8 +290,36 @@ contract RewardsCoordinator is
...
+
+        emit OperatorAVSSplitBipsSet(msg.sender, operator, avs, activatedAt, oldSplit, split);
+    }
+
+    /// @inheritdoc IRewardsCoordinator
+    function setOperatorPISplit(address operator, uint16 split) external onlyWhenNotPaused(PAUSED_OPERATOR_PI_SPLIT) {
+        require(msg.sender == operator, "RewardsCoordinator.setOperatorPISplit: caller is not the operator");
+        require(split <= ONE_HUNDRED_IN_BIPS, "RewardsCoordinator.setOperatorPISplit: split must be <= 10000 bips");
+
+        uint32 activatedAt = uint32(block.timestamp) + activationDelay;
```
gpsanant, line 318: move `activatedAt` logic into `_setOperatorSplit`?

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -267,8 +290,36 @@ contract RewardsCoordinator is
...
+        uint16 split
+    ) external onlyWhenNotPaused(PAUSED_OPERATOR_AVS_SPLIT) {
+        require(msg.sender == operator, "RewardsCoordinator.setOperatorAVSSplit: caller is not the operator");
+        require(split <= ONE_HUNDRED_IN_BIPS, "RewardsCoordinator.setOperatorAVSSplit: split must be <= 10000 bips");
+
+        uint32 activatedAt = uint32(block.timestamp) + activationDelay;
+        uint16 oldSplit = _getOperatorSplit(operatorAVSSplitBips[operator][avs]);
+        _setOperatorSplit(operatorAVSSplitBips[operator][avs], split, activatedAt);
+
+        emit OperatorAVSSplitBipsSet(msg.sender, operator, avs, activatedAt, oldSplit, split);
```
gpsanant, line 310: just checking with @bowenli86 that this flat event is preferred to emitting the structs?

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -340,6 +391,86 @@ contract RewardsCoordinator is
...
+            "RewardsCoordinator._validateOperatorDirectedRewardsSubmission: startTimestamp too far in the past"
+        );
+        require(
+            operatorDirectedRewardsSubmission.startTimestamp + operatorDirectedRewardsSubmission.duration <
+                block.timestamp,
+            "RewardsCoordinator._validateOperatorDirectedRewardsSubmission: operator-directed rewards submission is not retroactive"
+        );
+
+        // Require operatorDirectedRewardsSubmission is for whitelisted strategy or beaconChainETHStrategy
+        address currAddress = address(0);
```
gpsanant, line 457: can we abstract any of this logic internally to be used in the other rewards functions

src/contracts/core/RewardsCoordinator.sol
bowenli86: lgtm

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -430,30 +561,104 @@ contract RewardsCoordinator is
...
     function _setActivationDelay(uint32 _activationDelay) internal {
         emit ActivationDelaySet(activationDelay, _activationDelay);
         activationDelay = _activationDelay;
     }
 
-    function _setGlobalOperatorCommission(uint16 _globalCommissionBips) internal {
-        emit GlobalCommissionBipsSet(globalOperatorCommissionBips, _globalCommissionBips);
-        globalOperatorCommissionBips = _globalCommissionBips;
+    function _setDefaultOperatorSplit(uint16 split) internal {
+        emit DefaultOperatorSplitBipsSet(defaultOperatorSplitBips, split);
```
ypatil12, line 612: Style Nit: We emit after setting throughout repo

src/contracts/interfaces/IRewardsCoordinator.sol
```diff
@@ -327,11 +455,32 @@ interface IRewardsCoordinator {
...
+     * @dev The split will be activated after the activation delay
+     */
+    function setOperatorAVSSplit(address operator, address avs, uint16 split) external;
+
+    /**
+     * @notice Sets the split for a specific operator for Programmatic Incentives.
+     * @param operator The operator on behalf of which the split is being set.
+     * @param split The split for the operator for Programmatic Incentives in bips.
+     * @dev Only callable by the operator
+     * @dev Split has to be between 1000 and 10000 bips (inclusive)
```
ypatil12, line 480: Stale comment - no floor

src/contracts/core/RewardsCoordinator.sol
0xrajath: Not opinionated here. But this very same contract does `emit Event` and then set for all setters. So you let me know what you want.

src/contracts/core/RewardsCoordinator.sol
0xrajath: Oh you mean the validations. Yup can do that.

src/contracts/core/RewardsCoordinator.sol
0xrajath: I did think about it when writing this. But it's being emitted in the event below as well. Then that would need a separate `uint32(block.timestamp) + activationDelay` calculation. 

I think this is cleaner.

src/contracts/core/RewardsCoordinator.sol
```diff
@@ -118,11 +126,9 @@ contract RewardsCoordinator is
      */
 
     /// @inheritdoc IRewardsCoordinator
-    function createAVSRewardsSubmission(RewardsSubmission[] calldata rewardsSubmissions)
-        external
-        onlyWhenNotPaused(PAUSED_AVS_REWARDS_SUBMISSION)
-        nonReentrant
-    {
+    function createAVSRewardsSubmission(
+        RewardsSubmission[] calldata rewardsSubmissions
```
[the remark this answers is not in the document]
0xrajath: I would rather do that in the UAM release and not touch these functions as part of Rewards v2 (to keep prevent scope creep). The Operator-directed rewards submission already does take `avs` as a param.

src/contracts/interfaces/IRewardsCoordinator.sol
0xrajath: Done in [1d6abb0](https://github.com/Layr-Labs/eigenlayer-contracts/pull/837/commits/1d6abb05b9beb3637f32674f8b00a6763c5bbed9)

src/contracts/interfaces/IRewardsCoordinator.sol
```diff
@@ -150,13 +190,67 @@ interface IRewardsCoordinator {
...
+     * @param caller The address calling `createOperatorDirectedAVSRewardsSubmission`.
+     * @param avs The avs on behalf of which the operator-directed rewards are being submitted.
+     * @param submissionNonce Current nonce of the avs. Used to generate a unique submission hash.
+     * @param operatorDirectedRewardsSubmissionHash Keccak256 hash of (`avs`, `submissionNonce` and `operatorDirectedRewardsSubmission`).
+     * @param operatorDirectedRewardsSubmission The Operator-Directed Rewards Submission. Contains the token, start timestamp, duration, operator rewards, description and, strategy and multipliers.
+     */
+    event OperatorDirectedAVSRewardsSubmissionCreated(
+        address indexed caller,
+        address indexed avs,
+        uint256 submissionNonce,
```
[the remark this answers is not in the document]
0xrajath: Done in [1d6abb0](https://github.com/Layr-Labs/eigenlayer-contracts/pull/837/commits/1d6abb05b9beb3637f32674f8b00a6763c5bbed9)

src/contracts/core/RewardsCoordinator.sol
0xrajath: Done in [caf848c](https://github.com/Layr-Labs/eigenlayer-contracts/pull/837/commits/caf848cfa3a22b38d65085f6c4e48bdf23acce8e)

src/contracts/core/RewardsCoordinator.sol
0xrajath: We spoke offline about this. The reason these setters emit the event first is because they have both the oldvalue and newvalue. So it's cheaper to emit the event first and not have to store the old value in memory.

src/contracts/core/RewardsCoordinator.sol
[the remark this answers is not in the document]
ypatil12: SGTM

ypatil12 (approved):

PR #841, opened 2024-10-17
feat: restricted strategy list

0xClandestine:


src/contracts/core/AllocationManager.sol
```diff
@@ -53,13 +53,19 @@ contract AllocationManager is
...
         SlashingParams calldata params
     ) external onlyWhenNotPaused(PAUSED_OPERATOR_SLASHING) {
+        // Assert that the amount to slash is within bounds of 1 and WAD (1e18).
         require(0 < params.wadToSlash && params.wadToSlash <= WAD, InvalidWadToSlash());
 
-        // Check that the operator is registered and slashable
         OperatorSet memory operatorSet = OperatorSet({avs: msg.sender, operatorSetId: params.operatorSetId});
-        bytes32 operatorSetKey = _encodeOperatorSet(operatorSet);
+
+        // Assert that the provided strategies are all whitelisted for the given operator set.
```
0xClandestine, line 61: Not a fan of repeat external calls to avsd, hoping we can merge these contracts.

src/contracts/core/AVSDirectory.sol
```diff
@@ -413,6 +413,24 @@ contract AVSDirectory is
...
+        return _operatorSetStrategies[_encodeOperatorSet(operatorSet)].contains(address(strategy));
+    }
+
+    /// @inheritdoc IAVSDirectory
+    function isOperatorSetStrategyBatch(
+        OperatorSet calldata operatorSet,
+        IStrategy[] calldata strategies
+    ) external view returns (bool) {
+        for (uint256 i; i < strategies.length; ++i) {
+            if (!_operatorSetStrategies[_encodeOperatorSet(operatorSet)].contains(address(strategies[i]))) {
```
ypatil12, line 427: Nit: can we save gas by not encoding every time?

src/contracts/core/AllocationManager.sol
ypatil12: It's possible we can merge the _new_ AVSD functionality (ie. opset functionality) into the ALM. Unsure about merging ALM into AVSD

src/contracts/core/AllocationManager.sol
```diff
@@ -53,13 +53,19 @@ contract AllocationManager is
...
+        require(avsDirectory.isOperatorSetStrategyBatch(operatorSet, params.strategies), InvalidStrategy());
+        // Asser that the operator is slashable for the given operator set.
         require(avsDirectory.isOperatorSlashable(params.operator, operatorSet), InvalidOperator());
 
+        // Check that the operator is registered and slashable
+        bytes32 operatorSetKey = _encodeOperatorSet(operatorSet);
```
ypatil12, line 67: Nit: comment isn't doing this check? I think you mean to put above `require(avsDirectory.isOperatorSlashable(params.operator, operatorSet), InvalidOperator());`

src/contracts/core/AllocationManager.sol
```diff
@@ -164,6 +170,11 @@ contract AllocationManager is
                     // Add the operatorSet to the deallocation queue
                     deallocationQueue[msg.sender][allocation.strategy].pushBack(operatorSetKey);
                 } else if (info.pendingDiff > 0) {
+                    require(
+                        avsDirectory.isOperatorSetStrategy(allocation.operatorSets[j], allocation.strategy),
```
ypatil12, line 174: The net effect here is that you should ONLY be able to allocate to strategies that are whitelisted but can deallocate regardless. I think this is great. Can we add this assumption to the interface?

src/contracts/interfaces/IAVSDirectory.sol
```diff
@@ -323,6 +323,15 @@ interface IAVSDirectory is IAVSDirectoryEvents, IAVSDirectoryErrors, ISignatureU
         OperatorSet memory operatorSet
     ) external view returns (IStrategy[] memory strategies);
 
+    /// @notice Returns whether or not a strategy is allocatable/slashable for an operator set.
+    function isOperatorSetStrategy(OperatorSet calldata operatorSet, IStrategy strategy) external view returns (bool);
+
+    /// @notice Returns whether or not a list of strategies are all allocatable/slashable for an operator set.
```
ypatil12, line 329: Nit: should we natspec the params?

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -111,6 +111,10 @@ contract AllocationManagerUnitTests is EigenLayerUnitTestSetup, IAllocationManag
         // Set operatorSet to being valid
         avsDirectoryMock.setIsOperatorSetBatch(operatorSets, true);
 
+        // Set strategy to being valid for operator set
```
ypatil12, line 114: Can we add three tests: 

1. Revert when you try to slash for a strategy that does not exist. The operator should have allocated to the strategy, but then the AVS removes the strategy from their list of strategies we should revert the slash
2. Revert when the operator tries to allocate for a strategy not added to the operatorSet
3. Validate that we can deallocate, even if the strategy is not part of the operator set

src/contracts/core/AVSDirectory.sol
ypatil12, line 427: Can we add tests to the AVSD to validate this is calculated correctly? Just sanity checking the output from those fuzz tests

ypatil12 (commented):
Mainly want some more test coverage

src/contracts/core/AVSDirectory.sol
0xClandestine: noted, will fix if we don't move new avsd logic.

src/contracts/core/AllocationManager.sol
0xClandestine: You mean document it?

ypatil12: Ya in the interface comments

src/test/unit/AllocationManagerUnit.t.sol
ypatil12: I think we're still missing these?

ypatil12 (approved):

src/contracts/core/AllocationManager.sol
```diff
@@ -53,13 +53,18 @@ contract AllocationManager is
...
     ) external onlyWhenNotPaused(PAUSED_OPERATOR_SLASHING) {
+        // Assert that the amount to slash is within bounds of 1 and WAD (1e18).
         require(0 < params.wadToSlash && params.wadToSlash <= WAD, InvalidWadToSlash());
 
-        // Check that the operator is registered and slashable
         OperatorSet memory operatorSet = OperatorSet({avs: msg.sender, operatorSetId: params.operatorSetId});
-        bytes32 operatorSetKey = _encodeOperatorSet(operatorSet);
+
+        // Assert that the provided strategies are all whitelisted for the given operator set.
+        require(avsDirectory.isOperatorSetStrategyBatch(operatorSet, params.strategies), InvalidStrategy());
```
wadealexc, line 62: Can we not have this method accept an operator set as an argument and fetch strategies from the AVSD? That seems far preferable to passing in an array of strategies here.

src/contracts/core/AllocationManager.sol
```diff
@@ -53,13 +53,18 @@ contract AllocationManager is
...
+        // Asser that the operator is slashable for the given operator set.
```
wadealexc, line 63: `Asser` -> `Assert`

src/contracts/core/AllocationManager.sol
0xClandestine: FIxed in #855, but the fix introduces a via-ir compile error. Thinking we wait until #845 is merged which should reduce code complexity.

PR #843, opened 2024-10-18
POC: Key Rotation

ypatil12:
A POC of the key rotation functionality.

It is expected that each core contract will inherit the `PermissionControllerMixin` to interact with the `PermissionController`.

Assumptions
- PermissionController is _upgradeable_. We should have a deprecation pathway
- There is only 1 admin per account
- There is NOT a 1-to-1 relationship between admin and account. That is, the same address can be an admin for multiple accounts.


wadealexc:
Should `PermissionController` go in the `permissions/` folder? 

IMO this makes more sense than `core`, which is starting to just accumulate every new contract :grimacing:

PR #844, opened 2024-10-18
POC: Key Rotation Integration

ypatil12:
Examples of what the core protocol functions would look to satisfy:

1. Key rotation for operators
2. Forward-compatible delegated account interfaces 

See #843 for only the implementation

Breaking Interfaces:

1. `modifyOperatorDetails` - DM
2. `setMetadataURI` - DM
3. `createAVSRewardsSubmission` - RC. _Note: we likely can leave this the same as this reward type will be deprecated long term_
4. `cancelSalt` - AVSD
5. `setMetadataURI` - AVSD.  _Note: We may want to not break this interface for AVSs_

I specifically chose to leave out `setClaimer` in the RC and `undelegate` in the DM as they already have delegation enabled. An argument can be made to add `setClaimer`, but then this feature becomes usable by stakers too, which is not in scope.   

Core contracts without changes are: `EigenPodManager`, `StrategyManager`, `StrategyFactory`

### DIFF
--- a/src/contracts/core/AllocationManager.sol
+++ b/src/contracts/core/AllocationManager.sol
@@ -6,6 +6,7 @@ import "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
 import "@openzeppelin-upgrades/contracts/security/ReentrancyGuardUpgradeable.sol";
 
 import "../permissions/Pausable.sol";
+import "../mixins/PermissionControllerMixin.sol";
 import "../libraries/SlashingLib.sol";
 import "./AllocationManagerStorage.sol";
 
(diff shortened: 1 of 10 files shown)


src/contracts/core/AllocationManager.sol
```diff
@@ -207,18 +214,11 @@ contract AllocationManager is
...
         _setAllocationDelay(operator, delay);
     }
 
-    /// @inheritdoc IAllocationManager
-    function setAllocationDelay(
-        uint32 delay
-    ) external {
-        require(delegation.isOperator(msg.sender), OperatorNotRegistered());
-        _setAllocationDelay(msg.sender, delay);
-    }
```
MinisculeTarantula, line 220: it's unsafe to remove the check here, I think

ypatil12: yup, most definitely is - added

PR #845, opened 2024-10-21
feat: move new AVSD logic to ALM

0xClandestine:


src/contracts/core/AVSDirectory.sol
```diff
@@ -17,23 +17,20 @@ contract AVSDirectory is
...
-
     /**
      *
      *                         INITIALIZING FUNCTIONS
      *
      */
 
     /**
-     * @dev Initializes the immutable addresses of the strategy mananger, delegationManager,
+     * @dev Initializes the immutable addresses of the strategy mananger, delegationManage,
```
8sunyuan, line 27: Not: typo

8sunyuan: should just be mentioning DM and AM

PR #850, opened 2024-10-22
slashing: fully slashed strategies

8sunyuan:
This handles edge cases of 100% slashed scenarios where an operator is fully slashed for a strategy(maxMagnitude = 0)
and also when a native restaker has had enough accrued delta decreases in their beacon chain shares to the point where their beaconChainScalingFactor is 0. Either of these situations will result in a divide by 0 and has more transparent error revert checks when this occurs. 

Changes:
- Undelegation from a fully slashed operator is also fixed now as the queue withdrawal is ignored and deposit shares are fully decreased.
- Delegated deposits (`increaseDelegatedShares`) to an operator for a strategy that is fully slashed will revert
- Queue withdrawals for a strategy that is fully slashed will revert

Note: Complete queued withdrawals are unchanged but perhaps should revert if they were slashed during queue delay. It would avoid wasted gas for withdrawing 0 shares.

Tests:

- [x] As staker delegating to an operator who is slashed 100% for a strategy will revert
- [x] As existing staker delegated to an operator who is slashed 100% for a strategy, depositing more shares(`increaseDelegatedShares`) will revert
- [x] As existing staker with withdrawable shares delegated to an operator, after the operator was slashed 100%, the staker no longer has withdrawable shares
- [x] As staker delegated to operator who was slashed 100% for a strategy, the staker can still `increaseDelegatedShares` for other strategies with non-zero maxMagnitude for the operator.
- [x] Staker can undelegate even when slashed 100% (strategy that is 100% slashed is not queue withdrawn)
- [x] Staker cannot queue withdrawal for strategy slashed 100%

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -212,21 +212,36 @@ contract DelegationManager is
 
         for (uint256 i = 0; i < strategies.length; i++) {
             StakerScalingFactors storage ssf = stakerScalingFactor[staker][strategies[i]];
-            IStrategy[] memory singleStrategy = new IStrategy[](1);
-            uint256[] memory singleShares = new uint256[](1);
-            uint64[] memory singleMaxMagnitude = new uint64[](1);
-            singleStrategy[0] = strategies[i];
-            // TODO: this part is a bit gross, can we make it better?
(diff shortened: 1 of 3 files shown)


ypatil12 (approved):
This looks good for strategies, will need to figure out the EP edge cases in separate PR

PR #853, opened 2024-10-23
POC: Operator Example Contract

ypatil12:


src/contracts/periphery/Operator.sol
```diff
@@ -0,0 +1,30 @@
...
+
+import "./RolesAuthority.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+import "@openzeppelin/contracts/interfaces/IERC1271.sol";
+
+
+contract Operator is RolesAuthority {
+    constructor(address _owner) RolesAuthority(_owner) {}
+
+    function multicall(address[] memory targets, bytes[] memory data) external returns (bytes[] memory results) {
```
scotthconner, line 12: How does this impact ABI legibility and transaction construction? As a client, I now need to consume two ABIs to make a single call, and manually construct/encode calldata, no?

scotthconner: I also will need to, for a specific delegation, find the correct operator factory to multi-call through. How do I do this if my account can be a delegate of multiple operators? Or are we adding the 1:1 restriction back in?

And how does this approach different from using a module as part of a multi-sig?

src/contracts/periphery/Operator.sol
```diff
@@ -0,0 +1,30 @@
...
+        }
+    }
+
+    // TODO: implement 1271 support
+    /// @dev One idea here is to update the 712 typehash to include the function signature
+    /// of the function that is validating the signature. This way, users with a given role
+    /// can sign for the given (target, selector).
+    function isValidSignature(bytes32 hash, bytes memory signature) external view returns (bytes4 magicValue) {
+        return IERC1271.isValidSignature.selector;
+    }
```
scotthconner, line 29: Not entirely sure why this is needed.

src/contracts/periphery/OperatorFactory.sol
```diff
@@ -0,0 +1,26 @@
...
+
+    function getOperatorContract(
+        address owner
+    ) external view returns (Operator) {
+        Operator operator = Operator(Create2.computeAddress(
+            bytes32(uint256(uint160(owner))),
+            keccak256(type(Operator).creationCode)
+        ));
+    }
+}
```
scotthconner, line 26: This seems like it could be pure - it also seems like without storage, if the operator code changes or upgrades or is modified for any reason this won't work.

src/contracts/periphery/RolesAuthority.sol
```diff
@@ -0,0 +1,97 @@
...
+        bool enabled
+    ) public virtual onlyOwner {
+        if (enabled) {
+            getRolesWithCapability[target][functionSig] |= bytes32(1 << role);
+        } else {
+            getRolesWithCapability[target][functionSig] &= ~bytes32(1 << role);
+        }
+
+        emit RoleCapabilityUpdated(role, target, functionSig, enabled);
+    }
```
scotthconner, line 78: This goes much further in enshrining RBAC than uniform delegation. Having the additional role construct is now beginning to enter the problem space of account management. I'm not certain that is the goal here.

scotthconner (changes requested):
I have some blocking open questions around requirements, maintenance, integration, and feature support. I think we all need to align with Z and Matt about what the requirements for delegation are - as this solution removes some things (multiple admins), but also adds many others (full enshrined RBAC, plus two additional contracts and a full multi-call enabled virtual account.

We need to explore the impact and integration/assumptions on this with Operator Console, and need to consider the complexity and/or interoperability with the modularization of the AVS stack.

src/contracts/periphery/OperatorFactory.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.8.0;
+
+import "./Operator.sol";
+import "@openzeppelin/contracts/utils/Create2.sol";
+
+contract OperatorFactory {
+    function createOperator() external returns (Operator) {
+        return Operator(
+            Create2.deploy(
```
0xrajath, line 10: You don't need to use create2 library anymore. Solidity supports native create2 by just using `{salt: _salt}`. Example: https://solidity-by-example.org/new-contract/

src/contracts/periphery/OperatorFactory.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.8.0;
+
+import "./Operator.sol";
+import "@openzeppelin/contracts/utils/Create2.sol";
+
+contract OperatorFactory {
```
0xrajath, line 7: You might want to consider having every instance be a minimal proxy to reduce deploy costs. Tradeoff is the additional `delegatecall` for every call (May or may not be worth it). If deploys are decently cheap and a one time expense then we can keep as is.

src/contracts/periphery/Operator.sol
ypatil12: We'll need it to for legacy AVS registration

src/contracts/periphery/RolesAuthority.sol
ypatil12: Are roles the wrong abstraction here?

src/contracts/periphery/OperatorFactory.sol
ypatil12: True, figured we don't want the additional `delegateCall`

src/contracts/periphery/Operator.sol
ypatil12: > I also will need to, for a specific delegation, find the correct operator factory to multi-call through. How do I do this if my account can be a delegate of multiple operators?

Need to run an indexer for knowing all operators you can act on behalf of. Same problem exists in uniform delegation PR, unless we add additional storage. This isn't a requirement in PRD

src/contracts/periphery/RolesAuthority.sol
ypatil12: Synced with Z -> should be non roles based

ypatil12:
Close in favor of: https://github.com/Layr-Labs/eigenlayer-contracts/pull/865

PR #854, opened 2024-10-23
refactor: optimize pausable

0xClandestine:


ypatil12 (commented):
PR looks correct. How much are we saving storage-wise?

So the breaking change here is that we cannot reset the PauserRegistry without an upgrade. This _seems_ fine to me at first glance, will think more

ypatil12 (approved):

src/contracts/permissions/Pausable.sol
```diff
@@ -20,114 +19,105 @@ import "../interfaces/IPausable.sol";
...
-        require(address(pauserRegistry) == address(0) && address(_pauserRegistry) != address(0), InputAddressZero());
-        _paused = initPausedStatus;
-        emit Paused(msg.sender, initPausedStatus);
-        _setPauserRegistry(_pauserRegistry);
+    /// Construction
+
+    constructor(
+        IPauserRegistry _pauserRegistry
+    ) {
+        pauserRegistry = _pauserRegistry;
```
wadealexc, line 73: Let's add a check that this is nonzero

src/contracts/permissions/Pausable.sol
```diff
@@ -20,114 +19,105 @@ import "../interfaces/IPausable.sol";
...
+    /// @dev Returns a bitmap representing the paused status of the contract.
     uint256 private _paused;
 
-    uint256 internal constant UNPAUSE_ALL = 0;
-    uint256 internal constant PAUSE_ALL = type(uint256).max;
+    /// Modifiers
 
-    /// @notice
+    /// @dev Thrown if the caller is not a valid pauser according to the pauser registry.
     modifier onlyPauser() {
```
8sunyuan, line 43: Should we make `onlyPauser` and `onlyUnpauser` internal instead to save a little bit more gas?
`onlyPauser` is used twice in this contract but not something we ever use in inheriting contracts in core so the readability of having it as a modifier not as important imo.

8sunyuan (approved):

src/contracts/permissions/Pausable.sol
wadealexc: i checked, these don't save us bytecode

PR #858, opened 2024-10-28
feat: track staker withdrawals

0xClandestine:
### Changes

1) Some storage has been added.
```solidity
/// @notice Returns a list of  queued withdrawals for a given `staker`.
/// @dev Entrys are removed when the withdrawal is completed.
mapping(address staker => EnumerableSet.Bytes32Set withdrawalRoots) internal _stakerQueuedWithdrawalRoots;

/// @notice Returns the details of a queued withdrawal for a given `staker` and `withdrawalRoot`.
mapping(bytes32 withdrawalRoot => Withdrawal withdrawal) public queuedWithdrawals;
```
2) With the above storage additions `pendingWithdrawals` becomes redundant, but is needed to ensure legacy withdrawals are completable, thus it has been renamed to `_legacyPendingWithdrawals`.

**NOTE:** `_legacyPendingWithdrawals` should not longer be written to, only **read** or **deleted**.


src/contracts/core/DelegationManager.sol
```diff
@@ -298,11 +301,39 @@ contract DelegationManager is
...
     ) external onlyWhenNotPaused(PAUSED_EXIT_WITHDRAWAL_QUEUE) nonReentrant {
-        for (uint256 i = 0; i < withdrawals.length; ++i) {
+        uint256 n = withdrawals.length;
+        for (uint256 i; i < n; ++i) {
             _completeQueuedWithdrawal(withdrawals[i], tokens[i], receiveAsTokens[i]);
         }
     }
 
+    /// @inheritdoc IDelegationManager
+    function completeQueuedWithdrawal(
```
ypatil12, line 311: We only need the `numToComplete` variant

ypatil12 (commented):
Made a first pass

src/contracts/core/DelegationManager.sol
```diff
@@ -298,11 +301,39 @@ contract DelegationManager is
...
+        IERC20[] calldata tokens,
+        bool receiveAsTokens
+    ) external onlyWhenNotPaused(PAUSED_EXIT_WITHDRAWAL_QUEUE) nonReentrant {
+        EnumerableSet.Bytes32Set storage withdrawalRoots = _stakerQueuedWithdrawalRoots[msg.sender];
+        
+        // NOTE: This will revert if the staker does not have at least 1 pending queued withdrawal.
```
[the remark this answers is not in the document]
8sunyuan: Yeah would prefer an explicit withdrawalRoots.length() == 0 error check vs underflowing

src/contracts/core/DelegationManager.sol
```diff
@@ -298,11 +301,39 @@ contract DelegationManager is
...
+
+    /// @inheritdoc IDelegationManager
+    function completeQueuedWithdrawals(
+        IERC20[][] calldata tokens,
+        bool[] calldata receiveAsTokens,
+        uint256 numToComplete
+    ) external onlyWhenNotPaused(PAUSED_EXIT_WITHDRAWAL_QUEUE) nonReentrant {
+        EnumerableSet.Bytes32Set storage withdrawalRoots = _stakerQueuedWithdrawalRoots[msg.sender];
+        for (uint256 i; i < numToComplete; ++i) {
+            _completeQueuedWithdrawal(queuedWithdrawals[withdrawalRoots.at(i)], tokens[i], receiveAsTokens[i]);
```
8sunyuan, line 333: I believe this will also revert calling `withdrawalRoots.at(i)` if numToComplete is greater than the length of enumerableSet. Can we loop for number of iterations = `min(withdrawalRoots.length(), numToComplete)` to avoid this error.

src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -334,10 +334,37 @@ interface IDelegationManager is ISignatureUtils, IDelegationManagerErrors, IDele
...
+     * @notice Used to complete the lastest queued withdrawal.
+     * @param tokens Array in which the i-th entry specifies the `token` input to the 'withdraw' function of the i-th Strategy in the `withdrawal.strategies` array.
+     * @param receiveAsTokens If true, the shares calculated to be withdrawn will be withdrawn from the specified strategies themselves
+     * and sent to the caller, through calls to `withdrawal.strategies[i].withdraw`. If false, then the shares in the specified strategies
+     * will simply be transferred to the caller directly.
+     * @dev beaconChainETHStrategy shares are non-transferrable, so if `receiveAsTokens = false` and `withdrawal.withdrawer != withdrawal.staker`, note that
+     * any beaconChainETHStrategy shares in the `withdrawal` will be _returned to the staker_, rather than transferred to the withdrawer, unlike shares in
+     * any other strategies, which will be transferred to the withdrawer.
+     */
+    function completeQueuedWithdrawal(
```
8sunyuan, line 346: So all legacy withdrawals need to be completed with the old interface if they aren't queued with enumerable sets and new mappings. We should make it more clear here in the interface and on the view functions like `getQueuedWithdrawals` below

8sunyuan: I'm wondering if we should have a combined interface so that stakers can complete withdrawals of both legacy and post-slashing withdrawals in one call. At the moment, legacy withdrawals have to be completed with the old interface.

src/contracts/core/DelegationManagerStorage.sol
8sunyuan, line 1: nit: storage gap

src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -114,6 +117,13 @@ abstract contract DelegationManagerStorage is IDelegationManager {
     /// @dev We do not need the `beaconChainScalingFactor` for non-beaconchain strategies, but it's nicer syntactically to keep it.
     mapping(address staker => mapping(IStrategy strategy => StakerScalingFactors)) public stakerScalingFactor;
 
+    /// @notice Returns a list of  queued withdrawals for a given `staker`.
```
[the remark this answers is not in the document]
0xClandestine: Good thinking..

src/contracts/interfaces/IDelegationManager.sol
ypatil12: They can use the old interface to complete both types of withdrawals

src/contracts/core/DelegationManager.sol
```diff
@@ -542,8 +573,12 @@ contract DelegationManager is
             }
         }
 
-        // Remove `withdrawalRoot` from pending roots
+        _stakerQueuedWithdrawalRoots[withdrawal.staker].remove(withdrawalRoot);
+
+        delete queuedWithdrawals[withdrawalRoot];
```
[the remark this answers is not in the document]
0xClandestine: I think we would need additional storage to denote withdrawals have been completed.

Also `delete` provides a nice gas refund, which is nice given how expensive it is to store.

ypatil12 (commented):
Additional Comment

src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -334,10 +334,22 @@ interface IDelegationManager is ISignatureUtils, IDelegationManagerErrors, IDele
     ) external returns (bytes32[] memory);
 
     /**
-     * @notice Used to complete the specified `withdrawal`. The caller must match `withdrawal.withdrawer`
-     * Withdrawals remain slashable during the withdrawal delay period and the actual withdrawn shares are calculated
-     * based off the scaledShares.
-     * @param withdrawal The Withdrawal to complete.
+     * @notice Used to complete the all queued withdrawals.
```
ypatil12, line 337: Nit: can we add a comment that this only works for post-slashing withdrawals

src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -551,6 +563,11 @@ interface IDelegationManager is ISignatureUtils, IDelegationManagerErrors, IDele
     /// @notice Return the M2 minimum withdrawal delay in blocks for backwards compatability
     function minWithdrawalDelayBlocks() external view returns (uint256);
 
+    /// @notice Returns a list of pending queued withdrawals for a `staker`, and the `shares` to be withdrawn.
```
ypatil12, line 566: Nit: can we add a comment that this is relevant only for post-slashing queued withdrawals?

src/contracts/libraries/SlashingLib.sol
```diff
@@ -150,10 +151,11 @@ library SlashingLib {
 
         // Step 3: Calculate newStakerDepositScalingFactor
         /// forgefmt: disable-next-item
-        uint256 newStakerDepositScalingFactor = newShares
+        uint184 newStakerDepositScalingFactor = newShares
```
ypatil12, line 154: Why change this to uint184?

ypatil12 (commented):
Minor comments

src/contracts/libraries/SlashingLib.sol
0xClandestine: Changed type to pack the struct into a single slot, we could use a smaller type if it makes sense.

0xClandestine: may be a better idea to divide the extra bits to deposit scaler and beacon chain scaler

PR #860, opened 2024-10-28
feat: move operator set functionality to alm

wadealexc:
* some storage renaming and refactoring
* changes interface to accept allocation modifications in terms of `(operatorSet, []strategies)` rather than the other way around
* changes delays to use blocks rather than timestamps
* changes `modifyAllocations` to allow instant deallocations if an operator is not slashable
* AVSs can configure an AVSRegistrar that will be called when operators register for / deregister from operator sets. Defaults to the AVS's own address if not set.

### DIFF
--- a/src/contracts/core/AllocationManager.sol
+++ b/src/contracts/core/AllocationManager.sol
@@ -7,6 +7,7 @@ import "@openzeppelin-upgrades/contracts/security/ReentrancyGuardUpgradeable.sol
 
 import "../permissions/Pausable.sol";
 import "../libraries/SlashingLib.sol";
+import "../libraries/OperatorSetLib.sol";
 import "./AllocationManagerStorage.sol";
 
 contract AllocationManager is
@@ -16,8 +17,11 @@ contract AllocationManager is
     AllocationManagerStorage,
     ReentrancyGuardUpgradeable
 {
-    using Snapshots for Snapshots.DefaultWadHistory;
     using DoubleEndedQueue for DoubleEndedQueue.Bytes32Deque;
+    using EnumerableSet for *;
+
+    using Snapshots for Snapshots.DefaultWadHistory;
+    using OperatorSetLib for OperatorSet;
     using SlashingLib for uint256;
 
     /**
@@ -27,18 +31,16 @@ contract AllocationManager is
      */
 
     /**
-     * @dev Initializes the immutable addresses of the strategy mananger, delegationManage,
-     * and eigenpodManager contracts
+     * @dev Initializes the DelegationManager address, the deallocation delay, and the allocation configuration delay.
      */
     constructor(
         IDelegationManager _delegation,
-        IAVSDirectory _avsDirectory,
         IPauserRegistry _pauserRegistry,
         uint32 _DEALLOCATION_DELAY,
         uint32 _ALLOCATION_CONFIGURATION_DELAY
-    )
-        AllocationManagerStorage(_delegation, _avsDirectory, _DEALLOCATION_DELAY, _ALLOCATION_CONFIGURATION_DELAY)
-        Pausable(_pauserRegistry)
+    ) 
+        AllocationManagerStorage(_delegation, _DEALLOCATION_DELAY, _ALLOCATION_CONFIGURATION_DELAY)
+        Pausable(_pauserRegistry) 
     {
         _disableInitializers();
     }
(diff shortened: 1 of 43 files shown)


src/contracts/core/AllocationManager.sol
```diff
@@ -55,138 +63,131 @@ contract AllocationManager is
...
+        // For each strategy in the operator set, slash any existing allocation
+        for (uint256 i = 0; i < length; i++) {
+            // 1. Get the operator's allocation info for the strategy and operator set
+            IStrategy strategy = IStrategy(_operatorSetStrategies[operatorSet.key()].at(i));
+            (StrategyInfo memory info, Allocation memory allocation) =
+                _getUpdatedAllocation(msg.sender, operatorSet.key(), strategy);
+            strategiesSlashed[i] = strategy;
+
+            // 2. Skip if the operator is not allocated to the operator set from this strategy
+            if (allocation.currentMagnitude == 0) {
```
ypatil12, line 84: IMO should still have a hard requirement that `currentMagnitude > 0`. Offchain systems having to process a slashing where currentMag == 0 doesn't make sense

src/contracts/core/AllocationManager.sol
```diff
@@ -198,13 +199,77 @@ contract AllocationManager is
...
+            // forgefmt: disable-next-item
+            registrationStatus[params.operator][operatorSet.key()] = RegistrationStatus({
+                registered: false,
+                slashableUntil: uint32(block.number) + DEALLOCATION_DELAY
+            });
+
+            // TODO - should this instead do modifyAllocations + removing allocations?
+            // issue right now is if I call deregister, i have to actually go remove allocations.
+            // if i re-register, i immediately become slashable again
+        }
```
ypatil12, line 266: Heavily in favor of removing allocations once you deregister. 

If we go down the route of not removing allocations, we still need to enable proper introspection with something like: 

```solidity
struct Registration Status{
    bool registered;
    bool allocated;
    uint32 slashableUntil;
}
```

src/contracts/core/AllocationManager.sol
wadealexc: we can't have this requirement, because an operator may not have a strategy allocated! this would create unslashable operators (unless im misunderstanding)

src/contracts/core/AllocationManager.sol
wadealexc: ok that's what i was feeling too. lemme take a crack at it today...

src/contracts/core/AllocationManager.sol
wadealexc: one thing we could _maybe_ do is have an enumerable set of strategies that have existing allocations:

```
mapping(operator => operatorSetKey => EnumerableSet.AddressSet) allocationList;
```

... and we could iterate over this list when slashing.

src/contracts/core/AllocationManager.sol
```diff
@@ -55,138 +63,131 @@ contract AllocationManager is
...
-                    require(info.encumberedMagnitude <= maxMagnitude, InsufficientAllocatableMagnitude());
+        require(delegation.isOperator(msg.sender), InvalidOperator());
+
+        for (uint256 i = 0; i < params.length; i++) {
+            require(params[i].strategies.length == params[i].newMagnitudes.length, InputArrayLengthMismatch());
+
+            // Check that the operator set exists and whether the operator is slashable for the set
+            OperatorSet calldata operatorSet = params[i].operatorSet;
+            require(_operatorSets[operatorSet.avs].contains(operatorSet.id), InvalidOperatorSet());
+            bool isSlashable = isOperatorSlashable(msg.sender, operatorSet);
```
8sunyuan, line 145: Should we also have a IStrategy param passed into `isOperatorSlashable`?
Right now we allow deallocations to immediately decrement encumberedMagnitude if it is not slashable. But what if the AVS removes the specific strategy. We should let the operator deallocate immediately in this scenario too.

src/contracts/core/AllocationManager.sol
wadealexc: I guess the main question here is how we handle pending modifications.

Obviously the main case we want to handle is when we have existing magnitude, and no pending modification. Easy - create a pending deallocation for the current magnitude.

What about these cases, though?

1. `(curMagnitude: 0, pendingDiff > 0)`
2. `(curMagnitude > 0, pendingDiff > 0)`
3. `(curMagnitude > 0, pendingDiff < 0)`

src/contracts/core/AllocationManager.sol
wadealexc: Actually i dont think my solution is possible - we might end up removing allocatedStrategies in the middle of iterating over it, and the `.remove` method swaps indices around, so... dangerous!

src/contracts/core/AllocationManager.sol
wadealexc: Okay - should have addressed this here: [62a802a](https://github.com/Layr-Labs/eigenlayer-contracts/pull/860/commits/62a802a305f313d78edd7acce27fecadae611520)

src/contracts/core/AllocationManager.sol
wadealexc: ok i have a solution here -- we iterate in reverse order.

and in fact i think my solution is a requirement, because we need to be able to fetch allocations that aren't necessarily recognized by an operator set. currently we lack the ability to do this, so we need an `allocationList` type tracker for this.

ypatil12: > we can't have this requirement, because an operator may not have a strategy allocated

Ah true, I forgot that we have are not passing in a list of strategies. This is fine, just need offchain systems to properly handle the 0 slash case.

ypatil12: Interesting. Yes, we definitely do want a tracker for this because given the below discussion it sounds like we aren't going to go and clear all _pending_ modifications. 

Another edge case is when a strategy is removed from the operatorSet, operator is still allocated, strategy is re-added. The operator is instantly allocatable

src/contracts/core/AllocationManager.sol
gpsanant: I am opposed to removing allocations upon deregistration, primarily due to the case of ejection. Removing allocations upon ejection means it takes at least `DEALLOCATION_DELAY+allocationDelay(op)` in order to rejoin the operatorSet.

EigenDA recently made its ejection cooldown 3 days instead of 7 days so that ejected operators can join sooner.

wadealexc: [cacf8f9](https://github.com/Layr-Labs/eigenlayer-contracts/pull/860/commits/cacf8f97c1bb307bc45b1d8e103a7dadd8bcef37)

removed deallocations on deregistration here

src/contracts/core/AllocationManager.sol
wadealexc: [faf9a7b](https://github.com/Layr-Labs/eigenlayer-contracts/pull/860/commits/faf9a7b23e157652f16db8fee8c2f826b1aa1377)

added this enumeration here

src/contracts/core/AllocationManager.sol
wadealexc: > specifically we need to ensure that if the strategy is not a part of the set, your only action is to reclaim an existing allocation

note, i did not implement this portion -- `modifyAllocations` still allows a user to allocate to an operator set even if the set does not support the strategy. Was simpler to implement this way and should be instantly deallocatable after the allocation delay anyway, so nbd.

ypatil12 (commented):
We need to add metadataURI

0xClandestine (commented):
nits

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -93,365 +115,296 @@ contract AllocationManagerUnitTests is EigenLayerUnitTestSetup, IAllocationManag
...
+        return operatorSet;
     }
 
-    /// -----------------------------------------------------------------------
-    /// Generate random slashing parameters
-    /// -----------------------------------------------------------------------
+    function _newOperatorSet_SingleMockStrategy(
+        address avs
+    ) internal returns (OperatorSet memory) {
+        uint32 nextId = _opSetCount[avs];
```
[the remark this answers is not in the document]
wadealexc: we could but i wasnt explicitly looking for randomness here

src/test/tree/AllocationManagerUnit.tree
```diff
@@ -37,7 +37,7 @@
 │       ├── given that the latest effect timestamp has not been reached
```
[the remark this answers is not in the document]
wadealexc: [89c60f1](https://github.com/Layr-Labs/eigenlayer-contracts/pull/860/commits/89c60f12002bf986082fe88c32461562afde1add)

src/test/DevnetLifecycle.t.sol
```diff
@@ -137,79 +139,61 @@ contract Devnet_Lifecycle_Test is Test {
...
-        IAllocationManagerTypes.MagnitudeInfo[] memory infos = allocationManager.getAllocationInfo(operator, wethStrategy, _getOperatorSetsArray());
-        assertEq(infos[0].currentMagnitude, 0);
-        assertEq(infos[0].pendingDiff, int128(uint128(magnitudeToSet)));
-        assertEq(infos[0].effectTimestamp, block.number + 1);
+        IAllocationManagerTypes.Allocation memory info = allocationManager.getAllocation(operator, operatorSet, wethStrategy);
+        assertEq(info.currentMagnitude, 0);
+        assertEq(info.pendingDiff, int128(uint128(magnitudeToSet)));
+        assertEq(info.effectBlock, block.number + 1);
 
         // Warp to effect timestamp
```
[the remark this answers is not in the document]
wadealexc: [89c60f1](https://github.com/Layr-Labs/eigenlayer-contracts/pull/860/commits/89c60f12002bf986082fe88c32461562afde1add)

src/contracts/core/AllocationManager.sol
```diff
@@ -219,6 +277,65 @@ contract AllocationManager is
...
+
+    /// @inheritdoc IAllocationManager
+    function createOperatorSets(
+        CreateSetParams[] calldata params
+    ) external {
+        for (uint256 i = 0; i < params.length; i++) {
+            OperatorSet memory operatorSet = OperatorSet(msg.sender, params[i].operatorSetId);
+
+            // Create the operator set, ensuring it does not already exist
+            require(_operatorSets[msg.sender].add(operatorSet.id) == true, InvalidOperatorSet());
```
8sunyuan, line 303: can remove the == true

src/contracts/core/AllocationManager.sol
```diff
@@ -198,13 +201,68 @@ contract AllocationManager is
...
+    /// @inheritdoc IAllocationManager
+    function deregisterFromOperatorSets(
+        DeregisterParams calldata params
+    ) external onlyWhenNotPaused(PAUSED_OPERATOR_SET_REGISTRATION_AND_DEREGISTRATION) {
+        require(msg.sender == params.operator || msg.sender == params.avs, InvalidCaller());
+
+        for (uint256 i = 0; i < params.operatorSetIds.length; i++) {
+            // Check the operator set exists and the operator is registered to it
+            OperatorSet memory operatorSet = OperatorSet(params.avs, params.operatorSetIds[i]);
+            require(_operatorSets[params.avs].contains(operatorSet.id), InvalidOperatorSet());
```
8sunyuan, line 245: Noting since operatorSetIds cant be deleted, this opSetId existing check is kind of implied if the operator was able to register in the first place. Fine to leave as is but I don't think it's required

src/contracts/core/AllocationManager.sol
```diff
@@ -55,138 +57,139 @@ contract AllocationManager is
...
+            require(params[i].strategies.length == params[i].newMagnitudes.length, InputArrayLengthMismatch());
+
+            // Check that the operator set exists and get the operator's registration status
+            // Operators do not need to be registered for an operator set in order to allocate
+            // slashable magnitude to the set. In fact, it is expected that operators will
+            // allocate magnitude before registering, as AVS's will likely only accept
+            // registrations from operators that are already slashable.
+            OperatorSet calldata operatorSet = params[i].operatorSet;
+            bool isRegistered = _isRegistered(msg.sender, operatorSet);
+            require(_operatorSets[operatorSet.avs].contains(operatorSet.id), InvalidOperatorSet());
```
8sunyuan, line 145: Similar to other comment, the `_isRegistered(msg.sender, operatorSet)` call being true already implies existence of the operatorSet. We could remove this require check

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -3,22 +3,35 @@ pragma solidity ^0.8.27;
 
 import "src/contracts/core/AllocationManager.sol";
 import "src/test/utils/EigenLayerUnitTestSetup.sol";
+import "src/test/mocks/MockAVSRegistrar.sol";
```
ypatil12, line 6: We don't have tests for some view functions:

`getMaxMagnitudesAtBlock`
`getStrategyAllocations`

src/contracts/core/AllocationManager.sol
```diff
@@ -7,6 +7,7 @@ import "@openzeppelin-upgrades/contracts/security/ReentrancyGuardUpgradeable.sol
 
 import "../permissions/Pausable.sol";
 import "../libraries/SlashingLib.sol";
+import "../libraries/OperatorSetLib.sol";
```
ypatil12, line 10: No `setMetadataURI` method?

ypatil12 (commented):
Minor comments

src/contracts/core/AllocationManager.sol
```diff
@@ -256,82 +375,133 @@ contract AllocationManager is
...
+        Allocation memory allocation
     ) internal {
-        _operatorMagnitudeInfo[operator][strategy][operatorSetKey] = MagnitudeInfo({
-            currentMagnitude: info.currentMagnitude,
-            pendingDiff: info.pendingDiff,
-            effectTimestamp: info.effectTimestamp
-        });
-
+        // Update encumbered magnitude
         encumberedMagnitude[operator][strategy] = info.encumberedMagnitude;
```
8sunyuan, line 476: Note: multiple `_updateAllocationInfo` calls redundantly overwriting encumberedMagnitude in clearDeallocationQueue. Something to look at if `modifyAllocations` or `clearDeallocationQueue` are too expensive to call

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -2175,53 +2254,53 @@ contract AllocationManagerUnitTests_SetAllocationDelay is AllocationManagerUnitT
     }
 
     function test_fuzz_setDelay_multipleTimesWithinConfigurationDelay(
-        uint32 firstDelay, uint32 secondDelay
-    ) public {
-        firstDelay = uint32(bound(firstDelay, 1, type(uint32).max));
-        secondDelay = uint32(bound(secondDelay, 1, type(uint32).max));
+        Randomness r
+    ) public rand(r) {
+        uint32 firstDelay = uint32(r.Uint256(1, type(uint32).max));
```
8sunyuan, line 2259: Can we change this to 0 since allocationDelay of 0 is enabled now. These bounds were missed previously in that PR

src/test/tree/AllocationManagerUnit.tree
8sunyuan, line 1: Think this tree file needs some updating.
Examples:
1. "given that the delay is set to 0, it should revert"
We allow 0 allocationDelays now, looks like this was missed from previous change.
2. clearModificationQueue -> clearDeallocationQueue, some of the code paths are outdated too.
3. Add operatorSet related logic carried over from AVSD

src/contracts/core/AllocationManager.sol
ypatil12: Nvm its there

src/contracts/core/AllocationManager.sol
wadealexc: I removed a bunch of extraneous checks but i like this one for clarity. Good point though!

src/contracts/core/AllocationManager.sol
wadealexc: I like this on as-is because `isRegistered` isn't a requirement to allocate magnitude. However, the operator set existing should be a requirement!

src/test/unit/AllocationManagerUnit.t.sol
wadealexc: @0xClandestine can you add these to your testing / TODO PR and get them in a followup?

I'd like to get this merged as it's a big change and we've been sitting on it for a long time.

src/contracts/core/AllocationManager.sol
wadealexc: Yes, this is true. It was pretty hard to avoid the duplicate writes while keeping the code clean, unfortunately.

I felt it was probably okay given that SSTORE is significantly cheaper in these cases (only 100 gas for followup writes).

But yes, we can look at changing this if needed.

src/test/tree/AllocationManagerUnit.tree
wadealexc: cc @0xClandestine can i give this to you?

src/test/unit/AllocationManagerUnit.t.sol
wadealexc: [61772f2](https://github.com/Layr-Labs/eigenlayer-contracts/pull/860/commits/61772f2f930aa2a816f4ee3a1eb6416e505f8596)

src/contracts/core/AllocationManager.sol
wadealexc: [61772f2](https://github.com/Layr-Labs/eigenlayer-contracts/pull/860/commits/61772f2f930aa2a816f4ee3a1eb6416e505f8596)

ypatil12 (approved):

PR #861, opened 2024-10-29
slashing: revert timestamps delegation

8sunyuan:
Decided on keeping `struct Withdrawal` as using blocknumbers as opposed to converting to timestamps.

<s> However because the delay is going to change in the upcoming release to align with DEALLOCATION_DELAY in the AllocationManager, the withdrawal startBlock will need to be checked against the blocknumber of the Slashing upgrade. See `getCompletableTimestamp()` for details.
</s>
All legacy withdrawals queued prior to the upgrade will actually have their withdrawal delay extended by however long the new withdrawal delay period (17.5 day delay so extended by 10.5 days). Thus removing any branching logic for handling withdrawals and their timestamps. 

Additionally, `minWithdrawalDelayBlocks` will still be deprecated as having this lie in storage with a setter function is not practical. Any changes to this delay will require deeper consideration and likely an upgrade to the AllocationManager. Using the constant `MIN_WITHDRAWAL_DELAY_BLOCKS` for post-slashing queue withdrawals seems most reasonable here.

### DIFF
--- a/script/deploy/devnet/deploy_from_scratch.s.sol
+++ b/script/deploy/devnet/deploy_from_scratch.s.sol
@@ -365,10 +365,11 @@ contract DeployFromScratch is Script, Test {
         _verifyInitializationParams();
 
         // Check DM and AM have same withdrawa/deallocation delay
-        require(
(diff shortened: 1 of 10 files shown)


ypatil12 (approved):
LGTM

PR #864, opened 2024-10-30
feat: track staker withdrawals

0xClandestine:


src/contracts/libraries/SlashingLib.sol
```diff
@@ -22,12 +23,10 @@ uint64 constant WAD = 1e18;
  * Note that `withdrawal.scaledShares` is scaled for the beaconChainETHStrategy to divide by the beaconChainScalingFactor upon queueing
  * and multiply by the beaconChainScalingFactor upon withdrawal
  */
-
 struct StakerScalingFactors {
-    uint256 depositScalingFactor;
-    // we need to know if the beaconChainScalingFactor is set because it can be set to 0 through 100% slashing
-    bool isBeaconChainScalingFactorSet;
+    uint184 depositScalingFactor;
```
ypatil12, line 27: @8sunyuan are we okay with this bit packing? Seems like math def checks out

8sunyuan (approved):

PR #870, opened 2024-11-04
feat: user access management

ypatil12:
# Overview

User Access Management (UAM) adds a core contract, the `PermissionController` as a dependency for the `DelegationManager` , `AllocationManager`, and `RewardsCoordinator` . We enable this functionality for AVSs and Operators but not for stakers. There is no support for the `AVSDirectory` as we plan on deprecating its functionality eventually. 

Operators & AVSs can have multiple admins for their account. This enables key rotation for operators. 

Operators cannot appoint an address to _register_ on its behalf.  In other words, the address they register with is their account and they can then transfer admin of that account in a subsequent tx. Similarly, the account of an AVS is the address they use to initialize state in the `AllocationManager`. 

# Added functionality

We add a `registerAsAVS` function as the entrypoint to become an AVS in the `AllocationManager`

# Edge Cases

- An admin can only be removed if there is at least 1 admin remaining after removal 

# TODOs

- [x] Tests with other contracts

# Delegation Enabled Functions

We do NOT add UAM support for `processClaim` (ie. claim rewards) as that function is used by stakers, operators, and (in RewardsV2) AVSs. There is already delegation for this function via the `setClaimerFor`. The downside here is that there can only be 1 claimer for an operator or AVS, as opposed to multiple for other functions. This seems acceptable for now, but open to hearing any concerns. 

## Operators

- `registerForOperatorSets`
- `deregisterFromOperatorSets`
- `modifyAllocations`
- `setAllocationDelay`
- `setClaimerFor`
- `modifyOperatorDetails`
- `undelegate`
- `updateOperatorMetadataURI`

## AVSs

- `deregisterFromOperatorSets`
- `slashOperator`
- `setAVSRegistrar`
- `updateAVSMetadataURI`
- `createOperatorSets`
- `addStrategiesToOperatorSet`
- `removeStrategiesFromOperatorSet`
- `createAVSRewardsSubmission`
- `setClaimerFor`


src/contracts/mixins/PermissionControllerMixin.sol
```diff
@@ -0,0 +1,38 @@
...
+        _;
+    }
+
+    /**
+     * @notice Checks if the caller is allowed to call a function on behalf of an account.
+     * @param account the account to check
+     * @dev `msg.sender` is the caller to check that can call the function on behalf of `account`.
+     * @dev Returns a bool, instead of reverting
+     */
+    function _checkCanCall(
```
stevennevins, line 33: pass in the delegate vs inferring

src/contracts/mixins/PermissionControllerMixin.sol
```diff
@@ -0,0 +1,38 @@
...
+        IPermissionController _permissionController
+    ) {
+        permissionController = _permissionController;
+    }
+
+    /// @notice Checks if the caller (msg.sender) can call on behalf of an account.
+    modifier checkCanCall(
+        address account
+    ) {
+        require(permissionController.canCall(account, msg.sender, address(this), msg.sig), InvalidPermissions());
```
stevennevins, line 23: use internal function and pass msg.sender as delegate

0xClandestine (commented):
Just some renaming suggestions.

src/contracts/permissions/PermissionController.sol
```diff
@@ -0,0 +1,176 @@
...
+     */
+
+    /// @inheritdoc IPermissionController
+    function isAdmin(address account, address caller) public view returns (bool) {
+        if (_permissions[account].admins.length() == 0) {
+            // If the account does not have an admin, the caller must be the account
+            return account == caller;
+        } else {
+            // If the account has an admin, the caller must be an admin
+            return _permissions[account].admins.contains(caller);
```
8sunyuan, line 127: this is done in constant lookup time so probably can just remove the length == 0 check. Remove the extra storage check for users who have admins set

src/contracts/permissions/PermissionController.sol
```diff
@@ -0,0 +1,176 @@
...
+    function setAdmin(address account, address admin) external onlyAdmin(account) {
+        // Add the admin to the account's admins
+        // If the admin is already set, the set will fail
+        EnumerableSet.AddressSet storage admins = _permissions[account].admins;
+        require(admins.add(admin), AdminAlreadySet());
+
+        emit AdminSet(account, admin);
+    }
+
+    function removeAdmin(address account, address admin) external onlyAdmin(account) {
```
8sunyuan, line 44: So if you remove all admins, the original account becomes the admin again?
Is this desired? Or can we just prevent ever having 0 admins when you try to remove.

src/contracts/interfaces/IPermissionController.sol
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * @notice Set a delegate for a given account
+     * @param account to set delegate for
+     * @param delegate to set
+     * @param target to set delegate for
+     * @param selector to set delegate for
+     * @dev Only the admin of the account can set a delegate
+     */
+    function setDelegate(address account, address delegate, address target, bytes4 selector) external;
```
8sunyuan, line 56: Is this getting renamed? As discussed before we should not confuse this with staker<>operator delegation.
I suggested something like `Entrustee` or `Entrant`. Entrant seems like it describes the actor who can enter specific function selectors.

src/contracts/permissions/PermissionControllerStorage.sol
```diff
@@ -0,0 +1,30 @@
...
+
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+
+import "../interfaces/IPermissionController.sol";
+
+abstract contract PermissionControllerStorage is IPermissionController {
+    using EnumerableSet for EnumerableSet.Bytes32Set;
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    struct AccountPermissions {
```
8sunyuan, line 12: I quite like this storage design

src/contracts/interfaces/IPermissionController.sol
stevennevins: not a fan on `Entrant`

src/contracts/permissions/PermissionControllerStorage.sol
stevennevins: lets go the rest of the way with this pattern and used salted storage location. Then we don't have to worry about storage gaps

src/contracts/permissions/PermissionController.sol
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /**
+     *
+     *                         INTERNAL FUNCTIONS
+     *
+     */
+
+    /// @dev Encodes the target and selector into a single bytes32 value
+    function _encodeTargetSelector(address target, bytes4 selector) internal pure returns (bytes32) {
```
8sunyuan, line 99: Would be good to have reversability unit test. Where you get the same starting target,selector values.

_encodeTargetSelector(target, selector) => bytes32 targetSelector
_decodeTargetSelector(targetSelector) => (target, selector)

src/contracts/permissions/PermissionController.sol
```diff
@@ -0,0 +1,176 @@
...
+        // Remove the admin from the account's admins
+        // If the admin is not set, the remove will fail
+        require(admins.remove(admin), AdminNotSet());
+
+        emit AdminRemoved(account, admin);
+    }
+
+    /// @inheritdoc IPermissionController
+    function setDelegate(
+        address account,
```
stevennevins, line 56: can we update to:

delegator

delegatee

src/contracts/permissions/PermissionController.sol
```diff
@@ -0,0 +1,176 @@
...
+    function initialize() external initializer {}
+
+    /**
+     *
+     *                         EXTERNAL FUNCTIONS
+     *
+     */
+
+    /// @inheritdoc IPermissionController
+    function setAdmin(address account, address admin) external onlyAdmin(account) {
```
stevennevins, line 35: addAdmin

isAdmin(account)

src/contracts/permissions/PermissionController.sol
stevennevins: addDelegate

src/contracts/permissions/PermissionController.sol
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     *
+     *                         INTERNAL FUNCTIONS
+     *
+     */
+
+    /// @dev Encodes the target and selector into a single bytes32 value
+    function _encodeTargetSelector(address target, bytes4 selector) internal pure returns (bytes32) {
+        return bytes32(abi.encodePacked(target, uint96(bytes12((selector)))));
```
stevennevins, line 100: Pretty sure this could be simplified to
```
bytes32(abi.encodePacked(target, selector));
```
Decoding would be 

```
address target = address(uint160(uint256(targetSelector) >> 32));
bytes4 selector = bytes4(uint32(uint256(targetSelector));
```

or 
```
(address target, bytes4 selector) = abi.decode(bytes(targetSelector), (address, bytes4));
```

stevennevins: without casting in the decode
```
(address target, bytes4 selector) = abi.decode(targetSelector, (address, bytes4));
```

src/contracts/permissions/PermissionController.sol
```diff
@@ -0,0 +1,176 @@
...
+        AccountPermissions storage permissions = _permissions[account];
+
+        bytes32 targetSelector = _encodeTargetSelector(target, selector);
+        require(permissions.delegatePermissions[delegate].contains(targetSelector), DelegateNotSet());
+
+        // Remove the delegate from the account's permissions
+        permissions.delegatePermissions[delegate].remove(targetSelector);
+        permissions.permissionDelegates[targetSelector].remove(delegate);
+
+        emit DelegateRemoved(account, delegate, target, selector);
```
stevennevins, line 89: We should capture who added the permission since it might not be the `account`

stevennevins, line 89: We should capture who `msg.sender` removed the delegate

src/contracts/permissions/PermissionController.sol
```diff
@@ -0,0 +1,176 @@
...
+        AccountPermissions storage permissions = _permissions[account];
+
+        bytes32 targetSelector = _encodeTargetSelector(target, selector);
+        require(!permissions.delegatePermissions[delegate].contains(targetSelector), DelegateAlreadySet());
+
+        // Add the delegate to the account's permissions
+        permissions.delegatePermissions[delegate].add(targetSelector);
+        permissions.permissionDelegates[targetSelector].add(delegate);
+
+        emit DelegateSet(account, delegate, target, selector);
```
stevennevins, line 70: same here for who removed the delegate

src/contracts/permissions/PermissionController.sol
```diff
@@ -0,0 +1,176 @@
...
+            return admin;
+        } else {
+            return _permissions[account].admins.values();
+        }
+    }
+
+    /// @inheritdoc IPermissionController
+    function canCall(address account, address caller, address target, bytes4 selector) external view returns (bool) {
+        return isAdmin(account, caller)
+            || _permissions[account].delegatePermissions[caller].contains(_encodeTargetSelector(target, selector));
```
stevennevins, line 147: Can we make a view function

`isPermissioned(address account, address caller, address target, bytes4 selector);`

that does the logic after `||`

src/contracts/interfaces/IPermissionController.sol
ypatil12: How about agent??

src/contracts/mixins/PermissionControllerMixin.sol
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * @notice Checks if the caller is allowed to call a function on behalf of an account.
+     * @param account the account to check
+     * @dev `msg.sender` is the caller to check that can call the function on behalf of `account`.
+     * @dev Returns a bool, instead of reverting
+     */
+    function _checkCanCall(
+        address account
+    ) internal returns (bool) {
+        return permissionController.canCall(account, msg.sender, address(this), msg.sig);
```
wadealexc, line 36: Can this return `msg.sender == account || <external call>`? That way a short-circuit will prevent an external call for what feels like a likely common case.

src/contracts/permissions/PermissionController.sol
ypatil12: So we don't want to overload this term. How about delegate?

src/contracts/permissions/PermissionController.sol
wadealexc, line 44: Needs `inheritdoc`

wadealexc: Or as Michael said prevent having 0 when you remove

src/contracts/permissions/PermissionController.sol
```diff
@@ -0,0 +1,176 @@
...
+    /// @dev Decodes the target and selector from a single bytes32 value
+    function _decodeTargetSelector(
+        bytes32 targetSelector
+    ) internal view returns (address, bytes4) {
+        address target = address(uint160(uint256(targetSelector) >> 96));
+        // The selector is in the lower 32 bits of the targetSelector
+        bytes4 selector = bytes4(uint32(uint256(targetSelector) >> 64));
+
+        return (target, selector);
+    }
```
wadealexc, line 112: If we are using shifts to decode, we should consider using shifts to encode. Makes it easier to reason that they're inverse operations.

src/contracts/permissions/PermissionController.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/870/commits/cf26f4eb29af4a89763b2e26b99802c8fd04d090

src/contracts/interfaces/IPermissionController.sol
ypatil12: Went with appointee: https://github.com/Layr-Labs/eigenlayer-contracts/pull/870/commits/cf26f4eb29af4a89763b2e26b99802c8fd04d090

src/contracts/mixins/PermissionControllerMixin.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import "../interfaces/IPermissionController.sol";
+
+abstract contract PermissionControllerMixin {
```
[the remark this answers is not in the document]
ypatil12: We do Mixin in the SignatureChecker too? You mean `PermissionControlledMixin`

src/contracts/mixins/PermissionControllerMixin.sol
```diff
@@ -0,0 +1,38 @@
...
+    IPermissionController public immutable permissionController;
+
+    constructor(
+        IPermissionController _permissionController
+    ) {
+        permissionController = _permissionController;
+    }
+
+    /// @notice Checks if the caller (msg.sender) can call on behalf of an account.
+    modifier checkCanCall(
```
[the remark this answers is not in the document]
ypatil12: Hmm, to me `checkCanCall` is still the simplest to read

src/contracts/permissions/PermissionControllerStorage.sol
ypatil12: Prefer not using salted location for consistency with storage layouts across core

src/contracts/permissions/PermissionController.sol
ypatil12: Changed function to addAdmin. Kept onlyAdmin modifier given we have a function with same name: https://github.com/Layr-Labs/eigenlayer-contracts/pull/870/commits/ccb8c89bf32175dc20f9d2b77f179db465ab9f98

src/contracts/permissions/PermissionController.sol
ypatil12: Let's prevent 0. We want to specifically enable permanent key rotation

src/contracts/permissions/PermissionControllerStorage.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/870/commits/eac8bbe1116ca3cef86eaada1f9d6d666aff49f7

src/contracts/permissions/PermissionController.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/870/commits/eac8bbe1116ca3cef86eaada1f9d6d666aff49f7

src/contracts/permissions/PermissionController.sol
ypatil12: Reads cleaner: https://github.com/Layr-Labs/eigenlayer-contracts/pull/870/commits/cd629bed6e808752567b701a1fc607d0f9281d14

src/contracts/permissions/PermissionController.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/870/commits/ccb8c89bf32175dc20f9d2b77f179db465ab9f98

src/contracts/permissions/PermissionController.sol
ypatil12: Hm, not seeing the use case here

src/contracts/mixins/PermissionControllerMixin.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/870/commits/efe8e2e4ad33a3a1341ba9d62fe65d5e0afcff05

src/contracts/permissions/PermissionController.sol
ypatil12: Not sure why we would want to call a subset of this function?

src/contracts/permissions/PermissionController.sol
ypatil12: Don't think any of those work??

`bytes32(abi.encodePacked(target, selector))`

The encoded would be: [160 bits target][32 bits selector][64 bits padding]

`bytes4 selector = bytes4(uint32(uint256(targetSelector));`

Ends up being all zeros because we're getting the least significant 32 bits of the encoded value. 

Can't do abi.decode since it depends on bytes and would rather not do the bytes.concat(bytes32).

src/contracts/mixins/PermissionControllerMixin.sol
ypatil12: Will do for the function, but modifier is much cleaner without

ypatil12: Can do for the function, but the modifier is much cleaner without. Still think we should pass in the appointee (formerly called delegate)?

src/contracts/permissions/PermissionController.sol
ypatil12: Added basic reversibility: https://github.com/Layr-Labs/eigenlayer-contracts/pull/870/commits/343b44e11fe3b3824941fbb6fe05a19307b93c25

ypatil12:
> If an admin is set, neither the isAdmin or canCall methods give permission when account == caller.

Since granting permissions is not 2-step, it's possible to brick your delegation just by setting your first admin to an account that can't do anything.

Are we aware of this limitation? Could we consider allowing account == caller to act as an admin/universal permissions?

Yes we are aware. The reason why we don't allow `account == caller` to act as a universal admin is that we want to enable complete key rotation and doing so results in the original account key having perms. 

Adding a 2-step makes sense

src/contracts/core/AllocationManager.sol
```diff
@@ -536,6 +536,11 @@ contract AllocationManager is
      *
      */
 
+    /// @inheritdoc IAllocationManager
+    function getAVSSetCount(address avs) external view returns (uint256) {
+        return _operatorSets[avs].length();
+    }
```
wadealexc: `getOperatorSetCount`

wadealexc (approved):
Looks good except for the name. Approving so you can merge after changing!

src/contracts/core/AllocationManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/870/commits/c24a46b0b9e17eda0b254b59680b9babf143a751

src/contracts/core/DelegationManager.sol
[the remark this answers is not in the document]
ypatil12: Yup, very good call out. I'll add this to the ELIP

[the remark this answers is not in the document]
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/870/commits/a1c843d33f21478f4c16ffca7388d5a3fd66ce13

PR #872, opened 2024-11-04
refactor: require "name" to be populated for deployments

nadir-akhtar:
This change introduces the requirement for users to explicitly title their contract when performing deployments. This name will be used by Zeus for tracking.


gpsanant (approved):

jbrower95:
sucks that you have to specify manually. if you could have the name inferred that would be baller

PR #876, opened 2024-11-05
refactor: delegation refactor

wadealexc:


src/contracts/libraries/SlashingLib.sol
```diff
@@ -23,13 +23,11 @@ uint64 constant WAD = 1e18;
  * Note that `withdrawal.scaledShares` is scaled for the beaconChainETHStrategy to divide by the beaconChainScalingFactor upon queueing
  * and multiply by the beaconChainScalingFactor upon withdrawal
  */
-struct StakerScalingFactors {
-    uint184 depositScalingFactor;
-    uint64 beaconChainScalingFactor;
-    bool isBeaconChainScalingFactorSet;
+struct DepositScalingFactor {
```
8sunyuan, line 26: Could probably remove the struct

8sunyuan: Unless the clarity with the lib functions is preferred which I assume is why it was left in

wadealexc: Yeah this felt like a clarity thing, but I'm not sold one way or another. I may reconsider as I finish this refactor.

src/contracts/core/DelegationManager.sol
```diff
@@ -173,42 +190,31 @@ contract DelegationManager is
         uint64[] memory maxMagnitudes = allocationManager.getMaxMagnitudes(operator, strategies);
 
         for (uint256 i = 0; i < strategies.length; i++) {
-            StakerScalingFactors storage ssf = stakerScalingFactor[staker][strategies[i]];
-
             // If the operator was not slashed 100% for the strategy and the staker has not been fully slashed
```
8sunyuan, line 193: Can we remove/update this comment? This is no longer true and we will have withdrawal roots produced even if the operator is 100% slashed. ScaledShares will just be 0 in this case

src/contracts/core/DelegationManager.sol
```diff
@@ -725,6 +728,19 @@ contract DelegationManager is
         return _operatorDetails[operator].delegationApprover;
     }
 
+    /// @inheritdoc IDelegationManager
+    function depositScalingFactor(address staker, IStrategy strategy) public view returns (uint256) {
+        return _depositScalingFactor[staker][strategy].scalingFactor();
```
8sunyuan, line 733: This should return `SlashingLib.scalingFactor()`. Existing stakers who never redeposit after upgrade should read WAD calling this function.

src/contracts/libraries/SlashingLib.sol
```diff
@@ -58,130 +56,91 @@ library SlashingLib {
...
     // CONVERSION
 
-    function toShares(
+    function calcWithdrawable(
+        DepositScalingFactor memory dsf,
         uint256 depositShares,
-        StakerScalingFactors memory ssf,
-        uint64 magnitude
-    ) internal pure returns (uint256 shares) {
+        uint256 slashingFactor
```
gpsanant, line 135: why are we renaming this?

src/contracts/libraries/SlashingLib.sol
gpsanant: is custom type the move?

src/contracts/core/DelegationManager.sol
```diff
@@ -293,23 +283,21 @@ contract DelegationManager is
...
-                maxMagnitude: maxMagnitudes[0]
-            });
-        }
+        IStrategy[] memory strategies = new IStrategy[](1);
+        strategies[0] = strategy;
+        uint64 maxMagnitude = allocationManager.getMaxMagnitudes(operator, strategies)[0];
+
+        // Increase the staker's deposit scaling factor and delegate shares to the operator
+        _increaseDelegation({
+            operator: delegatedTo[staker],
```
8sunyuan, line 294: reuse `operator` instead of reading storage again

src/contracts/core/DelegationManager.sol
```diff
@@ -533,20 +521,21 @@ contract DelegationManager is
...
-        operatorShares[operator][strategy] += addedShares;
-        emit OperatorSharesIncreased(operator, staker, strategy, addedShares);
-
         // Update the staker's depositScalingFactor
         DepositScalingFactor storage dsf = _depositScalingFactor[staker][strategy];
         dsf.update(existingDepositShares, addedShares, slashingFactor);
         emit DepositScalingFactorUpdated(staker, strategy, dsf.scalingFactor());
+
+        // If the staker is delegated to an operator, update the operator's shares
+        if (isDelegated(staker)) {
```
8sunyuan, line 535: Less clear but more efficient, can just check `operator != address(0)`. Fine for either soln here

src/contracts/core/DelegationManager.sol
```diff
@@ -533,20 +521,21 @@ contract DelegationManager is
...
         require(slashingFactor != 0, FullySlashed());
 
-        // Increment operator shares
-        operatorShares[operator][strategy] += addedShares;
-        emit OperatorSharesIncreased(operator, staker, strategy, addedShares);
-
         // Update the staker's depositScalingFactor
         DepositScalingFactor storage dsf = _depositScalingFactor[staker][strategy];
         dsf.update(existingDepositShares, addedShares, slashingFactor);
         emit DepositScalingFactorUpdated(staker, strategy, dsf.scalingFactor());
```
8sunyuan, line 532: Noting: The DSF won't update because staker is undelegated => maxMagnitude is 0 => DSF is always equal to WAD on each new deposit. However, if it is the beaconChainStrategy and the BCSF is non-WAD then this will update.

src/contracts/core/DelegationManager.sol
wadealexc: Agreed... though I'm gonna keep this as is, since SLOAD will get cheaper when reading a slot twice in a txn.

It's nice that we use the same helper throughout the contract. I won't die on this hill though, so if we want to change this i'm game.

src/contracts/core/DelegationManager.sol
wadealexc: I'll add a comment but I'd rather keep the code path without a branch for the BCSF. Good note, ty

wadealexc: [34ba24b](https://github.com/Layr-Labs/eigenlayer-contracts/pull/876/commits/34ba24be2ffa25c33f8c45803a56bb9377010a31)

src/contracts/core/DelegationManager.sol
wadealexc: [34ba24b](https://github.com/Layr-Labs/eigenlayer-contracts/pull/876/commits/34ba24be2ffa25c33f8c45803a56bb9377010a31)

src/contracts/core/DelegationManager.sol
wadealexc: [2ea120f](https://github.com/Layr-Labs/eigenlayer-contracts/pull/876/commits/2ea120f280126f1f2d32e8baad70e859ecea0175)

src/contracts/libraries/SlashingLib.sol
```diff
@@ -58,130 +56,91 @@ library SlashingLib {
...
-        StakerScalingFactors storage ssf,
+    function update(
+        DepositScalingFactor storage dsf,
         uint256 existingDepositShares,
         uint256 addedShares,
-        uint64 maxMagnitude
+        uint256 slashingFactor
     ) internal {
         if (existingDepositShares == 0) {
             // if this is their first deposit for the operator, set the scaling factor to inverse of maxMagnitude
```
ypatil12, line 90: Comment NIT: Inverse of `slashingFactor`. This isn't necessarily the inverse of the `maxMagnitude`. For Native ETH, it's the inverse of (mag*bcsf)

src/contracts/core/DelegationManager.sol
```diff
@@ -538,29 +531,31 @@ contract DelegationManager is
      * @param strategy The strategy to increase the delegated delegatedShares and the depositScalingFactor for
      * @param existingDepositShares The number of deposit shares the staker already has in the strategy.
      * @param addedShares The shares added to the staker in the StrategyManager/EigenPodManager
-     * @param maxMagnitude The current max magnitude of the operator for the strategy
+     * @param slashingFactor The current slashing factor for the staker/operator/strategy
      */
     function _increaseDelegation(
```
ypatil12, line 536: This method name is a bit misleading now, since it's called anytime a staker's shares increase, regardless of being delegated

ypatil12: Previously we (incorrectly) only increased when a staker was delegated

wadealexc: i hear that.... but im not sure what a better name is. hard to capture both "update DSF" and "increase delegation" in one name

the simplest thing would be `updateShares` but i feel it's important to capture the delegation portion? idk. im gonna leave this for this PR, but im open to any changes you want to make?

ypatil12: Ya, more of a callout than anything. I don’t have a better name so fine keeping as-is

ypatil12 (approved):

PR #881, opened 2024-11-08
feat: burn erc20s on slashing

8sunyuan:
### Synchronous Burning
One solution for burning slashed shares is to simply take the existing logic, take the sharesToDecrement in `calcSlashedAmount` and just burn that amount. But for all shares that were withdrawn and are still slashable for `MIN_WITHDRAWAL_DELAY_BLOCKS`, they would have to be slashed and burned when those withdrawals are completed.

Instead, completed withdrawals don't burn and we calculate all the shares to burn upon slashing. We burn synchronously all those slashabe shares in the queue by reading all the cumulative withdrawn shares from the time of slashing and `MIN_WITHDRAWAL_DELAY_BLOCKS` into the past. Here is where checkpointed values come into play again.

Changes:
- `Snapshots.sol`: Added back OZ CheckpointsUpgradeable library for blocknumbers to the Snapshots library. The reasoning is because we want to keep track of cumulative queue withdrawn shares from an operator for a specific strategy and not using a default WAD value like we use for magnitudes.
- `DelegationManager.sol`: renamed function `decreaseAndBurnOperatorShares` which will now read all the slashable shares queue withdrawn from an operator for a specific strategy, calculate amount of shares to burn, and call `StrategyManager.burnShares`.
- `StrategyManager.sol`: burnShares function is exact same as the withdraw function but with the destination address being the `DEFAULT_BURN_ADDRESS` (address 0).
- `SlashingLib.sol`: Changed `calcSlashedAmount` back to taking in as input prevMaxMagnitude and newMaxMagnitude because the way that burned shares are calculated with cumulative scaled shares.

Note: The call trace begins from `ALM.slashOperator -> DM.decreaseAndBurnOperatorShares -> SM.burnShares -> Strategy.withdraw -> underlyingToken.transfer` Each of these calls are permissioned by the core contracts in the protocol with the exception of slashOperator which is only callable by the AVS. We may want to additionally consider for a reentrancy guard on `slashOperator` as a result. 

TODO:
- Added unit and integration tests


src/contracts/core/DelegationManagerStorage.sol
```diff
@@ -115,6 +119,10 @@ abstract contract DelegationManagerStorage is IDelegationManager {
     /// @dev This variable only reflects withdrawals that were made after the slashing release.
     mapping(bytes32 withdrawalRoot => Withdrawal withdrawal) public queuedWithdrawals;
 
+    /// @dev Contains history of the total cumulative staker withdrawals for a given strategy. Used to calculate
+    /// burned shares for StrategyManager shares.
+    mapping(IStrategy strategy => Snapshots.WithdrawalHistory) internal _cumulativeWithdrawalsHistory;
```
ypatil12, line 124: This needs to be storage on a per operator, strategy basis? Otherwise we burn all queued withdrawals that were slashed for a given strategy regardless of operator?

8sunyuan: Just fixed, realized this miss too

src/contracts/core/StrategyManagerStorage.sol
```diff
@@ -26,6 +26,9 @@ abstract contract StrategyManagerStorage is IStrategyManager {
     // index for flag that pauses deposits when set
     uint8 internal constant PAUSED_DEPOSITS = 0;
 
+    /// @notice default address for burning slashed shares and transferring underlying tokens
+    address public constant DEFAULT_BURN_ADDRESS = address(0);
```
gpsanant, line 30: can we do some analysis on if this is ok for LSTs?

src/contracts/libraries/Snapshots.sol
```diff
@@ -127,4 +130,132 @@ library Snapshots {
...
+        uint224 _value;
+    }
+
+    /**
+     * @dev Returns the value at a given block number. If a withdrawal snapshot is not available at that block, the closest one
+     * before it is returned, or zero otherwise. Similar to {upperLookup} but optimized for the case when the searched
+     * withdrawal snapshot is probably "recent", defined as being among the last sqrt(N) withdrawal snapshots where N is the number of
+     * withdrawal snapshots.
+     */
+    function getAtProbablyRecentBlock(
```
gpsanant, line 153: just do upperLookup imo. is this any different?

src/contracts/core/DelegationManager.sol
```diff
@@ -685,6 +699,26 @@ contract DelegationManager is
         _beaconChainSlashingFactor[staker] = bsf;
     }
 
+    /// @dev Calculate amount of withdrawable shares from an operator for a strategy that is still in the queue
+    /// and therefore slashable.
+    function _getSlashableSharesInQueue(address operator, IStrategy strategy) internal view returns (uint256) {
```
gpsanant, line 704: can we expose a public view func for this?

src/contracts/core/DelegationManager.sol
```diff
@@ -685,6 +699,26 @@ contract DelegationManager is
...
+        return currCumulativeWithdrawalShares - pastCumulativeWithdrawalShares;
+    }
+
+    /// @dev Update the cumulative withdrawn shares from an operator for a given strategy
+    function _updateCumulativeWithdrawnShares(
+        address operator,
+        IStrategy strategy,
+        uint256 queueWithdrawnShares
+    ) internal {
+        uint256 currCumulativeWithdrawalShares = uint256(_cumulativeWithdrawalsHistory[operator][strategy].latest());
```
gpsanant, line 718: this PR looks nice!

my main feedback is that we should store _cumulative scaled shares_ queued for withdrawal, not shares.

Upon burning we should multiply the scaled shares in the withdrawal queue by the slashed magnitude. This would make it so that view functions for introspection into the withdrawal queue automatically update to reflect correct values after slashing

src/contracts/core/StrategyManagerStorage.sol
8sunyuan: It's for sure not, placeholder value for now. Personally a fan of 0xdeadbeef

src/contracts/libraries/Snapshots.sol
8sunyuan: It's the same logic but named more clearly for blocknumber usage. I can change this back though to be more consistent with DefaultWadHistory.

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/138eef799d1ce06c60ca1f2b06b626ff2410dfa5 
Added, had to change calcSlashedAmount back to old interface but it is more consistent with the way scaledShares are also calculated.

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/138eef799d1ce06c60ca1f2b06b626ff2410dfa5

src/contracts/core/StrategyManagerStorage.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/91a806b48e0c665ef6389749b6582b3c7ec01c18 
0xDeaDBeef

src/contracts/core/DelegationManager.sol
```diff
@@ -625,6 +636,11 @@ contract DelegationManager is
 
             // Remove delegated shares from the operator
             if (operator != address(0)) {
+                // Staker was delegated and remains slashable during the withdrawal delay period
+                // Cumulative withdrawn scaled shares are updated for the strategy, this is for accounting
+                // purposes for burning shares if slashed
+                _updateCumulativeScaledShares(operator, strategies[i], scaledShares[i]);
```
gpsanant, line 642: we don't want to do this for beaconchain ETH, because this value is sorta useless since it doesn't take slashing factors into account

src/contracts/core/StrategyManagerStorage.sol
ypatil12: We should at least do `0x000000000000000000000000000000000000dEaD` or address(0) since other protocols use that. Working with Matt Nelson to get feedback

ypatil12: Looks like we cannot do address(0) given that stETH's contract

src/contracts/core/DelegationManager.sol
8sunyuan: Good catch, yes this only for SM Strategies

8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/b0ed123a0abfbec7c6f961097c74a850d708512d

src/contracts/core/AllocationManager.sol
```diff
@@ -111,13 +111,14 @@ contract AllocationManager is
...
-            // 6. Decrease operators shares in the DelegationManager
-            delegation.decreaseOperatorShares({
+            // 6. Decrease and burn operators shares in the DelegationManager
+            delegation.decreaseAndBurnOperatorShares({
                 operator: params.operator,
                 strategy: strategy,
-                wadSlashed: sharesWadSlashed
+                prevMaxMagnitude: prevMaxMagnitude,
+                newMaxMagnitude: info.maxMagnitude
             });
```
wadealexc, line 122: I think we could just call this `burnOperatorShares` for brevity.

src/contracts/core/AllocationManager.sol
```diff
@@ -111,13 +111,14 @@ contract AllocationManager is
 
             // 5. Update state
             _updateAllocationInfo(params.operator, operatorSet.key(), strategy, info, allocation);
-            _updateMaxMagnitude(params.operator, strategy, info.maxMagnitude);
+            uint64 prevMaxMagnitude = _updateMaxMagnitude(params.operator, strategy, info.maxMagnitude);
```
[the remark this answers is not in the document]
wadealexc: (or, its prior value, anyway)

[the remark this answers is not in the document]
wadealexc: feels much better to assign `prevMaxMagnitude` higher in the method near `slashedMagnitude` and `sharesWadSlashed` rather than giving this method a new return value.

src/contracts/core/DelegationManager.sol
```diff
@@ -732,6 +748,32 @@ contract DelegationManager is
...
+        uint256 slashableShareInQueue = SlashingLib.scaleSharesForBurningQueueWithdrawals(
+            pastCumulativeScaledShares, currCumulativeScaledShares, maxMagnitude
+        );
+
+        return slashableShareInQueue;
+    }
+
+    /// @dev Update the cumulative withdrawn scaled shares from an operator for a given strategy
+    function _updateCumulativeScaledShares(address operator, IStrategy strategy, uint256 scaledShares) internal {
+        if (strategy != beaconChainETHStrategy) {
```
wadealexc, line 771: Why would beacon chain stuff also not be added?

Do we add it to this array eventually when we have burning for native ETH?

src/contracts/core/DelegationManager.sol
```diff
@@ -732,6 +748,32 @@ contract DelegationManager is
...
+        uint64 maxMagnitude
+    ) internal view returns (uint256) {
+        uint256 currCumulativeScaledShares = uint256(_cumulativeScaledSharesHistory[operator][strategy].latest());
+        // Note: this will simply return 0 if no history exists, same for latest() as well
+        uint256 pastCumulativeScaledShares = _cumulativeScaledSharesHistory[operator][strategy].upperLookupRecent(
+            block.number - MIN_WITHDRAWAL_DELAY_BLOCKS
+        );
+        uint256 slashableShareInQueue = SlashingLib.scaleSharesForBurningQueueWithdrawals(
+            pastCumulativeScaledShares, currCumulativeScaledShares, maxMagnitude
+        );
```
wadealexc, line 764: Cleaner name - just `scaleSharesForBurning`

src/contracts/libraries/SlashingLib.sol
```diff
@@ -141,7 +149,18 @@ library SlashingLib {
             .mulWad(slashingFactor);
     }
 
-    function calcSlashedAmount(uint256 operatorShares, uint256 wadSlashed) internal pure returns (uint256) {
-        return operatorShares.mulWad(wadSlashed);
+    function calcSlashedAmount(
+        uint256 operatorShares,
+        uint256 queuedWithdrawnScaledShares,
```
wadealexc, line 154: `scaledSharesInQueue` - shorter

src/test/unit/DelegationUnit.t.sol
```diff
@@ -3944,6 +3952,333 @@ contract DelegationManagerUnitTests_completeQueuedWithdrawal is DelegationManage
...
+
+        uint256 queuedSlashableSharesAfter = delegationManager.getSlashableSharesInQueue(operator, strategyMock);
+        uint256 operatorSharesAfter = delegationManager.operatorShares(operator, strategyMock);
+        assertEq(queuedSlashableSharesBefore, 0, "there should be no slashable shares in queue");
+        assertEq(queuedSlashableSharesAfter, 0, "there should be no slashable shares in queue");
+        assertEq(operatorSharesAfter, operatorSharesBefore - sharesToBurn, "operator shares should be decreased by sharesToBurn");
+    }
+
+    /**
+     * @notice Test burning shares for an operator with no slashable queued withdrawals in past week.
```
ypatil12, line 4006: Nit: past 14 days?

src/test/unit/DelegationUnit.t.sol
```diff
@@ -3944,6 +3952,333 @@ contract DelegationManagerUnitTests_completeQueuedWithdrawal is DelegationManage
     }
 }
 
+contract DelegationManagerUnitTests_burningShares is DelegationManagerUnitTests {
+    using SingleItemArrayLib for *;
+
```
ypatil12, line 3957: One sanity test: withdraw all & after the withdrawal delay blocks have elapsed, nothing is burned

src/test/unit/StrategyManagerUnit.t.sol
```diff
@@ -1416,6 +1416,50 @@ contract StrategyManagerUnitTests_withdrawSharesAsTokens is StrategyManagerUnitT
...
+        address staker,
+        uint256 depositAmount,
+        uint256 sharesToBurn
+    ) external filterFuzzedAddressInputs(staker) {
+        cheats.assume(staker != address(0));
+        cheats.assume(sharesToBurn > 0 && sharesToBurn < dummyToken.totalSupply() && depositAmount >= sharesToBurn);
+        IStrategy strategy = dummyStrat;
+        IERC20 token = dummyToken;
+        _depositIntoStrategySuccessfully(strategy, staker, depositAmount);
+        uint256 balanceBefore = token.balanceOf(staker);
```
ypatil12, line 1455: balance before should be of default burn address?

src/contracts/core/DelegationManager.sol
8sunyuan: At minimum, in the `burnOperatorShares` method we wouldn't be able to use these cumulative scaled shares to efficiently burn beaconChain shares in a callback like the one we make to the StrategyManager.
However, the scaled shares does take into account slashing factors and therefore the bcsf. We could potentially store this value as well?

src/test/unit/DelegationUnit.t.sol
```diff
@@ -3944,6 +3952,333 @@ contract DelegationManagerUnitTests_completeQueuedWithdrawal is DelegationManage
...
+            uint256[] memory sharesArray = shares.toArrayU256();
+            uint256[] memory depositArray = depositAmount.toArrayU256();
+            strategyManagerMock.setDeposits(staker1, strategyArray, sharesArray);
+            // Set the second staker's deposits in the strategies
+            strategyManagerMock.setDeposits(staker2, strategyArray, depositArray);
+        }
+        _delegateToOperatorWhoAcceptsAllStakers(staker1, operator);
+        _delegateToOperatorWhoAcceptsAllStakers(staker2, operator);
+
+        // 3. Queue withdrawal for staker2 and roll blocks forward so that the withdrawal is not slashable
```
ypatil12, line 4123: Nit: typo, test not rolling blocks forward

src/contracts/libraries/Snapshots.sol
```diff
@@ -127,4 +130,129 @@ library Snapshots {
...
+     * Does not default to WAD, as the value is always 0 if there are no snapshots.
+     */
+    struct WithdrawalHistory {
+        WithdrawalSnapshot[] _snapshots;
+    }
+
+    struct WithdrawalSnapshot {
+        uint32 _blockNumber;
+        uint224 _value;
+    }
```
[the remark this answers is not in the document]
wadealexc: On this note, I don't think we should be including these `SafeCastUpgradable` calls in `Snapshots`. It's adding bytecode and boy that's at a huge premium rn :sweat:

src/contracts/libraries/Snapshots.sol
```diff
@@ -127,4 +130,129 @@ library Snapshots {
...
+
+    /**
+     * @dev Returns the value at a given block number. If a withdrawal snapshot is not available at that block, the closest one
+     * before it is returned, or zero otherwise. Similar to {upperLookup} but optimized for the case when the searched
+     * withdrawal snapshot is probably "recent", defined as being among the last sqrt(N) withdrawal snapshots where N is the number of
+     * withdrawal snapshots.
+     */
+    function upperLookupRecent(WithdrawalHistory storage self, uint256 blockNumber) internal view returns (uint256) {
+        require(blockNumber < block.number, BlocknumberDoesNotExist());
```
wadealexc, line 154: I don't think we need this check. We should validate this in the DM rather than the library

src/contracts/core/DelegationManager.sol
```diff
@@ -732,6 +751,32 @@ contract DelegationManager is
...
+    function _getSlashableSharesInQueue(
+        address operator,
+        IStrategy strategy,
+        uint64 maxMagnitude
+    ) internal view returns (uint256) {
+        uint256 currCumulativeScaledShares = uint256(_cumulativeScaledSharesHistory[operator][strategy].latest());
+        // Note: this will simply return 0 if no history exists, same for latest() as well
+        uint256 pastCumulativeScaledShares = _cumulativeScaledSharesHistory[operator][strategy].upperLookupRecent(
+            block.number - MIN_WITHDRAWAL_DELAY_BLOCKS
+        );
```
wadealexc, line 764: The completable block for a withdrawal and the block we use for the scaled share history lookup are the same block. 

There's a racing condition where i have a withdrawal that can be completed in the same block a slash occurs - if my withdrawal is completed first, then we burn too many shares when the slash lands.

Can you confirm this is an issue? If so, we need a patch + regression test.

src/contracts/core/AllocationManager.sol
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/173c411ca2bc66853c4f417d09557a35dc5a4935

src/contracts/core/DelegationManager.sol
```diff
@@ -732,6 +748,32 @@ contract DelegationManager is
...
+        return slashableShareInQueue;
+    }
+
+    /// @dev Update the cumulative withdrawn scaled shares from an operator for a given strategy
+    function _updateCumulativeScaledShares(address operator, IStrategy strategy, uint256 scaledShares) internal {
+        if (strategy != beaconChainETHStrategy) {
+            uint256 currCumulativeScaledShares = uint256(_cumulativeScaledSharesHistory[operator][strategy].latest());
+            _cumulativeScaledSharesHistory[operator][strategy].push(currCumulativeScaledShares + scaledShares);
+        }
+    }
```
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/173c411ca2bc66853c4f417d09557a35dc5a4935

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/173c411ca2bc66853c4f417d09557a35dc5a4935

src/contracts/libraries/Snapshots.sol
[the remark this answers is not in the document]
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/173c411ca2bc66853c4f417d09557a35dc5a4935 Implemented your suggestion for reusability. 

I used SafeCastUpgradeable in one call when casting uint256 to uint224 since thats used in the underlying storage. However, I've validated that `uint224` is plenty large enough to store cumulative withdrawals in the protocol so happy to remove this too.

src/contracts/libraries/SlashingLib.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/173c411ca2bc66853c4f417d09557a35dc5a4935 `slashableSharesInQueue` since the function just takes in the shares after 'descaling' the scaled shares

src/contracts/core/AllocationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/173c411ca2bc66853c4f417d09557a35dc5a4935

src/contracts/libraries/Snapshots.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/173c411ca2bc66853c4f417d09557a35dc5a4935
Removed upperLookupRecent and this check as well

src/test/unit/DelegationUnit.t.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/4799402ad08aacd77cf121280bb2e91e62e5dfe8

src/test/unit/StrategyManagerUnit.t.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/4799402ad08aacd77cf121280bb2e91e62e5dfe8

src/contracts/core/DelegationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/4799402ad08aacd77cf121280bb2e91e62e5dfe8#diff-b8d08460d9bf99760a27f1f51eb20e5293abe371cc45d3442b62180da942bd0aR4279

Tested to be working as intended here. Checks the following:

1. On the block right before the withdrawal is completable, attempting to complete withdrawal reverts and the slashable shares are in the queue.
2. On the block the withdrawal is completable, completing a withdrawal doesn't revert and the slashable shares in the queue are 0.
3. Burning shares after step 2: results in 0 shares burned since operatorShares are 0 and all the staker's shares were withdrawn and completed.

src/contracts/libraries/Snapshots.sol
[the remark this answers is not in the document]
wadealexc: I think we should do an unsafe case. We're doing unsafe casts for `uint64` and i don't see this as a different case - as long as we've validated we can't pass values larger than that.

It saves us ~150 bytes which we really need

src/contracts/core/DelegationManager.sol
wadealexc: Thanks for the test. I think this isn't a race condition because our `upperLookup` library methods only return if a key is strictly greater than the input key.

src/contracts/libraries/Snapshots.sol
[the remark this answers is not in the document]
wadealexc: [e804598](https://github.com/Layr-Labs/eigenlayer-contracts/pull/881/commits/e80459845c6b945779b79ba8ffccfb65e91d1f58)

wadealexc (approved):

PR #883, opened 2024-11-12
feat: prevent beacon chain allocs

0xClandestine:
### Changes

- ~~`require(strategies[i] != beaconChainETHStrategy)` check to `addStrategiesToOperatorSet` and `createOperatorSets`.~~ 

- `if(strategy == beaconChainETHStrategy) continue;` added to `modifyAllocations`. This prevents allocating to the beacon-chain ETH strategy, but does not revert if it's provided. Can alternatively modify this to a `require` statement.

- Test to ensure no storage is mutated if `beaconChainETHStrategy` is provided.

### DIFF
--- a/src/test/unit/AllocationManagerUnit.t.sol
+++ b/src/test/unit/AllocationManagerUnit.t.sol
@@ -2858,6 +2858,15 @@ contract AllocationManagerUnitTests_removeStrategiesFromOperatorSet is Allocatio
 
 contract AllocationManagerUnitTests_createOperatorSets is AllocationManagerUnitTests {
     using SingleItemArrayLib for *;
+    
+    function test_CannotIncludeBeaconChainETHStrategy() public {
+        IStrategy strategy = IStrategy(allocationManager.beaconChainETHStrategy());
+        cheats.prank(defaultAVS);
+        cheats.expectRevert(InvalidStrategy.selector);
+        allocationManager.createOperatorSets(
+            CreateSetParams(1, strategy.toArray()).toArray()
+        );
+    }
 
     function test_createOperatorSets_InvalidOperatorSet() public {
         cheats.prank(defaultAVS);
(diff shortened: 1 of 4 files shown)


src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -1729,6 +1729,31 @@ contract AllocationManagerUnitTests_ModifyAllocations is AllocationManagerUnitTe
         }
     }
 
+    function test_CannotIncludeBeaconChainETHStrategy() public {
```
ypatil12, line 1732: For sanity can we add a test that attempts to slash beaconChainETH in an operatorSet and does not? We need to ensure no storage mutated or events emitted with beaconChainETH on a slash

ypatil12, line 1732: For sanity can we add a test that attempts to slash beaconChainETH in an operatorSet and does not? We need to ensure no storage mutated or events emitted with beaconChainETH on a slash

ypatil12 (commented):
Can we add a sanity test?

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -411,6 +411,25 @@ contract AllocationManagerUnitTests_SlashOperator is AllocationManagerUnitTests
...
+    function test_revert_CannotSlashBeaconChainETHStrategy() public {
+        IStrategy[] memory strategies = allocationManager.beaconChainETHStrategy().toArray();
+
+        cheats.prank(defaultAVS);
+        allocationManager.addStrategiesToOperatorSet(defaultOperatorSet.id, strategies);
+
+        cheats.prank(defaultAVS);
+        allocationManager.slashOperator(SlashingParams(defaultOperator, defaultOperatorSet.id, WAD, ""));
+
+        // assertEq(
```
ypatil12: Can we slash this operator (mag < WAD) and then assert that the magnitude is still WAD

PR #889, opened 2024-11-13
feat: add getMinimumSlashableStake

wadealexc:


src/contracts/core/AllocationManager.sol
```diff
@@ -738,4 +738,42 @@ contract AllocationManager is
...
+                uint64 maxMagnitude = _maxMagnitudeHistory[operator][strategy].latest();
+                if (maxMagnitude == 0) {
+                    continue;
+                }
+
+                Allocation memory alloc = getAllocation(operator, operatorSet, strategy);
+
+                // If the pending change takes effect before `futureBlock`, include it in `currentMagnitude`
+                if (alloc.effectBlock <= futureBlock) {
+                    alloc.currentMagnitude = _addInt128(alloc.currentMagnitude, alloc.pendingDiff);
```
gpsanant, line 771: we should only do this if the increase is negative. otherwise this isn't a minimum. 

Since 
1. an operator may queue an allocation that takes effect at block `n` and increases slashable stake by 10%
2. an AVS calls `getMinimumSlashableStake` for `futureBlock=n` at block `n-10` and gets back the stake, factoring in the 10% increase
3. the AVS slashes the operator at `n-5`, before the increase, thinking that the increase is what's at stake

stevennevins (approved):
API change lgtm

src/contracts/core/AllocationManager.sol
wadealexc: [93aa2af](https://github.com/Layr-Labs/eigenlayer-contracts/pull/889/commits/93aa2afd48cc33f7c9a244640bc6b0f7d202fb4f)

PR #892, opened 2024-11-15
refactor: EigenPods and beacon chain slashing

wadealexc:
Changes:
* Removed `_beaconChainSlashingFactor` from `DelegationManager` and placed in `EigenPodManager`
* Altered equation used to decrease operator shares when beacon chain scaling factor is reduced (see `DM.decreaseDelegatedShares`)
* EigenPod checkpoints no longer deleted when complete. Checkpoint timestamps are reset to 0, but the struct itself is no longer deleted. This saves gas when creating a new checkpoint, because the state slot is not zeroed out.
  * Removed slashing math calculation from EigenPod in favor of implementing this in `EPM._reduceSlashingFactor`. This removes a branch from EigenPods.
* `EPM.recordBeaconChainETHBalanceUpdate` branches cleaned up:
  * Fixed bug where no update would occur when `podOwnerDepositShares == 0`
  * Explicitly follow `_addShares` path when `balanceDeltaWei == 0`

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -289,7 +289,7 @@ contract DelegationManager is
     function increaseDelegatedShares(
         address staker,
         IStrategy strategy,
-        uint256 existingDepositShares,
+        uint256 curDepositShares,
         uint256 addedShares
     ) external onlyStrategyManagerOrEigenPodManager {
         address operator = delegatedTo[staker];
@@ -301,42 +301,44 @@ contract DelegationManager is
             operator: operator,
             staker: staker,
             strategy: strategy,
-            existingDepositShares: existingDepositShares,
+            curDepositShares: curDepositShares,
             addedShares: addedShares,
             slashingFactor: slashingFactor
         });
     }
 
     /// @inheritdoc IDelegationManager
-    function decreaseBeaconChainScalingFactor(
+    function decreaseDelegatedShares(
         address staker,
-        uint256 existingDepositShares,
-        uint64 proportionOfOldBalance
+        uint256 curDepositShares,
+        uint64 prevBeaconChainSlashingFactor,
+        uint256 wadSlashed
     ) external onlyEigenPodManager {
-        // decrease the staker's beaconChainScalingFactor proportionally
-        address operator = delegatedTo[staker];
-        uint64 maxMagnitude = allocationManager.getMaxMagnitude(operator, beaconChainETHStrategy);
-
-        DepositScalingFactor memory dsf = _depositScalingFactor[staker][beaconChainETHStrategy];
-
-        uint256 slashingFactor = _getSlashingFactor(staker, beaconChainETHStrategy, maxMagnitude);
-        uint256 withdrawableBefore = dsf.calcWithdrawable(existingDepositShares, slashingFactor);
-
-        _decreaseBeaconChainSlashingFactor(staker, proportionOfOldBalance);
-
-        slashingFactor = _getSlashingFactor(staker, beaconChainETHStrategy, maxMagnitude);
-        uint256 withdrawableAfter = dsf.calcWithdrawable(existingDepositShares, slashingFactor);
-
-        // if the staker is delegated to an operators
-        if (isDelegated(staker)) {
-            // subtract strategy shares from delegated scaled shares
-            _decreaseDelegation({
-                operator: operator,
-                staker: staker,
-                strategy: beaconChainETHStrategy,
-                sharesToDecrease: withdrawableBefore - withdrawableAfter
-            });
+        if (!isDelegated(staker)) {
+            return;
         }
+
+        address operator = delegatedTo[staker];
+
+        // Calculate the previous slashing factor. This mirrors the calculation done
+        // in _getSlashingFactor
+        uint64 maxMagnitude = allocationManager.getMaxMagnitude(operator, beaconChainETHStrategy);
+        uint256 prevSlashingFactor = maxMagnitude.mulWad(prevBeaconChainSlashingFactor);
+
+        // Calculate the shares to remove from the operator by calculating the previously-withdrawable
+        // shares and applying slashing
+        DepositScalingFactor memory dsf = _depositScalingFactor[staker][beaconChainETHStrategy];
+        uint256 sharesToRemove = dsf.calcWithdrawable({
+            depositShares: curDepositShares,
+            slashingFactor: prevSlashingFactor
+        }).mulWad(uint256(WAD) - wadSlashed); // ? TODO verify
+
+        _decreaseDelegation({
+            operator: operator,
+            staker: staker,
+            strategy: beaconChainETHStrategy,
+            sharesToDecrease: sharesToRemove
+        });
     }
 
     /// @inheritdoc IDelegationManager
@@ -449,7 +451,7 @@ contract DelegationManager is
                 operator: operator, 
                 staker: staker, 
                 strategy: strategies[i],
-                existingDepositShares: uint256(0),
+                curDepositShares: uint256(0),
                 addedShares: depositedShares[i],
                 slashingFactor: slashingFactors[i]
             });
@@ -516,7 +518,7 @@ contract DelegationManager is
                 });
             } else {
                 // Award shares back in StrategyManager/EigenPodManager.
-                (uint256 existingDepositShares, uint256 addedShares) = shareManager.addShares({
+                (uint256 curDepositShares, uint256 addedShares) = shareManager.addShares({
                     staker: withdrawal.staker,
                     strategy: withdrawal.strategies[i],
                     token: tokens[i],
@@ -528,7 +530,7 @@ contract DelegationManager is
                     operator: newOperator,
                     staker: withdrawal.staker,
                     strategy: withdrawal.strategies[i],
-                    existingDepositShares: existingDepositShares,
+                    curDepositShares: curDepositShares,
                     addedShares: addedShares,
                     slashingFactor: newSlashingFactors[i]
                 });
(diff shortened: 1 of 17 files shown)


src/contracts/core/DelegationManager.sol
```diff
@@ -305,36 +305,36 @@ contract DelegationManager is
         });
     }
 
-    /// @inheritdoc IDelegationManager
-    function decreaseBeaconChainScalingFactor(
+    function decreaseDelegatedShares(
         address staker,
-        uint256 existingDepositShares,
-        uint64 proportionOfOldBalance
+        uint256 curDepositShares,
```
gpsanant, line 310: nit: `increaseDelegatedShares` calls this `existingDepositShares`. I have a slight preference for `cur`, but we should make consistent

src/contracts/pods/EigenPodManager.sol
```diff
@@ -124,10 +106,22 @@ contract EigenPodManager is
...
             });
-        } else if (sharesDelta < 0 && podOwnerDepositShares[podOwner] > 0) {
-            delegationManager.decreaseBeaconChainScalingFactor(
-                podOwner, uint256(podOwnerDepositShares[podOwner]), proportionOfOldBalance
-            );
+        } else if (sharesDelta < 0 && podOwnerDepositShares[podOwner] > 0) { // TODO >=
+            uint64 prevBeaconSlashingFactor = beaconChainSlashingFactor(podOwner);
+            uint64 newBeaconSlashingFactor = uint64(prevBeaconSlashingFactor.mulWad(wadSlashed));
+
+            emit BeaconChainSlashingFactorDecreased(podOwner, newBeaconSlashingFactor);
```
gpsanant, line 113: could we add wadsSlashed here for indexing purposes

src/contracts/core/DelegationManager.sol
wadealexc: ty, good callout. i meant to change all instances for consistency as i also prefer `cur`

src/contracts/pods/EigenPod.sol
```diff
@@ -626,35 +629,37 @@ contract EigenPod is Initializable, ReentrancyGuardUpgradeable, EigenPodPausingC
...
+            uint64 totalRestakedBeforeGwei = withdrawableRestakedExecutionLayerGwei + checkpoint.prevBeaconBalanceGwei;
+            uint64 totalRestakedAfterGwei = totalRestakedBeforeGwei - uint64(-totalShareDeltaGwei);
 
-            // Update pod owner's shares
-            eigenPodManager.recordBeaconChainETHBalanceUpdate(podOwner, totalShareDeltaWei, proportionOfOldBalance);
-            emit CheckpointFinalized(lastCheckpointTimestamp, totalShareDeltaWei);
-        } else {
-            _currentCheckpoint = checkpoint;
+            //TODO: analyze effects of rounding errors here see `test_VerifyWC_Slash_StartCP_VerifyWC_CompleteCP` for example
+            wadSlashed = uint256(totalRestakedAfterGwei).divWad(totalRestakedBeforeGwei);
```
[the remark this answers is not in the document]
wadealexc: I checked, same failing tests regardless. @8sunyuan could double check on precision loss tho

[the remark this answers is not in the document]
wadealexc: I will say there will never be a sub-gwei component to these numbers so I don't know where we'd get precision loss.

src/contracts/core/DelegationManager.sol
wadealexc: [4d5d8a8](https://github.com/Layr-Labs/eigenlayer-contracts/pull/892/commits/4d5d8a86da20c26e4ae7c3e500b24e0438a880c6)

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -129,4 +154,12 @@ interface IEigenPodManager is IEigenPodManagerErrors, IEigenPodManagerEvents, IS
 
     /// @notice returns canonical, virtual beaconChainETH strategy
     function beaconChainETHStrategy() external view returns (IStrategy);
+
+    /**
+     * @notice Returns the amount of reported slashing a staker has experienced on the
```
8sunyuan, line 159: "Amount" is a bit unclear here as its not actual ETH amounts. Maybe historical sum of proportions slashed is more correct but also a mouthful.

src/contracts/pods/EigenPodManager.sol
```diff
@@ -267,8 +259,42 @@ contract EigenPodManager is
...
+        /// forgefmt: disable-next-item
+        _beaconChainSlashingFactor[podOwner] = BeaconChainSlashingFactor({
+            slashingFactor: newBeaconSlashingFactor,
+            isSet: true
+        });
+
+        uint256 curDepositShares = uint256(podOwnerDepositShares[podOwner]);
+        // Note that underflow here should be impossible given
+        // `balanceRemainingWad` is guaranteed to be less than WAD.
+        uint256 wadSlashed = uint256(WAD) - balanceRemainingWad;
```
8sunyuan, line 291: https://github.com/Layr-Labs/eigenlayer-contracts/pull/901
Already made the relevant change in this PR but wanted to note I think we should round down the amount slashed by changing the `divWad` in calculating `balanceRemainingWad` to be rounded upwards.
We do this similarly in `SlashingLib.calcSlasheAmount` specifically for underflow edge cases

wadealexc: [3d4e2a6](https://github.com/Layr-Labs/eigenlayer-contracts/pull/892/commits/3d4e2a6d10792ebb36c436610ddbe0113084d373)

src/contracts/pods/EigenPodManager.sol
```diff
@@ -267,8 +259,42 @@ contract EigenPodManager is
...
+    function _reduceSlashingFactor(
+        address podOwner,
+        uint256 prevRestakedBalanceWei,
+        uint256 balanceDecreasedWei
+    ) internal returns (uint256, uint64, uint256) {
+        // Apply negative balance delta to calculate the proportion of the original
+        // balance that remains. Note that underflow here should be impossible given
+        // the invariants pods use to calculate these values.
+        uint256 newRestakedBalanceWei = prevRestakedBalanceWei - balanceDecreasedWei;
+        uint256 balanceRemainingWad = newRestakedBalanceWei.divWadRoundUp(prevRestakedBalanceWei);
```
ypatil12, line 274: Nit: The naming here is throwing me off since I view this as a proportion, but balanceRemaining implies a delta. Is the thinking that Wad reflects this term as a proportion?

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -49,14 +49,39 @@ interface IEigenPodManagerEvents {
         address withdrawer,
         bytes32 withdrawalRoot
     );
+
+    /// @notice Emitted when a staker's beaconChainSlashingFactor is updated
+    event BeaconChainSlashingFactorDecreased(address staker, uint64 newBeaconChainSlashingFactor);
```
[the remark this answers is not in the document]
ypatil12: `PodSharesUpdated` is unfortunately an exception to the rule, but changing that has tons of offchain implications now

src/contracts/pods/EigenPodStorage.sol
```diff
@@ -11,7 +11,7 @@ abstract contract EigenPodStorage is IEigenPod {
     uint64 internal __deprecated_mostRecentWithdrawalTimestamp;
 
     /// @notice the amount of execution layer ETH in this contract that is staked in EigenLayer (i.e. withdrawn from the Beacon Chain but not from EigenLayer),
-    uint64 public withdrawableRestakedExecutionLayerGwei;
+    uint64 internal restakedExecutionLayerGwei;
```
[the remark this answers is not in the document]
wadealexc: yes! but this isn't a breaking change because i added a `public` function with the same name as the state variable. 

So the UI doesn't need to know about this - it'll function exactly the same.

src/contracts/pods/EigenPodManager.sol
wadealexc: > Is the thinking that Wad reflects this term as a proportion?

yes that was the idea. What about `proportionRemainingWad`?

src/contracts/interfaces/IEigenPodManager.sol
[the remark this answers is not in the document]
wadealexc: Okay actually i am gonna leave the name as-is, because an earlier request from G was to add `wadSlashed` in as a parameter, which is the delta.

src/contracts/pods/EigenPodManager.sol
ypatil12: I like that one better :)

src/contracts/pods/EigenPodStorage.sol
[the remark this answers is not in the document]
ypatil12: Ah, did not see that one. Perfect

src/contracts/pods/EigenPodManager.sol
wadealexc: [678abaf](https://github.com/Layr-Labs/eigenlayer-contracts/pull/892/commits/678abaf2c00c20254daf9b4515dca86cae2495d3)

src/contracts/interfaces/IEigenPodManager.sol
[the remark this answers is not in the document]
wadealexc: [678abaf](https://github.com/Layr-Labs/eigenlayer-contracts/pull/892/commits/678abaf2c00c20254daf9b4515dca86cae2495d3)

src/contracts/interfaces/IEigenPodManager.sol
wadealexc: [678abaf](https://github.com/Layr-Labs/eigenlayer-contracts/pull/892/commits/678abaf2c00c20254daf9b4515dca86cae2495d3)

src/contracts/pods/EigenPodManager.sol
wadealexc: [678abaf](https://github.com/Layr-Labs/eigenlayer-contracts/pull/892/commits/678abaf2c00c20254daf9b4515dca86cae2495d3)

8sunyuan (approved):

PR #896, opened 2024-11-15
feat: regen align-environments branch

MinisculeTarantula:
an attempt to recreate and fix https://github.com/Layr-Labs/eigenlayer-contracts/pull/794

modifications had to be made due to changes to configs etc due to work on Zeus

this also notably excludes deploy scripts for the EIGEN token and TimelockControllers specifically for Holesky, where they were already used


ypatil12:
@MinisculeTarantula Do we need to merge this?

MinisculeTarantula:
> @MinisculeTarantula Do we need to merge this?

As far as I can tell, there is no _need_ to merge this. All operations it was concerned with were completed, and the switch over to using Zeus for these things seems quite complete.
This PR does accomplish some cleanup + correct the JSON config files in this repo, but it might be better to just make a new PR that more ruthlessly deletes things instead?

PR #902, opened 2024-11-22
docs: slashing AVS metadataURI standard

ypatil12:


bowenli86 (commented):
problem of storing operator set offchain:

- not able to replay history or restore state any a given past block
- current metadata uri convention is that avs/operator can update/change content in the uri without notifying eigenlayer or call `setMetadata()` b/c EL will periodically crawl them. Asking them to set operator set in uri will break this convention. (btw, i do think the convention should be broken and fixed, but not for reason of accomodating operator set)

can we store this state onchain instead of offchain?

ypatil12:
> can we store this state onchain instead of offchain?

What state do you want to store onchain? @bowenli86

bowenli86:
> > can we store this state onchain instead of offchain?
> 
> What state do you want to store onchain? Everything in the URI? Are you suggesting to not have an avs metadataURI? @bowenli86

meant section of "operatorSets" and which operator sets an AVS is in
 
to help me understand better, how will the new "operatorSets" be used E2E in the design? eg. AVS call setMetadataURI to provide uri of operator sets -> ?? (what's next?)

stevennevins (approved):

bowenli86 (approved):

PR #903, opened 2024-11-22
refactor: remove deprecated methods

wadealexc:


src/contracts/core/DelegationManager.sol
```diff
@@ -550,15 +521,15 @@ contract DelegationManager is
      * @param operator The operator to increase the delegated delegatedShares for
      * @param staker The staker to increase the depositScalingFactor for
      * @param strategy The strategy to increase the delegated delegatedShares and the depositScalingFactor for
-     * @param curDepositShares The number of deposit shares the staker already has in the strategy.
+     * @param prevDepositShares The number of deposit shares the staker had in the strategy prior to the increase
```
ypatil12, line 524: This comment isn't true in all cases. If the staker undelegates & then delegates, we pass in 0 depositShares but that isn't the number of deposit shares the staker had prior to the increase. It's the number of _delegated_ deposited shares the staker had in the strategy

ypatil12 (commented):
Looks good, just one minor comment

src/contracts/core/DelegationManager.sol
wadealexc: [55e1e7d](https://github.com/Layr-Labs/eigenlayer-contracts/pull/903/commits/55e1e7d060f70a3d27bcd9d4b5e235ed64319ac6)

ypatil12 (approved):

PR #905, opened 2024-11-25
feat: register as avs

ypatil12:
Adds a `registerAsAVS` function to the ALM as an entrypoint for AVSs to the system. As part of this change, we require that the avss's registrar is never set to `address(0)`, which is used by the `isAVS` function to know if an address is an AVS. 

We use `isAVS` in three places

- Gate avs operations such as creating operator sets, updating URI, etc
- Gate which address can call the UAM version of `setClaimerFor`


src/contracts/core/AllocationManager.sol
```diff
@@ -26,6 +26,13 @@ contract AllocationManager is
     using OperatorSetLib for OperatorSet;
     using SlashingLib for uint256;
 
+    modifier validateAVS(
+        address avs
+    ) {
+        require(isAVS(avs), InvalidAVS());
+        _;
+    }
+
```
wadealexc, line 35: Would call this `onlyAVS` since it basically works like `onlyOwner` or `onlyOperator` or whatever.

src/contracts/core/AllocationManager.sol
```diff
@@ -230,6 +239,7 @@ contract AllocationManager is
         RegisterParams calldata params
     ) external onlyWhenNotPaused(PAUSED_OPERATOR_SET_REGISTRATION_AND_DEREGISTRATION) checkCanCall(operator) {
         // Check that the operator exists
+        require(isAVS(params.avs), InvalidAVS());
```
8sunyuan, line 242: Use modifier to be consistent

src/contracts/core/AllocationManager.sol
```diff
@@ -292,40 +302,43 @@ contract AllocationManager is
...
-        emit AVSRegistrarSet(avs, getAVSRegistrar(avs));
+    function registerAsAVS(
+        IAVSRegistrar registrar,
+        string calldata metadataURI,
+        CreateSetParams[] calldata params
+    ) external {
+        _updateAVSRegistrar(msg.sender, registrar);
+        _updateAVSMetadataURI(msg.sender, metadataURI);
+        _createOperatorSets(msg.sender, params);
     }
```
wadealexc, line 313: So I think part of the appeal of this change would be to make "AVS" an "Operator" a unique distinction. As in, you can be one or the other, but not both. ... So I'd recommend adding a check that `msg.sender` is NOT an operator in the core contracts.

That said -- we would need to check whether any existing AVSs are also operators on mainnet, because we don't want this to brick anyone...

src/contracts/core/AllocationManager.sol
```diff
@@ -59,7 +66,9 @@ contract AllocationManager is
     function slashOperator(
         address avs,
         SlashingParams calldata params
-    ) external onlyWhenNotPaused(PAUSED_OPERATOR_SLASHING) checkCanCall(avs) {
+    ) external onlyWhenNotPaused(PAUSED_OPERATOR_SLASHING) {
```
8sunyuan, line 69: How explicit do we want to be with the `validateAVS` modifiers to check the avs input _is_ an AVS?
For example registerForOperatorSets has the isAVS check but slashOperator doesn't. For registering for an operatorSet, it has to exist implying it has to have been created, which also implies that the AVS is registered (since you must be a registered AVS to create operatorSets). Similarly, the `slashOperator` function also doesn't have the isAVS check but for there to be any slashable allocations in the first place, the AVS had to have been registered too.

src/contracts/core/DelegationManager.sol
```diff
@@ -115,12 +115,9 @@ contract DelegationManager is
     }
 
     /// @inheritdoc IDelegationManager
-    function modifyOperatorDetails(
-        address operator,
-        OperatorDetails calldata newOperatorDetails
-    ) external checkCanCall(operator) {
+    function modifyOperatorDetails(address operator, address newDelegationApprover) external checkCanCall(operator) {
```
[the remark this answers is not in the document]
ypatil12: Rebasing TODO

src/contracts/core/AllocationManager.sol
ypatil12: I actually removed this check we validate the opSet exists

src/contracts/core/AllocationManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/905/commits/4a8d82622e97b881672e3073f2e7796febf9200a

src/contracts/core/AllocationManager.sol
ypatil12: I figured we don't need it for `registerForOperatorSets` (this was erroneously checked before), `slashOperator` and `modifyAllocations` since we validate the existence of the operatorSet, which implies the avs must already have registered.

src/contracts/core/AllocationManager.sol
wadealexc: ... right. This is a lot of effort tbh :sweat:

src/contracts/core/AllocationManager.sol
ypatil12: So basically, if we validate the opSet exists, we don't need to validate that the avs IS an avs

PR #909, opened 2024-11-27
refactor: address informational security review findings

nadir-akhtar:
This PR addresses some of the informational findings found during the internal security review. Specifically, this PR:
* Arranges functions to match our EigenLayer style guide
* Makes a note on `processClaims()` about an array of too large a size potentially causing a user to hit gas limits, and recommending a smaller array in that circumstance

### DIFF
--- a/src/contracts/core/RewardsCoordinator.sol
+++ b/src/contracts/core/RewardsCoordinator.sol
@@ -341,6 +341,76 @@ contract RewardsCoordinator is
      */
 
     /**
+     * @notice Internal helper to process reward claims.
+     * @param claim The RewardsMerkleClaims to be processed.
+     * @param recipient The address recipient that receives the ERC20 rewards
+     */
+    function _processClaim(RewardsMerkleClaim calldata claim, address recipient) internal {
+        DistributionRoot memory root = _distributionRoots[claim.rootIndex];
+        _checkClaim(claim, root);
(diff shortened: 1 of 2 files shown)


0xrajath (commented):
I get why it should be reshuffled according to the Solidity guide. But I also don't think we should introduce such a large diff for audit purposes just to adhere to that.

nadir-akhtar:
I spoke with @wadealexc and refactored the PR accordingly. 

He told me that we don't actually follow the Solidity style guide for this repo, though no _documented_ internal style guide exists. After talking to him, I now understand that we order functions differently from the style guide, specifically in the following order:
* `external`/`public` functions grouped by relevance. Note that these are interchangable for style/ordering purposes: `external` and `public` functionscan be grouped together however it makes sense based on their purpose.
* `internal` functions. These are to be ordered by state-changing internal functions first, and `view`/`pure` functions interchangably after.
* `external`/`public` `view` functions, also interchangable.

As such, the PR diff as it stands now is smaller.

That said, this is a PR remedying an informational finding, the lowest possible severity. I recommend merging in the change so that auditors work off of the intended latest version. Moreover, there are no functionality changes with this PR, so it shouldn't be that disruptive, and it will ensure our codebase aligns with our general style guide for consistency and clarity. However, I also understand not wanting to disrupt auditors' understanding of the target commit. @0xrajath

0xrajath (approved):

PR #911, opened 2024-11-29
test: alm unit test refactor + coverage

ypatil12:
### Functionality Changes
- Fixes event emit logic -> bring `AllocationUpdated` into the parent `slashOperator` and `modifyAllocations` functions. Keeping it in the `_updateAllocationInfo` function would result in having to do complex if case handling whether called from the above 2 funcs OR `clearDeallocationQueue` 
- Prevents encumbered mag from being updated when it doesn't need to be, such as when a deallocation is queued

### Test Changes
- Remove `assertApproxEq` and replace with a slippage calculation (this is worst case at most 1wei)
- Clean up all slashing tests with slashing helpers from @0xClandestine 
- Added assertions around `getMinimumSlashableStake` 

### Doc Changes
- We need to properly document the rounding error (at most 1 Wei) when an AVS's operatorSet doesn't have stake slashed but the magnitude of their slashable stake goes down due to ANOTHER operatorSet being slashed. This is due to rounding down on operator shares
- Added commenting around usage of `encumberedMagnitude` -> integrators need to be careful about reading directly from it as it doesn't accurately reflect what is currently encumbered because items on the deallocation queue can be cleared


src/contracts/core/AllocationManager.sol
```diff
@@ -490,15 +508,18 @@ contract AllocationManager is
...
+        // The mapping should NOT be updated when there is a deallocation on a delay
+        if (encumberedMagnitude[operator][strategy] != info.encumberedMagnitude) {
+            encumberedMagnitude[operator][strategy] = info.encumberedMagnitude;
+            emit EncumberedMagnitudeUpdated(operator, strategy, info.encumberedMagnitude);
+        }
 
         // Update allocation for this operator set from the strategy
+        // We emit an `AllocationUpdated` from the `modifyAllocations` and `slashOperator` functions.
+        // `clearDeallocationQueue` does not emit an `AllocationUpdated` event since it was
+        // emitted when the deallocation was queued
```
8sunyuan, line 521: Ah this comment makes it clear to me why you removed the event from this internal function

src/contracts/core/AllocationManager.sol
8sunyuan, line 1: So the AllocationUpdated event on the deallocation on step (4) can occur multiple times for that same effect block if there are multiple slashings on the same deallocation. Something to keep in mind when trying to index events.

src/contracts/core/AllocationManagerStorage.sol
```diff
@@ -83,6 +83,8 @@ abstract contract AllocationManagerStorage is IAllocationManager {
         _maxMagnitudeHistory;
 
     /// @dev For a strategy, contains the amount of magnitude an operator has allocated to operator sets
+    /// @dev This value should be read with caution, as deallocations that are completable but not
+    ///      popped off the queue are still included in the encumbered magnitude
     mapping(address operator => mapping(IStrategy strategy => uint64)) public encumberedMagnitude;
```
8sunyuan, line 88: Good clarification, the "true" encumberedMagnitude can be read from calling 
`getMaxMagnitude(operator, strategy) - getAllocatableMagnitude(operator, strategy)` which could potentially give you a smaller value than what is in storage.

src/contracts/core/AllocationManager.sol
ypatil12: Yup, our subgraph and sidecar properly handle this

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -528,15 +662,30 @@ contract AllocationManagerUnitTests_SlashOperator is AllocationManagerUnitTests
...
         AllocateParams[] memory allocateParams = _newAllocateParams(defaultOperatorSet, WAD);
 
         cheats.prank(defaultOperator);
-        allocationManager.modifyAllocations(defaultOperator, allocateParams);
+        allocationManager.modifyAllocations(defaultOperator, allocateParams);    
+
+        uint64 encumberedMagnitudeBefore = allocationManager.encumberedMagnitude(defaultOperator, strategyMock);
+        uint64 maxMagnitudeBefore = allocationManager.getMaxMagnitudes(defaultOperator, strategyMock.toArray())[0];
+
+        // The only slash event we expect is the OperatorSlashed. There isn't a way to validate
```
8sunyuan, line 681: Was thinking about this too when writing DM unit tests as well and wanting to check for certain events
not being emitted (when not delegated for example).
Perhaps in a separate PR we can write a utils library using 
`vm.recordLogs()` and `vm.getRecordedLogs().length` before and after the call

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -643,50 +791,44 @@ contract AllocationManagerUnitTests_SlashOperator is AllocationManagerUnitTests
...
+            })
         });
+
+        uint256 slashedStake = DEFAULT_OPERATOR_SHARES.mulWad(expectedWadSlashed); // Wad is same as slashed mag since we start with max mag
         _checkSlashableStake({
             operatorSet: defaultOperatorSet,
             operator: defaultOperator,
             strategies: defaultStrategies,
-            expectedStake: (DEFAULT_OPERATOR_SHARES - slashedStake).mulWad(newSlashableMagnitude)
+            expectedStake: (DEFAULT_OPERATOR_SHARES - slashedStake).mulWad(expectedCurrentMag.divWad(expectedMaxMag))
```
8sunyuan, line 831: What is the mulWad and divWad used here for? 
Is the expectedStake input taking in the expected operatorShares? or the expected slashable operatorShares

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -1030,43 +1120,35 @@ contract AllocationManagerUnitTests_SlashOperator is AllocationManagerUnitTests
...
-        _checkSlashEvents({
-            operator: defaultOperator,
-            operatorSet: allocateParams[0].operatorSet,
-            strategies: allocateParams[0].strategies,
-            wadsToSlash: slashingParams.wadsToSlash,
-            description: "test"
-        });
+        // Manually check slash events since we have a deallocation pending
+        // Deallocation update is emitted first
+        cheats.expectEmit(true, true, true, true, address(allocationManager));
```
8sunyuan, line 1142: Possible to replace with `_checkSlashEvents`?

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -1174,12 +1244,12 @@ contract AllocationManagerUnitTests_SlashOperator is AllocationManagerUnitTests
...
     /**
-     * Allocates all magnitude to a single opSet. Deallocateas magnitude. Slashes al
+     * Allocates all magnitude to a single opSet. Deallocates magnitude. Slashes all
      * Asserts that:
      * 1. The Allocation is 0 after slash
-     * 2. Them sotrage post slash for encumbered and maxMags ais zero
+     * 2. Them storage post slash for encumbered and maxMags is zero
      */
-    function test_allocateAll_deallocateAll() public {
+    function test_slash_allocateAll_deallocateAll() public {
```
8sunyuan, line 1252: Would be good to complete the deallocation here and still see that the `allocatable` amount is 0

src/test/unit/AllocationManagerUnit.t.sol
8sunyuan, line 1: `testFuzz_revert_insufficientAllocatableMagnitude` and `testFuzz_revert_overAllocate` are equivalent. We can just keep one

8sunyuan, line 1: L3173: missed `cheats.expectEmit(true, true, false, false, address(allocationManager));`

src/contracts/core/AllocationManager.sol
```diff
@@ -292,7 +308,7 @@ contract AllocationManager is
     function setAllocationDelay(address operator, uint32 delay) external {
         if (msg.sender != address(delegation)) {
             require(_checkCanCall(operator), InvalidCaller());
-            require(delegation.isOperator(operator), OperatorNotRegistered());
+            require(delegation.isOperator(operator), InvalidOperator());
```
8sunyuan, line 311: Can we delete error `OperatorNotRegistered`?

src/test/unit/AllocationManagerUnit.t.sol
ypatil12: 100% , let me add to TODOs

ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/911/commits/03aa2f85a712f56093d9238934b5accb79e83a53

src/test/unit/AllocationManagerUnit.t.sol
ypatil12: Unfortunately we cannot since a deallocation also emits an additional event interspersed with the other events. Would need a helper but are only testing this twice so figure its not needed rn

src/test/unit/AllocationManagerUnit.t.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/911/commits/26112fe52d6a327ae7da00c4caad2ae227b64e16

src/test/unit/AllocationManagerUnit.t.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/911/commits/4ab0ee14860ebec5286f08510ac82d3e1c6e392f

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -46,8 +46,6 @@ uint8 constant PAUSED_NON_PROOF_WITHDRAWALS = 5;
 
 abstract contract IntegrationDeployer is ExistingDeploymentParser, Logger {
     using StdStyle for *;
-    using ArrayLib for *;
-    using ArrayLib for IStrategy[];
```
0xClandestine, line 50: Why was this removed?

src/test/unit/RewardsCoordinatorUnit.t.sol
```diff
@@ -2710,8 +2710,8 @@ contract RewardsCoordinatorUnitTests_processClaim is RewardsCoordinatorUnitTests
         }
 
         // Parse all 3 claim proofs for distributionRoots 0,1,2 respectively
-        IRewardsCoordinator.RewardsMerkleClaim[] memory claims = _parseAllProofs();
-        IRewardsCoordinator.RewardsMerkleClaim memory claim = claims[2];
+        IRewardsCoordinatorTypes.RewardsMerkleClaim[] memory claims = _parseAllProofs();
```
0xClandestine, line 2713: nit: inherit `IRewardsCoordinatorTypes` for cleaner syntax

src/contracts/core/AllocationManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/911/commits/aeb02ac411cbc1fb79350d64fcdc5c14bb77a998

0xClandestine (commented):
small nits

src/test/unit/AllocationManagerUnit.t.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/911/commits/7e6d1acee86fc7e026913d220c57fea859531adb

src/test/integration/IntegrationDeployer.t.sol
ypatil12: Think it was just a remanant of the rebase? Still compiles/passes

src/test/unit/RewardsCoordinatorUnit.t.sol
ypatil12: Good catch, updated entire file: https://github.com/Layr-Labs/eigenlayer-contracts/pull/911/commits/005d9e7879a3e1b65321203999480e86f9ad9188

0xClandestine (approved):

8sunyuan (approved):

PR #912, opened 2024-11-30
feat: upgrade compatibility for integration tests + basic upgrade tests

ypatil12:
This PR updates the integration testing framework for slashing by:

- Adding a new `User_M2` for M2 operations
- Getting all fork tests to pass in the CI (and updating the associated workflow)
- Adjusting assertions to handle M2 and slashing interfaces

It handles the following upgrade scenarios:

1. EigenPod migration: `Integration_EigenPod_Slashing_Migration` 
2. Queueing prior to the upgrade and completing after the upgrade: `testFuzz_delegate_deposit_queue_completeAsShares` and `testFuzz_delegate_deposit_queue_completeAsTokens`

TODOs in a separate PR: 

- [ ]  Get rid of `assertApproxEq`
- [ ] Move randomization for slashing into its own the IntegrationBase (instead of in the slash case)

### DIFF
--- a/src/test/integration/tests/eigenpod/EigenPod_Slashing_Migration.t.sol
+++ b/src/test/integration/tests/eigenpod/EigenPod_Slashing_Migration.t.sol
@@ -1,0 +1,75 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "src/test/integration/IntegrationChecks.t.sol";
+import "src/test/integration/users/User.t.sol";
+
+contract Integration_EigenPod_Slashing_Migration is IntegrationCheckUtils, EigenPodPausingConstants {
+    modifier r(uint24 _rand) {
+        _configRand({
+            _randomSeed: _rand,
+            _assetTypes: HOLDS_ETH,
+            _userTypes: DEFAULT
+        });
+
+        _;
+    }
+
+    /**
+     * 1. Verify validators' withdrawal credentials
+     *    -- earn rewards on beacon chain (withdrawn to pod)
+     * 2. Start a checkpoint
+     * 3. Pause starting checkpoints
+     * 4. Complete in progress checkpoint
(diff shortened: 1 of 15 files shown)


src/test/integration/tests/eigenpod/EigenPod_Slashing_Migration.t.sol
```diff
@@ -0,0 +1,75 @@
...
+        check_CompleteCheckpoint_WithPodBalance_State(staker, expectedWithdrawnGwei);
+
+        // 5. Upgrade Contracts for slashing
+        _upgradeEigenLayerContracts();
+
+        // 6. Exit validators
+        // Fully exit one or more validators and advance epoch without generating rewards
+        uint40[] memory subset = _choose(validators);
+        uint64 exitedBalanceGwei = staker.exitValidators(subset);
+        beaconChain.advanceEpoch_NoRewards();
```
8sunyuan, line 67: Thought of another migration test but not blocking right now. 
After upgrade, start/complete checkpoint with slashed validators and verify podDepositShares are unchanged

src/test/integration/users/User_M1.t.sol
8sunyuan, line 1: Seems we no longer use User_M1 anywhere, we could either delete or keep in case

8sunyuan (approved):
Comments aren't blocking and PR LGTM. Just need to rebase and fix conflicts

src/test/integration/tests/eigenpod/EigenPod_Slashing_Migration.t.sol
ypatil12: Added this to the integration testing plan

src/test/integration/users/User_M1.t.sol
ypatil12: Keeping just in case, we can deprecate after

PR #914, opened 2024-12-02
feat: add redelegate

wadealexc:
* add `redelegate`, a combination of `undelegate` and `delegateTo`
* refactors out common parts of each  method into helpers (`_undelegate` and `_checkApproverSignature`)
* reorders internal methods to make more sense (`queue` is now above `complete`)

### DIFF
--- a/src/contracts/interfaces/IDelegationManager.sol
+++ b/src/contracts/interfaces/IDelegationManager.sol
@@ -263,6 +263,22 @@ interface IDelegationManager is ISignatureUtils, IDelegationManagerErrors, IDele
     ) external returns (bytes32[] memory withdrawalRoots);
 
     /**
+     * @notice Undelegates the staker from their current operator, and redelegates to `newOperator`
+     * Queues a withdrawal for all of the staker's withdrawable shares.
+     * @dev This method acts like a call to `undelegate`, then `delegateTo`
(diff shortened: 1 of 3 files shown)


src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -262,6 +262,22 @@ interface IDelegationManager is ISignatureUtils, IDelegationManagerErrors, IDele
...
 
+    /**
+     * @notice Undelegates the staker from their current operator, and redelegates to `newOperator`
+     * Queues a withdrawal for all of the staker's withdrawable shares.
+     * @dev This method acts like a call to `undelegate`, then `delegateTo`
+     * @param newOperator the new operator that will be delegated all assets
+     * @dev NOTE: the following 2 params are ONLY checked if `newOperator` has a `delegationApprover`.
+     * If not, they can be left empty.
+     * @param newOperatorApproverSig A signature from the operator's `delegationApprover`
+     * @param approverSalt A unique single use value tied to the approver's signature
```
ypatil12, line 273: Nit: Can we make it clear that the new operator is delegated all of the assets AFTER the withdrawal is completed as shares

ypatil12 (approved):
LGTM, just one minor comment. Approving

PR #916, opened 2024-12-02
[1.0.0] slashing upgrade script

jbrower95:


script/releases/v1.0.0-slashing/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "solidity.defaultCompiler": "localNodeModule"
+}
```
jbrower95, line 3: this definitely shouldn't be committed

script/releases/v1.0.0-slashing/6-executeTimelock.s.sol
```diff
@@ -0,0 +1,57 @@
...
+        );
+    }
+
+    function testDeploy() override public {}
+
+    function testExecute() public {
+        // TODO: run 1-queueUpgrade + 2-queueUnpause, e.g via 
+    
+        // vm.startPrank(this._operationsMultisig());
+        // super._runAsMultisig();
```
jbrower95, line 39: @wadealexc / @ypatil12 i need syntax to call a method from another contract without inheritance. since we're testing two upgrade scripts here, we need access to `_runAsMultisig()` and also `_ _getMultisigTransactionCalldata` from both of those scripts...

jbrower95: or, we can do this as two separate phases, and just land one after the other with no changes

wadealexc: wouldn't this inherit from the queue script? i thought that's how we set the templates up

script/releases/v1.0.0-slashing/3-pause.s.sol
```diff
@@ -0,0 +1,31 @@
...
+    function options() internal override view returns (MultisigOptions memory) {
+        return MultisigOptions(
+            Env.pauserMultisig(),
+            Operation.Call
+        );
+    }
+
+    function runAsMultisig() internal virtual override {
+        // TODO: set: 
+        //    pause EigenPodManager
```
ypatil12, line 23: Can we also pause programmatic incentives as that has not been updated. 

`PAUSED_REWARD_ALL_STAKERS_AND_OPERATORS` in the RC

script/releases/v1.0.0-slashing/cleanup/EigenPod.abi.json
```diff
@@ -0,0 +1 @@
...atch","inputs":[]},{"type":"error","name":"InsufficientWithdrawableBalance","inputs":[]},{"type":"error","name":"InvalidEIP4788Response","inputs":[]},{"type":"error","name":"InvalidProof","inputs":[]},{"type":"error","name":"InvalidProofLength","inputs":[]},{"type":"error","name":"InvalidProofLength","inputs":[]},{"type":"error","name":"InvalidPubKeyLength","inputs":[]},{"type":"error","name":"InvalidValidatorFieldsLength","inputs":[]},{"type":"error","name":"MsgValueNot32ETH","inputs":[]},{"type":"error","name":"NoActiveCheckpoint","inputs":[]},{"type":"error","name":"NoBalanceToCheckpoint","inputs":[]},{"type":"error","name":"OnlyEigenPodManager","inputs":[]},{"type":"error","name":"OnlyEigenPodOwner","inputs":[]},{"type":"error","name":"OnlyEigenPodOwnerOrProofSubmitter","inputs":[]},{"type":"error","name":"TimestampOutOfRange","inputs":[]},{"type":"error","name":"ValidatorInactiveOnBeaconChain","inputs":[]},{"type":"error","name":"ValidatorIsExitingBeaconChain","inputs":[]},{"type":"error","name":"ValidatorNotActiveInPod","inputs":[]},{"type":"error","name":"ValidatorNotSlashedOnBeaconChain","inputs":[]},{"type":"error","name":"WithdrawalCredentialsNotForEigenPod","inputs":[]}]
```
ypatil12, line 1: Why not use bindings already in repo instead of json abi?

script/releases/v1.0.0-slashing/cleanup/script.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func panicOnError(msg string, err error) {
+	if err != nil {
+		fmt.Printf("Error: %s", msg)
+		panic(err)
+	}
+}
+
+func runScript(args TArgs) error {
```
ypatil12, line 78: For validation, I assume we plan to run this script twice - once to complete checkpoints, again to validate none are pending?

ypatil12 (commented):
Quick Review

script/releases/Env.sol
```diff
@@ -0,0 +1,266 @@
...
+
+    function timelockController() internal view returns (TimelockController) {
+        return TimelockController(payable(_envAddress("timelockController")));
+    }
+
+    function multiSendCallOnly() internal view returns (address) {
+        return _envAddress("MultiSendCallOnly");   
+    }
+
+    function EIGENPOD_GENESIS_TIME() internal view returns (uint64) {
```
8sunyuan, line 79: This can be removed entirely and also from EigenPod.sol itself

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
```diff
@@ -0,0 +1,187 @@
...
+            deployedTo: address(new AVSDirectory({
+                _delegation: Env.proxy.delegationManager(),
+                _pauserRegistry: Env.impl.pauserRegistry()
+            }))
+        });
+
+        deployImpl({
+            name: type(DelegationManager).name,
+            deployedTo: address(new DelegationManager({
+                _avsDirectory: Env.proxy.avsDirectory(),
```
ypatil12, line 81: Nit: rebase, so we don't have avsD

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
```diff
@@ -0,0 +1,187 @@
...
+        });
+
+        /// core/
+
+        deployImpl({
+            name: type(AllocationManager).name,
+            deployedTo: address(new AllocationManager({
+                _delegation: Env.proxy.delegationManager(),
+                _pauserRegistry: Env.impl.pauserRegistry(),
+                _permissionController: Env.proxy.permissionController(),
```
8sunyuan, line 55: How can we validate in Zeus this contract is correctly deployed and not some 0 address? Because this is being deployed from above from the `deployImpl` and `deployProxy` calls

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
8sunyuan, line 81: would be good to rebase since you will have to update constructors (avsd is removed here)

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
```diff
@@ -0,0 +1,187 @@
...
+                _ALLOCATION_CONFIGURATION_DELAY: Env.ALLOCATION_CONFIGURATION_DELAY()
+            }))
+        });
+
+        deployProxy({
+            name: type(AllocationManager).name,
+            deployedTo: address(new TransparentUpgradeableProxy({
+                _logic: address(Env.impl.allocationManager()),
+                admin_: Env.proxyAdmin(),
+                _data: ""
```
ypatil12, line 66: Initialize? cc @8sunyuan

script/releases/Env.sol
8sunyuan: I swear we removed this though way back

script/releases/v1.0.0-slashing/5-executeUpgradeAndUnpause.sol
```diff
@@ -0,0 +1,58 @@
...
+        assertFalse(timelock.isOperationDone(txHash), "Transaction should NOT be complete.");
+        
+        // 2- warp past delay
+        vm.warp(block.timestamp + timelock.getMinDelay()); // 1 tick after ETA
+        assertEq(timelock.isOperationReady(txHash), true, "Transaction should be executable.");
+        
+        // 3- execute
+        execute();
+
+        assertTrue(timelock.isOperationDone(txHash), "Transaction should be complete.");
```
ypatil12, line 56: Can we validate ALM has been initialized here?

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
```diff
@@ -0,0 +1,187 @@
...
+        deployImpl({
+            name: type(EigenPod).name,
+            deployedTo: address(new EigenPod({
+                _ethPOS: Env.ethPOS(),
+                _eigenPodManager: Env.proxy.eigenPodManager(),
+                _GENESIS_TIME: Env.EIGENPOD_GENESIS_TIME()
+            }))
+        });
+
+        /// strategies/
```
8sunyuan, line 137: Need to upgrade EigenStrategy. Unsure about Eigen

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
```diff
@@ -0,0 +1,187 @@
...
+
+        deployImpl({
+            name: type(StrategyBaseTVLLimits).name,
```
8sunyuan, line 140: Note: all LST Strategies with caps will point to this implementation

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
```diff
@@ -0,0 +1,187 @@
...
+
+        /// core/
+
+        deployImpl({
+            name: type(AllocationManager).name,
+            deployedTo: address(new AllocationManager({
+                _delegation: Env.proxy.delegationManager(),
+                _pauserRegistry: Env.impl.pauserRegistry(),
+                _permissionController: Env.proxy.permissionController(),
+                _DEALLOCATION_DELAY: Env.MIN_WITHDRAWAL_DELAY(),
```
ypatil12, line 56: image

These are returning different values from what we expect

ypatil12: 60 for withdrawal delay and 90 for allocation config delay: https://www.notion.so/eigen-labs/Slashing-Upgrade-14d13c11c3e080dca3abfb2eca7a8bf4?pvs=4#14d13c11c3e0803ca724c68d09638918

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
8sunyuan: Ah yes its added to the queueUpgrade script

script/releases/v1.0.0-slashing/2-queueUpgradeAndUnpause.s.sol
```diff
@@ -0,0 +1,148 @@
...
+        /// Add call to upgrade each pre-longtail strategy instance
+        uint count = Env.instance.strategyBaseTVLLimits_Count();
+        for (uint i = 0; i < count; i++) {
+            address proxyInstance = address(Env.instance.strategyBaseTVLLimits(i));
+
+            executorCalls.append({
+                to: Env.proxyAdmin(),
+                data: Encode.proxyAdmin.upgrade({
+                    proxy: proxyInstance,
+                    impl: address(Env.impl.strategyFactory())
```
ypatil12, line 114: Wrong upgrade implementation

ypatil12: Should be new `StrategyBase`

ypatil12 (changes requested):

script/releases/v1.0.0-slashing/2-queueUpgradeAndUnpause.s.sol
8sunyuan: This is why I find the `_verify...` functions in ExistingDeploymentParser to be useful as part of sanity check tests. We should port those over.

script/releases/v1.0.0-slashing/3-pause.s.sol
ypatil12: Nvm, going to hard fork

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
jbrower95: - You can check the deployed address, via the `deployments()` [struct](https://github.com/Layr-Labs/zeus-templates/blob/master/src/templates/EOADeployer.sol#L76)
- You can check the updated tracked value within zeus via the `Env` object (see how it fetches the contract address in there). In our tests, we typically grab the before and after addresses, assert that there was a change, and then test certain things about the deployed contract. (This is a forge fork test).

script/releases/v1.0.0-slashing/2-queueUpgradeAndUnpause.s.sol
jbrower95: Can you link? What do these do?

script/releases/Env.sol
```diff
@@ -0,0 +1,266 @@
...
+
+    function permissionController(DeployedImpl) internal view returns (PermissionController) {
+        return PermissionController(_deployedImpl(type(PermissionController).name)); 
+    }
+
+    /**
+     * pods/
+     */
+
+    function eigenPod(DeployedProxy) internal view returns (EigenPod) {
```
gpsanant, line 175: hm. this is gross. makes it weird to treat a beacon the same way as TUPs...

I need to read thru zeus

script/releases/v1.0.0-slashing/2-queueUpgradeAndUnpause.s.sol
gpsanant, line 114: can we change this to the Strategy factory beacon? That way we never need to special case old strategies

wadealexc: See Yash's comment above

script/releases/Env.sol
wadealexc: There's room to make this `Env.beacon.eigenPod` instead, and treat the two separately. I'm down.

script/releases/v1.0.0-slashing/2-queueUpgradeAndUnpause.s.sol
wadealexc: @ypatil12 new `StrategyBase`? Or new `StrategyBaseTVLLimits`?

I was gonna do the latter but did we discuss doing the former? (theoretically we aren't using and don't plan on using the "tvl limits" part of that contract, but i havent checked for storage compatibility)

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
wadealexc: Have fixed this in the metadata repo

script/releases/v1.0.0-slashing/5-executeUpgradeAndUnpause.sol
wadealexc: [15671ad](https://github.com/Layr-Labs/eigenlayer-contracts/pull/916/commits/15671adba6dcec1dc4ece525af44794792c775b7)

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
wadealexc: Will do this later today.

script/releases/v1.0.0-slashing/2-queueUpgradeAndUnpause.s.sol
wadealexc: okay chatted w/ G - the request is to make the instanced strategies proxy through a beacon proxy that points to the beacon used by the rest of our strategies

im tempted to leave this as-is. The existing strategies will still be edge cases in the StrategyFactory because of the blacklist/deploy system the factory uses -- so rather than try and fix this now, maybe we should leave as-is, given timelines are tight.

script/releases/v1.0.0-slashing/1-deployContracts.s.sol
wadealexc: Check the tests in this file now - they thoroughly test the newly-deployed contracts

script/releases/Env.sol
wadealexc: +1, but... it's not removed :D 

I'm gonna keep it here as removing it feels out of scope for this PR.

wadealexc (approved):

ypatil12 (approved):

PR #933, opened 2024-12-06
refactor: add `strategies` back to `SlashingParams`

0xClandestine:


src/contracts/core/AllocationManager.sol
```diff
@@ -59,28 +59,26 @@ contract AllocationManager is
     function slashOperator(
         address avs,
         SlashingParams calldata params
-    ) external onlyWhenNotPaused(PAUSED_OPERATOR_SLASHING) {
-        // Check that the msg.sender can call - we don't use a modifier to avoid `stack too deep` errors
-        require(_checkCanCall(avs), InvalidCaller());
-        require(0 < params.wadToSlash && params.wadToSlash <= WAD, InvalidWadToSlash());
+    ) external onlyWhenNotPaused(PAUSED_OPERATOR_SLASHING) checkCanCall(avs) {
+        // // Check that the msg.sender can call - we don't use a modifier to avoid `stack too deep` errors
+        // require(_checkCanCall(avs), InvalidCaller());
```
8sunyuan, line 64: delete

src/contracts/core/AllocationManager.sol
```diff
@@ -59,28 +59,26 @@ contract AllocationManager is
...
+        uint256[] memory wadSlashed = new uint256[](params.strategies.length);
 
         // For each strategy in the operator set, slash any existing allocation
-        for (uint256 i = 0; i < length; i++) {
+        for (uint256 i = 0; i < params.strategies.length; i++) {
+            // Check that `wadToSlash` is within acceptable bounds.
+            require(0 < params.wadsToSlash[i] && params.wadsToSlash[i] <= WAD, InvalidWadToSlash());
+
             // 1. Get the operator's allocation info for the strategy and operator set
-            IStrategy strategy = IStrategy(_operatorSetStrategies[operatorSet.key()].at(i));
```
8sunyuan, line 80: So we need to now require the strategy passed in is in the operatorSet because its no longer read from the enumerable set in storage

src/test/unit/AllocationManagerUnit.t.sol
8sunyuan, line 1: Could we add unit tests for the new requires in slashOperator?
1. The function should revert if strategy in calldata is not in the operatorSet
2. If there are duplicate strategies in calldata

0xClandestine: added in [07b9372](https://github.com/Layr-Labs/eigenlayer-contracts/pull/933/commits/07b93722d30c7bef90254ec572105180a9e13f7e)

8sunyuan (approved):
LGTM

PR #934, opened 2024-12-06
test: slashing integration todos

0xClandestine:
### Changes:

- adds `HOLDS_MAX` assetType (denotes a user that holds the maximum allowed strategies to test gas constraints).
- adds `print.gasUsed()` (prints the gas consumed from the last call).
- adds gas logging via `print.gasUsed()`.
- adds `PermissionController` support to `AVS` and `User` integration user types.
- removes **most** `assertApproxEqAbs` (1 single instance remains for beaconchain).

### TODOS:
- add remaining slashing related checks.
- add remaining slashing scenarios.


src/test/integration/IntegrationBase.t.sol
```diff
@@ -14,7 +14,7 @@ import "src/test/integration/TimeMachine.t.sol";
 import "src/test/integration/users/User.t.sol";
 import "src/test/integration/users/User_M1.t.sol";
 
-abstract contract IntegrationBase is IntegrationDeployer {
+abstract contract IntegrationBase is IntegrationDeployer, IAllocationManagerTypes {
```
8sunyuan, line 17: Can we remove IAllocationManagerTypes

src/test/integration/IntegrationBase.t.sol
```diff
@@ -351,6 +347,234 @@ abstract contract IntegrationBase is IntegrationDeployer {
...
+                );
+            }
+
+            // Check magnitudes
+            Magnitudes memory curMagnitude = curMagnitudes[i];
+            Magnitudes memory prevMagnitude = prevMagnitudes[i];
+
+            // if (isAllocation) ...
+            if (prevAlloc.currentMagnitude < allocateParams.newMagnitudes[i]) {
+                uint256 delta = allocateParams.newMagnitudes[i] - prevAlloc.currentMagnitude;
```
8sunyuan, line 401: Can we move this delta calculation above the if statement and then do the check. Would be a little easier to read 
and its being done in both if else blocks
```
// if (isAllocation)
if (delta > 0) {
...

// if (deallocation)
} else {
}

File: src/test/integration/IntegrationBase.t.sol [hunk 3]
```diff
@@ -351,6 +347,234 @@ abstract contract IntegrationBase is IntegrationDeployer {
...
+                
+                if (completed) {
+                    // FIXME: fix completed case
```
8sunyuan, line 404: If its an allocation, doesn't it add to encumbered magnitude(and therefore reducing allocatable by same amount)
regardless if the delay has passed? So we can remove the if check and the encumbered/allocatable values are always updated regardless if completed or not
https://github.com/Layr-Labs/eigenlayer-contracts/blob/d3109212d8869ebba791790bfe0e22bdfadd7e5f/src/contracts/core/AllocationManager.sol#L206-L211

src/test/integration/IntegrationBase.t.sol
```diff
@@ -351,6 +347,234 @@ abstract contract IntegrationBase is IntegrationDeployer {
...
+
+            uint64 slashedMagnitude = uint64(uint256(prevAlloc.currentMagnitude).mulWadRoundUp(slashingParams.wadsToSlash[i]));
+
+            assertEq(
+                curAlloc.currentMagnitude, 
+                prevAlloc.currentMagnitude - slashedMagnitude, 
+                string.concat(err, " (currentMagnitude)")
+            );
+
+            if (completed) {
```
8sunyuan, line 496: Even if minimal, very much appreciate the comments done in 
`assert_Snap_Allocations_Modified` which indicate what block of code is checking which values like 
`// Max magnitude should not have mutated.`, `// Check allocations`, etc

src/test/integration/IntegrationBase.t.sol
```diff
@@ -351,6 +347,234 @@ abstract contract IntegrationBase is IntegrationDeployer {
...
+                prevMagnitude.max, 
+                string.concat(err, " (max magnitude)")
+            );
+        }
+    }
+
+    function assert_Snap_Allocations_Slashed(
+        SlashingParams memory slashingParams,
+        OperatorSet memory operatorSet,
+        bool completed,
```
8sunyuan, line 472: What does completed mean in this context? When would this value be false?

0xClandestine: It's supposed to differentiate whether the allocation was completed before slashed or not.

src/test/integration/IntegrationBase.t.sol
```diff
@@ -351,6 +347,234 @@ abstract contract IntegrationBase is IntegrationDeployer {
...
+                        uint64(uint256(uint128(-prevAlloc.pendingDiff)).mulWadRoundUp(slashingParams.wadsToSlash[i]));
+                    
+                    assertEq(
+                        curAlloc.pendingDiff, 
+                        prevAlloc.pendingDiff + int128(uint128(slashedPending)), 
+                        string.concat(err, " (pendingDiff)")
+                    );
+                }
+
+                assertEq(
```
8sunyuan, line 517: I think the effectBlock here can vary depending on if we have a pending allocation or deallocation. Need to check for that

src/contracts/pods/EigenPod.sol
```diff
@@ -397,8 +397,8 @@ contract EigenPod is Initializable, ReentrancyGuardUpgradeable, EigenPodPausingC
      * @dev Reverts if `amountWei` is not a whole Gwei amount
      */
     function withdrawRestakedBeaconChainETH(address recipient, uint256 amountWei) external onlyEigenPodManager {
-        require(amountWei % GWEI_TO_WEI == 0, AmountMustBeMultipleOfGwei());
```
ypatil12, line 400: Can we get a unit test on this

ypatil12 (changes requested):
Need unit test on EP change

src/test/integration/IntegrationBase.t.sol
```diff
@@ -351,6 +347,234 @@ abstract contract IntegrationBase is IntegrationDeployer {
...
+
+            assertEq(
+                curMagnitude.max, 
+                prevMagnitude.max - slashedMagnitude, 
+                string.concat(err, " (max magnitude)")
+            );
+        }
+    }
+
+    function assert_HasUnderlyingTokenBalances_AfterSlash(
```
8sunyuan, line 547: For non native ETH Strategies, we should also have an assertion on the token balance of the Strategy being burned/updated after the `slashOperator` call and it should be sent to the burn address.
Perhaps it makes sense to be in this check or a separate assert function

src/test/integration/tests/Deposit_Delegate_Allocate.t.sol
```diff
@@ -27,21 +30,161 @@ contract Integration_Deposit_Delegate_Allocate is IntegrationCheckUtils {
...
+        for (uint256 i; i < operatorSets.length; ++i) {
+            assert_Snap_Allocations_Modified(
+                operator, allocateParams[i], true, "operator allocations should be updated after delay"
+            );
+        }
     }
-}
\ No newline at end of file
+
+    function testFuzz_deposit_delegate_allocate_slash_undelegate_completeAsTokens(
```
8sunyuan, line 62: To clarify, this is and the other slash + completeAsTokens integration tests failed due to the EigenPod gwei bug correct?

src/test/integration/tests/Deposit_Delegate_Allocate.t.sol
```diff
@@ -27,21 +30,161 @@ contract Integration_Deposit_Delegate_Allocate is IntegrationCheckUtils {
...
+        // 5. Undelegate from an operator
+        IDelegationManagerTypes.Withdrawal[] memory withdrawals = staker.undelegate();
+        bytes32[] memory withdrawalRoots = _getWithdrawalHashes(withdrawals);
+        // 6. Complete withdrawal
+        _rollBlocksForCompleteWithdrawals();
+        for (uint256 i = 0; i < withdrawals.length; ++i) {
+            uint256[] memory expectedTokens =
+                _calculateExpectedTokens(withdrawals[i].strategies, withdrawals[i].scaledShares);
+            staker.completeWithdrawalAsTokens(withdrawals[i]);
+            // FIXME: check_Withdrawal_AsTokens_State_AfterSlash(staker, operator, withdrawals[i], allocateParams, slashingParams, expectedTokens);
```
8sunyuan, line 112: Whats broken in this check atm?

0xClandestine: currently failing for staker shares, which should also be conditionally adjusted

src/test/integration/tests/Deposit_Delegate_Allocate.t.sol
0xClandestine: yes

src/test/integration/tests/Deposit_Delegate_Allocate.t.sol
```diff
@@ -4,21 +4,24 @@ pragma solidity ^0.8.27;
 import "src/test/integration/IntegrationChecks.t.sol";
 import "src/test/integration/users/User.t.sol";
 
-contract Integration_Deposit_Delegate_Allocate is IntegrationCheckUtils {
-    function testFuzz_deposit_delegate_allocate(uint24 _random) public {
+// TODO: move randomness from tests
+
+contract Integration_Deposit_Delegate_Allocate is IntegrationCheckUtils, IDelegationManagerTypes {
```
8sunyuan, line 9: Same thing here and the other test file, could we remove the IDelegationManagerTypes inheritance?

src/contracts/pods/EigenPod.sol
0xClandestine: Already done `testFuzz_withdrawRestakedBeaconChainETH_AmountGweiNotDivisibleByGwei` in `EigenPodUnit.t.sol`.

src/test/integration/IntegrationBase.t.sol
0xClandestine: will include in follow up pr

src/test/integration/tests/Deposit_Delegate_Allocate.t.sol
0xClandestine: will fix in follow up pr

src/test/integration/IntegrationBase.t.sol
0xClandestine: fixed in [d7a5d61](https://github.com/Layr-Labs/eigenlayer-contracts/pull/934/commits/d7a5d61437737a470dd4d22220e8a818d21c2ac1)

src/test/integration/IntegrationBase.t.sol
0xClandestine: fixed in [d7a5d61](https://github.com/Layr-Labs/eigenlayer-contracts/pull/934/commits/d7a5d61437737a470dd4d22220e8a818d21c2ac1)

src/test/integration/IntegrationBase.t.sol
0xClandestine: fixed in [d7a5d61](https://github.com/Layr-Labs/eigenlayer-contracts/pull/934/commits/d7a5d61437737a470dd4d22220e8a818d21c2ac1)

src/test/integration/tests/Deposit_Delegate_Allocate.t.sol
```diff
@@ -27,21 +30,161 @@ contract Integration_Deposit_Delegate_Allocate is IntegrationCheckUtils {
...
+
+        AllocateParams[] memory allocateParams = new AllocateParams[](operatorSets.length);
+
+        for (uint256 i; i < operatorSets.length; ++i) {
             uint256 len = allocationManager.getStrategiesInOperatorSet(operatorSets[i]).length;
-            operator.modifyAllocations(operatorSets[i], _randMagnitudes({ sum: 1 ether / uint64(operatorSets.length), len: len }));
-            (IStrategy[] memory strategiesToSlash, uint[] memory wadsToSlash) = _randStrategiesAndWadsToSlash(operatorSets[i]);
-            avs.slashOperator(operator, operatorSets[i].id, strategiesToSlash, wadsToSlash);
+            allocateParams[i] = operator.modifyAllocations(
+                operatorSets[i], _randMagnitudes({sum: 1 ether / uint64(operatorSets.length), len: len})
```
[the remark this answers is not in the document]
0xClandestine: fixed in [d7a5d61](https://github.com/Layr-Labs/eigenlayer-contracts/pull/934/commits/d7a5d61437737a470dd4d22220e8a818d21c2ac1)

src/test/integration/tests/Deposit_Delegate_Allocate.t.sol
0xClandestine: fixed in [d7a5d61](https://github.com/Layr-Labs/eigenlayer-contracts/pull/934/commits/d7a5d61437737a470dd4d22220e8a818d21c2ac1)

ypatil12:
Closing in favor of https://github.com/Layr-Labs/eigenlayer-contracts/pull/945

PR #935, opened 2024-12-06
feat: constrain maximum strategies per operator set

0xClandestine:


src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -3100,6 +3102,16 @@ contract AllocationManagerUnitTests_addStrategiesToOperatorSet is AllocationMana
         allocationManager.addStrategiesToOperatorSet(defaultAVS, defaultOperatorSet.id, defaultStrategies);
     }
 
+    function test_addStrategiesToOperatorSet_MaxStrategiesExceeded() public {
```
8sunyuan, line 3105: Could we modify this test to have fuzzed random number of existing strategies and then adding an amount that exceeds causes a revert?
Looks correct to me but good to have coverage on the line
`_operatorSetStrategies[operatorSetKey].length() + strategies.length <= MAX_OPERATOR_SET_STRATEGY_LIST_LENGTH`

0xClandestine: fixed in [a729b40](https://github.com/Layr-Labs/eigenlayer-contracts/pull/935/commits/a729b40fd6243e0fa3867a6c9b520a3f14ae169b)

8sunyuan (approved):

PR #936, opened 2024-12-09
test: delegationManager unit test refactor

8sunyuan:
### EigenPodManager refactor:
Refactored the calculation done for negative delta decreases for EigenPods.
This was due to failing fuzzed tests for `decreaseDelegatedShares` where the amount to decrement from operatorShares was exceeding the actual shares in storage and was underflowing. This change is logically equivalent and has less introduced error from using the `wadSlashed` paramater. The beaconChainSlashingFactor is effectively the "source of truth" for how much is slashed and we use that to calculate the operatorShares decreased. This is similar to how we calculate operatorShares to burn using prevMaxMagnitude and newMaxMagnitude. 
One thing to note is how when we calculate the newBeaconChainSlashingFactor, we are doing a mulDiv which rounds down. Meaning that eigenPod shares are rounded down, this is intentional.
Note: `event BeaconChainSlashingFactorDecreased` has been modified

### DelegationManager Test Changes:
- Using DelegationManagerHarness to access internal storage (queuedSlashableShares, depositScalingFactors, slashingFactors, etc)
- Rather than using the SlashingLib functions, copied the underlying functions into internal helpers to ensure any changes in the library are being checked against properly in tests. Examples include:
  - `_calcWithdrawableShares`
  - `_calcDepositScalingFactor`
- Event helper functions which check every emitted event and added to tests missing event checks
- Helper assertions such as:
  - `_assertDeposit`: checks for updated depositShares, withdrawableShares, operatorShares, depositScalingFactor based on passed in before values
  - `_assertWithdrawal`: checks for decremented depositShares, operatorShares,
  - `_assertCompletedWithdrawal`: checks correctly withdrawn shares based on slashed values and withdrawal roots completed and Withdrawal struct not in storage
  - `_assertQueuedWithdrawalExists`: checks withdrawal root set to true and Withdrawal struct exists in storage
- Using Random.sol library
- @nadir-akhtar Test for beaconChain and AVS slashing and asserting that performing the same slash in either order results in the same withdrawableShares for a staker. see `testFuzz_beaconSlashAndAVSSlash`
- Added tests:
  - `testFuzz_increaseDelegatedShares_beaconChainShares`
  - `testFuzz_increaseDelegatedShares_slashedOperatorAndBeaconChainShares`
  - `testFuzz_Revert_decreaseDelegatedShares_invalidCaller`
  - `testFuzz_decreaseDelegatedShares_noop`
  - `testFuzz_decreaseDelegatedShares_nonSlashedOperator`
  - `testFuzz_decreaseDelegatedShares_slashedOperator` (this was the test that was failing from rounding underflow)
  - completeQueuedWithdrawals tests, fuzzing receiveAsTokens and fuzzed magnitudes

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -342,8 +342,7 @@ contract DelegationManager is
             sharesToDecrease: sharesToDecrement
         });
 
-        /// TODO: implement EPM.burnShares interface. Likely requires more complex interface than just shares
-        /// so not adding a burnShares method in IShareManager
+        // NOTE: native ETH shares will be burned by a different mechanism in a future release
         if (strategy != beaconChainETHStrategy) {
             strategyManager.burnShares(strategy, sharesToBurn);
             emit OperatorSharesBurned(operator, strategy, sharesToBurn);
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -296,7 +296,7 @@ contract DelegationManager is
         }
         address operator = delegatedTo[staker];
 
-        // Calculate the shares to remove from the operator by calculating difference in shares 
+        // Calculate the shares to remove from the operator by calculating difference in shares
         // from the newly updated beaconChainSlashingFactor
         uint64 maxMagnitude = allocationManager.getMaxMagnitude(operator, beaconChainETHStrategy);
         DepositScalingFactor memory dsf = _depositScalingFactor[staker][beaconChainETHStrategy];
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -289,27 +289,21 @@ contract DelegationManager is
     function decreaseDelegatedShares(
         address staker,
         uint256 curDepositShares,
-        uint64 prevBeaconChainSlashingFactor,
-        uint256 wadSlashed
+        uint64 beaconChainSlashingFactorDecrease
     ) external onlyEigenPodManager {
         if (!isDelegated(staker)) {
             return;
         }
-
         address operator = delegatedTo[staker];
 
-        // Calculate the previous slashing factor. This mirrors the calculation done
-        // in _getSlashingFactor
+        // Calculate the shares to remove from the operator by calculating difference in shares 
+        // from the newly updated beaconChainSlashingFactor
         uint64 maxMagnitude = allocationManager.getMaxMagnitude(operator, beaconChainETHStrategy);
-        uint256 prevSlashingFactor = maxMagnitude.mulWad(prevBeaconChainSlashingFactor);
-
-        // Calculate the shares to remove from the operator by calculating the previously-withdrawable
-        // shares and applying slashing
         DepositScalingFactor memory dsf = _depositScalingFactor[staker][beaconChainETHStrategy];
         uint256 sharesToRemove = dsf.calcWithdrawable({
             depositShares: curDepositShares,
-            slashingFactor: prevSlashingFactor
-        }).mulWad(wadSlashed);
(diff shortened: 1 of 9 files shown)


src/contracts/pods/EigenPodManager.sol
```diff
@@ -273,31 +274,18 @@ contract EigenPodManager is
...
-        uint256 curDepositShares = uint256(podOwnerDepositShares[podOwner]);
-        return (curDepositShares, prevBeaconSlashingFactor, wadSlashed);
+        // newBeaconSlashingFactor is less than prevBeaconSlashingFactor because
+        // newRestakedBalanceWei < prevRestakedBalanceWei
+        uint64 newBeaconSlashingFactor =
+            uint64(prevBeaconSlashingFactor.mulDiv(newRestakedBalanceWei, prevRestakedBalanceWei));
+        uint64 beaconChainSlashingFactorDecrease = prevBeaconSlashingFactor - newBeaconSlashingFactor;
+        _beaconChainSlashingFactor[podOwner] =
+            BeaconChainSlashingFactor({slashingFactor: newBeaconSlashingFactor, isSet: true});
+        emit BeaconChainSlashingFactorDecreased(podOwner, beaconChainSlashingFactorDecrease, newBeaconSlashingFactor);
```
gpsanant, line 287: this second parameter should be the `prevBeaconSlashingFactor`
```suggestion
        emit BeaconChainSlashingFactorDecreased(podOwner, prevBeaconSlashingFactor, newBeaconSlashingFactor);
```

8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/936/commits/225e707681d42b55b9233cd2614b7440dc0cd09f

src/contracts/core/DelegationManager.sol
```diff
@@ -289,27 +289,21 @@ contract DelegationManager is
...
-
-        // Calculate the shares to remove from the operator by calculating the previously-withdrawable
-        // shares and applying slashing
         DepositScalingFactor memory dsf = _depositScalingFactor[staker][beaconChainETHStrategy];
         uint256 sharesToRemove = dsf.calcWithdrawable({
             depositShares: curDepositShares,
-            slashingFactor: prevSlashingFactor
-        }).mulWad(wadSlashed);
+            slashingFactor: maxMagnitude.mulWad(beaconChainSlashingFactorDecrease)
+        });
```
ypatil12, line 306: So IIUC, this is equivalent to the old calculation.

Old calculation is `depositShares*maxMag*prevFactor*wadSlashed`
New Calculation is `depositShares*maxMag*bsfDecrease`

Must be the case that `prevFactor*wadSlashed = bsfDecrease`

Splitting these up,

`pF*WS = pF - nF`, where pF is prevBeaconChainScalingFactor, WS is wadSlashed, and nf Is newBeaconChainScalingFactor.

`WS = -nF/pF`

ypatil12 (approved):
tests LGTM, appreciate the fully slashed one

src/contracts/core/DelegationManager.sol
8sunyuan: Old calculation is actually `depositShares*dsf*maxMag*wadSlashed`
because its using depositScalingFactor as well and prevSlashingFactor in the old code was just the maxMagnitude itself. 
New calculation is now `depositShares*dsf*maxMag*bsfDecrease`

Implying wadSlashed and bsfDecrease are equivalent which intuitively makes sense.

8sunyuan: Update: @ypatil12 is correct here actually, the prevBeaconChainSlashingFactor is used in the old calculation and
`prevFactor*wadSlashed = bsfDecrease` is true

PR #939, opened 2024-12-10
feat: update rc for UAM

ypatil12:
Updates the RewardsCoordiator to be UAM-compliant:
- Removes UAM from the original AVS rewards submission (keeping interface the same here, and we want users to eventually move to operatorSet-based AVSs)
- Adds UAM to: setting PI split, setting AVS split, submitting operator directed rewards

### DIFF
--- a/src/contracts/core/RewardsCoordinator.sol
+++ b/src/contracts/core/RewardsCoordinator.sol
@@ -96,9 +96,8 @@ contract RewardsCoordinator is
 
     /// @inheritdoc IRewardsCoordinator
     function createAVSRewardsSubmission(
-        address avs,
         RewardsSubmission[] calldata rewardsSubmissions
-    ) external onlyWhenNotPaused(PAUSED_AVS_REWARDS_SUBMISSION) checkCanCall(avs) nonReentrant {
+    ) external onlyWhenNotPaused(PAUSED_AVS_REWARDS_SUBMISSION) nonReentrant {
         for (uint256 i = 0; i < rewardsSubmissions.length; i++) {
             RewardsSubmission calldata rewardsSubmission = rewardsSubmissions[i];
             uint256 nonce = submissionNonce[msg.sender];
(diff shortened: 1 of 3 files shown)


src/contracts/core/RewardsCoordinator.sol
```diff
@@ -158,9 +157,7 @@ contract RewardsCoordinator is
     function createOperatorDirectedAVSRewardsSubmission(
         address avs,
         OperatorDirectedRewardsSubmission[] calldata operatorDirectedRewardsSubmissions
-    ) external onlyWhenNotPaused(PAUSED_OPERATOR_DIRECTED_AVS_REWARDS_SUBMISSION) nonReentrant {
-        require(msg.sender == avs, UnauthorizedCaller());
-
+    ) external onlyWhenNotPaused(PAUSED_OPERATOR_DIRECTED_AVS_REWARDS_SUBMISSION) checkCanCall(avs) nonReentrant {
```
0xrajath, line 160: Where is the `checkCanCall` modifier initialized?

0xrajath: Nvm it's in PermissionControllerMixin.

0xrajath (approved):

src/test/unit/RewardsCoordinatorUnit.t.sol
```diff
@@ -832,7 +883,7 @@ contract RewardsCoordinatorUnitTests_createAVSRewardsSubmission is RewardsCoordi
         // 3. Call createAVSRewardsSubmission() with expected revert
         cheats.prank(avs);
         cheats.expectRevert(IRewardsCoordinatorErrors.AmountExceedsMax.selector);
```
0xClandestine, line 885: nit: could inherit `IRewardsCoordinatorErrors`

0xClandestine (commented):
LGTM

8sunyuan:
While we're here, can we just remove     `function createRewardsForAllSubmission`, the old interface meant for foundation that ended up paused and unused?

ypatil12:
> While we're here, can we just remove  function createRewardsForAllSubmission, the old interface meant for foundation that ended up paused and unused?

I'd rather not, just given that we don't know if foundation will need it. I say let's deprecate in the future Rewards slashing code, when we have time to confirm that foundation will 100% never use it

src/test/unit/RewardsCoordinatorUnit.t.sol
ypatil12: Updated: https://github.com/Layr-Labs/eigenlayer-contracts/pull/939/commits/42812f581b8cb99cd96d71832574b6b0501cdb91

PR #951, opened 2024-12-12
feat: add back minWithdrawalDelayBlocks for introspection

ypatil12:
This is so we don't break the BLSSigChecker on preprod/testnet

### DIFF
--- a/src/contracts/interfaces/IDelegationManager.sol
+++ b/src/contracts/interfaces/IDelegationManager.sol
@@ -526,6 +526,9 @@ interface IDelegationManager is ISignatureUtils, IDelegationManagerErrors, IDele
     /// @notice return address of the beaconChainETHStrategy
     function beaconChainETHStrategy() external view returns (IStrategy);
 
+    /// @notice Backwards-compatible interface to return the `MIN_WITHDRAWAL_DELAY_BLOCKS` value
+    function minWithdrawalDelayBlocks() external view returns (uint32);
+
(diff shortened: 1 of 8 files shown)


stevennevins (approved):

wadealexc (approved):

src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -526,6 +526,10 @@ interface IDelegationManager is ISignatureUtils, IDelegationManagerErrors, IDele
     /// @notice return address of the beaconChainETHStrategy
     function beaconChainETHStrategy() external view returns (IStrategy);
 
+    /// @notice Backwards-compatible interface to return the `MIN_WITHDRAWAL_DELAY_BLOCKS` value
+    /// @dev Previous value in storage was deprecated. See `__deprecated_minWithdrawalDelayBlocks`
+    function minWithdrawalDelayBlocks() external view returns (uint32);
+
```
wadealexc, line 532: We need to get rid of the all-caps version of the method, or we'll have to support both permanently.

We can't afford the extra bytecode.

wadealexc (changes requested):
See above

src/contracts/interfaces/IDelegationManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/951/commits/48236927e0bfb33c73d063ad02654ddc532e02ae

wadealexc (approved):

PR #958, opened 2024-12-15
refactor: `foundry.toml`

0xClandestine:
### Changes:
- `forge fmt` only formats `src/contracts`.
- `FOUNDRY_PROFILE=test forge fmt` formats `script` and `src/test`.
- `make fmt` formats everything.
- `FOUNDRY_PROFILE=lite` added for faster fuzzing.
- Compile warnings now prevent an otherwise successful `forge b`.
- Imports are now sorted alphabetically.
- Format CI now checks `src/test` and `script`.

### NOTE
Checkout `foundry.toml` for suggested formatter changes.

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -1,46 +1,69 @@
 [profile.default]
-src = 'src'
-out = 'out'
-libs = ['lib']
-fs_permissions = [{ access = "read-write", path = "./"}]
-show_progress=true
-gas_reports = ["*"]
-# ignore upgrade testing in scripts by default
-no_match_test = "queueUpgrade"
-no_match_path = "script/releases/**/*.sol"
-remappings = [
-    "@openzeppelin/=lib/openzeppelin-contracts-v4.9.0/",
-    "@openzeppelin-upgrades/=lib/openzeppelin-contracts-upgradeable-v4.9.0/",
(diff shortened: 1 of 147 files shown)


foundry.toml
```diff
@@ -1,46 +1,120 @@
...
+        "./script/**/*",
+        "./src/test/**/*"
+    ]
+
+[profile.test.fmt]
+    int_types = "short"
+    line_length = 140
+    ignore = [
+        "./src/contracts/**/*"
+    ]
```
0xClandestine, line 109: Throughout out tests, we typically opt for short uint/int types, and prefer longer line length.

wadealexc (changes requested):
`foundry.toml` is indeed messy rn, and could use a cleanup. I like your refactor generally; the file is easier to read. However, a few things I'd like to see:

1. This is a _huge_ diff. It's really hard to get a sense for what all of these new options and changes do because this PR changes 120 files! Can we try doing this gradually? i.e. start with a `foundry.toml` refactor for cleanliness and add one or two of your favorite config changes so we can assess things in chunks.
2. I'm not sure how I feel about running any sort of formatter on our tests. I'm willing to consider it, but IMO that should be its own, separate PR because it'll be a big diff.

PR #959, opened 2024-12-16
refactor: small cleanup

0xClandestine:
### Changes
- removes unused `ISignatureUtils` import for `IAllocationManager`.
- add missing `InputArrayLengthMismatch()` check to `slashOperator()`.
- fix `getQueuedWithdrawals` bug + test

### DIFF
--- a/src/test/unit/DelegationUnit.t.sol
+++ b/src/test/unit/DelegationUnit.t.sol
@@ -68,7 +68,6 @@ contract DelegationManagerUnitTests is EigenLayerUnitTestSetup, IDelegationManag
     // Helper to use in storage
     DepositScalingFactor dsf;
     uint256 stakerDSF;
-    uint256 slashingFactor;
 
     /// @notice mappings used to handle duplicate entries in fuzzed address array input
     mapping(address => uint256) public totalSharesForStrategyInArray;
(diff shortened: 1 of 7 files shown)


src/test/unit/DelegationUnit.t.sol
```diff
@@ -8274,3 +8273,56 @@ contract DelegationManagerUnitTests_Lifecycle is DelegationManagerUnitTests {
...
+            staker: defaultStaker,
+            withdrawer: defaultStaker,
+            strategies: strategies,
+            depositWithdrawalAmounts: depositShares
+        });
+
+        cheats.prank(defaultStaker);
+        delegationManager.queueWithdrawals(queuedWithdrawalParams);
+
+        // Get queued withdrawals.
```
ypatil12, line 8317: As a regression test, the error we had is when `i > j`. Therefore, we want the length of `totalQueued` to be greater than the number of strategies in a withdrawal. In this case its equal

ypatil12: This could be an additional test to the correctness one we have here

ypatil12 (commented):
Looks good, comment on the regression test

8sunyuan (approved):
LGTM

PR #962, opened 2024-12-17
fix: complete withdrawals

8sunyuan:
Decided on removing numToComplete interface and this PR also fixes some broken tests.

Also fixes a race condition bug where a staker's withdrawal is still slashable in the same block that the withdrawal is completable, leading to possible race conditions. The regression test `test_getQueuedWithdrawals_SlashAfterWithdrawalCompletion` on the view function found this and passes as a result of the fix.

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -242,28 +242,6 @@ contract DelegationManager is
     }
 
     /// @inheritdoc IDelegationManager
-    function completeQueuedWithdrawals(
-        IERC20[][] calldata tokens,
-        bool[] calldata receiveAsTokens,
-        uint256 numToComplete
-    ) external onlyWhenNotPaused(PAUSED_EXIT_WITHDRAWAL_QUEUE) nonReentrant {
-        EnumerableSet.Bytes32Set storage withdrawalRoots = _stakerQueuedWithdrawalRoots[msg.sender];
-        uint256 length = withdrawalRoots.length();
-        numToComplete = numToComplete > length ? length : numToComplete;
-
-        // Read withdrawals to complete. We use 2 seperate loops here because the second
-        // loop will remove elements by index from `withdrawalRoots`.
-        Withdrawal[] memory withdrawals = new Withdrawal[](numToComplete);
-        for (uint256 i; i < withdrawals.length; ++i) {
-            withdrawals[i] = queuedWithdrawals[withdrawalRoots.at(i)];
-        }
-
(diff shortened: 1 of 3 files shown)


src/test/unit/DelegationUnit.t.sol
```diff
@@ -6619,6 +6539,8 @@ contract DelegationManagerUnitTests_completeQueuedWithdrawal is DelegationManage
                 strategyMock, 
                 depositSharesPerWithdrawal
             );
+            withdrawal.nonce = nonce;
+            nonce += 1;
```
8sunyuan, line 6543: Note: this was needed because it was looping over creating queuedWithdrawalParams but each one was having a nonce of 0 from reading directly off the contract. I overwrite this value here since the expectation is that the withdrawals are queued in order with nonces from 0-3 inclusive

src/contracts/core/DelegationManager.sol
```diff
@@ -959,7 +937,24 @@ contract DelegationManager is
...
             shares[i] = new uint256[](withdrawals[i].strategies.length);
 
-            uint256[] memory slashingFactors = _getSlashingFactors(staker, operator, withdrawals[i].strategies);
+            uint32 completableBlock = withdrawals[i].startBlock + MIN_WITHDRAWAL_DELAY_BLOCKS;
+
+            uint256[] memory slashingFactors;
+            // if the completion block is in the future, simply read current slashing factors
+            // still possible however for the slashing factors to change before the withdrawal is completable
+            // and the shares withdrawn to be less
+            if (completableBlock > uint32(block.number)) {
```
0xClandestine, line 946: nit: could be a ternary (no biggie tho)

0xClandestine (commented):
one nit, ci is failing too

0xClandestine:
> one nit, ci is failing too

noticed it's also failing upstream, can fix if needed

src/contracts/core/DelegationManager.sol
```diff
@@ -959,7 +937,24 @@ contract DelegationManager is
...
+                slashingFactors =
+                    _getSlashingFactors({staker: staker, operator: operator, strategies: withdrawals[i].strategies});
+                // Read slashing factors at the completable block
```
ypatil12, line 949: Can we remove these from this PR

ypatil12 (changes requested):
Can we simplify this PR?

ypatil12:
Closing in favor of https://github.com/Layr-Labs/eigenlayer-contracts/pull/965

PR #964, opened 2024-12-18
feat: add share helpers

gpsanant:
added a view func for converting deposit shares from withdrawable shares

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -949,6 +949,21 @@ contract DelegationManager is
     }
 
     /// @inheritdoc IDelegationManager
+    function convertToDepositShares(address staker, IStrategy[] memory strategies, uint256[] memory shares) external view returns (uint256[] memory) {
+        // Get the slashing factors for the staker/operator/strategies
+        address operator = delegatedTo[staker];
+        uint256[] memory slashingFactors = _getSlashingFactors(staker, operator, strategies);
+
+        // Calculate the deposit shares based on the slashing factor
+        uint256[] memory depositShares = new uint256[](strategies.length);
+        for (uint256 i = 0; i < strategies.length; ++i) {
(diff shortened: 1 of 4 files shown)


src/contracts/core/DelegationManager.sol
```diff
@@ -970,6 +970,21 @@ contract DelegationManager is
         }
     }
 
+    /// @inheritdoc IDelegationManager
+    function convertToDepositShares(address staker, IStrategy[] memory strategies, uint256[] memory shares) external view returns (uint256[] memory) {
```
8sunyuan, line 974: nit: use withdrawable shares as input

src/test/unit/DelegationUnit.t.sol
```diff
@@ -8274,3 +8274,98 @@ contract DelegationManagerUnitTests_Lifecycle is DelegationManagerUnitTests {
...
+            // so this value is sane to input into `completeQueuedWithdrawals`
+            assertLe(
+                depositShares[i],
+                expectedDepositShares[i],
+                "deposit shares should be less than or equal to expected deposit shares"
+            );
+        }
+
+        // get the deposit shares
+        uint256[] memory oneThirdWithdrawableShares = new uint256[](strategies.length);
```
8sunyuan, line 8356: Not quite sure this is being done here in the helper

gpsanant: it's just to make sure the thing properly scales down as well when called with smaller values that the max amount of shares

src/contracts/core/DelegationManager.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-contracts/pull/964/commits/36e8a850306ad710f3ac7499e1baccbd49a81d8b

8sunyuan (approved):

ypatil12 (approved):

PR #965, opened 2024-12-18
fix: slashable window boundaries

8sunyuan:
Fixing slashable windows here by ensuring that from
[block.number, block.number + MIN_WITHDRAWAL_DELAY_BLOCKS] and
[block.number, block.number + DEALLOCATION_DELAY] are both the same window. 
Since we want these two values `MIN_WITHDRAWAL_DELAY_BLOCKS` and `DEALLOCATION_DELAY` to be configured to be the same in deployments, we make the following changes:

1. Upon withdrawal completion, we now have a slashableUntil block which is `withdrawal.startBlock + MIN_WITHDRAWAL_DELAY_BLOCKS` which is the last block the staker's withdrawal is slashable for. That means the first block they can complete withdraw is the block following hence the strict inequality
```
uint32 slashableUntil = withdrawal.startBlock + MIN_WITHDRAWAL_DELAY_BLOCKS;
require(slashableUntil < uint32(block.number), WithdrawalDelayNotElapsed());
```

2. For deallocations, we have effectBlock set as `block.number + DEALLOCATION_DELAY` currently but that is the block that the deallocation is no longer slashable and can be completed. We fix this to ensure the slashable window of `[block.number, block.number + DEALLOCATION_DELAY]` by making the `effectBlock = block.number + DEALLOCATION_DELAY + 1`.

3.  For `DelegationManager._getSlashedSharesInQueue` we want also ensure a slashable window of [block.number - MIN_WITHDRAWAL_DELAY_BLOCKS, block.number] including blocknumber `block.number - MIN_WITHDRAWAL_DELAY_BLOCKS`. So since a subtraction is done of `curCumulativeScaledShares - prevCumulativeScaledShares`, we make the snapshot.lookup of prevCumulativeScaledShares to be `-1` to include that block itself.


ypatil12:
Closing in favor of: https://github.com/Layr-Labs/eigenlayer-contracts/pull/965

src/contracts/core/AllocationManager.sol
```diff
@@ -446,7 +448,9 @@ contract AllocationManager is
     function _isOperatorSlashable(address operator, OperatorSet memory operatorSet) internal view returns (bool) {
         RegistrationStatus memory status = registrationStatus[operator][operatorSet.key()];
 
-        return status.registered || block.number < status.slashableUntil;
+        // slashableUntil returns the last block the operator is slashable in so we check for
+        // than or equal to
+        return status.registered || block.number <= status.slashableUntil;
```
gpsanant, line 453: should redefine or rename slashableUntil since `<=` means a strict inequality

gpsanant: nah this is ok

src/contracts/core/DelegationManager.sol
```diff
@@ -937,7 +942,24 @@ contract DelegationManager is
...
             shares[i] = new uint256[](withdrawals[i].strategies.length);
 
-            uint256[] memory slashingFactors = _getSlashingFactors(staker, operator, withdrawals[i].strategies);
+            uint32 slashableUntil = withdrawals[i].startBlock + MIN_WITHDRAWAL_DELAY_BLOCKS;
+
+            uint256[] memory slashingFactors;
+            // if the slashableUntil block is in the future, simply read current slashing factors
+            // still possible however for the slashing factors to change before the withdrawal is completable
+            // and the shares withdrawn to be less
+            if (slashableUntil > uint32(block.number)) {
```
gpsanant, line 951: this should be >=

alternatively, making it `uint32(block.number) > slashableUntil` and switching the inners of the conditionals is cleaner

src/contracts/core/AllocationManager.sol
```diff
@@ -200,7 +200,9 @@ contract AllocationManager is
                         // the deallocation delay. This magnitude remains slashable until then.
                         deallocationQueue[operator][strategy].pushBack(operatorSet.key());
 
-                        allocation.effectBlock = uint32(block.number) + DEALLOCATION_DELAY;
+                        // deallocations are slashable in the window [block.number, block.number + deallocationDelay]
+                        // therefore, the effectBlock is set to the block right after the slashable window
+                        allocation.effectBlock = uint32(block.number) + DEALLOCATION_DELAY + 1;
```
ypatil12, line 205: Note: every comparison to `effectBlock` is `< effectBlock`. There is some minor difference in naming (and resulting comparisons) because we do:

- < effectBlock
- <= slashableUntil 

Nomenclature does make sense here though

ypatil12, line 205: Regression tests:

1. Assert that a deallocation is not slashable 1 block after the dealloc delay, but is slashable right on it
2. Assert that a registration is not slashable 1 block after the registration but slashable right on it

ypatil12: I can add these tmrw as well

src/contracts/core/DelegationManager.sol
8sunyuan: `_getSlashingFactors` gets the current slashing factor, so the logic here is that if slashableUntil is at some future block, then we simply read the current slashing factors. Else we read it at the exact slashableUntil block

src/contracts/core/AllocationManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/965/commits/339873cea7ebc156471924a2eb3737e5ff7f6968

src/contracts/core/DelegationManager.sol
```diff
@@ -550,16 +550,18 @@ contract DelegationManager is
 
         uint256[] memory prevSlashingFactors;
         {
-            uint32 completableBlock = withdrawal.startBlock + MIN_WITHDRAWAL_DELAY_BLOCKS;
-            require(completableBlock <= uint32(block.number), WithdrawalDelayNotElapsed());
+            // slashableUntil is block inclusive so we need to check if the current block is strictly greater than the slashableUntil block
+            // meaning the withdrawal can be completed.
+            uint32 slashableUntil = withdrawal.startBlock + MIN_WITHDRAWAL_DELAY_BLOCKS;
+            require(slashableUntil < uint32(block.number), WithdrawalDelayNotElapsed());
```
ypatil12, line 556: `test_Revert_WhenWithdrawalDelayNotPassed` is a regression here, so we're good (already had this unit test-case covered)

ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/965/commits/6ffb6c399498cad8fc13223a95a28cf398fb3712

src/contracts/core/DelegationManager.sol
```diff
@@ -761,9 +763,12 @@ contract DelegationManager is
     ) internal view returns (uint256) {
         // Fetch the cumulative scaled shares sitting in the withdrawal queue both now and before
         // the withdrawal delay.
+        // NOTE: We want all the shares in the window [block.number - MIN_WITHDRAWAL_DELAY_BLOCKS, block.number]
+        // as this is all slashable and since prevprevCumulativeScaledShares is being subtracted from curCumulativeScaledShares
+        // we do a -1 on the block number to also include (block.number - MIN_WITHDRAWAL_DELAY_BLOCKS) as slashable.
         uint256 curCumulativeScaledShares = _cumulativeScaledSharesHistory[operator][strategy].latest();
         uint256 prevCumulativeScaledShares = _cumulativeScaledSharesHistory[operator][strategy].upperLookup({
-            key: uint32(block.number) - MIN_WITHDRAWAL_DELAY_BLOCKS
+            key: uint32(block.number) - MIN_WITHDRAWAL_DELAY_BLOCKS - 1
```
ypatil12, line 771: Burning test: https://github.com/Layr-Labs/eigenlayer-contracts/pull/965/commits/8d380fa8fac7d391f6e664071735e1e7996f9115

src/contracts/core/DelegationManager.sol
ypatil12: Its clearer if we go >=, but the underlying logic is still the same

8sunyuan: I get now what you guys are saying now and agreed.
Updated now https://github.com/Layr-Labs/eigenlayer-contracts/pull/965/commits/e6c66cd4f62125f9b5058dcf59ea686d279ccc75

ypatil12 (approved):

src/contracts/core/AllocationManager.sol
```diff
@@ -446,7 +448,9 @@ contract AllocationManager is
     function _isOperatorSlashable(address operator, OperatorSet memory operatorSet) internal view returns (bool) {
         RegistrationStatus memory status = registrationStatus[operator][operatorSet.key()];
 
-        return status.registered || block.number < status.slashableUntil;
+        // slashableUntil returns the last block the operator is slashable in so we check for
+        // than or equal to
```
wadealexc, line 452: comment typo, "less than or equal to"

src/contracts/core/DelegationManager.sol
```diff
@@ -761,9 +763,12 @@ contract DelegationManager is
     ) internal view returns (uint256) {
         // Fetch the cumulative scaled shares sitting in the withdrawal queue both now and before
         // the withdrawal delay.
+        // NOTE: We want all the shares in the window [block.number - MIN_WITHDRAWAL_DELAY_BLOCKS, block.number]
+        // as this is all slashable and since prevprevCumulativeScaledShares is being subtracted from curCumulativeScaledShares
```
wadealexc, line 767: `prevprevCumulativeScaledShares` typo

8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/965/commits/4d27ea0980b1670d19d3f6fbea6f03972dfced12

src/contracts/core/AllocationManager.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/965/commits/4d27ea0980b1670d19d3f6fbea6f03972dfced12

PR #968, opened 2024-12-18
feat: remove max strategies list

ypatil12:
We do not need this parameter any more since on `slashOperator` we don't loop through the list of strats in an opSet


0xClandestine (approved):
Looks good.

0xClandestine:
> Looks good.

Should storage layout CI be failing?

wadealexc (approved):

PR #970, opened 2024-12-18
refactor: `foundry.toml` and CI

0xClandestine:
### Changes
- `foundry.toml` cleanup
- `forge fmt` no longer formats tests
- added `ci` profile (disables optimizer)
- added `intense` profile (continuous fuzzing)


0xClandestine:
CI renaming will require somebody with privs to adjust "required" workflows to the new names.

PR #976, opened 2024-12-19
feat: permission controller intents

0xClandestine:


0xClandestine:
EIP-7702 can provide the same functionality at a lower risk to the protocol.

PR #982, opened 2025-01-02
refactor: final slashing cleanup

wadealexc:
* clean up and update variable names / comments
* simplify `undelegate`: I removed the 0 address check in `undelegate` because it's implied by the `isDelegated` check, and i condensed the `msg.sender != staker` logic. 
* removed check for deposit shares in `completeQueuedWithdrawal`. each `shareManager` has its own version of this check, so we don't need to do this in the DM
* updated `completeQueuedWithdrawal` to use checks-effects-interactions.
* deprecated the `QueuedWithdrawalParams.withdrawer` field. The DM now completely ignores this field, rather than reverting.
* simplify `redelegate` from 17 LOC to 2
* add public getter for a staker's queued withdrawal roots (`getQueuedWithdrawalRoots`)


src/contracts/core/DelegationManager.sol
```diff
@@ -927,6 +902,13 @@ contract DelegationManager is
         return (strategies, shares);
     }
 
+    /// @inheritdoc IDelegationManager
+    function getQueuedWithdrawal(
```
ypatil12, line 906: Any reason to not use the public getter on the mapping? Will be another breaking change to communicate

ypatil12 (commented):
LGTM. Had one quick question

src/contracts/core/DelegationManager.sol
wadealexc: Yes, because the implicit public getter will return a tuple rather than a struct

implicit public getters on mappings that return structs are super lame for this reason :grimacing:

ypatil12 (approved):

PR #987, opened 2025-01-03
fix: rewards v2 audit fixes

0xrajath:
# Motivation

As part of the Rewards v2 SigmaPrime audit, we need to tackle issue `ELSC2-07: Operators Can Lock In Current Default Split By Front-Running Changes`

# Modifications

* Moved `split <= ONE_HUNDRED_IN_BIPS` logic into common internal function : `_setOperatorSplit`
* Updated `_setOperatorSplit` and `_getOperatorSplit` for edge case of time after first split initialization, but before activation. 
* Updated tests.

# Result

Rewards v2 audit fixes.

### DIFF
--- a/src/contracts/core/RewardsCoordinator.sol
+++ b/src/contracts/core/RewardsCoordinator.sol
@@ -301,7 +301,6 @@ contract RewardsCoordinator is
         uint16 split
     ) external onlyWhenNotPaused(PAUSED_OPERATOR_AVS_SPLIT) {
         require(msg.sender == operator, "RewardsCoordinator.setOperatorAVSSplit: caller is not the operator");
-        require(split <= ONE_HUNDRED_IN_BIPS, "RewardsCoordinator.setOperatorAVSSplit: split must be <= 10000 bips");
 
         uint32 activatedAt = uint32(block.timestamp) + activationDelay;
         uint16 oldSplit = _getOperatorSplit(operatorAVSSplitBips[operator][avs]);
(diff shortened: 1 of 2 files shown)


src/contracts/core/RewardsCoordinator.sol
```diff
@@ -395,13 +393,14 @@ contract RewardsCoordinator is
...
+        require(split <= ONE_HUNDRED_IN_BIPS, "RewardsCoordinator._setOperatorSplit: split must be <= 10000 bips");
         require(
             block.timestamp > operatorSplit.activatedAt,
             "RewardsCoordinator._setOperatorSplit: earlier split not activated yet"
         );
         if (operatorSplit.activatedAt == 0) {
-            // If the operator split has not been initialized yet, set the old split to the default split.
-            operatorSplit.oldSplitBips = defaultOperatorSplitBips;
+            // If the operator split has not been initialized yet, set the old split to `type(uint16).max` as a flag.
+            operatorSplit.oldSplitBips = type(uint16).max;
```
0xrajath, line 403: Decided to go with using `type(uint16).max` instead since the Bips variable will never have that value. And also allows us to proceed without using any additional storage.

nadir-akhtar (approved):

src/test/unit/RewardsCoordinatorUnit.t.sol
```diff
@@ -595,14 +609,28 @@ contract RewardsCoordinatorUnitTests_setOperatorPISplit is RewardsCoordinatorUni
...
+        // Change default split to check if it is returned before activation
+        rewardsCoordinator.setDefaultOperatorSplit(5000);
+        // Check that the split returns the default split before activation for the first time.
+        assertEq(
+            rewardsCoordinator.defaultOperatorSplitBips(),
+            rewardsCoordinator.getOperatorPISplit(operator),
+            "Operator split is not Default split before Activation for first time"
+        );
+
         cheats.warp(activatedAt);
```
ypatil12, line 634: Can we also add a regression test for the scenario described in audit report?

0xrajath: Yup it's already there within the `testFuzz_setOperatorAVSSplitFirstTime` and  `testFuzz_setOperatorPISplitFirstTime` tests.

0xrajath: https://github.com/Layr-Labs/eigenlayer-contracts/pull/987/files#diff-1ddf2647b6d1ee196f902307ebcac00fb15be8dd0c02acd2c89ad686cc94265eR476-R485

ypatil12 (approved):

PR #989, opened 2025-01-03
feat: async burning of shares

8sunyuan:
### Description
This PR removes synchronous burning and adding a separate permissionless function that can clear `burnableShares` for all the accumulated slashed shares for a StrategyManager Strategy. 
This implies that when AVSs slash by calling `AllocationManager.slashOperator`, the call stack no longer has the underlying ERC20 token transfer included and we avoid some of the edge cases revolving around the previous try/catch design.

Additionally:
~Removes the function `SlashingLib.scaleSharesForBurning` which is used for calculating the amount of shares to slash in the withdrawal queue. This was removed as its logic is equivalent to `calcSlashedAmount` and cleans up some of the code.~ This is actually not correct, operatorShares and scaledShares are not used equivalently because operatorShares decreases over time upon each slash so do a mulWad(maxMagnitude) is not equivalent to calculating the amount slashed. 
TL;DR the original implementation should remain as-is

### DIFF
--- a/src/contracts/core/StrategyManager.sol
+++ b/src/contracts/core/StrategyManager.sol
@@ -141,7 +141,18 @@ contract StrategyManager is
     }
 
     /// @inheritdoc IStrategyManager
-    function burnShares(IStrategy strategy, uint256 sharesToBurn) external onlyDelegationManager {
+    function increaseBurnableShares(IStrategy strategy, uint256 addedSharesToBurn) external onlyDelegationManager {
+        // burning shares is functionally the same as withdrawing but with different destination address
+        burnableShares[strategy] += addedSharesToBurn;
+        emit BurnableSharesIncreased(strategy, addedSharesToBurn);
+    }
+
+    /// @inheritdoc IStrategyManager
+    function burnShares(IStrategy strategy, uint256 sharesToBurn) external nonReentrant {
+        require(strategyIsWhitelistedForDeposit[strategy], StrategyNotWhitelisted());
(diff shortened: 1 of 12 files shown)


src/contracts/core/StrategyManager.sol
```diff
@@ -141,7 +141,18 @@ contract StrategyManager is
...
+        burnableShares[strategy] += addedSharesToBurn;
+        emit BurnableSharesIncreased(strategy, addedSharesToBurn);
+    }
+
+    /// @inheritdoc IStrategyManager
+    function burnShares(IStrategy strategy, uint256 sharesToBurn) external nonReentrant {
+        require(strategyIsWhitelistedForDeposit[strategy], StrategyNotWhitelisted());
+        require(sharesToBurn <= burnableShares[strategy], BurnSharesAmountTooHigh());
+        burnableShares[strategy] -= sharesToBurn;
+        emit BurnableSharesDecreased(strategy, sharesToBurn);
```
wadealexc: Shouldn't this just burn all available burnable shares?

8sunyuan: I was uncertain whether or not there would be liability issues so I was unsure about this too

wadealexc: I think more decision making ability typically correlates with more liability so my instinct would be to make it all or nothing

wadealexc:
Looks like rebasing went awry? The `docs/release` folder should not be in this PR

8sunyuan:
Rebasing is off, closing this in favor of https://github.com/Layr-Labs/eigenlayer-contracts/pull/1001

PR #1001, opened 2025-01-06
refactor: async burning

8sunyuan:
### Description
This PR removes synchronous burning and adding a separate permissionless function that can clear `burnableShares` for all the accumulated slashed shares for a StrategyManager Strategy. 
This implies that when AVSs slash by calling `AllocationManager.slashOperator`, the call stack no longer has the underlying ERC20 token transfer included and we avoid some of the edge cases revolving around the previous try/catch design.

### Changes
- Added `burnableShares` storage to StrategyManager and two separate functions `increaseBurnableShares`, `burnShares`
- Renamed `burnOperatorShares` -> `slashOperatorShares` since burning no longer occurs within the same transaction
- Interfaces, events
- `DelegationUnit.t.sol`: renaming to `slashOperatorShares`


wadealexc (approved):
LGTM. If we want to add in some EPM logic now's the time :)

ypatil12 (approved):
LGTM

PR #1004, opened 2025-01-07
docs: finish delegation manager docs

wadealexc:


docs/core/DelegationManager.md
8sunyuan, line 1: registerAsOperator is missing `allocationDelay` in the Effects section and needs to note that its actually stored in the ALM.

8sunyuan: Ah I guess not needed since it notes it in the params and its not stored in the DelegationManager contract storage

docs/core/DelegationManager.md
```diff
@@ -245,6 +239,8 @@ Of these mappings, only `pendingWithdrawals` and `cumulativeWithdrawalsQueued`
```
8sunyuan, line 239: "Of these mappings, only `pendingWithdrawals` and `cumulativeWithdrawalsQueued`"

Incomplete sentence. I think what is being meant to say is only these storage mappings are used in legacy withdrawals.

docs/core/DelegationManager.md
```diff
@@ -245,6 +239,8 @@ Of these mappings, only `pendingWithdrawals` and `cumulativeWithdrawalsQueued`
 
 #### Slashing Factors and Scaling Shares
 
+_See the [Shares Accounting](./accounting/SharesAccounting.md) doc for a more thorough explanation with examples._
+
 Throughout the `DelegationManager`, a staker's _deposit shares_ can be converted into their current _withdrawable shares_ by applying two factors: the _slashing factor_ and the _deposit scaling factor_. These two values are scaling factors that act as numerators or denominators when scaling shares. By default, these values start at `1 WAD` (`1e18`). A fundamental constraint of the system is that both _slashing factors_ and _deposit scaling factors_ are monotonically decreasing.
```
8sunyuan, line 244: Oh this is incorrect, _deposit scaling factors_ are actually montonically decreasing. I think this was mistaken for maxMagnitude

8sunyuan: > These two values are scaling factors that act as numerators or denominators when scaling shares.

We can remove the denominator part since scaledShares is refactored now.

docs/core/DelegationManager.md
wadealexc: oh ha, mb i forgot to finish this section. thanks for pointing that out

8sunyuan (approved):

PR #1012, opened 2025-01-10
test: self delegate after undelegation

0xClandestine:


src/test/integration/tests/Deposit_Delegate_Allocate.t.sol
```diff
@@ -132,6 +132,8 @@ contract Integration_Deposit_Delegate_Allocate is IntegrationCheckUtils {
         IDelegationManagerTypes.Withdrawal[] memory withdrawals = staker.undelegate();
```
ypatil12, line 132: The test name should be updated? Ideally we create a separate test because it's now a different scenario altogether

ypatil12 (commented):
Looks good, one organizational comment

PR #1013, opened 2025-01-10
fix: remove token param from Deposit event and related APIs

bowenli86:
fix: remove token param from Deposit event and related APIs, to remove redundancy and as a fix to the reported bug of token param is incorrectly populated


ypatil12 (approved):
LGTM, needs 1 more approval

8sunyuan:
I'm assuming no impact to EigenStrategy since that can either take a deposit of EIGEN or bEIGEN

8sunyuan (approved):

bowenli86:
> > I'm assuming no impact to EigenStrategy since that can either take a deposit of EIGEN or bEIGEN. If so then lgtm
> 
> We should be good here in terms of mechanics.
> 
> On deposits -> `SM` calls `strategyBase`, no interaction with DM or relevant interfaces removed.
> 
> On withdrawals via shares, we simply add shares back, bEIGEN strategy doesn't care here.
> 
> There are two drawbacks though:
> 
> 1. You cannot answer the question: are people depositing EIGEN or bEIGEN into the eigen strategy
> 2. In a multi-token strat world, we would have re-add the token parameter

chatted w/ Yash

EIGEN<>bEIGEN conversion is better tracked separately b/c there are multiple entry points to convert them (Deposit, Withdraw, or just contract calls), and we can associate transaction ids to the source entry points. We'd either 1/ add Wrapped/Unwrapped events to `Eigen.sol` (tbd w/ jeff) or 2/ use Transfer event in _mint/_burn which tho has too many transfers and is noisy

we don't need to worry about multi-token strat for now

bowenli86:
synced w/ Jeff, we'll add Wrapped/Unwrapped events to Eigen.sol as part of an upcoming ELIP

wadealexc (changes requested):
Needs `make bindings`

and updates to documentation:
* [EPM.addShares](https://github.com/Layr-Labs/eigenlayer-contracts/blob/removetoken/docs/core/EigenPodManager.md#addshares)
* [SM.addShares](https://github.com/Layr-Labs/eigenlayer-contracts/blob/removetoken/docs/core/StrategyManager.md#addshares)

wadealexc (approved):

PR #1016, opened 2025-01-11
fix: storage gaps

0xClandestine:
adds `make update-gaps`


src/contracts/strategies/StrategyFactoryStorage.sol
```diff
@@ -29,5 +29,5 @@ abstract contract StrategyFactoryStorage is IStrategyFactory {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[48] private __gap;
```
ypatil12, line 32: One thing I noticed is by updating the StrategyFactory 's gap from 48 -> 47, the layout in the storage report won't match what is actually deployed.
For example, the _initialized var will continue to be at slot 51 (instead of at slot 50 like the updated storage report would imply). This seems fine, but might be worth calling out somewhere

ypatil12, line 32: Can we also update the gaps as part of this PR for these two contracts?

ypatil12 (commented):
Looks good, left 2 comments

src/contracts/strategies/StrategyFactoryStorage.sol
0xClandestine: They're up to date.

ypatil12 (approved):
Approving for now, but would would be good to document discrepancy between true layout and that in the storage gap layout

ypatil12 (changes requested):
Given that CI storage checking is failing, let's do a fork test to validate that storage isn't being overwritten with this change?

PR #1020, opened 2025-01-13
feat: add EigenWrapped and EigenUnwrapped events

bowenli86:
to track wrap/unwrap actions

discussed with Jeff, this is technically small and we would want to include it into a bigger Eigen contract cleanup. Just put the PR out first for now


ypatil12:
Needs different target branch

PR #1024, opened 2025-01-13
fix: remove deprecated hardhat configs

bowenli86:
remove hardhat which has long been deprecated and not used in the repo


bowenli86:
> remove hardhat which has long been deprecated and not used in the repo

@0xClandestine thanks, moved to https://github.com/Layr-Labs/eigenlayer-contracts/pull/1090

PR #1025, opened 2025-01-14
feat: require avs register metadata in allocation manager

bowenli86:
require avs register metadata in allocation manager before they can create operatorset

### DIFF
--- a/src/contracts/core/AllocationManager.sol
+++ b/src/contracts/core/AllocationManager.sol
@@ -338,7 +338,7 @@ contract AllocationManager is
     /// @inheritdoc IAllocationManager
     function createOperatorSets(address avs, CreateSetParams[] calldata params) external checkCanCall(avs) {
         // Check that the AVS exists and has registered metadata
-        require(_avsRegisteredMetadata[avs], InvalidAVSWithNoMetadataRegistered());
+        require(_avsRegisteredMetadata[avs], NonexistentAVSMetadata());
 
         for (uint256 i = 0; i < params.length; i++) {
             OperatorSet memory operatorSet = OperatorSet(avs, params[i].operatorSetId);
--- a/src/contracts/core/AllocationManager.sol
+++ b/src/contracts/core/AllocationManager.sol
@@ -338,7 +338,7 @@ contract AllocationManager is
     /// @inheritdoc IAllocationManager
     function createOperatorSets(address avs, CreateSetParams[] calldata params) external checkCanCall(avs) {
         // Check that the AVS exists and has registered metadata
-        require(_avsRegisteredMetadata[params.avs], InvalidAVSWithNoMetadataRegistered());
+        require(_avsRegisteredMetadata[avs], InvalidAVSWithNoMetadataRegistered());
 
         for (uint256 i = 0; i < params.length; i++) {
             OperatorSet memory operatorSet = OperatorSet(avs, params[i].operatorSetId);
--- a/src/contracts/core/AllocationManager.sol
+++ b/src/contracts/core/AllocationManager.sol
@@ -288,8 +288,6 @@ contract AllocationManager is
     ) external onlyWhenNotPaused(PAUSED_OPERATOR_SET_REGISTRATION_AND_DEREGISTRATION) {
         // Check that the caller is either authorized on behalf of the operator or AVS
         require(_checkCanCall(params.operator) || _checkCanCall(params.avs), InvalidCaller());
-        // Check that the AVS exists and has registered metadata
-        require(_avsRegisteredMetadata[params.avs], InvalidAVSWithNoMetadataRegistered());
 
         for (uint256 i = 0; i < params.operatorSetIds.length; i++) {
             // Check the operator set exists and the operator is registered to it
--- a/src/contracts/core/AllocationManager.sol
+++ b/src/contracts/core/AllocationManager.sol
@@ -260,6 +260,8 @@ contract AllocationManager is
     ) external onlyWhenNotPaused(PAUSED_OPERATOR_SET_REGISTRATION_AND_DEREGISTRATION) checkCanCall(operator) {
         // Check that the operator exists
         require(delegation.isOperator(operator), InvalidOperator());
+        // Check that the AVS exists and has registered metadata
+        require(_avsRegisteredMetadata[params.avs], InvalidAVSWithNoMetadataRegistered());
 
         for (uint256 i = 0; i < params.operatorSetIds.length; i++) {
             // Check the operator set exists and the operator is not currently registered to it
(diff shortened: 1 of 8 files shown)


src/contracts/core/AllocationManager.sol
```diff
@@ -324,6 +328,10 @@ contract AllocationManager is
 
     /// @inheritdoc IAllocationManager
     function updateAVSMetadataURI(address avs, string calldata metadataURI) external checkCanCall(avs) {
+        if (!_avsRegisteredMetadata[avs]) {
```
ypatil12, line 331: Why not just store the URI?

src/contracts/core/AVSDirectory.sol
```diff
@@ -115,6 +118,8 @@ contract AVSDirectory is
             avsOperatorStatus[msg.sender][operator] == OperatorAVSRegistrationStatus.REGISTERED,
             OperatorNotRegisteredToAVS()
         );
+        // Assert that the AVS exist and has registered metadata
+        require(_avsRegisteredMetadata[msg.sender] == true, InvalidAVSWithNoMetadataRegistered());
```
ypatil12, line 122: IMO we shouldn't be updating the AVSD since we're planning on deprecating it

src/contracts/core/AllocationManager.sol
```diff
@@ -256,6 +256,8 @@ contract AllocationManager is
     ) external onlyWhenNotPaused(PAUSED_OPERATOR_SET_REGISTRATION_AND_DEREGISTRATION) checkCanCall(operator) {
         // Check that the operator exists
         require(delegation.isOperator(operator), InvalidOperator());
+        // Check that the AVS exists and has registered metadata
+        require(_avsRegisteredMetadata[params.avs], InvalidAVSWithNoMetadataRegistered());
```
ypatil12, line 260: Is your intent to only gate registration on having metadata?

src/contracts/core/AllocationManager.sol
```diff
@@ -282,6 +284,8 @@ contract AllocationManager is
     ) external onlyWhenNotPaused(PAUSED_OPERATOR_SET_REGISTRATION_AND_DEREGISTRATION) {
         // Check that the caller is either authorized on behalf of the operator or AVS
         require(_checkCanCall(params.operator) || _checkCanCall(params.avs), InvalidCaller());
+        // Check that the AVS exists and has registered metadata
+        require(_avsRegisteredMetadata[params.avs], InvalidAVSWithNoMetadataRegistered());
```
ypatil12, line 288: If we have the check on registration, it must be the case that on deregistration the metadata has been registered (since metadata cannot be de-registered). Thus, this check is unnecessary

ypatil12 (changes requested):

src/contracts/core/AllocationManager.sol
bowenli86: it can, but what's the goal of it? what will SC do with the uri?

src/contracts/core/AllocationManager.sol
bowenli86: yea, is there any more gatings in your mind?

src/contracts/core/AllocationManager.sol
ypatil12: If marketplace finds it useful we can, but you're right nothing else we can do

bowenli86: another concern is security in case of any injection attack vector from the uri eg malicious format which may exploit solidity

src/contracts/core/AllocationManager.sol
bowenli86: removed

src/contracts/core/AllocationManagerStorage.sol
```diff
@@ -43,6 +43,10 @@ abstract contract AllocationManagerStorage is IAllocationManager {
     /// Note: if set to 0, defaults to the AVS's address
     mapping(address avs => IAVSRegistrar) internal _avsRegistrar;
 
+    /// @dev Lists the AVSs who has registered metadata and claimed itself as an AVS
+    /// @notice bool is not used and if always true if the avs has registered metadata
+    mapping(address avs => bool) internal _avsRegisteredMetadata;
```
8sunyuan, line 48: This would cause storage issues with our existing `AllocationManager` on preprod and testnet. This added mapping should be defined below `_operatorSets`

bowenli86:
chatted with Scott the workflow it's good

@ypatil12 @8sunyuan i'm changing the assertion from when register operator to when create operator set so it can gate the workflow even eariler and more uptream

bowenli86:
@gpsanant @stevennevins fyi, we will merge this in soon

src/test/integration/tests/Deposit_Delegate_Allocate_Slash_Queue_Redeposit.t.sol
```diff
@@ -52,6 +52,9 @@ contract Integration_Deposit_Delegate_Allocate_Slash_Queue_Redeposit is Integrat
         staker.delegateTo(operator);
         check_Delegation_State(staker, operator, strategies, shares);
 
+        // update AVS metadata URI
+        avs.updateAVSMetadataURI("https://example.com");
```
ypatil12, line 56: This should be in the constructor of the AVS contract

ypatil12: No need to put this in the integration test itself

ypatil12 (approved):

ypatil12 (commented):
Approved, can you also update smart contract docs here with the new flow?

src/test/integration/users/AVS.t.sol
```diff
@@ -67,6 +67,18 @@ contract AVS is Logger, IAllocationManagerTypes, IAVSRegistrar {
     /// AllocationManager
     /// -----------------------------------------------------------------------
 
+    function updateAVSMetadataURI(
```
0xClandestine, line 70: This isn't useful for integration, we opted not to add it originally since it's not stateful and doesn't return anything.

src/contracts/interfaces/IAllocationManager.sol
```diff
@@ -24,6 +24,8 @@ interface IAllocationManagerErrors {
 
     /// @dev Thrown when an invalid operator is provided.
     error InvalidOperator();
+    /// @dev Thrown when an invalid avs whose metadata is not registered is provided.
+    error InvalidAVSWithNoMetadataRegistered();
```
0xClandestine, line 28: Naming could use some work.

```suggestion
    error NonexistentAVSMetadata();
```

ypatil12: Agree with this :)

wadealexc: +1

wadealexc (approved):
approved after greg changes the error name :)

0xClandestine (approved):

stevennevins (approved):

PR #1027, opened 2025-01-14
fix: overflow bug for pendingDiff input

8sunyuan:
Due to casting not causing a revert in `_addInt128`, its possible to cast and truncate a value that overflows. This introduces several problems but the one outlined in the added test shows you can manipulate encumberedMagnitude to be 0 and infinitely allocate instead of being capped by maxMagnitude. 
This simple fix will now safecast and cause a revert if overflow occurs when downcasting to a `uint64`. If needed, we can only add a check on the input param so that the new magnitude allocation is <= WAD


wadealexc (approved):
LGTM

ypatil12 (approved):

PR #1031, opened 2025-01-15
Alex/upgrade test refactor

wadealexc:


ypatil12 (approved):
Nice simplification. 

In ReadME: 
> Note how the initial staker actions are  NOT followed by check_X_State methods. This is because, before calling _upgradeEigenLayerContracts, the test is being run on old contracts. Adding invariant checks to old state transitions is not the point of this test and would add a lot of maintenance overhead.

This should mean we can simplify some checks as part of this PR:

1. https://github.com/Layr-Labs/eigenlayer-contracts/blob/7b8ebd105bf6bdca1f63509365d317f7b122e39f/src/test/integration/IntegrationBase.t.sol#L1489-L1491
2. https://github.com/Layr-Labs/eigenlayer-contracts/blob/7b8ebd105bf6bdca1f63509365d317f7b122e39f/src/test/integration/IntegrationBase.t.sol#L1659-L1661

Regardless will approve

wadealexc:
> Nice simplification.
> 
> In ReadME:
> 
> > Note how the initial staker actions are  NOT followed by check_X_State methods. This is because, before calling _upgradeEigenLayerContracts, the test is being run on old contracts. Adding invariant checks to old state transitions is not the point of this test and would add a lot of maintenance overhead.
> 
> This should mean we can simplify some checks as part of this PR:
> 
>     1. https://github.com/Layr-Labs/eigenlayer-contracts/blob/7b8ebd105bf6bdca1f63509365d317f7b122e39f/src/test/integration/IntegrationBase.t.sol#L1489-L1491
> 
>     2. https://github.com/Layr-Labs/eigenlayer-contracts/blob/7b8ebd105bf6bdca1f63509365d317f7b122e39f/src/test/integration/IntegrationBase.t.sol#L1659-L1661
> 
> 
> Regardless will approve

:+1: , I have already removed these in another PR I'm working on locally!

PR #1033, opened 2025-01-17
fix: ep negative shares bug

ypatil12:
Fixes the negative shares edge case where an operator is overrewarded shares if a staker, who previously had negative shares, completes a withdrawal as shares. Integration upgrade tests have been added to verify behavior.


ypatil12:
Proof of failure on bringing old logic back:

image

wadealexc (approved):
LGTM

Issue #513, opened 2024-04-15
stake() in eigenPodManager  will failed when msg.sender had not eigenPod


macrocan:
**Describe the bug**
stake() in eigenPodManager  will failed when msg.sender had not eigenPod 

**To Reproduce**
1. generate beacon stake deposit-data
2. https://holesky.etherscan.io/address/0x30770d7E3e71112d7A6b7259542D1f680a70e315#writeProxyContract
3. find stake()

**cause**
DepositDataRoot in deposit-data will digit serval elements which include withdrawal credentials, user can't know his  eigenpod(still not existed) address in advance.

PR #1035, opened 2025-01-17
Add check of deposit data in stake

louisbrown0212:
It resolves this bug #513


ypatil12:
We're not looking to resolve this bug at the moment, closing

PR #1037, opened 2025-01-17
feat: add `getAllocatedStake` and update `getMinimumSlashableStake`

8sunyuan:
`getMinimumSlashableStake` returns the amount of slashable stake for an OperatorSet and a respective list of operators. It is currently used by both AVSs (avs-sync) and operators (operator CLI). 

However, this view function doesn't consider whether or not the OperatorSet could actually _slash_ the operator. That is, if an operator registered and their slashableUntil block has passed then the OperatorSet cannot slash the Operator and their actual real slashable stake is 0. Thus, we add a `_isOperatorSlashable` check to this function. 

In the case where the operator wants to know how much slashable stake is allocated, irrespective of its slashable status, we add a `getAllocatedStake` helper function.


wadealexc (commented):
Needs `make bindings`

ypatil12 (approved):

PR #1038, opened 2025-01-17
feat(wip): better introspection for encumbered magnitude

wadealexc:


src/contracts/core/AllocationManagerStorage.sol
```diff
@@ -85,7 +85,7 @@ abstract contract AllocationManagerStorage is IAllocationManager {
     /// @dev For a strategy, contains the amount of magnitude an operator has allocated to operator sets
     /// @dev This value should be read with caution, as deallocations that are completable but not
```
ypatil12, line 86: Nit: can't be read publicly, so we can update comment

ypatil12 (approved):
LGTM, one non-blocking nit. Would also be good to have a test confirming expected behavior of the view func

wadealexc:
> LGTM, one non-blocking nit. Would also be good to have a test confirming expected behavior of the view func

So it doesn't really look like it because the updates I made to tests were very small, but the updated unit tests actually do check that the new function behaves correctly. 

I was also planning to further dig into this in integration tests -- this version of getting encumbered magnitude makes integration test invariant/assertion writing much easier :)

src/contracts/core/AllocationManagerStorage.sol
wadealexc: [919389d](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1038/commits/919389d42b75286d1092f5baa7499828ce05016d)

8sunyuan (approved):
LGTM, the only code smell to me is how these view functions are implemented differently from the internal logic but thats been the case for a while. A bit late for a refactor.

PR #1043, opened 2025-01-22
feat: github action for testnet deploy with zeus 1.2.0

jbrower95:
image

Github Action to support deploying testnets.

### DIFF
--- a/.github/workflows/deploy-testnet.yml
+++ b/.github/workflows/deploy-testnet.yml
@@ -1,0 +1,50 @@
+name: Deploy Tenderly Testnet
+
+on:
+  workflow_dispatch:
+    inputs:
+      upgrade_name:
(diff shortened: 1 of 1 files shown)


ypatil12 (approved):

0xrajath (approved):

.github/workflows/deploy-testnet.yml
```diff
@@ -0,0 +1,50 @@
...
+          - preprod
+          - testnet
+          - mainnet
+
+# Ensures only those with write access can trigger this workflow.
+permissions:
+  contents: write
+
+jobs:
+  deploy:
```
0xClandestine, line 24: can we add a name field so it doesn't default to lowercase (for consistency between other workflows).

0xClandestine (changes requested):

PR #1045, opened 2025-01-22
fix: delegate shares

eigenmikem:
**Motivation:**

Fixes an issue where stakers delegating Beacon Chain ETH from slashed Eigen Pods were able to delegate more shares than they should. Specifically, operators now are delegated a staker's `withdrawableShares` rather than their `depositShares`.

**Modifications:**

- Changed accounting logic on delegation in `DelegationManger.sol`
- `DepositScalingFactor` now resets when a staker withdraws all their shares, whether through undelegation, redelegation, or a simple withdrawal
- Changes in `StrategyManager.sol`, `IShareManager.sol`, `SlashingLib.sol`, and `EigenPodManager.sol` to accommodate new accounting
- New test files and changes to others to reflect new accounting and invariants
- Updated `docs/SharesAccounting.md`

**Result:**

System is now robust to stakers with arbitrary EigenPod states


wadealexc (changes requested):
Overall, pretty clean. Main feedback:

First, this adds 1kB of bytecode :grimacing:. Given that `getDelegatableShares` is basically just `getDepositedShares` with extra steps, can we find a way to factor out some of the shared logic between the two?

Second, I'm not a _huge_ fan of the way the `_delegate` method reads now:
1. `slashingFactors` is retrieved differently compared to everywhere else in the contract.
2. We're actually still using all the components of the `_getSlashingFactors` method, just applying them differently. This is a subtle difference that needs to be highlighted more in the code. i.e. it's not clear that we are actually making use of _both_ the operator's slashing factors _and_ the beacon chain slashing factor.

If `_delegate` is going to be an odd egg in the contract, the differences should pop out and be apparent at the top level.

src/contracts/core/DelegationManager.sol
```diff
@@ -340,24 +341,38 @@ contract DelegationManager is
...
+            IStrategy[] memory newStrategies = new IStrategy[](strategies.length + 1);
+            newStrategies[strategies.length] = beaconChainETHStrategy;
+
+            // Copy any strategy manager shares to complete array
+            for (uint256 i = 0; i < strategies.length; i++) {
+                newStrategies[i] = strategies[i];
+            }
+            strategies = newStrategies;
+        }
+        (uint256[] memory withdrawableShares,) = getWithdrawableShares(staker, strategies);
```
8sunyuan, line 359: I can see in this section you want the withdrawableShares but you also want the list of strategies that `getDepositedShares` returns so you can get the slashingFactors. I think it would be cleaner to just have another similar view function to `getWithdrawableShares` that returns the list of respective strategies the staker has deposits and withdrawable shares in.

src/contracts/libraries/SlashingLib.sol
```diff
@@ -136,6 +136,11 @@ library SlashingLib {
         dsf._scalingFactor = newDepositScalingFactor;
     }
 
+    function updateNewDelegation(DepositScalingFactor storage dsf, uint256 slashingFactor) internal {
+        //On delegation, we use the old deposit scaling factor in case that it is non-WAD
+        dsf._scalingFactor = dsf.scalingFactor().divWad(slashingFactor);
```
8sunyuan, line 141: I'm not entirely sure if this is correct and if the BCSF should be part of this calculation.

8sunyuan: Yeah the usage of slashingFactors throughout the DM includes both maxMagnitude and BCSF so its a little inconsistent/confusing here. Perhaps just a natspec comment on this function noting that it doesn't include the BCSF and its _only_ the maxMagnitude.

src/contracts/core/DelegationManager.sol
```diff
@@ -340,24 +341,38 @@ contract DelegationManager is
...
+            // Allocate extra space for beaconChainETHStrategy and shares
+            IStrategy[] memory newStrategies = new IStrategy[](strategies.length + 1);
+            newStrategies[strategies.length] = beaconChainETHStrategy;
+
+            // Copy any strategy manager shares to complete array
+            for (uint256 i = 0; i < strategies.length; i++) {
+                newStrategies[i] = strategies[i];
+            }
+            strategies = newStrategies;
+        }
```
wadealexc, line 358: Can we not just turn this whole thing into `(IStrategy[] memory strategies, ) = getDepositedShares(staker)`?

Saves half a kb

src/contracts/core/DelegationManager.sol
wadealexc: See above comment, we basically already have this

src/contracts/core/DelegationManager.sol
wadealexc: it's definitely at a premium -- the other code is only saving maybe a few hundred gas at most, which is pennies compared to the bytecode space we have left. 100% we are optimizing for codesize, even if we spend more gas.

src/contracts/libraries/SlashingLib.sol
```diff
@@ -136,6 +145,12 @@ library SlashingLib {
         dsf._scalingFactor = newDepositScalingFactor;
     }
 
+    function reset(
```
ypatil12, line 148: Nit: Might want to add a comment & docs on resetting on undelegation. The reason this works is that:

1. When queueing, we store the DSF at the time via `scaledShares`
2. For ERC-20s, the DSF is only normalized for the operator, so can reset. 
3. For BC ETH, we are able to reset because undelegate queues a withdrawal for _all_ shares, resulting in future native eth deposits (and slashes) between queue and complete to use the "reset" scaling factor. Upon completing as shares, we end up doing an increase in depositShares with all slashings applied.

src/contracts/core/DelegationManager.sol
```diff
@@ -412,6 +412,10 @@ contract DelegationManager is
                 depositSharesToWithdraw: singleDepositShares,
                 slashingFactors: singleSlashingFactor
             });
+
+            //Reset DepositScalingFactor
+            DepositScalingFactor storage dsf = _depositScalingFactor[staker][strategies[i]];
+            dsf.reset();
```
wadealexc, line 418: Doesn't this mean we can condense the logic in the `update` method, so that the first two branches are combined?

`if (newDelegation || prevDepositShares == 0) dsf = dsf.divWad(slashingFactor);`

ypatil12: Agree. We should reset DSF even if a user queues a withdrawal for all shares (no undelegation)

wadealexc: what does this look like?

im imagining on this line: https://github.com/layr-labs/eigenlayer-contracts/blob/57883c15864a2f84908fecf349d15989e248f290/src/contracts/core/DelegationManager.sol#L488

we would have `removeDepositShares` return the number of deposit shares remaining, and if it's zero we reset?

ypatil12: Correct

ypatil12: Generally pro doing this because of the consistency for undelegation & queue for all shares in the state model

wadealexc: BTW re: my simplification in the OP, if we're condensing those branches we can actually just remove `newDelegation` entirely

We're back to just `if (prevDepositShares == 0)`

wadealexc: I'm looking at the places `newDelegation` is used and it's only used when `prevDepositShares == 0`

src/contracts/core/DelegationManager.sol
```diff
@@ -477,6 +482,12 @@ contract DelegationManager is
                 });
             }
 
+            if (depositSharesToWithdraw[i] == shareManager.stakerDepositShares(staker, strategies[i])) {
+                //Reset DepositScalingFactor
+                DepositScalingFactor storage dsf = _depositScalingFactor[staker][strategies[i]];
+                dsf.reset();
```
wadealexc, line 488: Nit: can we make these one-liners? since the method doesnt have any args i dont think there's value in splitting it into 2 LOC

src/contracts/core/DelegationManager.sol
```diff
@@ -477,6 +490,11 @@ contract DelegationManager is
                 });
             }
 
+            //Reset DepositScalingFactor since deposit shares are equal to withdrawable shares after full withdrawal
+            if (depositSharesToWithdraw[i] == shareManager.stakerDepositShares(staker, strategies[i])) {
+                _depositScalingFactor[staker][strategies[i]].reset();
+            }
```
wadealexc, line 496: I would rather have `removeDepositShares` return the new number of deposit shares, and compare that against 0

This way feels more messy.

src/test/integration/tests/Slashed_Eigenpod.t.sol
```diff
@@ -0,0 +1,267 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "src/test/integration/IntegrationChecks.t.sol";
+
+contract SlashedEigenpod is IntegrationCheckUtils {
+    using ArrayLib for *;
```
wadealexc, line 7: All integration tests should start with `Integration_Name`

... that way we can run integration tests separately by doing `forge t --mc Integration`

src/contracts/core/DelegationManager.sol
```diff
@@ -490,13 +490,13 @@ contract DelegationManager is
...
             }
 
+            // Remove deposit shares from EigenPodManager/StrategyManager
+            uint256 sharesAfter = shareManager.removeDepositShares(staker, strategies[i], depositSharesToWithdraw[i]);
+
             //Reset DepositScalingFactor since deposit shares are equal to withdrawable shares after full withdrawal
-            if (depositSharesToWithdraw[i] == shareManager.stakerDepositShares(staker, strategies[i])) {
+            if (depositSharesToWithdraw[i] == sharesAfter) {
                 _depositScalingFactor[staker][strategies[i]].reset();
             }
```
wadealexc: Is this right? I thought it should be `sharesAfter == 0`?

ypatil12 (approved):
Minor nits, but non-blocking. Happy with the outcome here

src/contracts/core/DelegationManager.sol
```diff
@@ -343,24 +343,37 @@ contract DelegationManager is
...
-        uint256[] memory slashingFactors = _getSlashingFactors(staker, operator, strategies);
-
         for (uint256 i = 0; i < strategies.length; ++i) {
+            // Special case for beacon chain slashing - ensure the staker's beacon chain slashing is
+            // reflected in the number of shares they delegate.
+            if (strategies[i] == beaconChainETHStrategy) {
+                uint64 stakerBeaconChainSlashing = eigenPodManager.beaconChainSlashingFactor(staker);
+
+                DepositScalingFactor memory dsf = _depositScalingFactor[staker][strategies[i]];
+                withdrawableShares[i] = dsf.calcWithdrawable(withdrawableShares[i], stakerBeaconChainSlashing);
```
8sunyuan, line 366: Note: since undelegated, staker maxMagnitude is 0 so their slashingFactor only needs to use the BCSF for calculating withdrawable.

src/contracts/libraries/SlashingLib.sol
```diff
@@ -94,10 +94,11 @@ library SlashingLib {
...
         uint256 slashingFactor
     ) internal {
-        // If this is the staker's first deposit, set the scaling factor to
-        // the inverse of slashingFactor
         if (prevDepositShares == 0) {
-            dsf._scalingFactor = uint256(WAD).divWad(slashingFactor);
+            // If this is the staker's first deposit or is delegating to an operator, set the scaling factor to
+            // the inverse of slashingFactor. In the case of delegation slashingFactor is just
+            // maxMagnitude because BCSF is already accounted for in the dsf
+            dsf._scalingFactor = dsf.scalingFactor().divWad(slashingFactor);
```
8sunyuan, line 101: Should we keep this as `uint256(WAD).divWad(slashingFactor);` now that we are resetting DSF to WAD upon undelegation? I think it would be a bit clearer here because using the previous DSF makes it seem like the old value can be any value.

8sunyuan: If we keep as is, `SharesAccountingEdgeCases.md` needs to be updated under the `Upper bound on Residual Operator Shares` section

wadealexc: This would fail to account for the following scenario:
0. Staker is not delegated to an operator
1. Staker is slashed on beacon chain
2. Staker verifies more validators. This sets their DSF to a non-WAD value
3. Staker delegates to an operator. With your change, the non-WAD value is ignored when calculating the number of shares to delegate, and the staker's DSF is reset to WAD.

This scenario is implemented in test: `testFuzz_delegateSlashedStaker_dsfNonWad`

wadealexc: If docs need to be updated, could you take care of that?

src/test/integration/IntegrationBase.t.sol
```diff
@@ -1193,7 +1193,7 @@ abstract contract IntegrationBase is IntegrationDeployer, TypeImporter {
 
         // For each strategy, check (prev - removed == cur)
         for (uint i = 0; i < strategies.length; i++) {
-            assertEq(prevShares[i] - removedShares[i], curShares[i], err);
+            assertEq(prevShares[i], curShares[i] + removedShares[i], err);
```
wadealexc, line 1196: oof, i know exactly how annoying these can be to debug.

it feels very counterintuitive to write the assertions like this, though :sweat:

src/test/integration/IntegrationChecks.t.sol
```diff
@@ -104,8 +104,8 @@ contract IntegrationCheckUtils is IntegrationBase {
 
         assert_Snap_Unchanged_Staker_DepositShares(staker, "staker shares should not have decreased");
         assert_Snap_Removed_Staker_WithdrawableShares_AtLeast(staker, BEACONCHAIN_ETH_STRAT, slashedAmountGwei * GWEI_TO_WEI, "should have decreased withdrawable shares by at least slashed amount");
-        assert_Snap_Removed_ActiveValidatorCount(staker, slashedValidators.length, "should have decreased active validator count");
-        assert_Snap_Removed_ActiveValidators(staker, slashedValidators, "exited validators should each be WITHDRAWN");
+        //assert_Snap_Removed_ActiveValidatorCount(staker, slashedValidators.length, "should have decreased active validator count");
+        //assert_Snap_Removed_ActiveValidators(staker, slashedValidators, "exited validators should each be WITHDRAWN");
```
wadealexc, line 108: What's up with these?

src/test/integration/IntegrationChecks.t.sol
```diff
@@ -189,8 +189,9 @@ contract IntegrationCheckUtils is IntegrationBase {
         assertEq(address(operator), delegationManager.delegatedTo(address(staker)), "staker should be delegated to operator");
         assert_HasExpectedShares(staker, strategies, shares, "staker should still have expected shares after delegating");
         assert_Snap_Unchanged_Staker_DepositShares(staker, "staker shares should be unchanged after delegating");
-        assert_Snap_Unchanged_Staker_WithdrawableShares(staker, "withdrawable shares should be unchanged after delegating");
-        assert_Snap_Added_OperatorShares(operator, strategies, shares, "operator should have received shares");
+        //assert_Snap_Unchanged_Staker_WithdrawableShares_Delegation(staker, "withdrawable shares should be unchanged after delegating");
```
wadealexc, line 192: Uncommenting this doesn't cause any tests to fail. Is there a reason it's commented out?

src/test/integration/users/User.t.sol
```diff
@@ -237,6 +237,34 @@ contract User is Logger, IDelegationManagerTypes, IAllocationManagerTypes {
...
+                : IERC20Metadata(address(strat.underlyingToken())).name();
+            
+            console.log(
+                "   Expecting withdrawal with nonce %s of %s for %s scaled shares.", 
+                expectedWithdrawals[i].nonce,
+                name,
+                expectedWithdrawals[i].scaledShares[0]
+            );
+        }
+        return expectedWithdrawals;
```
wadealexc, line 265: mmk... time for a helper method for these :D 

can leave for another PR

src/test/integration/users/User.t.sol
```diff
@@ -237,6 +237,34 @@ contract User is Logger, IDelegationManagerTypes, IAllocationManagerTypes {
         return expectedWithdrawals;
     }
 
+    /// @dev Redelegate to a new operator
+    function redelegate(
+        User newOperator
+    ) public virtual createSnapshot returns (Withdrawal[] memory) {
+        print.method("redelegate", newOperator.NAME_COLORED());
+        Withdrawal[] memory expectedWithdrawals = _getExpectedWithdrawalStructsForStaker(address(this));
```
wadealexc, line 245: While im on the topic, this is useless because we now return withdrawal roots -- so we can query these after the fact.

src/test/integration/IntegrationChecks.t.sol
wadealexc: Re-added here: [86c255d](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1045/commits/86c255d601f1d7c96dc7428c8f2a3aee47ba5f0b)

I know this can be problematic because we dont have good handling for rounding issues -- i have had to tweak this in a branch i'm working on. But I'm gonna leave it uncommented unless it's actively breaking. That way when it does break, we can fix it :)

src/test/integration/IntegrationChecks.t.sol
wadealexc: Okay, figured out what's wrong with these. TLDR this assertion is only used when we do a slash that doesn't withdraw validators. I added a clarifying comment here [86c255d](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1045/commits/86c255d601f1d7c96dc7428c8f2a3aee47ba5f0b)

wadealexc (approved):
Nice work. There's test cleanup to be done in followup PRs, but this looks great otherwise.

src/contracts/libraries/SlashingLib.sol
8sunyuan: Ah good callout, always the beaconChain scenarios. Will update this in another PR.

src/test/integration/users/User.t.sol
wadealexc: If that is the purpose, I would say that code should exist at the test level rather in this contract. I think user-type methods should be stupid simple and should be (wherever possible) like just interacting with the contracts directly.

wadealexc: Otherwise we start building up business logic in the user contract, and it gets harder to maintain

0xClandestine:
flaky ci fixed upstream

PR #1051, opened 2025-01-25
feat: add OperatorSharesSlashed event to track shares slashed per operator

bowenli86:
add OperatorSharesSlashed event to track shares slashed per operator

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -307,6 +307,9 @@ contract DelegationManager is
             sharesToDecrease: operatorSharesSlashed
         });
 
+        // Emit event for operator shares being slashed
+        emit OperatorSharesSlashed(operator, strategy, totalDepositSharesToBurn);
+
         IShareManager shareManager = _getShareManager(strategy);
         // NOTE: for beaconChainETHStrategy, increased burnable shares currently have no mechanism for burning
         shareManager.increaseBurnableShares(strategy, totalDepositSharesToBurn);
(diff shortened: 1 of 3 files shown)


src/contracts/core/DelegationManager.sol
```diff
@@ -307,6 +307,9 @@ contract DelegationManager is
             sharesToDecrease: operatorSharesSlashed
         });
 
+        // Emit event for operator shares being slashed
+        emit OperatorSharesSlashed(operator, strategy, totalDepositSharesToBurn);
```
ypatil12, line 311: To confirm, you are trying to get what was slashed of the operator that was also in the queue? In this case, the `OperatorSharesDecreased` event will not map 1:1 with the `OperatorSharesSlashed` event in terms of magnitude of burned shares (since some could have been in the queue).

ypatil12 (commented):
Clarification

src/contracts/core/DelegationManager.sol
bowenli86: yes, that'll have more meanings so we are not simply having value-duplicated events

ypatil12, line 311: Can we add a test for the event emitting in our current lifecycle burning tests?

ypatil12 (commented):
Let's test the event is being emitted with expected values

ypatil12 (approved):
LGTM

8sunyuan (approved):
LGTM, noting that it looks slightly weird the naming being depositShares rather than just `operatorShares`/`delegatedShares` but that was the already existing code.

PR #1052, opened 2025-01-27
fix: overwriting existing pending modifications

8sunyuan:
As a result of an existing deallocation being slashed such that its pendingDiff becomes an amount 0, it is possible
to call modifyAllocations and create a new allocation/deallocation overwriting the old deallocation in storage.
This could result in the deallocationQueue no longer being in ascending order by the deallocation's respective effectBlocks.

This prevents this edge case altogether by simply changing the check to use the effectBlock instead `require(allocation.effectBlock == 0, ModificationAlreadyPending());`

### DIFF
--- a/src/test/unit/AllocationManagerUnit.t.sol
+++ b/src/test/unit/AllocationManagerUnit.t.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.27;
 
-import "src/contracts/core/AllocationManager.sol";
+import "src/test/harnesses/AllocationManagerHarness.sol";
 import "src/test/utils/EigenLayerUnitTestSetup.sol";
 import "src/test/mocks/MockAVSRegistrar.sol";
 
@@ -33,7 +33,7 @@ contract AllocationManagerUnitTests is EigenLayerUnitTestSetup, IAllocationManag
     /// Mocks
     /// -----------------------------------------------------------------------
 
-    AllocationManager allocationManager;
+    AllocationManagerHarness allocationManager;
     ERC20PresetFixedSupply tokenMock;
     StrategyBase strategyMock;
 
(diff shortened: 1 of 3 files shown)


8sunyuan:
image
Expected failed test cases from https://github.com/Layr-Labs/eigenlayer-contracts/pull/1052/commits/52c1dfbe508bc0cb8b5620b442baa854c7438f3e before the fix is implemented

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -1916,6 +1933,194 @@ contract AllocationManagerUnitTests_ModifyAllocations is AllocationManagerUnitTe
...
+        // overwriting/increasing the effectBlock
+        // roll blocks forward so that deallocations have different effectBlocks
+        cheats.roll(block.number + 100);
+        // Note: this should revert but previously it would not prior to the bugfix
+        AllocateParams[] memory newAllocateParams1 = _newAllocateParams(operatorSet1, 400);
+        cheats.prank(defaultOperator);
+        cheats.expectRevert(ModificationAlreadyPending.selector);
+        allocationManager.modifyAllocations(defaultOperator, newAllocateParams1);
+
+        // Assert that the deallocationQueue is unordered for the 2 deallocations in queue
```
ypatil12, line 2054: Nit: is `ordered`

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -1916,6 +1933,194 @@ contract AllocationManagerUnitTests_ModifyAllocations is AllocationManagerUnitTe
...
+
+    /**
+     * @notice Regression tests for the bugfix where pending modifications were checked by
+     * require(allocation.pendingDiff == 0, ModificationAlreadyPending()); 
+     * which would overwrite the effectBlock, pendingDiff if a pendingDiff
+     * of a deallocation was slashed to become 0.
+     *
+     * This test checks that the deallocationQueue is ascending ordered by effectBlocks
+     * In this case the new modifyAllocations call is an allocation
+     * where the effectBlock is increased and the deallocationQueue is unordered as well because the operator
```
ypatil12, line 2066: Nit: is `ordered`

ypatil12, line 2066: I'm a bit confused by this test since the allocation in step 6 reverts, meaning that we aren't testing an allocation with the new allocation delay

ypatil12 (approved):
LGTM, had a test question

src/test/unit/AllocationManagerUnit.t.sol
8sunyuan: It was testing the previous code where you can overwrite the deallocation with a new allocation. Then asserting the deallocationQueue was sorted which it wasn't. After the fix, it should revert altogether from trying to modify in the first place

ypatil12: Ahh okay, gotcha, see the 5000 allocation here and the 4000 allocation previously

PR #1053, opened 2025-01-28
feat: pectra compatibility

ypatil12:
Updates checkpoint proof system to be Pectra compatible. The breaking change to EigenPods is the `BeaconState` container increasing to have 37 fields, which results in the tree height to be > 5. 

### Overview
We need to solve for the following cases:

- Prevent deneb proofs from being submitted to pectra blocks
- Ensure that the PECTRA_FORK_TIMESTAMP is the first timestamp at or after the pectra hard fork for which there is a non missed slot

To do this, here is the upgrade process:

1. Pause checkpoint starting & credential proofs
2. Upgrade after fork is hit
3. Run script to detect the first timestamp at or after the pectra hard fork for which there is a non missed slot
4. Set pectra fork timestamp to the first timestamp at which there is a pectra block header
5. Unpause

### EigenPod
- Updated balance container and validator container proofs to pass in a proof timestamp & pectra fork timestamp to check against which tree height to use for the beacon state
- Modify storing variables in memory to handle stack too deep errors
- Note that since the 4788 oracle returns the PARENT beacon block root, our check against the pectra fork timestamp returns the previous tree length for proofs that are <= `pectraForkTimestamp`

### Deprecation Plan
- Post pectra, we can upgrade the EigenPod to deprecate the fork timestamp case handling once all in progress pre-Pectra checkpoints have been completed 

### TODOs
- [x] Unit Tests
- [x] Integration Tests simulating upgrade
- [x] Mekong Deployment
- [x] Update Integration Test User to use validators >32 ETH

### DIFF
--- a/src/test/integration/IntegrationDeployer.t.sol
+++ b/src/test/integration/IntegrationDeployer.t.sol
@@ -212,10 +212,10 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
         timeMachine = new TimeMachine();
         beaconChain = new BeaconChainMock(eigenPodManager, BEACON_GENESIS_TIME);
 
-        // Set the `pectraForkTimestamp` on the EigenPodManager
+        // Set the `pectraForkTimestamp` on the EigenPodManager. Use pectra state
         cheats.startPrank(executorMultisig);
         eigenPodManager.setProofTimestampSetter(executorMultisig);
-        eigenPodManager.setPectraForkTimestamp(BEACON_GENESIS_TIME);
+        eigenPodManager.setPectraForkTimestamp(BEACON_GENESIS_TIME); 
         cheats.stopPrank();
     }
 
@@ -257,6 +257,12 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
         // prior to test. `isUpgraded` is true by default, but is set to false in `UpgradeTest.t.sol`
         if (isUpgraded) {
             _upgradeMainnetContracts();
+
+            // Set the `pectraForkTimestamp` on the EigenPodManager. Use pectra state
+            cheats.startPrank(executorMultisig);
+            eigenPodManager.setProofTimestampSetter(executorMultisig);
+            eigenPodManager.setPectraForkTimestamp(BEACON_GENESIS_TIME); 
+            cheats.stopPrank();
         }
     }
 


src/contracts/pods/EigenPod.sol
```diff
@@ -254,6 +255,7 @@ contract EigenPod is Initializable, ReentrancyGuardUpgradeable, EigenPodPausingC
         for (uint256 i = 0; i < validatorIndices.length; i++) {
             // forgefmt: disable-next-item
             totalAmountToBeRestakedWei += _verifyWithdrawalCredentials(
+                beaconTimestamp,
```
wadealexc, line 258: Hmm, I wonder about the usage of `getParentBlockRoot` above.

Can `beaconTimestamp` belong to the block _after_ the fork, but the proof itself is over the parent block (aka pre-fork)?

wadealexc (commented):
Is this PR targeting testnet?

We're going to need a different PR for mainnet, since that's landing before slashing, right? (This EigenPod file is compatible with testnet slashing, but not mainnet)

ypatil12:
> Is this PR targeting testnet?

Yup, targeting testnet. There will be a separate mainnet PR. cc @wadealexc

src/contracts/pods/EigenPod.sol
wadealexc: For the record I think I'm correct here. 2 scenarios:

1. I try to verify withdrawal credentials the block after the hard fork. `getParentBlockRoot` grabs a block from before the hard fork; that's what my proof needs to be against.
2. I start a checkpoint just after the hard fork. The checkpoint block root used for proofs is from a block just before the hard fork.

Naive solution: have the fork selector logic in the proofs library check `proofTimestamp - 12`. However, this runs into potential issues with skipped slots -- I might have a proof timestamp 2 blocks after the hard fork, but there was a skipped slot right at the fork, so the proof block is from before the fork.

I think, unfortunately, we need to pause proofs just before the fork, and unpause just after the fork (once we see the first valid block). Although I'm not sure this entirely fixes the issue... will think more.

wadealexc: Maybe we need to pause + upgrade just after the fork + unpause? ugh.

ypatil12: Synced offline, here is what we need to do:

- Prevent deneb proofs from being submitted to pectra blocks
- Checkpoint proofs store the proof timestamp. If there are missed slots at the hard fork timestamp, it is possible, like Alex mentions above, that the `beaconTimestamp` is post pectra but the block header is deneb. Thus, we need to ensure that the `PECTRA_FORK_TIMESTAMP` is the first timestamp at which the 4788 oracle returns a pectra state root. 

To do this, here is the process:

1. Pause checkpoint & validator proofs
2. Upgrade after fork is hit
3. Set pectra fork timestamp to the first timestamp at which there is a pectra block header 
4. Unpause

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -312,4 +321,16 @@ library BeaconChainProofs {
     ) internal pure returns (uint64) {
         return Endian.fromLittleEndianUint64(validatorFields[VALIDATOR_EXIT_EPOCH_INDEX]);
     }
+
+    /// @notice The timestamp of the Pectra hard fork - this is updated on a per-network basis
+    uint64 public constant PECTRA_FORK_TIMESTAMP = 1_739_352_768;
```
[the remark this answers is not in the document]
ypatil12: Setting manually now, resolving

src/contracts/interfaces/IEigenPod.sol
```diff
@@ -351,4 +352,7 @@ interface IEigenPod is IEigenPodErrors, IEigenPodEvents {
     function getParentBlockRoot(
         uint64 timestamp
     ) external view returns (bytes32);
+
+    /// @notice Returns the timestamp of the Pectra fork, read from the `EigenPodManager` contract
+    function getPectraForkTimestamp() external view returns (uint64);
```
wadealexc, line 357: Do we need this added to the EigenPod interface?

It already exists on the EPM - why duplicate?

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -120,6 +128,16 @@ interface IEigenPodManager is
         int256 balanceDeltaWei
     ) external;
 
+    /// @notice Sets the address that can set proof timestamps
+    function setProofTimestampSetter(
+        address newProofTimestampSetter
+    ) external;
```
wadealexc, line 134: `forkTimestampSetter` feels like a better name

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -312,4 +323,13 @@ library BeaconChainProofs {
...
     }
+
+    /// @dev We check if the proofTimestamp is <= pectraForkTimestamp because a `proofTimestamp` at the `pectraForkTimestamp`
+    ///      is considered to be Pre-Pectra given the EIP-4788 oracle returns the parent block.
+    function getBeaconStateTreeHeight(
+        uint64 proofTimestamp,
+        uint64 pectraForkTimestamp
+    ) internal pure returns (uint256) {
+        return proofTimestamp <= pectraForkTimestamp ? DENEB_BEACON_STATE_TREE_HEIGHT : PECTRA_BEACON_STATE_TREE_HEIGHT;
+    }
```
wadealexc, line 334: Just making some notes:

Before we set `pectraForkTimestamp` in the EPM, it returns 0. This will cause all proofs to be interpreted as pectra proofs. However, we expect contracts to be paused if this value is unset, so that's fine :+1: 

Additionally, if the contracts are unpaused but the value is still unset, this logic means all proofs are interpreted as pectra proofs, even if we're working with a dencun block. This is good fallback behavior, as the worst case means some proofs will revert until we set the timestamp. :+1: 

It's the reverse behavior we wanted to avoid - where dencun proof sizes are allowed for pectra blocks. That shouldn't be possible here :+1:

src/contracts/pods/EigenPod.sol
wadealexc: image

wadealexc: Using the image above, we want to determine what proof type to use, given the proof timestamp _tp_. Because _tp_ is used to look up a _parent block_ in the EIP-4788 oracle, its proof type corresponds to the last non-skipped block. So, if:

* _tp > t1_, use pectra logic
* _tp <= t1_, use dencun logic

Given our beacon state tree height getter: https://github.com/layr-labs/eigenlayer-contracts/blob/b4852c74cdbe43fea2f7330ea0dc752fcf10b6e9/src/contracts/libraries/BeaconChainProofs.sol#L327-L334

... `pectraForkTimestamp` should be set to _the first valid, non-skipped block after the Pectra hard fork_.

src/contracts/interfaces/IEigenPod.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1053/commits/8d09664ff954cec1e9beba5dddaffa5dcec752f1

src/contracts/pods/EigenPod.sol
```diff
@@ -378,6 +383,7 @@ contract EigenPod is Initializable, ReentrancyGuardUpgradeable, EigenPodPausingC
     }
 
     /// @notice Called by EigenPodManager when the owner wants to create another ETH validator.
+    /// @dev This function only supports staking to a 0x01 validator. For compounding validators, please interact directly with the deposit contract.
```
8sunyuan, line 386: Can be backlog to add stake function with compounding validators when we implement eip7002

src/test/integration/UpgradeTest.t.sol
```diff
@@ -35,4 +37,22 @@ abstract contract UpgradeTest is IntegrationCheckUtils {
         isUpgraded = true;
         emit log("_upgradeEigenLayerContracts: slashing upgrade complete");
     }
+
+    uint64 constant PECTRA_FORK_TIMESTAMP = 1739352768;
```
wadealexc, line 41: Where are you getting this value? Is this the expected timestamp of the holesky fork?

Wouldn't we need to wait to determine this until after the fork? We need the first non-skipped slot, which may not be this specific value.

src/contracts/libraries/BeaconChainProofs.sol
```diff
@@ -312,4 +323,13 @@ library BeaconChainProofs {
     ) internal pure returns (uint64) {
         return Endian.fromLittleEndianUint64(validatorFields[VALIDATOR_EXIT_EPOCH_INDEX]);
     }
+
+    /// @dev We check if the proofTimestamp is <= pectraForkTimestamp because a `proofTimestamp` at the `pectraForkTimestamp`
+    ///      is considered to be Pre-Pectra given the EIP-4788 oracle returns the parent block.
+    function getBeaconStateTreeHeight(
```
wadealexc, line 329: Let's consider moving this logic into `EigenPod`, since it's specific to the EIP-4788 oracle (and the EPM getter), which this library doesn't handle.

We can pass the proofs library an enum (`DENEB|ELECTRA`), instead.

src/contracts/pods/EigenPodManagerStorage.sol
```diff
@@ -102,5 +108,5 @@ abstract contract EigenPodManagerStorage is IEigenPodManager {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[42] private __gap;
+    uint256[41] private __gap;
```
wadealexc, line 111: Should this be 40? we added two variables.

src/test/integration/IntegrationBase.t.sol
```diff
@@ -639,6 +639,9 @@ abstract contract IntegrationBase is IntegrationDeployer {
 
         // For each strategy, check (prev + added == cur)
         for (uint i = 0; i < strategies.length; i++) {        
+            console.log("prevShares[i]", prevShares[i]);
+            console.log("addedShares[i]", addedShares[i]);
+            console.log("curShares[i]", curShares[i]);
```
wadealexc, line 644: nit: can remove this

src/test/unit/EigenPodUnit.t.sol
```diff
@@ -675,11 +675,12 @@ contract EigenPodUnitTests_verifyWithdrawalCredentials is EigenPodUnitTests, Pro
 
     /// @notice beaconTimestamp must be after the current checkpoint
     function testFuzz_revert_beaconTimestampInvalid(uint256 rand) public {
+        rand = 10;
```
wadealexc, line 678: nit: nix this?

src/test/integration/mocks/BeaconChainMock_Pectra.t.sol
```diff
@@ -0,0 +1,264 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "src/test/integration/mocks/BeaconChainMock.t.sol";
+
+/// @notice A backwards-compatible BeaconChain Mock that updates containers & proofs for the Pectra upgrade
+contract BeaconChainMock_PectraForkable is BeaconChainMock {
```
wadealexc, line 7: This file/naming seems misleading -- is it more accurate to call this `BeaconChainMock_Deneb`?

src/test/integration/mocks/BeaconChainMock_Pectra.t.sol
```diff
@@ -0,0 +1,264 @@
...
+        // Pre-populate leaves with dummy values so sibling/parent tracking is correct
+        for (uint i = 0; i < leaves.length; i++) {
+            leaves[i] = bytes32(i + 1);
+        }
+
+        // Place validatorsRoot and balancesRoot into tree
+        leaves[BeaconChainProofs.VALIDATOR_CONTAINER_INDEX] = validatorsRoot;
+        leaves[BeaconChainProofs.BALANCE_CONTAINER_INDEX] = balancesRoot;
+        return leaves;
+    }
```
wadealexc, line 236: Did this need an override? It looks identical

src/test/integration/tests/upgrade/Prooftra.t.sol
```diff
@@ -0,0 +1,131 @@
...
+        check_CompleteCheckpoint_EarnOnBeacon_State(staker, expectedEarnedGwei);
+    }
+
+    function _pauseForkAndUpgrade(uint64 pectraForkTimestamp) internal {
+        // 1. Pause starting checkpoint, completing, and credential proofs
+        cheats.prank(pauserMultisig);
+        eigenPodManager.pause(
+            2 ** PAUSED_START_CHECKPOINT |
+            2 ** PAUSED_EIGENPODS_VERIFY_CREDENTIALS
+        );
```
wadealexc, line 99: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1067/files#r1966170120

wadealexc: (I think we should pause all proofs)

src/test/integration/tests/upgrade/Prooftra.t.sol
```diff
@@ -0,0 +1,131 @@
...
+
+        // 2. Fork to Pectra
+        BeaconChainMock_PectraForkable(address(beaconChain)).forkToPectra(pectraForkTimestamp);
+
+        // 3. Upgrade EigenPodManager & EigenPod
+        _upgradeEigenLayerContracts();
```
wadealexc, line 105: This already calls `_handlePectraFork`, so you're duplicating actions here to some extent. Can we pick whether we're adding special fork logic in `UpgradeTest`, or in this file?

src/test/integration/tests/upgrade/Prooftra.t.sol
```diff
@@ -0,0 +1,131 @@
...
+            BeaconChainMock_PectraForkable(address(beaconChain)).pectraForkTimestamp()
+        );
+        cheats.stopPrank();
+
+        // 2. Randomly warp to just after the fork timestamp
+        // If we do not warp, proofs will be against deneb state
+        if (_randBool()) {
+            // If we warp, proofs will be against electra state
+            cheats.warp(block.timestamp + 1); 
+        }
```
wadealexc, line 125: i'm pretty sure this is testing against electra regardless of whether you warp or not

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -562,7 +562,7 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
 
     function _genRandUser(string memory name, uint userType) internal returns (User user) {
         // Create User contract based on userType:
-        if (forkType == LOCAL) {
+        if (forkType == LOCAL || (forkType == MAINNET && isUpgraded)) {
```
wadealexc, line 565: Seems this has some cherry picked stuff from another branch?

src/test/integration/UpgradeTest.t.sol
wadealexc: Actually - since this is a timestamp from a week ago, this is probably messing with tests? 

We should maybe just have fork timestamp be current timestamp + 12 or something - that way we know we're moving forward in time if we warp to it.

src/contracts/pods/EigenPod.sol
```diff
@@ -731,4 +744,10 @@ contract EigenPod is Initializable, ReentrancyGuardUpgradeable, EigenPodPausingC
         require(success && result.length > 0, InvalidEIP4788Response());
         return abi.decode(result, (bytes32));
     }
+
+    /// @notice Returns the timestamp of the Pectra fork, read from the `EigenPodManager` contract
+    /// @dev Specifically, this returns the timestamp of the first non-missed slot at or after the Pectra hard fork
+    function _getPectraForkTimestamp() internal view returns (uint64) {
```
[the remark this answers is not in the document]
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1053/commits/032ba90f4b4eafe2e7fb3a9a2a47386688e5a697

src/contracts/pods/EigenPodManagerStorage.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1053/commits/e9a4a6daa4cf871a7d4021af8a85938090a37115

ypatil12: Good catch

src/contracts/libraries/BeaconChainProofs.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1053/commits/af3c486f7698bd608efdd4bb708369eb95007c8a

src/test/integration/mocks/BeaconChainMock_Pectra.t.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1053/commits/621c8c925a4d5d6cab70b2480d385d7ac26ec43d

src/test/integration/mocks/BeaconChainMock_Pectra.t.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1053/commits/621c8c925a4d5d6cab70b2480d385d7ac26ec43d

src/test/integration/IntegrationDeployer.t.sol
ypatil12: Yup, will be resolved once I merge this into `feat/prooftra` and rebase that off of `dev`

src/test/unit/EigenPodUnit.t.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1053/commits/8558842fe1d453847453b4956fa76b8d2feed39b

src/test/integration/IntegrationBase.t.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1053/commits/8558842fe1d453847453b4956fa76b8d2feed39b

src/test/integration/tests/upgrade/Prooftra.t.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1053/commits/4d1b1816d4b411ab75d4c4be3591fd88fe32ef0f

src/test/integration/IntegrationDeployer.t.sol
ypatil12: Resolved now :)

src/test/integration/tests/upgrade/Prooftra.t.sol
ypatil12: True. All non-prooftra tests will use deneb: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1053/commits/62f294328b3252c3e498090772bf02e383f463b2#diff-aa98f20f9910b62427a9fec255d7e69dc360707c27b15e7d05a9a27aacb44f73

src/test/integration/UpgradeTest.t.sol
ypatil12: Yup, update: https://github.com/Layr-Labs/eigenlayer-contracts/commit/62f294328b3252c3e498090772bf02e383f463b2

ypatil12:
Deneb tests against pectra: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1053/commits/a8dd876db4cbfe6a84e098260f9f3a09c0d5eb12

src/contracts/pods/EigenPodManagerStorage.sol
8sunyuan: Isn't the address and uint64 packed in 1 slot in storage? I think 42 -> 41 is correct

src/contracts/libraries/BeaconChainProofs.sol
8sunyuan: _getProofVersion refactor LGTM

8sunyuan (approved):
LGTM

PR #1064, opened 2025-02-03
feat: simplify and enhance EIGEN and bEIGEN token contracts

bowenli86:
**Motivation:**

**Modifications:**

**Result:**


ypatil12:
Please change name to `feat/token-upgrade`

PR #1067, opened 2025-02-04
feat: prooftra upgrade script

ypatil12:
**Motivation:**

Upgrade script for Prooftra

**Modifications:**

`v1.2.0` upgrade script. Key check is on the upgrade timestamp

**Result:**

Prooftra upgrade.

### DIFF
--- a/script/releases/v1.2.0-slashing/1-eoa.s.sol
+++ b/script/releases/v1.2.0-slashing/1-eoa.s.sol
@@ -1,0 +1,103 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.12;
+
+import {EOADeployer} from "zeus-templates/templates/EOADeployer.sol";
+import "../Env.sol";
+
+import "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
+import "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import "src/contracts/libraries/BeaconChainProofs.sol";
+
+contract Deploy is EOADeployer {
+    using Env for *;
+
+    function _runAsEOA() internal override {
+        vm.startBroadcast();
+
+        // Deploy EigenPodBeacon Impl
+        deployImpl({
+            name: type(EigenPod).name,
+            deployedTo: address(new EigenPod({
+                _ethPOS: Env.ethPOS(),
+                _eigenPodManager: Env.proxy.eigenPodManager(),
+                _GENESIS_TIME: Env.EIGENPOD_GENESIS_TIME()
+            }))
+        });
+
+        vm.stopBroadcast();
+    }
+
+    function testDeploy() public virtual {
+        _runAsEOA();
+        _validateNewImplAddresses(false);
+        _validateImplConstructors();
+    }
+
+    /// @dev Validate that the `Env.impl` addresses are updated to be distinct from what the proxy
+    /// admin reports as the current implementation address.
+    ///
+    /// Note: The upgrade script can call this with `areMatching == true` to check that these impl
+    /// addresses _are_ matches.
+    function _validateNewImplAddresses(bool areMatching) internal view {
+        function(address, address, string memory)
+            internal
+            pure assertion = areMatching ? _assertMatch : _assertNotMatch;
+
+        assertion(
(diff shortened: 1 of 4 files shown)


bowenli86:
QQs: who will run the script and decide when to broadcast vs just dryrun it locally? or it it triggered via a CICD job?

ypatil12:
> QQs: who will run the script and decide when to broadcast vs just dryrun it locally? or it it triggered via a CICD job?

We do deployments in warroom call. It is not run in CI, run locally.

script/releases/v1.2.0-slashing/3-queueUnpause.s.sol
```diff
@@ -0,0 +1,104 @@
...
+        prank(Env.opsMultisig())
+    {
+        bytes memory calldata_to_executor = _getCalldataToExecutor_queueUnpause();
+
+        TimelockController timelock = Env.timelockController();
+        timelock.schedule({
+            target: Env.executorMultisig(),
+            value: 0,
+            data: calldata_to_executor,
+            predecessor: QueueUpgradeAndTimestampSetter.getTimelockId(),
```
wadealexc, line 34: oo, nice usage of predecessor! this is a cool feature.

script/releases/v1.2.0-slashing/4-pause.s.sol
```diff
@@ -0,0 +1,44 @@
...
+ * Purpose: Enqueue a transaction which immediately sets
+ *  `EigenPodManager.PAUSED_START_CHECKPOINT=true` and
+ *  `EigenPodManager.PAUSED_EIGENPODS_VERIFY_CREDENTIALS=true`
+ */
+contract Pause is MultisigBuilder, EigenPodPausingConstants {
+    using Env for *;
+
+    function _runAsMultisig() prank(Env.pauserMultisig()) internal virtual override {
+        uint mask = 2 ** PAUSED_START_CHECKPOINT |
+            2 ** PAUSED_EIGENPODS_VERIFY_CREDENTIALS;
```
wadealexc, line 18: I don't think we should allow existing checkpoints to be proven, just to be safe. Leaving `verifyCheckpointProofs` open after the upgrade will default to Pectra logic. Theoretically this should be safe as any proofs will be invalid, but I think we should lock down all proofs until after we have set a fork timestamp and unpaused.

```solidity
uint mask =  2 ** PAUSED_START_CHECKPOINT |
    2 ** PAUSED_EIGENPODS_VERIFY_CREDENTIALS |
    2 ** PAUSED_EIGENPODS_VERIFY_CHECKPOINT_PROOFS |
    2 ** PAUSED_VERIFY_STALE_BALANCE;

Comment (ypatil12) [reply script/releases/v1.2.0-slashing/4-pause.s.sol]: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1067/commits/e0aa4c618a5cd25cbdd27a5d42289c941ce5c1ed

File: script/releases/v1.2.0-slashing/6-script/script.go [hunk 3]
```diff
@@ -0,0 +1,101 @@
...
+	}
+	defer resp.Body.Close()
+
+	var blockResponse SignedBeaconBlockResponse
+	if err := json.NewDecoder(resp.Body).Decode(&blockResponse); err != nil {
+		return fmt.Errorf("failed to decode response: %v", err)
+	}
+
+	timestamp := blockResponse.Data.Message.Body.ExecutionPayload.Timestamp
+	fmt.Printf("Slot timestamp: %s\n", timestamp)
```
8sunyuan, line 98: Good sanity test could be to assert this parsed out timestamp is >= to the fork timestamp based on the respective `forkSlot`

8sunyuan (approved):

script/releases/v1.2.0-slashing/6-script/script.go
ypatil12: Will add post testnet :)

script/releases/v1.2.0-slashing/6-script/script.go
```diff
@@ -0,0 +1,101 @@
...
+	slotNum := args.ForkSlot
+
+	for {
+		opts := &api.BeaconBlockHeaderOpts{Block: strconv.FormatUint(slotNum, 10)}
+		_, err := httpClient.BeaconBlockHeader(ctx, opts)
+		if err != nil {
+			fmt.Printf("Slot %d was missed, checking next slot...\n", slotNum)
+			slotNum++
+			continue
+		}
```
wadealexc, line 72: are there any other error types we should filter for here? i feel like we upgraded our error handling for the slashing mainnet release

PR #1077, opened 2025-02-10
feat: add `getWithdrawableShares` alias

0xClandestine:
**Motivation:**

Users want a way to query withdrawable shares without needing to provide several params (`staker`, `strategies`). 

**Modifications:**

- Moved `getWithdrawableShares` logic to a helper `_getWithdrawableShares`.
- Added `getWithdrawableShares(bytes32 withdrawalRoot)` fn.

**Result:**

Users can now query withdrawal shares using only a withdrawal root.

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -776,6 +776,30 @@ contract DelegationManager is
         }
     }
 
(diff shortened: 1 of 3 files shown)


0xClandestine:
@ypatil12 607 bytes of codesize remaining for DM...

PR #1078, opened 2025-02-10
feat: add `getSharesFromQueuedWithdrawal`

0xClandestine:
**Motivation:**

Users want a cost effective way to query withdrawable shares.

**Modifications:**

- added `getSharesFromQueuedWithdrawal(bytes32 withdrawalRoot)`
- renamed `getQueuedWithdrawals` to `getSharesFromQueuedWithdrawals`.

**Result:**

Users can query withdrawable shares with `withdrawalRoot` alone.

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -777,7 +777,7 @@ contract DelegationManager is
     }
 
     /// @dev Get the shares from a queued withdrawal.
-    function _getSharesFromQueuedWithdrawal(
+    function _getSharesByWithdrawalRoot(
         bytes32 withdrawalRoot
     ) internal view returns (Withdrawal memory withdrawal, uint256[] memory shares) {
         withdrawal = queuedWithdrawals[withdrawalRoot];
@@ -953,11 +953,11 @@ contract DelegationManager is
     function getSharesFromQueuedWithdrawal(
         bytes32 withdrawalRoot
     ) external view returns (Withdrawal memory withdrawal, uint256[] memory shares) {
-        (withdrawal, shares) = _getSharesFromQueuedWithdrawal(withdrawalRoot);
+        (withdrawal, shares) = _getSharesByWithdrawalRoot(withdrawalRoot);
     }
 
     /// @inheritdoc IDelegationManager
-    function getSharesFromQueuedWithdrawals(
+    function getQueuedWithdrawals(
         address staker
     ) external view returns (Withdrawal[] memory withdrawals, uint256[][] memory shares) {
         bytes32[] memory withdrawalRoots = getQueuedWithdrawalRoots(staker);
@@ -967,7 +967,7 @@ contract DelegationManager is
         shares = new uint256[][](totalQueued);
 
         for (uint256 i; i < totalQueued; ++i) {
-            (withdrawals[i], shares[i]) = _getSharesFromQueuedWithdrawal(withdrawalRoots[i]);
+            (withdrawals[i], shares[i]) = _getSharesByWithdrawalRoot(withdrawalRoots[i]);
         }
     }
 
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -784,7 +784,7 @@ contract DelegationManager is
         shares = new uint256[](withdrawal.strategies.length);
 
         uint32 slashableUntil = withdrawal.startBlock + MIN_WITHDRAWAL_DELAY_BLOCKS;
-        
+
         // If the slashableUntil block is in the past, read the slashing factors at that block.
         // Otherwise, read the current slashing factors. Note that if the slashableUntil block is the current block
         // or in the future, then the slashing factors are still subject to change before the withdrawal is completable,
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -952,8 +952,8 @@ contract DelegationManager is
     /// @inheritdoc IDelegationManager
     function getSharesFromQueuedWithdrawal(
         bytes32 withdrawalRoot
-    ) external view returns (Withdrawal memory withdrawal, uint256[] memory shares) {
-        (withdrawal, shares) = _getSharesByWithdrawalRoot(withdrawalRoot);
+    ) external view returns (uint256[] memory shares) {
+        (, shares) = _getSharesByWithdrawalRoot(withdrawalRoot);
     }
 
     /// @inheritdoc IDelegationManager
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -776,6 +776,41 @@ contract DelegationManager is
         }
     }
 
+    /// @dev Get the shares from a queued withdrawal.
+    function _getSharesFromQueuedWithdrawal(
+        bytes32 withdrawalRoot
+    ) internal view returns (Withdrawal memory withdrawal, uint256[] memory shares) {
+        withdrawal = queuedWithdrawals[withdrawalRoot];
+        shares = new uint256[](withdrawal.strategies.length);
+
+        address operator = delegatedTo[withdrawal.staker];
+        uint32 slashableUntil = withdrawal.startBlock + MIN_WITHDRAWAL_DELAY_BLOCKS;
+
+        uint256[] memory slashingFactors;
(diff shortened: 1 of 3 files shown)


src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -477,8 +477,13 @@ interface IDelegationManager is ISignatureUtils, IDelegationManagerErrors, IDele
         bytes32 withdrawalRoot
     ) external view returns (Withdrawal memory);
 
-    /// @notice Returns a list of pending queued withdrawals for a `staker`, and the `shares` to be withdrawn.
-    function getQueuedWithdrawals(
+    /// @notice Returns the shares from a queued withdrawal.
```
ypatil12, line 480: Would be more specific with a dev comment: `The shares are what a user would receive from completing a queued withdrawal, assuming all slashings are applied`. Can also add to the array-ified version

src/contracts/interfaces/IDelegationManager.sol
```diff
@@ -477,8 +477,13 @@ interface IDelegationManager is ISignatureUtils, IDelegationManagerErrors, IDele
         bytes32 withdrawalRoot
     ) external view returns (Withdrawal memory);
 
-    /// @notice Returns a list of pending queued withdrawals for a `staker`, and the `shares` to be withdrawn.
-    function getQueuedWithdrawals(
```
ypatil12, line 481: Changing this name has a lot of offchain implications (LRTs, CLI that completes withdrawals, AVSs that have already integrated). Would prefer keeping interface the same if possible

src/test/unit/DelegationUnit.t.sol
```diff
@@ -1114,7 +1114,7 @@ contract DelegationManagerUnitTests is EigenLayerUnitTestSetup, IDelegationManag
                 delegationManager.pendingWithdrawals(withdrawalRootToCheck),
                 "withdrawalRoot not pending"
             );
-            (Withdrawal[] memory withdrawalsInStorage, ) = delegationManager.getQueuedWithdrawals(staker);
```
ypatil12, line 1117: Can we add a unit test that specifically checks `getSharesFromQueuedWithdrawal` (single version). Will ensure coverage doesn't go down

ypatil12 (changes requested):

src/contracts/core/DelegationManager.sol
```diff
@@ -776,6 +776,41 @@ contract DelegationManager is
...
+                _getSlashingFactors({staker: withdrawal.staker, operator: operator, strategies: withdrawal.strategies});
+        }
+
+        for (uint256 j; j < withdrawal.strategies.length; ++j) {
+            shares[j] = SlashingLib.scaleForCompleteWithdrawal({
+                scaledShares: withdrawal.scaledShares[j],
+                slashingFactor: slashingFactors[j]
+            });
+        }
+    }
```
bowenli86, line 812: my concern is that this view function is essentially doing some heavy computing on the fly, which is really inefficient, eg consecutive read with same root would require it to compute again and again


Whereas a more proper way is to precompute everything and the view function only does a quick lookup read

is there a way to make it the 2nd pattern?

src/test/unit/DelegationUnit.t.sol
0xClandestine: oh rip, had unit tests in the original pr.

src/contracts/core/DelegationManager.sol
0xClandestine: Are you suggesting tracking this in storage?

ypatil12: Ya, a bit confused on the ask here @bowenli86? This value can't be stored because its dynamic (slashing while in the queue)

ypatil12: Remember, not all view functions need to be called on-chain, some are designed to be called off-chain too, where the gas limit is not the same

src/contracts/interfaces/IDelegationManager.sol
0xClandestine: [238b386](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1078/commits/238b386e06e135d977ac856221bf816f4fe63f2f)

src/contracts/interfaces/IDelegationManager.sol
0xClandestine: [9698172](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1078/commits/9698172b7f2d9287ddde703d57d8b619748b789e)

src/test/unit/DelegationUnit.t.sol
0xClandestine: [fd23e3b](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1078/commits/fd23e3b9135443d8cde720a14c3d5858601b5eda)

src/contracts/core/DelegationManager.sol
```diff
@@ -776,6 +776,41 @@ contract DelegationManager is
         }
     }
 
+    /// @dev Get the shares from a queued withdrawal.
+    function _getSharesFromQueuedWithdrawal(
```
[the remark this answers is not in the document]
0xClandestine: [9698172](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1078/commits/9698172b7f2d9287ddde703d57d8b619748b789e)

src/contracts/core/DelegationManager.sol
bowenli86: yea the concern is this seems not to be a view function for either onchain or offchain, b/c it's doing a bunch of complex logic rather than just get sth out of shelf.

but i see your point of it being dynamic, especially it's expensive to update the value every block

not sure what's the best design pattern for such cases. 
not a blocker

ypatil12: There are many view functions that do complex calculations on chain, there are just different flavors of it

src/contracts/core/DelegationManager.sol
```diff
@@ -776,6 +776,41 @@ contract DelegationManager is
...
+        bytes32 withdrawalRoot
+    ) internal view returns (Withdrawal memory withdrawal, uint256[] memory shares) {
+        withdrawal = queuedWithdrawals[withdrawalRoot];
```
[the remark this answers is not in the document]
0xClandestine: If a `Withdrawal` does not exist for it's corresponding `withdrawalRoot`, a null shares array will be returned.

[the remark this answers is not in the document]
0xClandestine: [169e6f9](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1078/commits/169e6f91154fab0246bacb34fd653980f23d1a4a)

src/contracts/core/DelegationManager.sol
```diff
@@ -914,6 +949,13 @@ contract DelegationManager is
         return queuedWithdrawals[withdrawalRoot];
     }
 
+    /// @inheritdoc IDelegationManager
+    function getSharesFromQueuedWithdrawal(
+        bytes32 withdrawalRoot
+    ) external view returns (Withdrawal memory withdrawal, uint256[] memory shares) {
+        (withdrawal, shares) = _getSharesByWithdrawalRoot(withdrawalRoot);
```
ypatil12, line 956: Can we just return the shares? If the user has the withdrawal root then they wouldn't need the withdrawal

ypatil12: This also works for Renzo

src/contracts/core/DelegationManager.sol
```diff
@@ -924,37 +966,8 @@ contract DelegationManager is
         withdrawals = new Withdrawal[](totalQueued);
         shares = new uint256[][](totalQueued);
 
-        address operator = delegatedTo[staker];
```
8sunyuan, line 927: Looks like this was a bug thats fixed now. If you had redelegated your old queued withdrawals would be looking at the slashingFactors of your current operator.

ypatil12: We should add a regression test for this. cc @0xClandestine

ypatil12: It's still looking at the queued withdrawals of the current operator? We should change to using `withdraw.delegatedTo` in the helper function and add a regression test

ypatil12 (approved):

8sunyuan (approved):

PR #1085, opened 2025-02-11
fix: avs registrar interface

8sunyuan:
**Motivation:**

Currently there is no way for an AVS implementing these callback functions of the `IAVSRegistrar.sol` to verify that the operatorSetIds belong to their operatorSets. This is because AVSRegistrar configuration does not have uniqueness and implementing uniqueness in the AllocationManager is not feasible due to frontrunning DOS vectors for existing AVSs.

**Modifications:**

`IAVSRegistrar.sol` interface

**Result:**

- Allowing AVSs to verify the correct operatorSet is being passed into their register/deregister callback functions and whether or not to proceed with the Operator action.
- Downstream breaking interface changes for AVSs

### DIFF
--- a/script/tasks/register_operator_to_operatorSet.s.sol
+++ b/script/tasks/register_operator_to_operatorSet.s.sol
@@ -11,8 +11,8 @@ import "forge-std/Test.sol";
 
 // Define dummy AVSRegistrar contract to prevent revert
 contract AVSRegistrar is IAVSRegistrar {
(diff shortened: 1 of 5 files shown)


ypatil12 (approved):

8sunyuan:
> LGTM - needs to pass CI + tests though. Is there a MW PR too?

https://github.com/Layr-Labs/eigenlayer-middleware/pull/398

ypatil12:
Seems like we're going with this one now?

PR #1092, opened 2025-02-12
fix: set registrar check

8sunyuan:
**Motivation:**

https://github.com/Layr-Labs/eigenlayer-contracts/pull/1085
An alternate solution to this problem with less intrusive changes to AVS integrations.

**Modifications:**

Adding a `avs()` view function to `IAVSRegistrar.sol` and an additional require check in `setAVSRegistrar`

**Result:**

Uniqueness of AVS registrars set per AVS.


8sunyuan:
hotfix to dev here https://github.com/Layr-Labs/eigenlayer-contracts/pull/1102

PR #1094, opened 2025-02-13
feat: parallelize all test suites to expedite CI run time

bowenli86:
**Motivation:**

parallelize unit, integration and fork tests to make CI time significantly faster

note currently tho the github action is called "testinparallel", unit, integration and fork tests are not parallelized

their time:
- unit test: 2-3 min
- integration test: 9-10 min
- fork test: 12-13 min

they run in sequence so the total time is sum(all tests) which is roughly 25-26 min. See https://github.com/Layr-Labs/eigenlayer-contracts/actions/runs/13208807194/job/36878141793?pr=1025 for example

by parallelizing the three test suites, we shrink the CI test to ~11-12 min in total https://github.com/Layr-Labs/eigenlayer-contracts/actions/runs/13209105775?pr=1074

- unit test: 2-3 min
- integration test: 9-10 min (total CI time is bounded by this test now)
- fork test: 6-7 min (not exactly sure why this got 2x improvement down from 12-13 min, my theory is that when running in sequence previously, integration tests run before fork test, and integration tests may hv resources not shut down completely, leading to significantly slow execution of fork test)

**Modifications:**

parallelize unit, integration and fork tests to make CI time significantly faster

**Result:**

### DIFF
--- a/.github/workflows/parallel_tests.yml
+++ b/.github/workflows/parallel_tests.yml
@@ -1,0 +1,40 @@
+name: Run Parallel Tests
+
+on:
+  push:
+  pull_request:
+    types: [opened, reopened]
+
+jobs:
+  test-suite:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        suite: [unit, integration, fork]
+    
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: stable
+
+      - name: Forge build
+        run: |
+          forge --version
+          forge build
+
+      - name: Run ${{ matrix.suite }} tests
+        run: |
+          case "${{ matrix.suite }}" in
+            unit) forge test --no-match-contract Integration ;;
+            integration) forge test --match-contract Integration ;;
+            fork) forge test --match-contract Integration ;;
+          esac
+        env:
+          FOUNDRY_PROFILE: ${{ matrix.suite == 'fork' && 'forktest' || '' }}
+          RPC_MAINNET: ${{ secrets.RPC_MAINNET }}
+          RPC_HOLESKY: ${{ secrets.RPC_HOLESKY }}
+          CHAIN_ID: ${{ secrets.CHAIN_ID }}


bowenli86:
synced with @0xClandestine and have migrated to a new PR in order to target dev branch

ypatil12:
The CI now re-runs parallelized and non-parallelized tests, we should do one or the other

.github/workflows/parallel_tests.yml
```diff
@@ -0,0 +1,40 @@
...
+    types: [opened, reopened]
+
+jobs:
+  test-suite:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        suite: [unit, integration, fork]
+    
+    steps:
```
0xClandestine, line 15: Rather than adding another file, you should modify `.github/workflows/foundry.yml`.

0xClandestine: Can we also get some up to date time comparisons for this branch (since the old ones were based on `slashing-magnitudes-fixes`.

0xClandestine (changes requested):
Nice work, just need to reorganize a bit.

.github/workflows/parallel_tests.yml
0xClandestine: image

Also ideally we only trigger one workflow, for either `push` or `pull request`.

bowenli86:
image

looks good

i've given it a larger runner with 16cpu, seems have brought down integration tests further down from 12min to 6min, so the total speedup is 4X now with 24min -> 6min

.github/workflows/parallel_tests.yml
0xClandestine: image

Can we rename such that it reads like:

```
Foundry / Test (Unit)
Foundry / Test (Integration)
Foundry / Test (Fork)
```

to maintain naming consistency with original implementation.

0xClandestine: [cc129e1](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1094/commits/cc129e108ed43bd563b2f35b57044d79d17b2ba1)

0xClandestine (dismissed):

0xClandestine (dismissed):

0xClandestine (approved):

ypatil12 (approved):

PR #1101, opened 2025-02-14
refactor: limit continuous fuzzing to dev branch

bowenli86:
**Motivation:**

per chat w/ @0xClandestine , given continuous fuzzing tests takes too long (currently ~2.5h ish), we wanna make it only run when a commit is merged to dev branch, instead of every PR change

it should make the CI passes as a whole a lot faster, and save us huge money spent on github actions

**Modifications:**

make continuous fuzzing only run when a commit is merged to dev branch, instead of every PR change

**Result:**

### DIFF
--- a/.github/workflows/foundry.yml
+++ b/.github/workflows/foundry.yml
@@ -67,6 +67,8 @@ jobs:
   continuous-fuzzing:
     name: Test (Intense)
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


bowenli86:
image

you can see as part of this PR that continuous fuzzing has been skipped

0xClandestine (approved):
lgtm

PR #1106, opened 2025-02-14
fix: dont call strategy if we have no burnable shares

wadealexc:
**Motivation:**

Fixes an issue arbitrary external contracts could be called via `StrategyManager.burnShares`. (Certora L-04)

**Modifications:**

`StrategyManager.burnShares` does not do an external call if the burnable share amount is zero

**Result:**

Should no longer be possible to call untrusted code directly through `burnShares`


ypatil12 (approved):

8sunyuan (approved):
Just realized this would prevent reverts from Strategies that have a revert on 0 amount transfers

PR #1107, opened 2025-02-14
fix: reentrant guards

8sunyuan:
**Motivation:**

Concerns about reentrancy in the DelegationManager and interactions of completed withdrawals which can call untrusted ERC20 transfers

**Modifications:**

Added reentrant guards across external functions

**Result:**

Preventing cross-function reentrancy in the DelegationManager


wadealexc (approved):

0xClandestine (approved):
LGTM!

PR #1112, opened 2025-02-15
perf: run Foundry CI tests only for relevant code changes

bowenli86:
**Motivation:**  

Foundry tests only need to run when specific codebase paths are modified, rather than for all changes.  

For example, tests are not required when only documentation or deployment scripts are updated.  

**Modifications:**  

- Added a list of code paths that trigger Foundry tests when modified.  

**Result:**  

- Speeds up CI for non-core contract changes.  
- Reduces CI costs.

### DIFF
--- a/.github/workflows/foundry.yml
+++ b/.github/workflows/foundry.yml
@@ -3,9 +3,21 @@ name: Foundry
 on:
   workflow_dispatch:
   pull_request:
+    paths:
+      - '.github/**'
+      - 'src/**'
+      - 'lib/**'
+      - 'certora/**'
+      - 'foundry.toml'
   push:
     branches:
       - "dev"
+    paths:
+      - '.github/**'
+      - 'src/**'
+      - 'lib/**'
+      - 'certora/**'
+      - 'foundry.toml'
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/foundry.yml
```diff
@@ -3,9 +3,23 @@ name: Foundry
 on:
   workflow_dispatch:
   pull_request:
+    paths:
+      - '.github/**'
+      - 'bin/**'
+      - 'src/**'
+      - 'lib/**'
+      - 'certora/**'
+      - 'foundry.toml'
```
0xClandestine, line 12: Would suggest filtering by file type, rather than directory. We should always rerun ci if solidity is modified:

```suggestion
      - '.github/**'
      - './**/**.sol'
```

0xClandestine (changes requested):
Great idea, small change requested 🙏

0xClandestine:
Why aren't the foundry tests being skipped?

0xClandestine:
> Why aren't the foundry tests being skipped?

Oh because .github/ is in path

ypatil12 (approved):

PR #1115, opened 2025-02-15
perf: speed up test coverage job in CI

bowenli86:
**Motivation:**  

After multiple improvements to other parts of our CI, the coverage job has become the longest-running task, limiting the total CI runtime.  

This PR significantly reduces the coverage runtime from **16-18 minutes** (already down from **26 minutes** thanks to a previous improvement using an x64 16-core runner) to just **2 minutes**—an **8-9x speedup**.  

**Modifications:**  

- Parallelized coverage report generation.  
- Cached Foundry and Cargo artifacts.  
- Optimized Forge arguments.

**Result:**  

The coverage runtime is reduced from **16-18 minutes** (previously **26 minutes**) to **2 minutes**, achieving an **8-9x improvement**.

### DIFF
--- a/.github/workflows/foundry.yml
+++ b/.github/workflows/foundry.yml
@@ -183,7 +183,7 @@ jobs:
       # Run Forge coverage with LCOV report format, excluding test and script files
       - name: Forge Coverage
         run: |
-          FOUNDRY_DENY_WARNINGS=false FOUNDRY_PROFILE=ci forge coverage --report lcov --report summary --no-match-coverage "script|test"
+          FOUNDRY_DENY_WARNINGS=false FOUNDRY_PROFILE=ci forge coverage --report lcov --report summary --no-match-coverage "script|test" -j $(nproc)
           genhtml -q -o report ./lcov.info
 
       # Upload coverage report as artifact before potential failure
--- a/.github/workflows/foundry.yml
+++ b/.github/workflows/foundry.yml
@@ -204,11 +204,6 @@ jobs:
           FOUNDRY_CACHE_PATH=cache \
           forge coverage --report lcov --report summary --no-match-coverage "script|test" -j $(nproc)
           genhtml -q -o report ./lcov.info
-        env:
-          FOUNDRY_FUZZ_RUNS: 1  # Reduce fuzz runs for coverage
-          FOUNDRY_MATCH_TEST: ""  # Don't filter tests
-          FOUNDRY_INVARIANT_RUNS: 1  # Reduce invariant test runs
-          FOUNDRY_FUZZ_MAX_TEST_REJECTS: 65536
 
       # Upload coverage report as artifact before potential failure
       - name: Upload Coverage Report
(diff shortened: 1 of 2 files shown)


bowenli86:
image

.github/workflows/foundry.yml
```diff
@@ -167,15 +180,27 @@ jobs:
...
+          FOUNDRY_DENY_WARNINGS=false \
+          FOUNDRY_PROFILE=ci \
+          FOUNDRY_OPTIMIZER=true \
+          FOUNDRY_OPTIMIZER_RUNS=1000000 \
+          FOUNDRY_CACHE=true \
+          FOUNDRY_CACHE_PATH=cache \
+          forge coverage --report lcov --report summary --no-match-coverage "script|test" -j $(nproc)
           genhtml -q -o report ./lcov.info
+        env:
+          FOUNDRY_FUZZ_RUNS: 1  # Reduce fuzz runs for coverage
```
0xClandestine, line 192: This should be defined in foundry.toml as a profile.

0xClandestine:
Surprised this worked, pre parallelism caching led to longer ci times.

0xClandestine (changes requested):
One suggestion, looks good otherwise.

bowenli86:
the most expensive part now of coverage is now installing loc

.github/workflows/foundry.yml
```diff
@@ -167,7 +180,13 @@ jobs:
       # Run Forge coverage with LCOV report format, excluding test and script files
       - name: Forge Coverage
         run: |
-          FOUNDRY_DENY_WARNINGS=false FOUNDRY_PROFILE=ci forge coverage --report lcov --report summary --no-match-coverage "script|test"
+          FOUNDRY_DENY_WARNINGS=false \
```
0xClandestine, line 183: Any environment variables that aren't `FOUNDRY_PROFILE` should be in a foundry profile.

0xClandestine, line 183: Any environment variables that aren't FOUNDRY_PROFILE should be in a foundry profile.

0xClandestine (changes requested):

foundry.toml
```diff
@@ -91,9 +91,10 @@
 #         "./src/contracts/**/*"
 #     ]
 
-[profile.ci.fuzz]
-    optimizer = false
-    runs = 256
```
0xClandestine, line 96: We still want a somewhat large number of runs for ci generally, for coverage a lower amount makes sense though.

0xClandestine (changes requested):

foundry.toml
bowenli86: make sense, we can split it into 2 profiles: 'ci-coverage' and 'ci-test' with different runs value

0xClandestine (approved):

ypatil12 (approved):

ypatil12:
🔥🔥🔥

PR #1118, opened 2025-02-17
perf: use gcovr instead of lcov

0xClandestine:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/.github/workflows/foundry.yml
+++ b/.github/workflows/foundry.yml
@@ -167,11 +167,11 @@ jobs:
         with:
           version: stable
 
-      # Install LCOV for coverage report generation.
-      - name: Install LCOV
+      # Install gcovr for coverage report generation
+      - name: Install gcovr
         run: |
-          sudo apt-get install lcov
-        id: lcov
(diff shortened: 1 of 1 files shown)


.github/workflows/foundry.yml
```diff
@@ -167,24 +167,18 @@ jobs:
...
+          sudo apt-get install -y gcovr
+        id: gcovr
 
       # Run Forge coverage with LCOV report format, excluding test and script files
       - name: Forge Coverage
         run: |
+          mkdir -p report
           FOUNDRY_DENY_WARNINGS=false FOUNDRY_PROFILE=ci forge coverage --report lcov --report summary --no-match-coverage "script|test"
-          genhtml -q -o report ./lcov.info
+          gcovr --html-details report/index.html -r . --exclude "script/.*" --exclude "test/.*" --fail-under-line 90 --fail-under-function 90 lcov.info
```
0xClandestine, line 181: Shouldn't fail under 90 here, since that logic is handled below.

PR #1121, opened 2025-02-17
perf: correctly set profile optimizer

bowenli86:
**Motivation:**

there's no setting of `<profile>.fuzz.optimizer`, only `<profile>.optimizer`

see
- https://book.getfoundry.sh/reference/config/solidity-compiler#optimizer
- https://book.getfoundry.sh/reference/config/testing#fuzz

**Modifications:**

move 'fuzz' out of profile to be prefix of 'runs' vars

**Result:**

make '<profile>.optimizer' setting work

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -91,21 +91,21 @@
 #         "./src/contracts/**/*"
 #     ]
 
-[profile.forktest.fuzz]    
+[profile.forktest]    
     optimizer = false
-    runs = 16
+    fuzz.runs = 16
 
-[profile.coverage.fuzz]
+[profile.coverage]
     optimizer = false
-    runs = 1
+    fuzz.runs = 1
 
-[profile.medium.fuzz]
+[profile.medium]
     optimizer = false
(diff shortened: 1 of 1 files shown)


bowenli86:
hmm, looks like 1st commit https://github.com/Layr-Labs/eigenlayer-contracts/pull/1121/commits/a5cc181965421c20b75ff895e7b25c5be145cadb does not work, and somehow broke `forge build`

thus, i'm removing optimizer entirely, given it by default is false, according to foundry doc above

0xClandestine:
> hmm, looks like 1st commit [a5cc181](https://github.com/Layr-Labs/eigenlayer-contracts/commit/a5cc181965421c20b75ff895e7b25c5be145cadb) does not work, and somehow broke `forge build`
> 
> thus, i'm removing optimizer entirely, given it by default is false, according to foundry doc above

We set it to `true` in our foundry.toml, thus I think we should be turning it off in these profiles.

PR #1123, opened 2025-02-18
perf: install lcov via prebuilt binary, instead of via apt

bowenli86:
**Motivation:**

after a few rounds of optimization, the whole CI suite is now bounded by `coverage` job which is then bounded by the `install lcov` step that takes 3.5min to install lcov via apt

this PR changes to install lcov via prebuilt binary, instead of via apt, which shortens the installation from 3.5min to 1s

**Modifications:**

this PR changes to install lcov via prebuilt binary, instead of via apt

**Result:**

shortens the installation from 3.5min to 1s

### DIFF
--- a/.github/workflows/foundry.yml
+++ b/.github/workflows/foundry.yml
@@ -181,10 +181,11 @@ jobs:
           version: stable
 
       # Install LCOV for coverage report generation.
-      - name: Install LCOV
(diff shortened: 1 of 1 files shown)


0xClandestine (approved):
Sick, I was exploring docker containers with the binary pre-installed, this is even better. LGTM!

PR #1105, opened 2025-02-14
docs: overhaul README.md

0xClandestine:
**Motivation:**  

The previous README lacked organization and visual clarity, making it harder for users to quickly grasp the project's purpose and key details. Enhancing its structure and aesthetics improves accessibility and engagement.  

**Modifications:**  

- Added badges for key project stats (e.g., build status, coverage, license).  
- Reorganized sections for better flow and readability.  
- Included a banner for a more professional and visually appealing look.  

**Result:**  

A clearer, more structured, and visually engaging README that makes it easier for users and contributors to understand and navigate the project. 

NOTE: Checkout the new [README](https://github.com/Layr-Labs/eigenlayer-contracts/tree/docs/overhaul-readme?tab=readme-ov-file#-eigenlayer).

PR #1124, opened 2025-02-18
docs: added license shield

operagxsasha:
**Motivation:**

**Modifications:**

**Result:**


0xClandestine:
#1105

PR #1127, opened 2025-02-19
docs: bc/avs slashing edge case

ypatil12:
**Motivation:**

**Modifications:**

**Result:**


docs/core/accounting/SlashingEdgeCase.md
```diff
@@ -0,0 +1,117 @@
...
+    <details>
+    <summary>Calculation</summary>
+
+    * `depositShares: 32` 
+    * `maxMagnitude: 0.25`
+    * `BCSF: 1`
+    * `DSF: 1`
+    * `withdrawable = 32 * 0.25 * 1 * 1 = 8 ETH`
+    </details>
+3. Alice is slashed by 16 ETH on the beacon chain
```
0xClandestine, line 26: This renders as a drop down, kinda odd.

0xClandestine (approved):
One suggestion, looks good otherwise.

docs/core/accounting/SlashingEdgeCase.md
ypatil12: Yup, was by design. Wanted to not clutter the PR

wadealexc (approved):

PR #1133, opened 2025-02-19
refactor: rename inconsistent `getQueuedWithdrawal` alias

0xClandestine:
**Motivation:**

Naming between the `getQueuedWIthdrawal` aliases was inconsistent.

**Modifications:**

- ~made `queuedWithdrawals[withdrawalRoot]` mapping public.~
- renamed `queuedWithdrawals` -> `_queuedWithdrawals`.
- added `_queuedWithdrawals` getter
- removed previous `getQueuedWithdrawal` alias.
- renamed `getQueuedWithdrawalFromRoot` to `getQueuedWithdrawal`. 

**Result:**

Consistent function naming.

### DIFF
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -963,16 +963,9 @@ contract DelegationManager is
 
         return (strategies, shares);
     }
-
+    
     /// @inheritdoc IDelegationManager
     function getQueuedWithdrawal(
-        bytes32 withdrawalRoot
-    ) external view returns (Withdrawal memory) {
-        return queuedWithdrawals[withdrawalRoot];
-    }
-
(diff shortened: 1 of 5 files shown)


ypatil12 (approved):

wadealexc (changes requested):
Public mappings return tuples rather than structs

If we're good with this route, let's explicitly define a public function:

```
function queuedWithdrawals(bytes32 root) public view returns (Withdrawal) {
    return _queuedWithdrawals[root];
}
```

0xClandestine:
> function queuedWithdrawals(bytes32 root) public view returns (Withdrawal) {
>     return _queuedWithdrawals[root];
> }

Ahh, that's where the bytecode savings were coming from.

wadealexc (approved):

PR #1136, opened 2025-02-19
test: integration multi staker

8sunyuan:
**Motivation:**

Currently integration tests mainly simulate generating a single staker delegated to an operator.

**Modifications:**

Added deploy functions and modified User.t.sol

**Result:**

Draft for writing a test with User[].
TODO: 
- Updated assertions used in clean way
- Updated beaconChainMock to account for validators and multiple stakers

### DIFF
--- a/src/test/integration/IntegrationBase.t.sol
+++ b/src/test/integration/IntegrationBase.t.sol
@@ -69,6 +69,23 @@ abstract contract IntegrationBase is IntegrationDeployer, TypeImporter {
         return (staker, strategies, tokenBalances);
     }
 
+    /// @dev creates `numNewStakers` new stakers with random strategies and token balances
+    function _newRandomStakers(uint numNewStakers) internal returns (User[] memory, IStrategy[][] memory, uint[][] memory) {
+        User[] memory stakers = new User[](numNewStakers);
+        IStrategy[][] memory strategies = new IStrategy[][](numNewStakers);
+        uint[][] memory tokenBalances = new uint[][](numNewStakers);
+
+        for (uint i = 0; i < numNewStakers; i++) {
+            (stakers[i], strategies[i], tokenBalances[i]) = _newRandomStaker();
+        }
+
+        numStakers += numNewStakers;
+    
+        return (stakers, strategies, tokenBalances);
+    }
+
+    /// @dev Exactly like _newRandomStaker above but sets strategies and tokenBalances arrays to length 1
+    /// to use a single strategy and balance
     function _newBasicStaker() internal returns (User, IStrategy[] memory, uint[] memory) {
         string memory stakerName;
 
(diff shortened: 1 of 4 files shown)


src/test/integration/users/User.t.sol
```diff
@@ -275,9 +297,35 @@ contract User is Logger, IDelegationManagerTypes, IAllocationManagerTypes {
...
+    /// @dev NOTE: do NOT call this function directly. This is simply implemented to manage the
+    /// delegatedStakers enum set. This function is only meant to be called on the Operator User contract.
+    function handleUndelegatingStaker(address staker) public virtual {
+        // check that the current User is an operator
+        require(
+            delegationManager.isOperator(address(this)),
+            "User is not an operator"
+        );
+        delegatedStakers.remove(staker);
+    }
```
wadealexc, line 327: I would call these `add/remove` rather than "handle"

src/test/integration/users/User.t.sol
```diff
@@ -275,9 +297,35 @@ contract User is Logger, IDelegationManagerTypes, IAllocationManagerTypes {
...
     }
 
+    /// @dev NOTE: do NOT call this function directly. This is simply implemented to manage the
+    /// delegatedStakers enum set. This function is only meant to be called on the Operator User contract.
+    function handleDelegatingStakers(address staker) public virtual {
+        // check that the current User is an operator
+        require(
+            delegationManager.isOperator(address(this)),
+            "User is not an operator"
+        );
```
wadealexc, line 314: Let's leave the checks out of this. If you're failing these, you'll see failures elsewhere - so let's keep these methods as simple as possible.

src/test/integration/IntegrationBase.t.sol
```diff
@@ -2307,6 +2324,21 @@ abstract contract IntegrationBase is IntegrationDeployer, TypeImporter {
         return expectedShares;
     }
 
+    /// @dev For some strategies/underlying token balances, calculate the expected shares received
+    /// Exact same as _calculateExpectedShares but for multiple stakers
+    function _calculateExpectedShares(
+        IStrategy[][] memory strategies,
+        uint[][] memory tokenBalances
```
wadealexc, line 2331: The arrayification of integration tests is getting unwieldy. It might be time to start introducing some helper structs.

src/test/integration/tests/multiStaker/Deposit_Delegate_Queue_Complete.t.sol
```diff
@@ -0,0 +1,53 @@
...
+        // 2. Delegate to an operator
+        stakers.delegateTo(operator);
+        // check_Delegation_State(staker, operator, strategies, shares);
+
+        // 3. Queue Withdrawals
+        Withdrawal[][] memory withdrawals = stakers.queueWithdrawals(strategiesPerStaker, sharesPerStaker);
+
+        // 4. Complete withdrawal
+        _rollBlocksForCompleteWithdrawals(withdrawals[0]);
+        stakers.completeWithdrawalsAsTokens(withdrawals);
```
wadealexc, line 51: Hmm. The problem with this is the snapshot system is only going to capture the final staker to make this call. So if we did an invariant check after this, the Snap assertions will be using the last staker in the list.

I think it might be better to continue handling each staker individually in terms of the actions they take. After all, we want to check invariants on each, right?

PR #1137, opened 2025-02-20
test: more comprehensive invariant assertions

ypatil12:
**Motivation:**

Update invariants to be more comprehensive

**Modifications:**

- Redelegate -> check new operator shares
- Queue -> check DSF reset or unchanged

**Result:**

More comprehensive assertions


ypatil12:
Looks like @eigenmikem has added these, will keep up till that is merged

PR #1149, opened 2025-02-20
test: improve slashing invariants and refactor user generation logic

wadealexc:
**Motivation:**

Improve slashing invariants, fix existing DM/ALM invariants, and refactor user generation logic

**Modifications:**

* Improve `IntegrationDeployer._randUser` generation logic so that additional helpers can be implemented to generate users with specific assets
* Update some withdrawal invariants to distinguish between deposit and withdrawable shares. More work to be done here.
* Implement several new tests in `SlashingWithdrawals`
* Removes the vscode settings file since it was autoformatting tests. Spoke with Rajath about this - i'm fine to revisit this after slashing, but the current impact of this file is that it will create massive diffs across the most active part of the codebase while we're trying to get critical last-mile testing finished. We'll readd this after slashing :+1: 

**Followup**:

* Additional cleanup for queue/complete withdrawal invariants and calling conventions
* Additional tests needed in `SlashingWithdrawals`
* Additional staker-level invariants needed when slashing an operator


ypatil12 (approved):

src/test/integration/IntegrationBase.t.sol
```diff
@@ -69,6 +69,31 @@ abstract contract IntegrationBase is IntegrationDeployer, TypeImporter {
...
+    function _newStaker(IStrategy[] memory strategies) internal returns (User, uint[] memory) {
+        string memory stakerName;
+
+        User staker;
+        uint[] memory tokenBalances;
+
+        if (!isUpgraded) {
+            stakerName = string.concat("M2Staker", cheats.toString(numStakers));
+
+            (staker, tokenBalances) = _randUser(stakerName, strategies);
```
8sunyuan, line 82: nit: User is the same here can move this line out of this if block

src/test/integration/IntegrationBase.t.sol
```diff
@@ -671,6 +696,22 @@ abstract contract IntegrationBase is IntegrationDeployer, TypeImporter {
...
     }
 
+    function assert_Snap_Removed_AllocatedStrats(
+        User operator,
+        OperatorSet memory operatorSet,
+        IStrategy[] memory removedStrats,
+        string memory err
+    ) internal {
+        IStrategy[] memory curAllocatedStrats = _getAllocatedStrats(operator, operatorSet);
+        IStrategy[] memory prevAllocatedStrats = _getPrevAllocatedStrats(operator, operatorSet);
```
8sunyuan, line 706: Why do we need prevAllocatedStrats here?

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -547,49 +554,56 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
...
+        // and deal the user some corresponding underlying token balances
+        uint assetType = _randAssetType();
+        IStrategy[] memory strategies = _selectRandAssets(assetType);
+        uint[] memory tokenBalances = _dealRandAmounts(user, strategies);
 
         print.user(name, assetType, userType, strategies, tokenBalances);
         return (user, strategies, tokenBalances);
     }
 
+    function _randUser(
```
8sunyuan, line 571: Note: this and `_newStaker(strategies)` can generate assets for specified strategies vs just random

src/test/integration/IntegrationChecks.t.sol
```diff
@@ -349,7 +355,7 @@ contract IntegrationCheckUtils is IntegrationBase {
             if (operator != staker) {
                 assert_Snap_Unchanged_TokenBalances(operator, "operator should not have any change in underlying token balances");
             }
-            assert_Snap_Added_OperatorShares(operator, withdrawal.strategies, withdrawal.scaledShares, "operator should have received shares");
+            assert_Snap_Added_OperatorShares(operator, strategies, shares, "operator should have received shares");
```
8sunyuan, line 358: Maybe in followup PR, but the `check_Withdrawal_AsShares_*` functions could probably include this call
`uint[] memory expectedShares = _calculateExpectedShares(withdrawals[i]);`
and not have to pass in shares as input. Since we are already passing in the withdrawal

8sunyuan (approved):

src/test/integration/IntegrationChecks.t.sol
wadealexc: yep, good callout. there's a decent bit of cleanup to be done in these assertions

src/test/integration/IntegrationBase.t.sol
wadealexc: Did some cleanup here: [5680efb](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1149/commits/5680efbc991bfba17dbde6171d88afe94e0138db)

PR #1150, opened 2025-02-20
test: slashing invariants

eigenmikem:
**Motivation:**

Implement new slashing invariants and add existing ones to checks.

**Modifications:**

- Explicit calculation of expected withdrawable shares and dsf after delegation and on completion of full withdrawal
- Add "unchanged dsf" assertion
- Add increased/decreased minSlashableStake assertions
- Overload minSlashableStake assertions to handle address input
- Add assertions to slashing flow state checks
- Fix `_choose` helper function

**Result:**

More complete coverage for slashing invariants


ypatil12 (approved):
LGTM, seems like just the slashable shares in queue remaining?

PR #1165, opened 2025-02-25
chore: add color to ci

0xClandestine:
**Motivation:**

**Modifications:**

**Result:**


0xClandestine:
https://github.com/foundry-rs/foundry/issues/7322

Not currently supported...

PR #1175, opened 2025-02-26
test: adding slashable shares in queue

eigenmikem:
**Motivation:**

Slashable shares in queue aren't currently checked by any invariants.

**Modifications:**

- added assertion that slashable shares in queue decrease on slash if a withdrawal is queued
- added assertion that slashable shares in queue increase on queued withdrawal
- changed IntegrationDeployer to initialize starting block as nonzero to address edge cases

**Result:**

Invariant coverage for slashable shares in queue state

### DIFF
--- a/src/test/integration/IntegrationChecks.t.sol
+++ b/src/test/integration/IntegrationChecks.t.sol
@@ -279,6 +279,8 @@ contract IntegrationCheckUtils is IntegrationBase {
             "check_QueuedWithdrawal_State: failed to remove staker shares");
         assert_Snap_Removed_Staker_WithdrawableShares(staker, strategies, withdrawableShares,
             "check_QueuedWithdrawal_State: failed to remove staker withdrawable shares");
+        assert_Snap_Increased_SlashableSharesInQueue(operator, withdrawals,
+            "check_QueuedWithdrawal_State: failed to increase slashable shares in queue");
         check_Decreased_SlashableStake(operator, withdrawableShares, strategies);
         // Check that the dsf is either reset to wad or unchanged
         for (uint i = 0; i < strategies.length; i++) {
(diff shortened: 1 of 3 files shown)


src/test/integration/IntegrationChecks.t.sol
```diff
@@ -997,6 +999,35 @@ contract IntegrationCheckUtils is IntegrationBase {
         // assert_Snap_Unchanged_AllocatedStrats(operator, operatorSet, "should not have updated allocated strategies");
     }
 
+    function check_Base_Slashing_State(
+        User operator,
+        AllocateParams memory allocateParams,
+        SlashingParams memory slashParams,
+        Withdrawal[] memory withdrawals
+    ) internal {
+        OperatorSet memory operatorSet = allocateParams.operatorSet;
```
ypatil12, line 1008: Nit: this can just call the previous function, and then we add the withdrawal specific check

ypatil12 (commented):
Looks good, one minor comment

src/test/integration/IntegrationChecks.t.sol
```diff
@@ -997,6 +999,16 @@ contract IntegrationCheckUtils is IntegrationBase {
         // assert_Snap_Unchanged_AllocatedStrats(operator, operatorSet, "should not have updated allocated strategies");
```
ypatil12, line 999: How can we assert that the slashable shares in queue decreases after a withdrawal is completable ?

ypatil12: Down to add it to roll blocks

ypatil12: Nvm, discussed offline, can put it in `check_Complete_Withdrawal_AsX_State`. We call this function right when withdrawal is completable so this is fine

ypatil12 (approved):
LGTM, we'll add the invariant for slashable shares in queue decreasing after a withdrawal is completable separately (if possible)

PR #1176, opened 2025-02-26
fix: DSF increase on checkpointed nonzero shares

ypatil12:
**Motivation:**

There exists an edge case where:

1. Alice queues a withdrawal for all her shares
2. Alice's operator is slashed
3. Alice checkpoints (no rewards or slashes). This actually increases her DSF, even though there wasn't a balance increase on the beacon chain
4. Alice completes withdrawal as tokens
5. Alice redeposits

After step 5, Alice's withdrawable shares are overcounted due to increasing the DSF.  Note that in step 4 if Alice completes her withdrawal as shares she can still inflate her withdrawable shares. 

**Modifications:**

The fix is to prevent any state updates if `addedShares` is 0. In addition, we also update `recordBeaconChainETHBalanceUpdate` to only call the DM if there is a nonzero `balanceDeltaWei`, which is symmetric to the behavior in the `StrategyManager`. 

Regression tests:

- [x] testFuzz_noCall_zeroBalanceUpdate
- [x] testFuzz_deposit_delegate_allocate_slash_checkpointZeroBalance
- [x] testFuzz_deposit_delegate_allocate_slashAndQueue_checkpoint_redeposit
- [x] testFuzz_deposit_delegate_allocate_slashAndQueue_completeAsTokens_redeposit
- [x] testFuzz_deposit_delegate_allocate_slashAndQueue_checkPoint_completeAsShares

**Result:**

Inflating withdrawable shares edge case fixed.


8sunyuan (dismissed):

src/test/integration/tests/Slashed_Eigenpod_AVS.t.sol
```diff
@@ -0,0 +1,167 @@
...
+
+        // 10. Redeposit
+        cheats.deal(address(staker), 32 ether);
+        (uint40[] memory newValidators, uint64 addedBeaconBalanceGwei) = staker.startValidators();
+        beaconChain.advanceEpoch_NoRewards();
+        staker.verifyWithdrawalCredentials(newValidators);
+        check_VerifyWC_State(staker, newValidators, addedBeaconBalanceGwei);
+    }
+
+    /// @dev Asserts that the staker cannot inflate withdrawable shares after checkpointing & completing as shares
```
0xClandestine, line 145: _randBool with asTokens path would be a nice to have

0xClandestine (changes requested):

src/test/integration/tests/Slashed_Eigenpod_AVS.t.sol
ypatil12: The issue here is that we do a separate flow if it's as tokens or as shares

0xClandestine (approved):

wadealexc (dismissed):

nadir-akhtar (dismissed):

8sunyuan (approved):

PR #1179, opened 2025-02-26
feat: add check for labels on PRs

nadir-akhtar:
**Motivation:**

PRs now expect labels, but there is no technical enforcement. Naturally, any manual process can easily slip, so this CI addition enforces that at least one label be added to the PR.

**Modifications:**

Added a new CI check, `Labels`, which flags any PR that does not have a label present.

**Result:**

PRs without labels will be blocked, encouraging PRs to be labelled and preventing unlabelled PRs.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -2,6 +2,7 @@ name: Check
 
 on:
   push:
+  pull_request:
   workflow_dispatch: {}
 
 permissions:
@@ -41,3 +42,14 @@ jobs:
           make docker
           docker run -v `pwd`:/build -w /build --rm -i eigenlayer-contracts:latest bash -c "make gha"
           if [ ! -z "$(git status --porcelain)" ]; then git diff; git status; exit 1; fi
+  
+  labels:
+    name: Labels
+    # Only run on pull requests that don't have any labels
+    if: github.event_name == 'pull_request' && github.event.pull_request.labels[0] == null
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail if this PR has no labels
+        run: |
+          echo "This PR has no labels. Please add the appropriate labels to the PR."
+          exit 1


.github/workflows/checks.yml
```diff
@@ -2,6 +2,7 @@ name: Check
 
 on:
   push:
```
0xClandestine, line 4: This check should only be triggered on PRs, not both.

.github/workflows/checks.yml
```diff
@@ -40,4 +41,15 @@ jobs:
         run: |
           make docker
           docker run -v `pwd`:/build -w /build --rm -i eigenlayer-contracts:latest bash -c "make gha"
-          if [ ! -z "$(git status --porcelain)" ]; then git diff; git status; exit 1; fi
\ No newline at end of file
+          if [ ! -z "$(git status --porcelain)" ]; then git diff; git status; exit 1; fi
+
```
0xClandestine, line 45: Rather than creating another job, this should just be part of the commit lint job, as not to muddle the GH workflows section of a PR.

0xClandestine (changes requested):

.github/workflows/checks.yml
```diff
@@ -2,6 +2,7 @@ name: Check
```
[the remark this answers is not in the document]
nadir-akhtar: Yep, rebased to dev

.github/workflows/checks.yml
nadir-akhtar: Gotcha. Is your recommendation we remove the `push:` trigger?

.github/workflows/checks.yml
nadir-akhtar: Fair enough. It does a distinctly different job than the commit linting, since it's looking at the PR as a whole rather than the commits themselves, but not super opinionated

nadir-akhtar:
Closing this PR in favor of #1177 which automatically labels PRs. Could be nice to add a requirement that double checks for the presence of labels, but #1177 will do this with high reliability

PR #1191, opened 2025-02-28
test: checkpoint with no added shares

8sunyuan:
**Motivation:**

`EigenPodManager.recordBeaconChainETHBalanceUpdate` currently has a bug where shares increases of 0 addedShares will forward a call to the DelegationManager and updates the DSF. This can be manipulated to increase your DSF and inflate a staker's inflatable shares. The test in this PR demonstrates a possible exploit scenario.

**Modifications:**

Added test file `Integration_CompleteCP_WithNoAddedShares` and integration test `test_completeCP_withNoAddedShares`

Setup is as follows:
1. Setup a Staker EigenPod where they are initially slashed and have a non-WAD BCSF
2. Queue withdraw all deposit shares setting it to 0
3. Start/Complete checkpoints with 0 addedShares
    - This triggers the following code block in `SlashingLib.sol` when updating the DSF and increases the staker's DSF
```
if (prevDepositShares == 0) {
    // If this is the staker's first deposit or they are delegating to an operator,
    // the slashing factor is inverted and applied to the existing DSF. This has the
    // effect of "forgiving" prior slashing for any subsequent deposits.
    dsf._scalingFactor = dsf.scalingFactor().divWad(slashingFactor);
    return;
}
```
4. Deposit again (in this case complete withdrawals as shares from step 3)
5. Delegate to an Operator

Staker ends up having inflated withdrawable shares and as a result is able to delegate way more share than they should be able to for an operator.

**Result:**

After this issue is fixed, this integration test should pass


ypatil12:
Can we merge this into the fix PR?

8sunyuan:
https://github.com/Layr-Labs/eigenlayer-contracts/pull/1176
@ypatil12 this one right?

src/test/integration/tests/eigenpod/CompleteCP_WithNoAddedShares.t.sol
```diff
@@ -0,0 +1,95 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "src/test/integration/IntegrationChecks.t.sol";
+
+contract Integration_CompleteCP_WithNoAddedShares is IntegrationCheckUtils {
```
ypatil12, line 6: Can we add this to `SlashedEigenpod_BC.t.sol`, just so we have one less file? wdyt

8sunyuan: Sounds good, will also move EigenPod related tests to the `eigenpod` folder

src/test/integration/tests/Slashed_Eigenpod_BC.t.sol
```diff
@@ -106,7 +109,7 @@ contract Integration_SlashedEigenpod_BC is IntegrationCheckUtils {
 
         (uint256[] memory withdrawableSharesAfter, uint256[] memory depositSharesAfter) = delegationManager.getWithdrawableShares(address(staker), strategies);
         assertEq(depositSharesAfter[0], initDelegatableShares[0], "Deposit shares should reset to reflect slash(es)");
-        assertApproxEqAbs(withdrawableSharesAfter[0], depositSharesAfter[0], 100, "Withdrawable shares should equal deposit shares after withdrawal");
+        assertApproxEqAbs(withdrawableSharesAfter[0], depositSharesAfter[0], 1000, "Withdrawable shares should equal deposit shares after withdrawal");
```
ypatil12, line 112: Assume this check is sanity? Since we do `assert_Snap_Added_Staker_WithdrawableShares` in the complete as shares check.

ypatil12 (approved):
LGTM, one clarifying comment

src/test/integration/tests/Slashed_Eigenpod_BC.t.sol
8sunyuan: There were some instances where it was failing with a real delta of 103. Perhaps because my added contract is inheriting the other tests

PR #1197, opened 2025-03-02
test: add invariant check for staker strategy list

ypatil12:
**Motivation:**

The staker strategy list should properly account for strategies when deposited or queued completely for withdrawal. 

**Modifications:**

Added the following checks:

- assert_StrategyNotInStakerStrategyList
- assert_StrategyInStakerStrategyList
- assert_StakerStrategyListEmpty

The first two are used on deposits or withdrawals. The last is used on undelegations. 

**Result:**

More comprehensive invariant checks

### DIFF
--- a/src/test/integration/IntegrationBase.t.sol
+++ b/src/test/integration/IntegrationBase.t.sol
@@ -314,6 +314,42 @@ abstract contract IntegrationBase is IntegrationDeployer, TypeImporter {
         assertEq(withdrawalRoot, delegationManager.calculateWithdrawalRoot(withdrawal), err);
     }
 
+    function assert_StakerStrategyListEmpty(
+        User staker,
+        string memory err
+    ) internal view {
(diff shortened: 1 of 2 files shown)


src/test/integration/IntegrationBase.t.sol
```diff
@@ -3058,6 +3100,10 @@ abstract contract IntegrationBase is IntegrationDeployer, TypeImporter {
         return curShares;
     }
 
+    function _getStakerStrategyList(User staker) internal view returns (IStrategy[] memory) {
```
0xClandestine, line 3103: thought this was already a getter at the bottom, could be wrong

0xClandestine (approved):

src/test/integration/IntegrationBase.t.sol
ypatil12: Just checked, don't think there is

PR #1206, opened 2025-03-04
refactor: remove dead code

0xClandestine:
**Motivation:**

We've got dead code throughout the repo.

**Modifications:**

- Used this [tool]( https://github.com/0xClandestine/wand-rs) to programmatically remove dead Solidity.

**Result:**

Dead code has been removed.

### DIFF
--- a/certora/harnesses/DelegationManagerHarness.sol
+++ b/certora/harnesses/DelegationManagerHarness.sol
@@ -8,5 +8,21 @@ contract DelegationManagerHarness is DelegationManager {
     constructor(IStrategyManager _strategyManager, ISlasher _slasher, IEigenPodManager _eigenPodManager)
         DelegationManager(_strategyManager, _slasher, _eigenPodManager) {}
 
+    function get_operatorShares(address operator, IStrategy strategy) public view returns (uint256) {
+        return operatorShares[operator][strategy];
+    }
 
+    function get_stakerDelegateableShares(address staker, IStrategy strategy) public view returns (uint256) {
+        // this is the address of the virtual 'beaconChainETH' strategy
+        if (address(strategy) == 0xbeaC0eeEeeeeEEeEeEEEEeeEEeEeeeEeeEEBEaC0) {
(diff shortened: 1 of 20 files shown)


0xClandestine:
Some of this shouldn't be removed, just think it illustrates what's not used.

wadealexc (changes requested):
Files that we should NOT tweak:
* `script/releases/Env.sol`
  * This library interacts with Zeus ENV variables, so anytime we have release scripts that aren't using some ENV variables, the methods that fetch those will be "dead". Let's not remove this code; we'll want it when writing future release scripts.
* `src/contracts/libraries/BytesLib.sol`
  * This isn't our code, it's a common library written by Goncalo Sa. Let's not alter the code; it's easier to validate that it's the "standard version" if we keep it as-is.
* Anything in `test/integration/deprecatedInterfaces`
  * It's dead now, but if we decide we need it, it's much more of a pain in the ass to re-implement/lookup old interfaces

certora/harnesses/DelegationManagerHarness.sol
```diff
@@ -8,21 +8,5 @@ contract DelegationManagerHarness is DelegationManager {
     constructor(IStrategyManager _strategyManager, ISlasher _slasher, IEigenPodManager _eigenPodManager)
         DelegationManager(_strategyManager, _slasher, _eigenPodManager) {}
 
-    function get_operatorShares(address operator, IStrategy strategy) public view returns (uint256) {
```
0xClandestine, line 11: @8sunyuan Are these changes to certora harnesses problematic to remove?

8sunyuan: Yeah lets leave this folder untouched for now. We will have a PR to our contracts repo with updated certora specs from their team

0xClandestine: [3936827](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1206/commits/393682778a2f8941d28772f074cd39a5e2b40da2)

PR #1218, opened 2025-03-06
test: Certora prover specs and CI

aehyvari:
Motivation:

The slashing-magnitude update requires updates for the specs and new rules for the certora prover

Modifications:

Update the existing rules and add new rules, add Certora's CI

Result:

The certora prover will run on the updated code with updated specs.


certora/specs/core/DelegationManager.spec
```diff
@@ -220,6 +256,8 @@ rule sharesBecomeDelegatedWhenStakerDelegates(address operator, address staker,
     method f;
     env e;
     calldataarg arg;
+    // Certora: The rule does not hold if uncommented
```
8sunyuan, line 259: will fix this in followup PR

8sunyuan (approved):

PR #1223, opened 2025-03-06
refactor: integration tests

0xClandestine:
**Motivation:**

This PR refactors the integration test framework to improve performance, maintainability, and organization. The existing integration test structure had grown complex with multiple responsibilities mixed in different files, making it harder to understand and maintain. The goal is to create a more modular, better organized test framework with clear separation of concerns to make writing and maintaining integration tests easier.

**Modifications:**

1. Created two new utility files to better organize the codebase:
   - `IntegrationGetters.t.sol`: Centralizes all state-fetching functionality, providing methods to get both current and previous state snapshots.
   - `IntegrationUtils.t.sol`: Contains helper functions for creating users, generating allocation parameters, and other common tasks.

2. Restructured folder organization:
   - Moved deprecated interfaces from `deprecatedInterfaces/` to `deprecated/`.
   - Relocated utility files from integration to `/test/utils/` folder.
   - Moved mocks from `integration/mocks/` to `test/mocks/`.

3. Renamed test files to follow a more consistent naming convention:
   - e.g., `ALM_RegisterAndModify.t.sol` → `Deposit_Delegate_Modify.t.sol`.

4. Refactored the `TimeMachine.t.sol` contract with improved functionality and moved it to `utils/` folder to make it accessible beyond integration tests.

5. Added a new `Constants.t.sol` file to centralize commonly used constants.

6. Added a `FOUNDRY_FUZZ_SEED` to fork tests that updates weekly (this significantly reduces our RPC usage).

7. Moved `_init` into `setUp()` (this means setup happens once per fuzz run, rather than once per test).

8. Sort `strategies` array once, rather than JIT (sorting an array in place is time intensive).

9. Refactored randomness sourcing to use Vm.
- Previously, we were using a `uint24 _random` variable combined with hashing to generate subsequent random values. 

- While this pattern simplified test reruns (allowing the reuse of the same _random value for consistency), it has notable drawbacks. Randomness isn't unevenly distributed, leading to modulo bias, and the process is computationally expensive.

Additionally, the same functionality can now be easily achieved through Foundry flags.

**Result:**

After these changes, the integration test framework will be more modular with clear separation of concerns, making it easier to understand and maintain.

Integration CI times has decreased from 15-25mins -> 5-6mins.

NOTE: There's still quite a bit that could be done, don't want this PR getting too big.


src/test/integration/tests/Deposit_Delegate_Modify.t.sol
```diff
@@ -3,27 +3,16 @@ pragma solidity ^0.8.27;
 
 import "src/test/integration/IntegrationChecks.t.sol";
```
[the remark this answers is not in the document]
0xClandestine: Goal is to combine all the deposit, delegate scenarios into a single contract, have noticed it's pretty costly in both compile time and test time having them all seperated when they all have minor differences in setup.

Agree though, current naming is pretty bad.

src/test/integration/tests/eigenpod/Slashed_Eigenpod_BC.t.sol
```diff
@@ -144,9 +134,8 @@ contract Integration_SlashedEigenpod_BC is IntegrationCheckUtils {
         _rollBlocksForCompleteAllocation(operator, operatorSet, strategies);
 
         //Slash operator before delegation
-        IAllocationManagerTypes.SlashingParams memory slashingParams;
-        uint wadToSlash = _randWadToSlash();
-        slashingParams = avs.slashOperator(operator, operatorSet.id, strategies, wadToSlash.toArrayU256());
+        SlashingParams memory slashingParams = _genSlashing_Half(operator, operatorSet);
```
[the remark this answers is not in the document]
0xClandestine: Plan to revamp `src/utils/Random.t.sol` and use it for more varied randomness. I don't think we should be slashing by half anywhere really, originally we were using _randWadToSlash everywhere. Agree it provides better coverage.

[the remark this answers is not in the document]
ypatil12: Confused why we had to use a half slashing to get this test to pass with the new framework

src/test/integration/tests/Deposit_Delegate_Modify.t.sol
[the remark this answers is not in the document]
ypatil12: Old name is more descriptive IMO

src/test/integration/IntegrationGetters.t.sol
```diff
@@ -0,0 +1,651 @@
...
+
+    function _getSlashingFactor(User staker, IStrategy strategy) internal view returns (uint) {
+        address operator = delegationManager.delegatedTo(address(staker));
+        uint64 maxMagnitude = allocationManager.getMaxMagnitudes(operator, strategy.toArray())[0];
+        if (strategy == BEACONCHAIN_ETH_STRAT) return maxMagnitude.mulWad(eigenPodManager.beaconChainSlashingFactor(address(staker)));
+        return maxMagnitude;
+    }
+
+    /// @dev Assumes that the staker has one withdrawal queued
+    function _getWithdrawableSharesAfterCompletion(User staker) internal view returns (uint[] memory withdrawableShares) {
```
ypatil12, line 634: So much cleaner to separate out these getters :)

src/test/integration/IntegrationChecks.t.sol
```diff
@@ -1,23 +1,65 @@
...
+contract IntegrationChecks is IntegrationUtils {
     using ArrayLib for *;
-    using SlashingLib for *;
     using StdStyle for *;
 
+    AVS avs;
+    SlashingParams slashParams;
+    SlashingParams slashingParams;
+
+    User operator;
```
ypatil12, line 18: Why do we need these vars in storage?

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -347,92 +305,53 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
...
 
+    /// @dev Initializes all proxies.
     function _initializeProxies() public noTracing {
+        allocationManager.initialize({initialOwner: executorMultisig, initialPausedStatus: 0});
+        avsDirectory.initialize({initialOwner: executorMultisig, initialPausedStatus: 0});
         delegationManager.initialize({initialOwner: executorMultisig, initialPausedStatus: 0});
-
+        eigenPodManager.initialize({initialOwner: executorMultisig, _initPausedStatus: 0});
+        // TODO Why is this missing?
+        // rewardsCoordinator.initialize({initialOwner: executorMultisig, initialPausedStatus: 0});
```
ypatil12, line 332: Technically RC isn't used in integration tests, but we should still initialize for completeness

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -623,128 +467,54 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
...
+            if (assetTypes & HOLDS_ETH != 0) options[count++] = HOLDS_ETH;
+            if (assetTypes & HOLDS_ALL != 0) options[count++] = HOLDS_ALL;
+            if (assetTypes & HOLDS_MAX != 0) options[count++] = HOLDS_MAX;
+            return options[cheats.randomUint(0, count - 1)];
         }
-
-        return strats;
     }
 
-    function _randomStrategies() internal returns (IStrategy[][] memory strategies) {
```
ypatil12, line 710: Reason why we don't need this? I think we were using this when creating an AVS with a random set of Strats

src/test/integration/tests/Deposit_Delegate_Modify.t.sol
```diff
@@ -119,7 +108,11 @@ contract Integration_InitRegistered is Integration_ALMBase {
         check_FullyDeallocated_State(operator, allocateParams, deallocateParams);
     }
 
-    function testFuzz_deregister_allocate_waitAllocate_deallocate(uint24 _r) public rand(_r) {
+    function testFuzz_deposit_delegate_register_deregister_allocate_waitAllocate_deallocate(uint24) public {
```
ypatil12, line 111: Where are we sourcing the randomness from?

src/test/integration/tests/Deposit_Delegate_Modify.t.sol
```diff
@@ -37,20 +26,12 @@ contract Integration_ALMBase is IntegrationCheckUtils {
         // 3. Create an operator set containing the strategies held by the staker
         operatorSet = avs.createOperatorSet(strategies);
     }
-}
-
-contract Integration_InitRegistered is Integration_ALMBase {
-    /// @dev Integration test variants that start with the operator being registered
-    /// for the operator set
-    function _init() internal virtual override {
```
ypatil12, line 45: Why can't  we use `super._init` and then init here? Prefer having the base in that regard, since now we have to register in every step for every test in this contract

ypatil12 (commented):
First pass, looking good

src/test/integration/IntegrationChecks.t.sol
0xClandestine: Plan is to eventually define a single `Scenario` struct, started by first moving everything to storage. Will be improved in followup PR.

src/test/integration/tests/Deposit_Delegate_Modify.t.sol
0xClandestine: We need to reduce the amount of test contracts given how large they are, goal is to reduce the LOC needed for a single test so we can consolidate as many tests into as few contracts as possible.

src/test/integration/tests/Deposit_Delegate_Modify.t.sol
0xClandestine: We now source randomness from Foundry using cheats.randomUint(). The main reason we initially used the hash-based approach was to ensure test rerun-ability. For example, you could retrieve the uint24 randomness value from a failed test run and rerun the test with that fixed value. However, Foundry already supports this exact functionality through flags available for forge t (see `--rerun` `--fuzz-seed` etc.).

src/test/integration/IntegrationDeployer.t.sol
0xClandestine: I believe it was unused code, removed using wand-rs.

src/test/integration/tests/Deposit_Delegate_Modify.t.sol
0xClandestine: Removed `cache/` from  .gitignore, the last failing fuzz seed will always be added to `cache/fuzz/failures`. You can re-run failed tests using that seed with `--fuzz-seed`.

script/output/devnet/SLASHING_deploy_from_scratch_deployment_data.json
```diff
@@ -1,33 +1,31 @@
...
-           "token_address": "0x94373a4919B3240D86eA41593D5eBa789FEF3848",
-           "token_symbol": "WETH"
-      }
-    ],
-    "strategyBeacon": "0x957c04A5666079255fD75220a15918ecBA6039c6",
-    "strategyFactory": "0x09F8f1c1ca1815083a8a05E1b4A0c65EFB509141",
-    "strategyFactoryImplementation": "0x8b1F09f8292fd658Da35b9b3b1d4F7d1C0F3F592",
-    "strategyManager": "0x70f8bC2Da145b434de66114ac539c9756eF64fb3",
-    "strategyManagerImplementation": "0x1562BfE7Cb4644ff030C1dE4aA5A9aBb88a61aeC",
+    "TestToken": "0xcae746B46013Bf4D43B8E2B9e98F9DBE2461532A",
```
ypatil12, line 3: Any reason this is updated?

ypatil12 (commented):
Solid refactor, had one question

src/test/integration/tests/Deposit_Delegate_Modify.t.sol
ypatil12: Can we add this process of using `--fuzz-seed` to the ReadME?

PR #1233, opened 2025-03-11
docs: add StrategyBase accounting doc

nadir-akhtar:
**Motivation:**

This PR adds documentation about the `StrategyBase` contract and the shares model, as well as answers to some common questions about the accounting model.

Specifically, it covers:
* Why shares are needed for EigenLayer accounting, and rebase tokens
* The impact of direct transfers into a `StrategyBase` contract
* Inflation attacks, and an exploit scenario
* The mitigations built into the `StrategyBase` contract and any residual risk & side effects

**Modifications:**

* Added the StrategyBaseAccounting doc
* Renamed `docs/core/accounting/SlashingEdgeCase.md` to `docs/core/accounting/DualSlashingEdgeCase.md` (for clarity)

**Result:**

More clarity on how `StrategyBase` does shares accounting!


MinisculeTarantula (dismissed):
I like this!
I think it is well-targeted to an audience that seeks to understand the basic shares accounting model. Additionally, since the use (at least so far) has really just been to handle rebasing tokens, the focus on explaining a bit about them + using them as motivation seems apt.
This does not cover "share-inflation" attacks or the chosen mitigation strategy, but I suspect that those topics may be of a greater technical depth than you were aiming for here.
So depending on the target audience, I think this either seems done or like more could be added.

nadir-akhtar:
@MinisculeTarantula Thank you for the review! The first iteration covered the basics, and I've now added the technical depth on the attack vectors you reference. Please have another look at the newly added sections 🙏

docs/core/accounting/StrategyBaseAccounting.md
```diff
@@ -0,0 +1,336 @@
...
+
+For example, say that Alice sends a *million* tokens to the `StrategyBase` contract. She does not go through the `deposit` function; she does not receive any additional shares for depositing these tokens into the contract. The exchange rate is now 1e6 tokens : 1 share.
+
+The current state:
+* `StrategyBase` total shares: 1
+* `StrategyBase` token balance: **1e6 + 1**
+* Alice's shares: **1**
+* Alice's "deserved" token balance: **1e6 + 1**
+
+Note the large difference between the `StrategyBase`'s token balance and the number of shares. As mentioned before, the exchange rate is now 1e6 tokens : 1 share.
```
ChaoticWalrus, line 199: tiny nit: it's actually (1e6 + 1) tokens : 1 share
but idk, probably it's clearer without this change

docs/core/accounting/StrategyBaseAccounting.md
```diff
@@ -0,0 +1,336 @@
...
+
+When Bob deposits 1000 tokens, he is depositing less than 1e6 tokens, meaning that *he receives no shares for his deposit*. Calculating Bob's total shares:
+
+* Bob's 1000 tokens * 1 share / 1e6 tokens = 1e-3 shares = 0 shares
+
+Due to the large divisor, Bob's received shares are now 1e-3, a very small number, which is *effectively 0* due to EVM division. Thus, **Bob receives no shares for his token deposit**. This is a problem, as he is entitled to 1000 tokens, but has no shares for withdrawing it.
```
ChaoticWalrus, line 205: suggestion for a change: `which is *effectively 0* due to EVM division` => `which is rounded to zero due to EVM division`

nadir-akhtar: [7d4e647](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1233/commits/7d4e647af93dc42ecef43d624a066617cbfdd1b0)

docs/core/accounting/StrategyBaseAccounting.md
nadir-akhtar: [7d4e647](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1233/commits/7d4e647af93dc42ecef43d624a066617cbfdd1b0)

nadir-akhtar: While I agree the exchange rate is a bit awkward, I think that readers will strongly prefer the accuracy! Thanks for the callout

8sunyuan (approved):

PR #1241, opened 2025-03-13
feat: remove transfer restriction code in EIGEN

bowenli86:
**Motivation:**

per ELIP5, remove transfer restriction code from eigen

mint function will be removed in next PR so we can limit change scope in each PR, and make each change more readable

**Modifications:**

remove transfer restriction code from eigen

**Result:**

remove transfer restriction code from eigen


src/contracts/token/Eigen.sol
```diff
@@ -63,59 +61,19 @@ contract Eigen is OwnableUpgradeable, ERC20VotesUpgradeable {
...
+        require(__deprecated_mintingAllowance[msg.sender] > 0, "Eigen.mint: msg.sender has no minting allowance");
+        require(
+            block.timestamp > __deprecated_mintAllowedAfter[msg.sender],
+            "Eigen.mint: msg.sender is not allowed to mint yet"
+        );
+        uint256 amount = __deprecated_mintingAllowance[msg.sender];
+        __deprecated_mintingAllowance[msg.sender] = 0;
         _mint(msg.sender, amount);
         emit Mint(msg.sender, amount);
     }
```
MinisculeTarantula, line 79: this function should be deleted entirely, along with the Mint event definition and all the minting-related initializer inputs. The function's existence predates bEIGEN. It is unusable in existing deployments and any new deployment should not use it.

MinisculeTarantula:
The same transfer restriction code can also be removed from bEIGEN (but it should keep its minting-related code, as minting is now done entirely via bEIGEN)

src/contracts/token/Eigen.sol
bowenli86: yeah, as mentioned in description, mint function and transfer function in beigen will be removed in next few PRs so we can limit change scope in each PR, and make each change more readable. otherwise too many lines of changes and too many tests to fix at one time

MinisculeTarantula: ahhh, caught out in my own lack of reading! Will resolve this thread.

MinisculeTarantula (approved):
LGTM. My apologies for not fully reading/grasping the PR description on first pass. This all seems appropriate for the carefully-stepped approach you are taking.

bowenli86:
> LGTM. My apologies for not fully reading/grasping the PR description on first pass. This all seems appropriate for the carefully-stepped approach you are taking.

no worry, thanks for reviewing!

PR #1242, opened 2025-03-13
Clarifying the reasoning for starting with base e18 in the multiplier

wesfloyd:
**Motivation:**

The reasoning for choosing a base e18 multiplier is not obvious for the reader. It would help to add context to explain why this is valuable for the developer - eg the "lower level processing of rewards distribution" causes a loss of precision if small numbers are used.

**Modifications:**

Added a small addition to the markdown to explain the behavior of the multiplier value.

**Result:**

Only additional docs content.


wesfloyd:
Closing b/c this should be obvious for sc developers:
>generally large numbers are used to prevent precision loss. I don’t think we have to over explain. It’s fine to just recommend using numbers like 1e18.

PR #1246, opened 2025-03-15
feat: consolidate mint function to BackingEigen, and change BackingEigen constructor to break circular dependency on Eigen

bowenli86:
**Motivation:**

- consolidate mint function to BackingEigen
- change BackingEigen constructor to break circular dependency on Eigen

**Modifications:**

- remove mint function in Eigen
- move all tests to use mint function in BackingEigen
- change integration tests setup to have minters mint their own supply
- change BackingEigen constructor to break circular dependency on Eigen

**Result:**

- only BackingEigen has mint capability
- no more circular dependency between BackingEigen and Eigen


src/test/token/EigenWrapping.t.sol
```diff
@@ -49,13 +43,35 @@ contract EigenWrappingTests is Test {
...
+
+        // The Eigen contract has all the bEIGEN tokens from initialization
+        // First approve itself to spend the tokens
+        vm.startPrank(address(eigen));
+        bEIGEN.approve(address(eigen), totalSupply);
+        // Then wrap the tokens, which will mint EIGEN tokens to itself
+        eigen.wrap(totalSupply);
+        // Finally distribute the EIGEN tokens to minters
+        eigen.transfer(minter1, totalSupply / 2);
+        eigen.transfer(minter2, totalSupply / 2);
```
MinisculeTarantula, line 74: This sequence is definitely weird.
We discussed offline that it makes sense to remove the minting-on-initialization behavior from the bEIGEN code (which is itself a relic particular to EIGEN historically being deployed first). In concert with this change (although it can be made before the change) I would recommend instead having one or both of `minter` and `minter2` directly mint the bEIGEN tokens and then wrap them into EIGEN tokens themselves.

MinisculeTarantula (approved):
This PR LGTM now. Thanks!

PR #1265, opened 2025-03-31
Fix/audit cast bug

8sunyuan:
**Motivation:**

Fix for high sev bug found in Cantina competition.

**Modifications:**

Contract Changes (EPM):
- Safecasting uint256 `depositSharesToRemove` to int256 since this input is provided by the user through `DelegationManager.queueWithdrawals`

Tests:
- Unit test `testFuzz_removeDepositShares_revert_depositSharesOverflow` which does not bound fuzzed values of `depositSharesToRemove` and checks for casting overflow, negative shares, or a valid input
- Removed all `uint224` usage in the EPM unit tests
- Regression test POC showing the inflation of staker shares based on original finding `testFuzz_deposit_revertOverflow_queueWithdrawal`

Screenshot below is test with the expectedRevert message commented out and with original unsafe casting implementation in the EPM. We can clearly see the depositSharesAfter is much larger than originally...
image

Updated Certora Spec:
-  Fixed certora parametric rule `integrityOfRemoveDepositShares` to not constrain input values and assert invalid input does not affect depositShares. Updated rule passes here https://prover.certora.com/output/748112/73606141d5914f93802e9d8551f0927e?anonymousKey=c2ad21a3e9796b8586d51bfb5bfaad8e560ad6bb 

image

**Result:**

Fix for bug


src/test/unit/EigenPodManagerUnit.t.sol
```diff
@@ -581,8 +612,8 @@ contract EigenPodManagerUnitTests_BeaconChainETHBalanceUpdateTests is EigenPodMa
 
     function testFuzz_recordPositiveBalanceUpdate(uint sharesBefore, uint sharesDelta, uint prevRestakedBalanceWei) public {
         // Constrain inputs
-        sharesBefore = bound(sharesBefore, 0, type(uint224).max) * uint(GWEI_TO_WEI);
-        sharesDelta = bound(sharesDelta, 1, type(uint224).max) * uint(GWEI_TO_WEI);
```
wadealexc, line 585: I think this is your issue - we're no longer submitting share amounts divisible by GWEI. So this test is reverting -- guess that's why we were using u224 here.

Should be a simple enough fix, I think you can revert these lines and it should work as intended!

wadealexc (approved):
LGTM, just left a comment for the test bugfix!

wadealexc (approved):

ypatil12 (approved):

wadealexc (approved):

0xClandestine (approved):

PR #1266, opened 2025-04-01
docs: post-cantina doc updates

ypatil12:
**Motivation:**

Address doc updates from Cantina Audit. 

**Modifications:**

- `deregisterFromOperatorSets` does not revert due to no try-catch
- Permissionless Strategies added to audit report
- We don't do `+1` on the `effectBlock` for allocations for instantaneous allocations
- New deposits are immediately subject to slashing
- Removing a strategy from an operatorSet does not allow a pending deallocation to be instantly completable
- Completing a withdrawal as shares is OK even if the strategy is not on the whitelist
- Delegating to operators with ultra-low magnitudes can result in loss of funds
- Edge case regarding withdrawals for an operator blocked by owning negative shares & a fully slashed strategy
- Negative rebase for strategies

**Result:**

More comprehensive docs.


wadealexc (approved):
LGTM!

PR #1270, opened 2025-04-01
chore: add event to dsf reset on full withdrawal

eigenmikem:
**Motivation:**

DSF reset isn't emitting event on a full withdrawal.

**Modifications:**

Added event and shadowed declaration to resolve stack too deep

**Result:**

Easier for front-end to track dsf changes.


ypatil12 (commented):
If I do this:

```solidity
            if (sharesAfter == 0) {
                DepositScalingFactor storage _dsf = _depositScalingFactor[staker][strategies[i]];
                _dsf.reset();
                emit DepositScalingFactorUpdated(staker, strategies[i], _dsf.scalingFactor());
            }
```

I don't get a stack too deep

wadealexc (approved):
LGTM. Agree with Yash, but IMO it's a bit cleaner to just use a storage pointer at the top of the method.

ypatil12 (approved):

0xClandestine (approved):

PR #1272, opened 2025-04-01
chore: address ep/epm informational

ypatil12:
**Motivation:**

Address cosmetic issues brought up for the native restaking system

**Modifications:**

1. Updates `withdrawRestakedBeaconChainETH` to note that we don't have to pass in a whole-gwei amount
2. Removes `bytesLib` from EigenPod. Also moves the file into the test folder
3. Marks `PAUSED_WITHDRAW_RESTAKED_ETH` as deprecated in `EigenPodPausingConstants`

**Result:**

Cleaner Contracts.


0xClandestine (approved):
Sick so we no longer use BytesLib is core?

wadealexc (approved):

PR #1287, opened 2025-04-01
test: zeus queued deployments + arbitrary envs

0xClandestine:
**Motivation:**

We want to support Zeus config parsing to enable testing of queued deployments. Our original deployment parser lacked the flexibility needed to support Zeus without requiring changes to Zeus itself. Additionally, the `User` and `AVS` contracts deployed by the `IntegrationDeployer` aren't able to properly read config from `IntegrationDeployer` leading to bandaid solutions.

**Modifications:**

- Removed `InitialDeploymentParser`: About 1k lines -> 100 for parsing.
- Added `ConfigParser`: A parser that supports both fast TOML parsing and Zeus-style parsing.
- Added `ConfigGetters`: A abstraction that allows contracts deployed by the `IntegrationDeployer` to easily access the current configuration.

**Result:**

Contracts deployed through the `IntegrationDeployer` can now reliably access their configuration, and Zeus integration is possible without needing to modify Zeus.


src/test/ParseMainnetConfig.t.sol
```diff
@@ -0,0 +1,109 @@
...
+    using stdToml for string;
+
+    Vm constant vm = Vm(address(uint160(uint(keccak256("hevm cheat code")))));
+    bytes32 constant ERC1967_IMPLEMENTATION_SLOT = bytes32(uint(keccak256("eip1967.proxy.implementation")) - 1);
+    bytes32 constant ERC1967_BEACON_SLOT = bytes32(uint(keccak256("eip1967.proxy.beacon")) - 1);
+    bytes32 constant ERC1967_ADMIN_SLOT = bytes32(uint(keccak256("eip1967.proxy.admin")) - 1);
+
+    function impl(address proxy) internal view returns (bytes32) {
+        return vm.load(proxy, ERC1967_IMPLEMENTATION_SLOT);
+    }
```
0xClandestine, line 54: No need to track implementations in config, as we can deduce it from onchain state.

PR #1315, opened 2025-04-17
fix: remove panics from mainnet script

ypatil12:
**Motivation:**

The mainnet deploy script panics the moment a checkpoint fails. We want to continue and complete in a subsequent script. 

**Modifications:**

Don't panic on `completeCheckpointForEigenpod`

**Result:**

Saved time :)


bowenli86:
since v1.3.0 is already cut from main, new commits should go into `main`, then cherry pick to v1.3.0

PR #1250, opened 2025-03-17
feat: add TokenWrapped and TokenUnwrapped events in Eigen for observability

bowenli86:
**Motivation:**

add TokenWrapped and TokenUnwrapped events in Eigen for observability

**Modifications:**

- add TokenWrapped and TokenUnwrapped events in Eigen
- add corresponding tests

**Result:**

we can track wrapping/unwrapping actions between Eigen and BackingEigen

PR #1331, opened 2025-04-21
feat: add TokenWrapped and TokenUnwrapped events in Eigen for observability

bowenli86:
**Motivation:**

add TokenWrapped and TokenUnwrapped events in Eigen for observability

**Modifications:**

add TokenWrapped and TokenUnwrapped events in Eigen
add corresponding tests

**Result:**

we can track wrapping/unwrapping actions between Eigen and BackingEigen


bowenli86:
this is re-apply https://github.com/Layr-Labs/eigenlayer-contracts/pull/1250, which i might hv accidentally eliminated...

PR #1337, opened 2025-04-23
release(refactor): integration overhaul

0xClandestine:
**Motivation:**

**Modifications:**

**Result:**


script/output/devnet/SLASHING_deploy_from_scratch_deployment_data.json
```diff
@@ -1,33 +1,31 @@
...
-           "token_address": "0x94373a4919B3240D86eA41593D5eBa789FEF3848",
-           "token_symbol": "WETH"
-      }
-    ],
-    "strategyBeacon": "0x957c04A5666079255fD75220a15918ecBA6039c6",
-    "strategyFactory": "0x09F8f1c1ca1815083a8a05E1b4A0c65EFB509141",
-    "strategyFactoryImplementation": "0x8b1F09f8292fd658Da35b9b3b1d4F7d1C0F3F592",
-    "strategyManager": "0x70f8bC2Da145b434de66114ac539c9756eF64fb3",
-    "strategyManagerImplementation": "0x1562BfE7Cb4644ff030C1dE4aA5A9aBb88a61aeC",
+    "TestToken": "0xcae746B46013Bf4D43B8E2B9e98F9DBE2461532A",
```
wadealexc, line 3: Can we nuke this file? It keeps popping up in places...

What's creating it?

0xClandestine: Yeah pretty sure I removed the logic that generates this, think it just needs to be removed now.

snapshots/Integration_ALM_Multi.json
```diff
@@ -1,3 +1,3 @@
 {
-  "gasUsed": "47579"
+  "gasUsed": "137087"
```
wadealexc, line 2: What is this file? Can we gitignore snapshots, or figure out what's generating this?

(is it useful/used somewhere?)

src/test/Config.t.sol
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    function totalStrategies() public view virtual returns (uint) {
+        return config.strategies.strategyAddresses.length;
+    }
+}
+
+contract ConfigGettersTest is ConfigGetters, Test {
+    function setUp() public {
+        vm.createSelectFork(vm.rpcUrl("mainnet"), 22_181_590);
```
wadealexc, line 162: Let's pull this block value out into a constant and leave a comment explaining what it is. 

(i assume this is post-slashing upgrade?)

src/test/ConfigParser.t.sol
```diff
@@ -0,0 +1,252 @@
...
+        vm.store(proxy, ERC1967_BEACON_SLOT, bytes32(uint(uint160(beacon))));
+        return proxy;
+    }
+
+    function setAdmin(address proxy, address admin) internal returns (address) {
+        vm.store(proxy, ERC1967_ADMIN_SLOT, bytes32(uint(uint160(admin))));
+        return proxy;
+    }
+
+    function emptyContract() external returns (address deployed) {
```
0xClandestine, line 197: This should be removed, doesn't work.

src/test/Config.t.sol
wadealexc: Also I think this just gets run with our standard local test suite, right?

That might be annoying if someone's not trying to run fork tests, but this test fails?

0xClandestine: Good point, will move/refactor these tests a bit.

src/test/ConfigParser.t.sol
```diff
@@ -0,0 +1,252 @@
...
+        }
+        require(deployed != address(0), "Deployment failed");
+    }
+}
+
+contract ConfigParserTest is Test {
+    using ConfigParser for *;
+
+    function test_ParseTOML() public {
+        vm.createSelectFork(vm.rpcUrl("mainnet"), 22_181_590);
```
wadealexc, line 210: See previous comments about this block value and running this by default

src/test/integration/users/User_M1.t.sol
```diff
@@ -1,21 +1,10 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.27;
 
-import "src/test/integration/deprecatedInterfaces/mainnet/IEigenPod.sol";
-import "src/test/integration/deprecatedInterfaces/mainnet/IEigenPodManager.sol";
-import "src/test/integration/deprecatedInterfaces/mainnet/IStrategyManager.sol";
+import "src/test/integration/deprecated/mainnet/IEigenPod.sol";
```
wadealexc, line 4: Let's delete User_M1 entirely. It's not used anywhere.

src/test/integration/users/User_M2.t.sol
```diff
@@ -2,28 +2,11 @@
 pragma solidity ^0.8.27;
 
 import "forge-std/Test.sol";
-
-import "src/test/integration/deprecatedInterfaces/mainnet/IEigenPod.sol";
-import "src/test/integration/deprecatedInterfaces/mainnet/IEigenPodManager.sol";
-import "src/test/integration/deprecatedInterfaces/mainnet/IStrategyManager.sol";
-import "src/test/integration/deprecatedInterfaces/mainnet/IDelegationManager.sol";
-
+import "src/test/integration/deprecated/mainnet/IDelegationManager.sol";
```
wadealexc, line 5: I think we _can_ delete this, but it might require just a little bit more work because it's still used (though we don't need to be using it).

Let's mark this work for a followup PR.

src/test/integration/IntegrationChecks.t.sol
```diff
@@ -1,23 +1,65 @@
...
+    uint[] withdrawableShares;
+    // Withdrawal[] withdrawals; // cannot copy from memory to storage easily
+    bytes32[] withdrawalRoots;
+    uint[] numTokensRemaining;
+
+    uint64 beaconBalanceGwei;
+    uint40[] validators;
+    uint40[] slashedValidators;
+    uint64 slashedGwei;
+    uint64 slashedAmountGwei;
```
wadealexc, line 36: What are all these state variables?

They're shadowed throughout the contract -- can we get rid of them? Tests that need persistent state across setups can just declare it themselves.

(I know we have a bunch of tests that declare similar storage, but people are gonna declare storage variables no matter what - I think it's good to have that at the surface level)

src/test/integration/IntegrationChecks.t.sol
```diff
@@ -1,23 +1,65 @@
...
+        for (uint i = 0; i < withdrawals.length; i++) {
+            if (_randBool()) {
+                staker.completeWithdrawalAsTokens(withdrawals[i]);
+                check_Withdrawal_AsTokens_State(staker, operator, withdrawals[i], withdrawalShares, expectedTokens);
+            } else {
+                staker.completeWithdrawalAsShares(withdrawals[i]);
+                check_Withdrawal_AsShares_State(staker, operator, withdrawals[i], withdrawals[i].strategies, withdrawalShares);
+            }
+        }
+    }
```
wadealexc, line 55: Is this supposed to be in this contract? It feels out of place

src/test/integration/IntegrationChecks.t.sol
```diff
@@ -1,23 +1,65 @@
...
+
     /**
      *
      *                              EIGENPOD CHECKS
      *
      */
-    function check_VerifyWC_State(User staker, uint40[] memory validators, uint64 beaconBalanceGwei) internal {
+    function check_VerifyWC_State(User staker, uint40[] memory validators, uint64 beaconBalanceGwei) public {
```
wadealexc, line 62: Putting a pin in this, why are these public now?

wadealexc: ok i figured this meant tests are calling them like `checker.check_VerifyWC_State()`, but they're not.

could we just keep these internal to keep the diff down, then? maybe im missing where this is needed?

src/test/integration/IntegrationChecks.t.sol
```diff
@@ -245,8 +283,8 @@ contract IntegrationCheckUtils is IntegrationBase {
             staker, strategies, shares, "staker should expected shares in each strategy after depositing"
         );
 
-        if (delegationManager.isDelegated(address(staker))) {
-            User operator = User(payable(delegationManager.delegatedTo(address(staker))));
+        if (delegationManager().isDelegated(address(staker))) {
```
wadealexc, line 286: Just noting, I feel like making these functions (instead of variables) is a slight downgrade for brevity. But I'm fine with it atm...

src/test/integration/IntegrationChecks.t.sol
```diff
@@ -558,14 +598,16 @@ contract IntegrationCheckUtils is IntegrationBase {
...
+    ) public {
         /// Complete withdrawal(s):
         // The staker will complete the withdrawal as shares
         //
         // ... check that the withdrawal is not pending, that the token balances of the staker and operator are unchanged,
         //     that the withdrawer received the expected shares, and that that the total shares of each o
         //     strategy withdrawn remains unchanged
-        assert_WithdrawalNotPending(delegationManager.calculateWithdrawalRoot(withdrawal), "staker withdrawal should no longer be pending");
+        assert_WithdrawalNotPending(
+            delegationManager().calculateWithdrawalRoot(withdrawal), "staker withdrawal should no longer be pending"
```
wadealexc, line 609: Followup work - we should be able to only pass in `withdrawal` and have the underlying method get the root.

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -131,503 +94,445 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
...
-    /**
-     * env FOUNDRY_PROFILE=forktest forge t --mc Integration
-     *
-     * Running foundry like this will trigger the fork test profile,
-     * lowering fuzz runs and using a remote RPC to test against mainnet state
-     */
-    function isForktest() public view returns (bool) {
-        return _hash("forktest") == _hash(cheats.envOr(string("FOUNDRY_PROFILE"), string("default")));
+    /// @dev Configures the possible asset and user types for the random users.
+    function _configRand(uint _assetTypes, uint _userTypes) internal virtual {
```
wadealexc, line 135: Seems like this method isn't used anymore - so let's just get rid of it?

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -131,503 +94,445 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
...
         for (uint i = 1; i < NUM_LST_STRATS + 1; ++i) {
             string memory name = string.concat("LST-Strat", cheats.toString(i), " token");
             string memory symbol = string.concat("lstStrat", cheats.toString(i));
-            // Deploy half of the strategies using the factory.
             _newStrategyAndToken(name, symbol, 10e50, address(this), i % 2 == 0);
         }
 
+        // Whitelist the strategies
+        cheats.prank(strategyManager().strategyWhitelister());
+        strategyManager().addStrategiesToDepositWhitelist(allStrats);
```
wadealexc, line 195: You're adding the beacon chain strat to the deposit whitelist here

src/test/mocks/BeaconChainMock.t.sol
```diff
@@ -7,9 +7,10 @@ import "src/contracts/libraries/BeaconChainProofs.sol";
 import "src/contracts/libraries/Merkle.sol";
```
wadealexc, line 7: Request - let's please move all the beacon chain / EIP mock contracts to a subfolder, for organization's sake.

e.g:

```
mocks/beacon/
 - BeaconChainMock.t.sol
 - BeaconChainMock_Deneb.t.sol
 - EIP_4788_Oracle_Mock.t.sol
```


I'm adding 2 more mocks for Pectra predeploys in MOOCOW, and having all of them in the same place is super helpful.

wadealexc: oh yeah, and i also did some refactoring of my own and added two libraries:

image

The libraries also need to sit next to the beacon chain stuff. So, a subfolder is needed :D

src/test/integration/tests/eigenpod/EigenPod.t.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "src/test/integration/IntegrationChecks.t.sol";
+
+contract EigenPodTest is IntegrationChecks {
+    function _init() internal virtual override {
+        if (!forkConfig.supportEigenPodTests) cheats.skip(true);
+    }
```
wadealexc, line 9: Why would we not support these?

src/test/integration/tests/upgrade/EigenPod.t.sol
```diff
@@ -1,192 +1,182 @@
...
-            // If we had a positive balance update, then the expected token delta is the withdrawal scaled shares
-            expectedTokenDeltas[0] = _withdrawal.scaledShares[0];
-        } else {
-            // If we had a negative balance update, then the expected token delta is the withdrawal scaled shares plus the negative share delta
-            expectedTokenDeltas[0] = uint(int(_withdrawal.scaledShares[0]) + _balanceUpdateShareDelta);
-        }
-        return expectedTokenDeltas;
-    }
-}
+// // SPDX-License-Identifier: BUSL-1.1
```
wadealexc, line 1: Gotta fix this before we merge

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -131,503 +94,445 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
...
-        console.log("Setting up `%s` integration tests:", "MAINNET_FORK".green().bold());
-        console.log("RPC:", cheats.rpcUrl("mainnet"));
-        console.log("Block:", mainnetForkBlock);
+    function _setUpFork() public virtual {
+        _setUpFork(true);
+    }
 
-        cheats.createSelectFork(cheats.rpcUrl("mainnet"), mainnetForkBlock);
+    /// @dev Sets up the integration tests for mainnet.
+    function _setUpFork(bool upgrade) public virtual {
```
wadealexc, line 216: This function is pretty messy, can we clean it up?

- Confusion between `bool upgrade` vs `forkConfig.upgradeBeforeTesting` vs `isUpgraded`
- `forkConfig.supportEigenPodTests` evaluated in multiple places; let's please keep that to 1 (im not sure what this is for in the first place though)
- First part of the method does either `_executeTimelockUpgrade` or `_deployProxies + _upgradeProxies`. Later in the method, we might also do `_upgradeMainnetContracts`. Is that mutually exclusive, or can that happen in addition to the earlier upgrade stuff? Logic should be more clear.

wadealexc: Oh, I realized `_upgradeMainnetContracts` is just `_upgradeProxies`, but disguised... let's collapse these definitions and condense upgrade logic pls

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -131,503 +94,445 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
...
+    /// @dev Sets up the integration tests for local.
+    function _setUpLocal() public virtual {
+        console.log("Setting up `%s` integration tests:", "LOCAL".yellow().bold());
+
+        // Deploy ProxyAdmin, PauserRegistry, and executorMultisig.
+        config.governance.proxyAdmin = new ProxyAdmin();
+        config.governance.pauserRegistry = new PauserRegistry(PAUSER.toArray(), UNPAUSER);
+        config.governance.executorMultisig = address(proxyAdmin().owner());
 
         _deployProxies();
```
wadealexc, line 177: I see we got rid of `_deployImplementations` here, which is fine - I think moving this to `_upgradeProxies` is clean.

That said, it feels little inconsistent that the `Config` contract has fields for "impls" - but I don't think we set them anywhere. IIUC, these are only set if we're doing a fork test and they get read from config?

We should either:
* Remove impls from config/parsing entirely. Maybe they're just not needed!
* Make sure config is updated with impls no matter what env you're running in

src/test/integration/IntegrationDeployer.t.sol
```diff
@@ -131,503 +94,445 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
...
+        // string memory profile = FOUNDRY_PROFILE();
+
+        user = userType == DEFAULT ? new User(name) : User(new User_AltMethods(name));
+
+        assertTrue(address(user) != address(0), "User is not initialized");
+        // if (eq(profile, "default") || eq(profile, "mainnet") || (eq(profile, "forktest") && isUpgraded)) {
+        //     user = userType == DEFAULT ? new User(name) : User(new User_AltMethods(name));
+        // } else if (eq(profile, "forktest") && !isUpgraded) {
+        //     user = User(new User_M2(name));
+        // }
```
wadealexc, line 499: If these are fully disabled lets just completely get rid of em.

src/test/integration/tests/eigenpod/FullySlashed_EigenPod.t.sol
```diff
@@ -1,22 +1,22 @@
...
+
         _configAssetTypes(HOLDS_ETH);
         (staker, strategies, initTokenBalances) = _newRandomStaker();
-        cheats.assume(initTokenBalances[0] >= 64 ether);
+
+        // Ensure the staker has at least 64 ETH to deposit.
+        if (initTokenBalances[0] < 64 ether) {
+            initTokenBalances[0] = 64 ether;
+            cheats.deal(address(staker), 64 ether);
+        }
```
wadealexc, line 19: we just need better "create user" syntax tbh. can leave that for a followup PR though.

src/test/integration/users/User.t.sol
```diff
@@ -56,20 +25,12 @@ contract User is Logger, IDelegationManagerTypes, IAllocationManagerTypes {
...
-        delegationManager = deployer.delegationManager();
-        strategyManager = deployer.strategyManager();
-        eigenPodManager = deployer.eigenPodManager();
-
-        timeMachine = deployer.timeMachine();
-        beaconChain = deployer.beaconChain();
-
-        _createPod();
+    constructor(string memory name) {
+        config = ConfigGetters(address(msg.sender));
```
wadealexc, line 31: Why not just inherit ConfigGetters?

I'm actually a little lost on what's going on here - are we calling methods on User when we access `config.delegationManager()`?

src/test/mocks/BeaconChainMock.t.sol
```diff
@@ -176,7 +176,7 @@ contract BeaconChainMock is Logger {
     /// - Setting their current/effective balance
     /// - Assigning them a new, unique index
     function newValidator(bytes memory withdrawalCreds) public payable returns (uint40) {
-        print.method("newValidator");
+        console.log("BeaconChain.newValidator()");
```
wadealexc, line 179: Why get rid of Logger?

Feels like we print inconsistently in a lot of places,,

wadealexc (changes requested):
See questions/comments above.

Overall nice work, there's a lot of good cleanup here.

PR #1346, opened 2025-04-28
feat(draft): `AllocationManager` redistribution support

0xClandestine:
**Motivation:**

Add support for redistributing slashed funds to specified recipients instead of burning them, while optimizing code size and improving test coverage.

**Modifications:**

Added a new redistribution system that allows AVSs to specify recipients for slashed funds. Code optimizations were made to reduce codesize, including consolidating repeated logic and a small `PermissionControllerMixin` refactor.

**Result:**

Slashed funds can now be redirected to specified recipients instead of being burned, with improved code efficiency and test coverage.

PR #1355, opened 2025-05-06
feat(draft): `AllocationManager` redistribution support (#1346)

0xClandestine:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/src/contracts/core/AllocationManager.sol
+++ b/src/contracts/core/AllocationManager.sol
@@ -64,14 +64,24 @@ contract AllocationManager is
     function slashOperator(
         address avs,
         SlashingParams calldata params
-    ) external onlyWhenNotPaused(PAUSED_OPERATOR_SLASHING) checkCanCall(avs) {
+    )
+        external
+        onlyWhenNotPaused(PAUSED_OPERATOR_SLASHING)
+        checkCanCall(avs)
+        returns (uint256 slashId, uint256[] memory shares)
+    {
         // Check that the operator set exists and the operator is registered to it
         OperatorSet memory operatorSet = OperatorSet(avs, params.operatorSetId);
-        require(params.strategies.length == params.wadsToSlash.length, InputArrayLengthMismatch());
-        require(_operatorSets[operatorSet.avs].contains(operatorSet.id), InvalidOperatorSet());
+        _checkArrayLengthsMatch(params.strategies.length, params.wadsToSlash.length);
+        _checkIsOperatorSet(operatorSet);
         require(isOperatorSlashable(params.operator, operatorSet), OperatorNotSlashable());
 
         uint256[] memory wadSlashed = new uint256[](params.strategies.length);
+
+        // Increment the slash count for the operator set, first `slashId` equals 1.
+        slashId = ++_slashCount[operatorSet.key()];
+        // Initialize the shares array.
+        shares = new uint256[](params.strategies.length);
 
         // For each strategy in the operator set, slash any existing allocation
         for (uint256 i = 0; i < params.strategies.length; i++) {
@@ -155,7 +165,7 @@ contract AllocationManager is
     ) external onlyWhenNotPaused(PAUSED_MODIFY_ALLOCATIONS) {
         // Check that the caller is allowed to modify allocations on behalf of the operator
         // We do not use a modifier to avoid `stack too deep` errors
-        require(_checkCanCall(operator), InvalidCaller());
+        _checkCanCall(operator);
 
         // Check that the operator exists and has configured an allocation delay
         uint32 operatorAllocationDelay;
@@ -166,7 +176,7 @@ contract AllocationManager is
         }
 
         for (uint256 i = 0; i < params.length; i++) {
-            require(params[i].strategies.length == params[i].newMagnitudes.length, InputArrayLengthMismatch());
+            _checkArrayLengthsMatch(params[i].strategies.length, params[i].newMagnitudes.length);
 
             // Check that the operator set exists and get the operator's registration status
             // Operators do not need to be registered for an operator set in order to allocate
@@ -174,7 +184,7 @@ contract AllocationManager is
             // allocate magnitude before registering, as AVS's will likely only accept
             // registrations from operators that are already slashable.
             OperatorSet memory operatorSet = params[i].operatorSet;
-            require(_operatorSets[operatorSet.avs].contains(operatorSet.id), InvalidOperatorSet());
+            _checkIsOperatorSet(operatorSet);
 
             bool _isOperatorSlashable = isOperatorSlashable(operator, operatorSet);
 
@@ -246,8 +256,7 @@ contract AllocationManager is
         IStrategy[] calldata strategies,
         uint16[] calldata numToClear
     ) external onlyWhenNotPaused(PAUSED_MODIFY_ALLOCATIONS) {
-        require(strategies.length == numToClear.length, InputArrayLengthMismatch());
-
+        _checkArrayLengthsMatch(strategies.length, numToClear.length);
         for (uint256 i = 0; i < strategies.length; ++i) {
             _clearDeallocationQueue({operator: operator, strategy: strategies[i], numToClear: numToClear[i]});
         }
@@ -258,13 +267,13 @@ contract AllocationManager is
         address operator,
         RegisterParams calldata params
     ) external onlyWhenNotPaused(PAUSED_OPERATOR_SET_REGISTRATION_AND_DEREGISTRATION) checkCanCall(operator) {
-        // Check that the operator exists
-        require(delegation.isOperator(operator), InvalidOperator());
+        // Check if the operator has registered.
+        _checkIsOperator(operator);
 
         for (uint256 i = 0; i < params.operatorSetIds.length; i++) {
             // Check the operator set exists and the operator is not currently registered to it
             OperatorSet memory operatorSet = OperatorSet(params.avs, params.operatorSetIds[i]);
-            require(_operatorSets[operatorSet.avs].contains(operatorSet.id), InvalidOperatorSet());
+            _checkIsOperatorSet(operatorSet);
             require(!isOperatorSlashable(operator, operatorSet), AlreadyMemberOfSet());
 
             // Add operator to operator set
@@ -285,12 +294,12 @@ contract AllocationManager is
         DeregisterParams calldata params
     ) external onlyWhenNotPaused(PAUSED_OPERATOR_SET_REGISTRATION_AND_DEREGISTRATION) {
         // Check that the caller is either authorized on behalf of the operator or AVS
-        require(_checkCanCall(params.operator) || _checkCanCall(params.avs), InvalidCaller());
+        require(_canCall(params.operator) || _canCall(params.avs), InvalidPermissions());
 
         for (uint256 i = 0; i < params.operatorSetIds.length; i++) {
             // Check the operator set exists and the operator is registered to it
             OperatorSet memory operatorSet = OperatorSet(params.avs, params.operatorSetIds[i]);
-            require(_operatorSets[params.avs].contains(operatorSet.id), InvalidOperatorSet());
+            _checkIsOperatorSet(operatorSet);
             require(registrationStatus[params.operator][operatorSet.key()].registered, NotMemberOfSet());
 
             // Remove operator from operator set
@@ -313,8 +322,8 @@ contract AllocationManager is
     /// @inheritdoc IAllocationManager
     function setAllocationDelay(address operator, uint32 delay) external {
         if (msg.sender != address(delegation)) {
-            require(_checkCanCall(operator), InvalidCaller());
-            require(delegation.isOperator(operator), InvalidOperator());
+            _checkCanCall(operator);
+            _checkIsOperator(operator);
         }
         _setAllocationDelay(operator, delay);
     }
@@ -330,31 +339,28 @@ contract AllocationManager is
 
     /// @inheritdoc IAllocationManager
     function updateAVSMetadataURI(address avs, string calldata metadataURI) external checkCanCall(avs) {
-        if (!_avsRegisteredMetadata[avs]) {
-            _avsRegisteredMetadata[avs] = true;
-        }
-
+        if (!_avsRegisteredMetadata[avs]) _avsRegisteredMetadata[avs] = true;
         emit AVSMetadataURIUpdated(avs, metadataURI);
     }
 
     /// @inheritdoc IAllocationManager
     function createOperatorSets(address avs, CreateSetParams[] calldata params) external checkCanCall(avs) {
-        // Check that the AVS exists and has registered metadata
         require(_avsRegisteredMetadata[avs], NonexistentAVSMetadata());
-
         for (uint256 i = 0; i < params.length; i++) {
-            OperatorSet memory operatorSet = OperatorSet(avs, params[i].operatorSetId);
+            _createOperatorSet(avs, params[i], DEFAULT_BURN_ADDRESS);
+        }
+    }
 
-            // Create the operator set, ensuring it does not already exist
-            require(_operatorSets[avs].add(operatorSet.id), InvalidOperatorSet());
-            emit OperatorSetCreated(OperatorSet(avs, operatorSet.id));
-
-            // Add strategies to the operator set
-            bytes32 operatorSetKey = operatorSet.key();
-            for (uint256 j = 0; j < params[i].strategies.length; j++) {
-                _operatorSetStrategies[operatorSetKey].add(address(params[i].strategies[j]));
-                emit StrategyAddedToOperatorSet(operatorSet, params[i].strategies[j]);
-            }
+    /// @inheritdoc IAllocationManager
+    function createRedistributingOperatorSets(
+        address avs,
+        CreateSetParams[] calldata params,
+        address[] calldata redistributionRecipients
+    ) external checkCanCall(avs) {
+        _checkArrayLengthsMatch(params.length, redistributionRecipients.length);
+        require(_avsRegisteredMetadata[avs], NonexistentAVSMetadata());
+        for (uint256 i = 0; i < params.length; i++) {
+            _createOperatorSet(avs, params[i], redistributionRecipients[i]);
         }
     }
 
@@ -365,10 +371,8 @@ contract AllocationManager is
         IStrategy[] calldata strategies
     ) external checkCanCall(avs) {
         OperatorSet memory operatorSet = OperatorSet(avs, operatorSetId);
+        _checkIsOperatorSet(operatorSet);
         bytes32 operatorSetKey = operatorSet.key();
-
-        require(_operatorSets[avs].contains(operatorSet.id), InvalidOperatorSet());
-
         for (uint256 i = 0; i < strategies.length; i++) {
             require(_operatorSetStrategies[operatorSetKey].add(address(strategies[i])), StrategyAlreadyInOperatorSet());
             emit StrategyAddedToOperatorSet(operatorSet, strategies[i]);
@@ -382,8 +386,7 @@ contract AllocationManager is
         IStrategy[] calldata strategies
     ) external checkCanCall(avs) {
         OperatorSet memory operatorSet = OperatorSet(avs, operatorSetId);
-        require(_operatorSets[avs].contains(operatorSet.id), InvalidOperatorSet());
-
+        _checkIsOperatorSet(operatorSet);
         bytes32 operatorSetKey = operatorSet.key();
         for (uint256 i = 0; i < strategies.length; i++) {
             require(_operatorSetStrategies[operatorSetKey].remove(address(strategies[i])), StrategyNotInOperatorSet());
@@ -396,6 +399,41 @@ contract AllocationManager is
      *                         INTERNAL FUNCTIONS
      *
      */
+
+    /**
+     * @notice Creates a new operator set for an AVS.
+     * @param avs The AVS address that owns the operator set.
+     * @param params The parameters for creating the operator set.
+     * @param redistributionRecipient Address to receive redistributed funds when operators are slashed.
+     * @dev If `redistributionRecipient` is address(0), the operator set is considered non-redistributing
+     * and slashed funds are sent to the `DEFAULT_BURN_ADDRESS`.
+     * @dev Providing `BEACONCHAIN_ETH_STRAT` as a strategy will revert since it's not currently supported.
+     */
+    function _createOperatorSet(
+        address avs,
+        CreateSetParams calldata params,
+        address redistributionRecipient
+    ) internal {
+        OperatorSet memory operatorSet = OperatorSet(avs, params.operatorSetId);
+
+        // Create the operator set, ensuring it does not already exist.
+        require(_operatorSets[avs].add(operatorSet.id), InvalidOperatorSet());
+        emit OperatorSetCreated(operatorSet);
+
+        bool isRedistributing = redistributionRecipient != DEFAULT_BURN_ADDRESS;
+
+        if (isRedistributing) {
+            _redistributionRecipients[operatorSet.key()] = redistributionRecipient;
+            emit RedistributionAddressSet(operatorSet, redistributionRecipient);
+        }
+
+        // Add strategies to the operator set
+        for (uint256 j = 0; j < params.strategies.length; j++) {
+            if (isRedistributing && params.strategies[j] == BEACONCHAIN_ETH_STRAT) revert InvalidStrategy();
+            _operatorSetStrategies[operatorSet.key()].add(address(params.strategies[j]));
+            emit StrategyAddedToOperatorSet(operatorSet, params.strategies[j]);
+        }
+    }
 
     /**
      * @dev Clear one or more pending deallocations to a strategy's allocated magnitude
@@ -603,6 +641,25 @@ contract AllocationManager is
     /// @dev Use safe casting when downcasting to uint64
     function _addInt128(uint64 a, int128 b) internal pure returns (uint64) {
         return uint256(int256(int128(uint128(a)) + b)).toUint64();
+    }
+
+    /// @dev Reverts if the operator set does not exist.
+    function _checkIsOperatorSet(
+        OperatorSet memory operatorSet
+    ) internal view {
+        require(_operatorSets[operatorSet.avs].contains(operatorSet.id), InvalidOperatorSet());
+    }
+
+    /// @dev Reverts if the provided arrays have different lengths.
+    function _checkArrayLengthsMatch(uint256 left, uint256 right) internal pure {
+        require(left == right, InputArrayLengthMismatch());
+    }
+
+    /// @dev Reverts if the operator is not registered.
+    function _checkIsOperator(
+        address operator
+    ) internal view {
+        require(delegation.isOperator(operator), InvalidOperator());
     }
 
     /**
@@ -904,5 +961,28 @@ contract AllocationManager is
         // slashableUntil returns the last block the operator is slashable in so we check for
         // less than or equal to
         return status.registered || block.number <= status.slashableUntil;
+    }
+
+    /// @inheritdoc IAllocationManager
+    function getRedistributionRecipient(
+        OperatorSet memory operatorSet
+    ) external view returns (address) {
+        // Load the redistribution recipient and return it if set, otherwise return the default burn address.
+        address redistributionRecipient = _redistributionRecipients[operatorSet.key()];
+        return redistributionRecipient == address(0) ? DEFAULT_BURN_ADDRESS : redistributionRecipient;
+    }
+
+    /// @inheritdoc IAllocationManager
+    function isRedistributingOperatorSet(
+        OperatorSet memory operatorSet
+    ) external view returns (bool) {
+        return _redistributionRecipients[operatorSet.key()] != address(0);
+    }
+
+    /// @inheritdoc IAllocationManager
+    function getSlashCount(
+        OperatorSet memory operatorSet
+    ) external view returns (uint256) {
+        return _slashCount[operatorSet.key()];
     }
 }
--- a/src/contracts/core/AllocationManagerStorage.sol
+++ b/src/contracts/core/AllocationManagerStorage.sol
@@ -14,6 +14,12 @@ abstract contract AllocationManagerStorage is IAllocationManager {
 
     // Constants
 
+    /// @dev The default burn address for slashed funds.
+    address internal constant DEFAULT_BURN_ADDRESS = 0x00000000000000000000000000000000000E16E4;
+
+    /// @dev The beacon chain ETH strategy.
+    IStrategy internal constant BEACONCHAIN_ETH_STRAT = IStrategy(0xbeaC0eeEeeeeEEeEeEEEEeeEEeEeeeEeeEEBEaC0);
+
     /// @dev Index for flag that pauses operator allocations/deallocations when set.
     uint8 internal constant PAUSED_MODIFY_ALLOCATIONS = 0;
 
@@ -93,6 +99,16 @@ abstract contract AllocationManagerStorage is IAllocationManager {
     /// @notice bool is not used and is always true if the avs has registered metadata
     mapping(address avs => bool) internal _avsRegisteredMetadata;
 
+    /// @notice Returns the number of slashes for a given operator set.
+    /// @dev This is also used as a unique slash identifier.
+    /// @dev This tracks the number of slashes after the redistribution release.
+    mapping(bytes32 operatorSetKey => uint256 slashId) public _slashCount;
+
+    /// @notice Returns the address where slashed funds will be sent for a given operator set.
+    /// @dev For redistributing Operator Sets, returns the configured redistribution address set during Operator Set creation.
+    ///      For non-redistributing or non-existing operator sets, returns `address(0)`.
+    mapping(bytes32 operatorSetKey => address redistributionAddr) internal _redistributionRecipients;
+
     // Construction
 
     constructor(IDelegationManager _delegation, uint32 _DEALLOCATION_DELAY, uint32 _ALLOCATION_CONFIGURATION_DELAY) {
@@ -106,5 +122,5 @@ abstract contract AllocationManagerStorage is IAllocationManager {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[36] private __gap;
+    uint256[34] private __gap;
 }
--- a/src/contracts/core/DelegationManager.sol
+++ b/src/contracts/core/DelegationManager.sol
@@ -158,7 +158,7 @@ contract DelegationManager is
         if (msg.sender != staker) {
             address operator = delegatedTo[staker];
 
-            require(_checkCanCall(operator) || msg.sender == delegationApprover(operator), CallerCannotUndelegate());
+            require(_canCall(operator) || msg.sender == delegationApprover(operator), CallerCannotUndelegate());
             emit StakerForceUndelegated(staker, operator);
         }
 
--- a/src/contracts/interfaces/IAllocationManager.sol
+++ b/src/contracts/interfaces/IAllocationManager.sol
@@ -20,8 +20,8 @@ interface IAllocationManagerErrors {
 
     /// Caller
 
-    /// @dev Thrown when caller is not authorized to call a function.
-    error InvalidCaller();
+    /// @dev Thrown when an invalid strategy is provided.
+    error InvalidStrategy();
 
     /// Operator Status
 
@@ -209,6 +209,9 @@ interface IAllocationManagerEvents is IAllocationManagerTypes {
     /// @notice Emitted when an operator is removed from an operator set.
     event OperatorRemovedFromOperatorSet(address indexed operator, OperatorSet operatorSet);
 
+    /// @notice Emitted when a redistributing operator set is created by an AVS.
+    event RedistributionAddressSet(OperatorSet operatorSet, address redistributionRecipient);
+
     /// @notice Emitted when a strategy is added to an operator set.
     event StrategyAddedToOperatorSet(OperatorSet operatorSet, IStrategy strategy);
 
@@ -244,7 +247,10 @@ interface IAllocationManager is IAllocationManagerErrors, IAllocationManagerEven
      *      rounding, which will result in small amounts of tokens locked in the contract
      *      rather than fully burning through the burn mechanism.
      */
-    function slashOperator(address avs, SlashingParams calldata params) external;
+    function slashOperator(
+        address avs,
+        SlashingParams calldata params
+    ) external returns (uint256 slashId, uint256[] memory shares);
 
     /**
      * @notice Modifies the proportions of slashable stake allocated to an operator set from a list of strategies
@@ -324,6 +330,20 @@ interface IAllocationManager is IAllocationManagerErrors, IAllocationManagerEven
      * @notice Allows an AVS to create new operator sets, defining strategies that the operator set uses
      */
     function createOperatorSets(address avs, CreateSetParams[] calldata params) external;
+
+    /**
+     * @notice Allows an AVS to create new Redistribution operator sets.
+     * @param avs The AVS creating the new operator sets.
+     * @param params An array of operator set creation parameters.
+     * @param redistributionRecipients An array of addresses that will receive redistributed funds when operators are slashed.
+     * @dev Same logic as `createOperatorSets`, except `redistributionRecipients` corresponding to each operator set are stored.
+     *      Additionally, emits `RedistributionOperatorSetCreated` event instead of `OperatorSetCreated` for each created operator set.
+     */
+    function createRedistributingOperatorSets(
+        address avs,
+        CreateSetParams[] calldata params,
+        address[] calldata redistributionRecipients
+    ) external;
 
     /**
      * @notice Allows an AVS to add strategies to an operator set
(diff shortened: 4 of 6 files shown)


src/contracts/interfaces/IStrategyManager.sol
non-fungible-nelson, line 1: Should we remove burn shares here? or unnecessary optimization?

script/releases/v1.4.1-slashing/1-deployContracts.s.sol
```diff
@@ -144,7 +144,7 @@ contract Deploy is EOADeployer {
 
         DelegationManager delegation = Env.impl.delegationManager();
         vm.expectRevert(errInit);
-        delegation.initialize(address(0), 0);
+        delegation.initialize(0);
```
ypatil12, line 147: WE can delete these deploy scripts

src/contracts/core/DelegationManager.sol
```diff
@@ -848,6 +884,37 @@ contract DelegationManager is
             : IShareManager(address(strategyManager));
     }
 
+    function _checkCallerIsStrategyManagerOrEigenPodManager() internal view {
```
ypatil12, line 887: Might not be worth to do anymore for space/minimal diff, since deprecating ownable is doing a lot of the heavy lifting

src/contracts/core/DelegationManager.sol
```diff
@@ -683,6 +673,52 @@ contract DelegationManager is
...
+    function _slashOperatorShares(
+        address operator,
+        OperatorSet memory operatorSet,
+        uint256 slashId,
+        IStrategy strategy,
+        uint64 prevMaxMagnitude,
+        uint64 newMaxMagnitude
+    ) internal returns (uint256 totalDepositSharesToBurn) {
+        // Avoid emitting events if nothing has changed for sanitization.
+        if (prevMaxMagnitude != newMaxMagnitude) {
```
ypatil12, line 687: Style: generally we continue on if, instead of having large if blocks

src/contracts/core/DelegationManager.sol
```diff
@@ -542,19 +525,26 @@ contract DelegationManager is
...
+        }
         bytes32 withdrawalRoot = calculateWithdrawalRoot(withdrawal);
         require(pendingWithdrawals[withdrawalRoot], WithdrawalNotQueued());
 
         uint256[] memory prevSlashingFactors;
         {
             // slashableUntil is block inclusive so we need to check if the current block is strictly greater than the slashableUntil block
             // meaning the withdrawal can be completed.
-            uint32 slashableUntil = withdrawal.startBlock + MIN_WITHDRAWAL_DELAY_BLOCKS;
+            // TODO: update delay blocks for redistribution + EIGEN_REDISTRIBUTION_DELAY_BLOCKS
```
ypatil12, line 545: This is from POC, can get rid of this

src/contracts/core/AllocationManagerStorage.sol
```diff
@@ -6,14 +6,19 @@ import "@openzeppelin/contracts/utils/structs/DoubleEndedQueue.sol";
...
 abstract contract AllocationManagerStorage is IAllocationManager {
     using Snapshots for Snapshots.DefaultWadHistory;
 
     // Constants
 
+    /// @dev The default burn address for slashed funds.
+    address internal constant DEFAULT_BURN_ADDRESS = 0x00000000000000000000000000000000000E16E4;
+
+    /// @dev The beacon chain ETH strategy.
+    IStrategy internal constant BEACONCHAIN_ETH_STRAT = IStrategy(0xbeaC0eeEeeeeEEeEeEEEEeeEEeEeeeEeeEEBEaC0);
```
ypatil12, line 20: We're not using this anywhere. I believe the original intention was to prevent creation of redistributable operator sets with nativeETH. I'm OK with preventing this.

src/contracts/interfaces/IAllocationManager.sol
```diff
@@ -176,6 +176,9 @@ interface IAllocationManagerEvents is IAllocationManagerTypes {
     /// @notice Emitted when operator updates their allocation delay.
     event AllocationDelaySet(address operator, uint32 delay, uint32 effectBlock);
 
+    /// @notice Emitted when an AVS updates the redistribution delay for a given strategy.
+    event RedistributionDelaySet(address avs, IStrategy strategy, uint32 delay);
```
ypatil12, line 180: Not used

src/contracts/core/SlashingWithdrawalRouter.sol
```diff
@@ -0,0 +1,406 @@
...
+    /// @inheritdoc ISlashingWithdrawalRouter
+    function getBurnOrRedistributionStartBlock(
+        OperatorSet calldata operatorSet,
+        uint256 slashId
+    ) public view returns (uint256) {
+        return _slashIdToStartBlock[operatorSet.key()][slashId];
+    }
+
+    /// @inheritdoc ISlashingWithdrawalRouter
+    function getStrategyBurnOrRedistributionDelay(
```
ypatil12, line 390: We should make sure we have a unit test where:

1. Two strategies, one with a strategy delay, one without. Complete first, warp, complete second

ypatil12 (changes requested):

script/releases/v1.4.1-slashing/1-deployContracts.s.sol
ypatil12: (I can do so in the upgrade script PR)

src/contracts/interfaces/IStrategyManager.sol
```diff
@@ -45,10 +45,10 @@ interface IStrategyManagerEvents {
     event StrategyRemovedFromDepositWhitelist(IStrategy strategy);
 
     /// @notice Emitted when an operator is slashed and shares to be burned are increased
-    event BurnableSharesIncreased(IStrategy strategy, uint256 shares);
+    event BurnableSharesIncreased(OperatorSet operatorSet, uint256 slashId, IStrategy strategy, uint256 shares);
```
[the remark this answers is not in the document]
ypatil12: (let's wait till fixing the shares bug to decide)

src/contracts/interfaces/IStrategyManager.sol
```diff
@@ -45,10 +45,10 @@ interface IStrategyManagerEvents {
...
 
     /// @notice Emitted when shares are burned
```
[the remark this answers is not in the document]
ypatil12: (let's wait till fixing the shares bug to decide)

src/contracts/core/StrategyManager.sol
```diff
@@ -143,12 +147,21 @@ contract StrategyManager is
...
+        uint256 sharesToBurn
     ) external onlyDelegationManager nonReentrant {
-        (, uint256 currentShares) = EnumerableMap.tryGet(burnableShares, address(strategy));
-        EnumerableMap.set(burnableShares, address(strategy), currentShares + addedSharesToBurn);
-        emit BurnableSharesIncreased(strategy, addedSharesToBurn);
+        emit BurnableSharesIncreased(operatorSet, slashId, strategy, sharesToBurn);
+
+        if (sharesToBurn != 0) {
+            // Withdraw the shares to the EigenLayer `SlashingWithdrawalRouter` contract.
+            uint256 amountOut =
```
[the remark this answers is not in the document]
0xClandestine: Oh rip, that complicates things for sure.

src/contracts/core/AllocationManager.sol
```diff
@@ -55,97 +53,31 @@ contract AllocationManager is
...
     function slashOperator(
         address avs,
         SlashingParams calldata params
-    ) external onlyWhenNotPaused(PAUSED_OPERATOR_SLASHING) checkCanCall(avs) {
+    )
+        external
+        onlyWhenNotPaused(PAUSED_OPERATOR_SLASHING)
+        checkCanCall(avs)
+        returns (uint256 slashId, uint256[] memory shares)
+    {
```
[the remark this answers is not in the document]
0xClandestine: That's fine, they were mostly added to be removed at the end if no longer needed.

src/contracts/interfaces/IAllocationManager.sol
0xClandestine: removed in test pr

src/contracts/core/SlashingWithdrawalRouter.sol
0xClandestine: added in test pr already

0xClandestine: 4180e5f

src/contracts/interfaces/IAllocationManager.sol
0xClandestine: 4180e5f

src/contracts/core/DelegationManager.sol
```diff
@@ -521,13 +496,21 @@ contract DelegationManager is
             scaledShares: scaledShares
         });
 
+        bytes32 withdrawalRoot = _addWithdrawalToQueue(withdrawal);
+
+        emit SlashingWithdrawalQueued(withdrawalRoot, withdrawal, withdrawableShares);
+        return withdrawalRoot;
+    }
+
+    function _addWithdrawalToQueue(
```
ypatil12, line 505: Let's remove this

src/contracts/interfaces/IAllocationManager.sol
wadealexc: What's the deal with this? It doesn't seem removed, and the commit you linked (seems) to be from a branch that was merged into this branch?

Not sure what I'm missing, but wanted to flag that this event is still here.

wadealexc (changes requested):
Before I forget, we need some stuff ASAP (before audit at the very least):
- Update function docs for ALM.slashOperator to reflect any functional changes as well as the new return values
- Update contract docs for ALM/DM/StrategyManager/etc
- Create contract docs for SlashingWithdrawalRouter or any other new contracts
-

src/contracts/strategies/StrategyBase.sol
```diff
@@ -154,7 +154,7 @@ contract StrategyBase is Initializable, Pausable, IStrategy, SemVerMixin {
         address recipient,
         IERC20 token,
         uint256 amountShares
-    ) external virtual override onlyWhenNotPaused(PAUSED_WITHDRAWALS) onlyStrategyManager {
+    ) external virtual override onlyWhenNotPaused(PAUSED_WITHDRAWALS) onlyStrategyManager returns (uint256 amountOut) {
```
wadealexc, line 157: Note - remember we'll need to upgrade not just the beacon, but also the individual strategies

src/test/integration/IntegrationBase.t.sol
```diff
@@ -2186,7 +2189,7 @@ abstract contract IntegrationBase is IntegrationDeployer, TypeImporter {
         }
     }
 
-    function _genSlashing_Half(User operator, OperatorSet memory operatorSet) internal view returns (SlashingParams memory params) {
+    function _genSlashing_Half(User operator, OperatorSet memory operatorSet) internal returns (SlashingParams memory params) {
```
wadealexc, line 2192: Where did `view` go here? Can we put these back for each of the functions in this file, to keep unneeded diff out?

src/test/integration/IntegrationBase.t.sol
```diff
@@ -2199,7 +2202,7 @@ abstract contract IntegrationBase is IntegrationDeployer, TypeImporter {
         }
     }
 
-    function _genSlashing_Full(User operator, OperatorSet memory operatorSet) internal view returns (SlashingParams memory params) {
+    function _genSlashing_Full(User operator, OperatorSet memory operatorSet) internal returns (SlashingParams memory params) {
```
wadealexc, line 2205: Revert change

src/test/integration/IntegrationBase.t.sol
```diff
@@ -2214,7 +2217,6 @@ abstract contract IntegrationBase is IntegrationDeployer, TypeImporter {
 
     function _genSlashing_Custom(User operator, OperatorSet memory operatorSet, uint wadsToSlash)
         internal
-        view
```
wadealexc, line 2217: Revert change

src/test/integration/IntegrationBase.t.sol
```diff
@@ -2234,7 +2236,6 @@ abstract contract IntegrationBase is IntegrationDeployer, TypeImporter {
 
     function _strategiesAndWadsForFullSlash(OperatorSet memory operatorSet)
         internal
-        view
```
wadealexc, line 2237: Revert change

src/test/integration/IntegrationChecks.t.sol
```diff
@@ -1065,7 +1065,7 @@ contract IntegrationCheckUtils is IntegrationBase {
         assert_Snap_Slashed_SlashableStake(operator, operatorSet, slashParams, "slash should lower slashable stake");
         assert_Snap_Slashed_OperatorShares(operator, slashParams, "slash should remove operator shares");
         assert_Snap_Slashed_Allocation(operator, operatorSet, slashParams, "slash should reduce current magnitude");
-        assert_Snap_Increased_BurnableShares(operator, slashParams, "slash should increase burnable shares");
+        // assert_Snap_Increased_BurnableShares(operatorSet, operator, slashParams, "slash should increase burnable shares");
```
wadealexc, line 1068: Adding a comment here to make sure this doesn't stay commented out :smile:

src/contracts/mixins/Deprecated_OwnableUpgradeable.sol
```diff
@@ -0,0 +1,26 @@
...
+ * This is useful in cases where a contract previously used OwnableUpgradeable but no longer
+ * needs ownership functionality, yet must maintain the same storage slots to ensure
+ * compatibility with existing deployed proxies.
+ *
+ * The contract preserves the same storage layout as OwnableUpgradeable:
+ * - It keeps the `_owner` storage variable in the same slot
+ * - It maintains the same storage gap for future upgrades
+ */
+abstract contract Deprecated_OwnableUpgradeable {
+    address private _owner;
```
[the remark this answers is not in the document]
wadealexc: Oh, yep, just realized our CI is _very_ unhappy about this change. Looks like it definitely broke storage layouts in the DelegationManager.

I don't see any storage layout checks for the ALM, so - requesting we add checks for that in this PR, please!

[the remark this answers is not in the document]
wadealexc: Looks like `.github/configs/storage-diff.json` doesn't include:
- AllocationManager
- Eigen
- BackingEigen
- SlashingWithdrawalRouter

I know `SlashingWithdrawalRouter` isn't on mainnet yet... adding it here because it will need to be added to this list eventually.

How do we want to keep this list up to date, going forward? I feel like missing the ALM here was a "close one" (though, granted: manual review, the DM checks, and integration fork tests all got our back)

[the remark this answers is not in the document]
0xClandestine: Was using fork PRs before so storage tests weren't running at all. Great catch ser!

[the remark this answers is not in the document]
0xClandestine: CI is updated in my integration refactor, it's just not merged yet.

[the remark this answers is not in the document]
0xClandestine: Think we plan to create a registry contract to track our deployments on any given chain. Could maybe plug into that to deduce what contracts to run the tests on.

src/test/integration/IntegrationChecks.t.sol
0xClandestine: We no longer increase og burnable shares, but will be adding a new variable to track on an operator set strategy basis so this will need to be added back.

src/test/integration/IntegrationBase.t.sol
0xClandestine: New foundry cheatcodes vm.sort() and vm.shuffle() require the fn is not view.

src/test/integration/IntegrationBase.t.sol
0xClandestine: see above

src/test/integration/IntegrationBase.t.sol
0xClandestine: see above

src/test/integration/IntegrationBase.t.sol
0xClandestine: see above

src/contracts/strategies/StrategyBase.sol
0xClandestine: Have added a note in the abacus, good point though.

src/contracts/core/AllocationManager.sol
```diff
@@ -605,6 +656,41 @@ contract AllocationManager is
...
+        address operator
+    ) internal view {
+        require(delegation.isOperator(operator), InvalidOperator());
+    }
+
+    function _checkAVSMetadata(
+        address avs
+    ) internal view {
+        require(_avsRegisteredMetadata[avs], NonexistentAVSMetadata());
+    }
```
wadealexc, line 682: Request - can we rename this to `_checkAVSExists`? Or at least comment that this is an existence check?

src/contracts/interfaces/IAllocationManager.sol
0xClandestine: weird, rebase error (had to dupe a pr) will fix.

0xClandestine: just pushed downstream

src/contracts/core/AllocationManager.sol
```diff
@@ -396,6 +319,134 @@ contract AllocationManager is
...
+        emit StrategyAddedToOperatorSet(operatorSet, strategy);
+    }
+
+    /**
+     * @notice Creates a new operator set for an AVS.
+     * @param avs The AVS address that owns the operator set.
+     * @param params The parameters for creating the operator set.
+     * @param redistributionRecipient Address to receive redistributed funds when operators are slashed.
+     * @dev If `redistributionRecipient` is address(0), the operator set is considered non-redistributing
+     * and slashed funds are sent to the `DEFAULT_BURN_ADDRESS`.
```
wadealexc, line 424: Update comment - the non-redistribution creation method passes in the DEFAULT_BURN_ADDRESS.

What do we want to do if someone uses `createRedistributingOperatorSets(avs, params, address(0))`? Do we:
- reject `0` input
- set `0` as the recipient (this is what we currently do, not sure if that's desired)
- convert `0` to `DEFAULT_BURN_ADDRESS`

0xClandestine: Might be a good idea to sanitize, no negative impact otherwise though since if the value is set to address(0) we default to the burn address.

src/contracts/core/AllocationManager.sol
```diff
@@ -396,6 +319,134 @@ contract AllocationManager is
...
+
+        bool isRedistributing = redistributionRecipient != DEFAULT_BURN_ADDRESS;
+
+        if (isRedistributing) {
+            _redistributionRecipients[operatorSet.key()] = redistributionRecipient;
+            emit RedistributionAddressSet(operatorSet, redistributionRecipient);
+        }
+
+        for (uint256 j = 0; j < params.strategies.length; j++) {
+            require(!isRedistributing || params.strategies[j] != BEACONCHAIN_ETH_STRAT, InvalidStrategy());
```
wadealexc, line 446: Double negations / ORs are a little hard to parse. Can we add a clarifying comment here that explains that we are not supporting beacon chain ETH for redistributing opsets?

wadealexc: Bug: we don't have this check for `addStrategiesToOperatorSet`.

So... probably even better would be to add this check to `_addStrategyToOperatorSet`?

wadealexc: While I'm at it - the AND version of this requirement goes way harder in terms of legibility:

```
require(!(isRedistributing && params.strategies[j] == BEACONCHAIN_ETH_STRAT));
```

Makes it much clearer that there's a specific check we're making for the BC_ETH_STRAT

src/test/integration/IntegrationBase.t.sol
wadealexc: ah, all g then. ill resolve

src/contracts/core/AllocationManager.sol
wadealexc: Where do we default? And if that's true, it's another reason to clear this up - I can emit an event that something is a redistributing operator set, but pass in 0 as the recipient.

It ends up not being a redistributing operator set, but it's treated like one.

0xClandestine: Going to prevent zero address input

0xClandestine: [727a689](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1355/commits/727a689a6a0c9821a6b459e75a539a6e8d0b7376)

src/contracts/core/AllocationManager.sol
```diff
@@ -905,4 +1008,45 @@ contract AllocationManager is
...
+        address operator
+    ) external view returns (bool) {
+        EnumerableSet.Bytes32Set storage allocatedSets = allocatedSets[operator];
+        EnumerableSet.Bytes32Set storage registeredSets = registeredSets[operator];
+
+        uint256 allocatedLength = allocatedSets.length();
+        for (uint256 i = 0; i < allocatedLength; ++i) {
+            bytes32 operatorSetKey = allocatedSets.at(i);
+            // Check if the operator is both allocated and registered to this set
+            if (registeredSets.contains(operatorSetKey) && _redistributionRecipients[operatorSetKey] != address(0)) {
```
wadealexc, line 1046: Bug: see my previous comment. https://github.com/Layr-Labs/eigenlayer-contracts/pull/1355/files#r2105325283

This should call `isRedistributingOperatorSet(operatorSetKey)` rather than checking the `_redistributionRecipient` manually.

src/contracts/core/AllocationManager.sol
```diff
@@ -905,4 +1008,45 @@ contract AllocationManager is
...
+        // Load the redistribution recipient and return it if set, otherwise return the default burn address.
+        address redistributionRecipient = _redistributionRecipients[operatorSet.key()];
+        return redistributionRecipient == address(0) ? DEFAULT_BURN_ADDRESS : redistributionRecipient;
+    }
+
+    /// @inheritdoc IAllocationManager
+    function isRedistributingOperatorSet(
+        OperatorSet memory operatorSet
+    ) public view returns (bool) {
+        return _redistributionRecipients[operatorSet.key()] != address(0);
```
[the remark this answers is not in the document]
wadealexc: Actually, I'm wrong - newly-created non-redistributing operator sets have this set to 0.

src/contracts/core/AllocationManager.sol
wadealexc: Actually, I'm wrong - newly-created non-redistributing operator sets have this set to 0.

src/contracts/core/AllocationManager.sol
wadealexc, line 1025: `_redistributionRecipients` is a... "weird" storage slot, because we might transform the value stored in it. (i.e. if the value stored in the mapping is "0", we transform the value to `DEFAULT_BURN_ADDRESS`.

We have a method that specifically performs this transformation just above - `getRedistributionRecipient`.

When we have special storage slots like this, the best practice I'd like us to follow is to _NOT_ reference the underlying mapping, and instead, use the high-level transformation function wherever possible. This prevents possible footguns and allows us to only use the transformation logic in a single place - the specialized function. (another example in this contract is the way we use `_getUpdatedAllocation` wherever possible, rather than referencing the underlying `allocations` mapping).

So for this method, let's instead do this:

`return getRedistributionRecipient(operatorSet) != DEFAULT_BURN_ADDRESS;`

src/contracts/core/AllocationManager.sol
```diff
@@ -905,4 +1008,45 @@ contract AllocationManager is
...
+    ) external view returns (uint256) {
+        return _slashCount[operatorSet.key()];
+    }
+
+    /// @inheritdoc IAllocationManager
+    function isOperatorRedistributable(
+        address operator
+    ) external view returns (bool) {
+        EnumerableSet.Bytes32Set storage allocatedSets = allocatedSets[operator];
+        EnumerableSet.Bytes32Set storage registeredSets = registeredSets[operator];
```
wadealexc, line 1040: You are shadowing storage variables here. If you want to use storage pointers, please rename them to avoid shadowing. Prefacing with an underscore is fine.

src/contracts/core/DelegationManager.sol
```diff
@@ -282,42 +280,19 @@ contract DelegationManager is
...
+        uint256 slashId,
+        IStrategy[] calldata strategies,
+        uint64[] calldata prevMaxMagnitudes,
+        uint64[] calldata newMaxMagnitudes
+    ) external onlyAllocationManager nonReentrant returns (uint256[] memory totalDepositSharesToBurn) {
+        totalDepositSharesToBurn = new uint256[](strategies.length);
+        for (uint256 i = 0; i < strategies.length; i++) {
+            totalDepositSharesToBurn[i] = _slashOperatorShares(
+                operator, operatorSet, slashId, strategies[i], prevMaxMagnitudes[i], newMaxMagnitudes[i]
+            );
```
wadealexc, line 293: Please use named parameters here, we have 6 parameters and some shared types.

src/contracts/mixins/Deprecated_OwnableUpgradeable.sol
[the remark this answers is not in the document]
wadealexc: TODO - still needs to inherit from ContextUpgradable, just like the original contract does.

src/contracts/core/AllocationManagerStorage.sol
wadealexc: +1, can we remove this if it's unused?

wadealexc: Future work item -- `libraries/Constants.sol` because we frequently add/remove this value to/from immutable variables. Would be nice to collect things like this in a single place so all contracts can use them.

src/contracts/core/DelegationManager.sol
wadealexc: +1, no need for a diff here.

src/contracts/core/DelegationManager.sol
```diff
@@ -542,19 +525,25 @@ contract DelegationManager is
...
 
         uint256[] memory prevSlashingFactors;
         {
-            // slashableUntil is block inclusive so we need to check if the current block is strictly greater than the slashableUntil block
-            // meaning the withdrawal can be completed.
-            uint32 slashableUntil = withdrawal.startBlock + MIN_WITHDRAWAL_DELAY_BLOCKS;
+            // The slashableUntil block is inclusive, so we verify that the current block number exceeds it
+            // before allowing the withdrawal to be completed.
+            uint32 delayBlocks = withdrawal.delegatedTo == address(this) ? 4 days : MIN_WITHDRAWAL_DELAY_BLOCKS;
+            uint32 slashableUntil = withdrawal.startBlock + delayBlocks;
```
wadealexc, line 546: What's going on with these changes? This is super messy - are we handling moving tokens to escrow in this method?

For the record - if we're overloading `withdrawal.delegatedTo` with `address(this)`, we need to:
1. prevent standard delegation to `address(this)`, because that would apparently allow people a much shorter withdrawal queue
2. validate that no existing stakers have set the DM as their operator

wadealexc: 3. Special handling for the EigenPodManager?

src/contracts/core/StrategyManager.sol
```diff
@@ -157,7 +199,7 @@ contract StrategyManager is
     ) external nonReentrant {
         (, uint256 sharesToBurn) = EnumerableMap.tryGet(burnableShares, address(strategy));
         EnumerableMap.remove(burnableShares, address(strategy));
-        emit BurnableSharesDecreased(strategy, sharesToBurn);
+        emit BurnOrRedistributableSharesDecreased(OperatorSet(address(this), 0), 0, strategy, sharesToBurn);
```
wadealexc, line 202: Can we clarify that `burnShares` (and its underlying `burnableShares` mapping) is basically being deprecated here?

It's probably worth changing the storage name to denote that it's deprecated, and running a script to clear the mapping just after the release. That way we can delete this method in a future release.

src/contracts/core/StrategyManager.sol
```diff
@@ -143,12 +148,49 @@ contract StrategyManager is
...
-        uint256 addedSharesToBurn
+        uint256 sharesToBurn
     ) external onlyDelegationManager nonReentrant {
-        (, uint256 currentShares) = EnumerableMap.tryGet(burnableShares, address(strategy));
-        EnumerableMap.set(burnableShares, address(strategy), currentShares + addedSharesToBurn);
-        emit BurnableSharesIncreased(strategy, addedSharesToBurn);
+        emit BurnOrRedistributableSharesIncreased(operatorSet, slashId, strategy, sharesToBurn);
+
+        EnumerableMap.AddressToUintMap storage operatorSetBurnableShares =
+            _operatorSetBurnableShares[operatorSet.key()][slashId];
```
wadealexc, line 159: I don't think "burnable shares" is really the right naming to use for this concept anymore, right?

Can we change this? It feels too similar to the existing "burn shares" concept.

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -0,0 +1,429 @@
...
+     *                         PAUSABLE ACTIONS
+     *
+     */
+
+    /// @inheritdoc ISlashEscrowFactory
+    function pauseRedistribution(OperatorSet calldata operatorSet, uint256 slashId) external virtual onlyPauser {
+        _checkNewPausedStatus(operatorSet, slashId, true);
+        _paused[operatorSet.key()][slashId] = true;
+        emit RedistributionPaused(operatorSet, slashId);
+    }
```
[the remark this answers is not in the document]
wadealexc: I will say, it might be annoying if an AVS decides to slash one strategy at a time. We'd need to initiate pauses for each slashId... have we considered pausing the entire operatorSet? Or the AVS as a whole? :grimacing:

src/contracts/core/StrategyManager.sol
```diff
@@ -143,12 +148,49 @@ contract StrategyManager is
...
+
+        // Cache length to avoid sloads.
+        uint256 length = operatorSetBurnableShares.length();
+
+        // Iterate over all strategies that have burnable shares.
+        for (uint256 i = 0; i < length; ++i) {
+            (address strategy, uint256 sharesToBurn) = operatorSetBurnableShares.at(i);
+
+            // Remove the strategy from the operator set burnable shares.
+            operatorSetBurnableShares.remove(address(strategy));
```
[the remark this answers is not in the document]
wadealexc: In any case, we're calling untrusted code at the end of this loop...

So my preference would be:
1. collect all the keys/values with one for loop
2. delete the map or mark it as "used" or something
3. call `withdraw` in a second loop

Yes, it's gas intensive, but it's super low risk (and we don't expect this to be called often).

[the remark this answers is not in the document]
wadealexc: One thing we want to check, no matter what implementation path we take here -- what's the gas limit for strategies in an operator set?

The implementation here assumes we can clear the entire array in one go - but AFAICT our ALM makes no restrictions on the number of strategies in an operator set. Is this maybe a dangerous assumption?

src/contracts/core/StrategyManager.sol
```diff
@@ -280,7 +322,7 @@ contract StrategyManager is
         address staker,
         IStrategy strategy,
         uint256 depositSharesToRemove
-    ) internal returns (bool, uint256) {
+    ) internal returns (uint256) {
```
wadealexc, line 325: Update natspec to reflect removed return value.

wadealexc (changes requested):
I will not have time to review `SlashEscrowFactory` and `SlashEscrow` before I leave.

Please see above comments - several changes requested (and tests needed for each). We also have a lot of out-of-date documentation and commenting to address.

src/contracts/mixins/Deprecated_OwnableUpgradeable.sol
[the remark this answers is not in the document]
wadealexc: FWIW the storage slot checker is still throwing errors, so it may well be that `ContextUpgradable` is in fact necessary.

src/contracts/core/DelegationManager.sol
ypatil12: This should have been removed? Was POC in a diff PR

src/contracts/core/SlashEscrowFactory.sol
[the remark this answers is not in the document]
ypatil12: We have a global slash too. That should be sufficient? Pause per slash. Pause globally

src/contracts/core/StrategyManager.sol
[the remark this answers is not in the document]
ypatil12: To solve both of these:

1. Iterate over the enumerableMap in reverse order (or have a while loop on the length of the enumerable map)
2. Have a function to remove an individual index from the enumerable map

src/contracts/mixins/Deprecated_OwnableUpgradeable.sol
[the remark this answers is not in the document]
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1355/commits/fad87238f26606882b58f9d923ca67c0e2e328f8 - fixed now :)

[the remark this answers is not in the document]
ypatil12: ALM is also in the checker

src/contracts/core/AllocationManager.sol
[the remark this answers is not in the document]
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1355/commits/a5132ffec643fec11afbcb5313a2e71187a5d04d

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -0,0 +1,429 @@
...
+        for (uint256 i = 0; i < operatorSetKeys.length; ++i) {
+            operatorSets[i] = operatorSetKeys[i].decode();
+        }
+
+        return operatorSets;
+    }
+
+    /// @inheritdoc ISlashEscrowFactory
+    function getTotalPendingOperatorSets() external view returns (uint256) {
+        return _pendingOperatorSets.length();
```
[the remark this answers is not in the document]
0xClandestine: Renamed downstream

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -0,0 +1,429 @@
...
+
+        strategies = new IStrategy[][](length);
+
+        for (uint256 i = 0; i < length; ++i) {
+            strategies[i] = getPendingStrategiesForSlashId(operatorSet, pendingSlashIds.at(i));
+        }
+    }
+
+    /// @inheritdoc ISlashEscrowFactory
+    function getPendingStrategiesForSlashIds() public view returns (IStrategy[][][] memory strategies) {
```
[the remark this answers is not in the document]
0xClandestine: Agreed, removed downstream. We should consider returning slashIds[] in return value for fn above.

src/contracts/core/SlashEscrowFactory.sol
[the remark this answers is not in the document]
0xClandestine: Discussed with @non-fungible-nelson, given we only want to pause in the event of a widespread protocol bug, this give us enough levers. In the worse case, we would likely just pause globally. This is mostly for super isolated cases...

src/contracts/core/DelegationManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1355/commits/817aad6caaafa311571b2fd2b1d6cdf4e1f616a4

src/contracts/core/DelegationManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1355/commits/817aad6caaafa311571b2fd2b1d6cdf4e1f616a4

src/contracts/core/AllocationManager.sol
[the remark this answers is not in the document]
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1355/commits/817aad6caaafa311571b2fd2b1d6cdf4e1f616a4

src/contracts/core/StrategyManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1355/commits/43749cfde7898f4fb726e67bd2decfef40de214f

src/contracts/core/DelegationManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1355/commits/43749cfde7898f4fb726e67bd2decfef40de214f

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -0,0 +1,482 @@
...
+    /**
+     *
+     *                         OWNER ACTIONS
+     *
+     */
+
+    /// @inheritdoc ISlashEscrowFactory
+    function setGlobalEscrowDelay(
+        uint32 delay
+    ) external onlyOwner {
```
nadir-akhtar, line 181: Nit: this should fit on one line (and surprised `forge fmt` didn't catch it)

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -0,0 +1,482 @@
...
+        _setGlobalEscrowDelay(delay);
```
nadir-akhtar, line 182: Curious why we have a helper function for this, but not for the similar two-line function `setStrategyEscrowDelay` below

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -0,0 +1,482 @@
...
+        // If the redistribution recipient is not the default burn address...
+        if (redistributionRecipient != DEFAULT_BURN_ADDRESS) {
+            require(msg.sender == redistributionRecipient, OnlyRedistributionRecipient());
+        }
+
+        // Assert that the slash ID is not paused
+        require(!isEscrowPaused(operatorSet, slashId), IPausable.CurrentlyPaused());
+
+        // Assert that the escrow delay has elapsed
+        require(block.number >= getEscrowCompleteBlock(operatorSet, slashId), EscrowDelayNotElapsed());
```
nadir-akhtar, line 212: I think, for most of our delays, we wait until _after_ the delay has passed. Here I would think we would do `>` rather than `>=`

nadir-akhtar (commented):
First pass, will revisit the `SlashEscrowFactory` and `SlashEscrow` tomorrow

src/contracts/core/AllocationManager.sol
```diff
@@ -605,6 +662,24 @@ contract AllocationManager is
...
+     * @param operator The operator to check
+     * @param operatorSets The list of operator sets to check
+     * @return True if the operator is redistributable from any of the operator sets, false otherwise
+     */
+    function _isOperatorRedistributable(
+        address operator,
+        OperatorSet[] memory operatorSets
+    ) internal view returns (bool) {
+        for (uint256 i = 0; i < operatorSets.length; ++i) {
+            if (isOperatorSlashable(operator, operatorSets[i]) && isRedistributingOperatorSet(operatorSets[i])) {
```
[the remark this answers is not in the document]
ypatil12: That function handles the stake delay window though right. So if a user is deregistered but still allocated, we would want to know if the stake delay window has elapsed and then return false

src/contracts/core/StrategyManager.sol
```diff
@@ -143,12 +147,21 @@ contract StrategyManager is
...
         IStrategy strategy,
-        uint256 addedSharesToBurn
+        uint256 sharesToBurn
     ) external onlyDelegationManager nonReentrant {
-        (, uint256 currentShares) = EnumerableMap.tryGet(burnableShares, address(strategy));
-        EnumerableMap.set(burnableShares, address(strategy), currentShares + addedSharesToBurn);
-        emit BurnableSharesIncreased(strategy, addedSharesToBurn);
+        emit BurnableSharesIncreased(operatorSet, slashId, strategy, sharesToBurn);
+
+        if (sharesToBurn != 0) {
```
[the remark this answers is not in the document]
ypatil12: We ended up keeping if statement to have same behavior as old `burnShares` function. Moved event to be within block too

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -0,0 +1,475 @@
...
+            strategy: strategy
+        });
+
+        // Update the slash escrow storage.
+        _updateSlashEscrowStorage(operatorSet, slashId);
+    }
+
+    /**
+     *
+     *                         PAUSABLE ACTIONS
```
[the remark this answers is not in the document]
ypatil12: EnumerableSet takes up two storage slots -> can you double check this?

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -0,0 +1,482 @@
...
+     *                         ACTIONS
+     *
+     */
+
+    /// @inheritdoc ISlashEscrowFactory
+    function initiateSlashEscrow(
+        OperatorSet calldata operatorSet,
+        uint256 slashId,
+        IStrategy strategy
+    ) external onlyStrategyManager {
```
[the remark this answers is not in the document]
ypatil12: Same call stack as `slashOperator`

src/contracts/core/AllocationManager.sol
```diff
@@ -290,7 +214,7 @@ contract AllocationManager is
         for (uint256 i = 0; i < params.operatorSetIds.length; i++) {
             // Check the operator set exists and the operator is registered to it
             OperatorSet memory operatorSet = OperatorSet(params.avs, params.operatorSetIds[i]);
-            require(_operatorSets[params.avs].contains(operatorSet.id), InvalidOperatorSet());
+            require(_operatorSets[operatorSet.avs].contains(operatorSet.id), InvalidOperatorSet());
```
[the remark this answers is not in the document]
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1355/commits/3871d2e30c6aefb54a065dfec9b18696bea2aa6a

src/contracts/core/AllocationManager.sol
```diff
@@ -905,4 +980,41 @@ contract AllocationManager is
...
+    }
+
+    /// @inheritdoc IAllocationManager
+    function isOperatorRedistributable(
+        address operator
+    ) external view returns (bool) {
+        // Get the registered and allocated sets for the operator.
+        // We get both sets, since upon deregistration the operator is removed from RegisteredSets, but is still allocated.
+        OperatorSet[] memory registeredSets = getRegisteredSets(operator);
+        OperatorSet[] memory allocatedSets = getAllocatedSets(operator);
```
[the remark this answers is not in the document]
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1355/commits/3871d2e30c6aefb54a065dfec9b18696bea2aa6a

ypatil12 (approved):
LGTM

PR #1358, opened 2025-05-07
feat: add `SlashingWithdrawalRouter`

0xClandestine:
**Motivation:**

We want to hold slashed funds in escrow temporarily to maintain protocol security (prevent malicious actors from draining the contracts).

**Modifications:**

- Add `SlashingWithdrawalRouter`.

**Result:**

The `SlashingWithdrawalRouter` contract is added for managing slashed funds in the EigenLayer protocol. It provides a mechanism to handle the burning or redistribution of slashed shares after a delay period, ensuring proper handling of slashed funds while maintaining protocol security.


bowenli86:
where's the new event that we discussed yesterday? will it be on the next PR?

src/contracts/core/DelegationManager.sol
```diff
@@ -315,9 +317,13 @@ contract DelegationManager is
         // Emit event for operator shares being slashed
         emit OperatorSharesSlashed(operator, strategy, totalDepositSharesToBurn);
 
-        IShareManager shareManager = _getShareManager(strategy);
-        // NOTE: for beaconChainETHStrategy, increased burnable shares currently have no mechanism for burning
-        shareManager.increaseBurnableShares(strategy, totalDepositSharesToBurn);
+        // This conditional is needed since EigenPodManager will not support redistribution initially.
+        if (address(_getShareManager(strategy)) == address(eigenPodManager)) {
+            eigenPodManager.increaseBurnableShares(strategy, totalDepositSharesToBurn);
```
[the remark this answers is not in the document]
0xClandestine: Totally agree, issue is that EPM inherits it. Modifying `IShareManager` would mean we'd need to upgrade EPM.

0xClandestine:
> where's the new event that we discussed yesterday? will it be on the next PR?

Rather than an additional event, I decided to modify `DecreaseBurnableShares` to include `recipient`.

src/contracts/core/AllocationManager.sol
```diff
@@ -948,7 +966,7 @@ contract AllocationManager is
         OperatorSet memory operatorSet,
         address[] memory operators,
         IStrategy[] memory strategies
-    ) public view returns (uint256[][] memory) {
+    ) external view returns (uint256[][] memory) {
```
0xClandestine, line 969: Wasn't used internally.

src/contracts/core/StrategyManager.sol
```diff
@@ -339,6 +419,25 @@ contract StrategyManager is
...
+    function _checkBurnOrRedistributionDelay(
+        OperatorSet calldata operatorSet,
+        IStrategy strategy,
+        uint256 slashId
+    ) internal view {
+        require(
+            block.timestamp >= allocationManager.getBurnOrRedistributionTimestamp(operatorSet, strategy, slashId),
+            BurnOrRedistributionDelayNotElapsed()
+        );
+    }
```
0xClandestine, line 439: Be nice if this could be a single call to ALM.

src/contracts/core/StrategyManagerStorage.sol
```diff
@@ -71,17 +74,22 @@ abstract contract StrategyManagerStorage is IStrategyManager {
...
-    /// @notice Returns the amount of `shares` that have been slashed on EigenLayer but not burned yet. Takes 3 storage slots.
-    EnumerableMap.AddressToUintMap internal burnableShares;
+    /// @dev Returns the amount of `shares` that have been slashed on EigenLayer but not burned yet. Takes 3 storage slots.
+    /// NOTE: Only tracks pre-distribution burns, in the future this mapping will be deprecated.
+    EnumerableMap.AddressToUintMap internal _burnableShares;
+
+    /// @dev Given an operator set and slashId, returns an enumerable mapping of strategies to the amount
+    /// of shares for each strategy that have been slashed but not burned/redistributed yet.
+    mapping(bytes32 operatorSetKey => mapping(uint256 slashId => EnumerableMap.AddressToUintMap)) internal
+        _operatorSetBurnableShares;
```
0xClandestine, line 84: Need to decide if we want to move this, there should be enough room.

0xClandestine: DM getting thick though, only 150 bytes remaining.

src/contracts/core/DelegationManager.sol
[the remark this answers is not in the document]
ypatil12: That's fine

src/contracts/core/AllocationManager.sol
```diff
@@ -239,7 +162,7 @@ contract AllocationManager is
                 _updateAllocationInfo(operator, operatorSet.key(), strategy, info, allocation);
 
                 // 6. Emit an event for the updated allocation
-                emit AllocationUpdated(
+                _emitAllocationUpdated(
```
ypatil12, line 165: Does emitting event in internal actually save that much? Prefer not doing that if possible

src/contracts/core/AllocationManager.sol
```diff
@@ -399,6 +320,105 @@ contract AllocationManager is
      *                         INTERNAL FUNCTIONS
      *
      */
+    function _slashOperator(
+        SlashingParams calldata params,
+        OperatorSet memory operatorSet
+    ) internal returns (uint256 slashId, uint256[] memory shares) {
```
ypatil12, line 326: Needs to return shares & slashID?

src/contracts/core/AllocationManager.sol
```diff
@@ -399,6 +320,105 @@ contract AllocationManager is
...
+        // Cannot realistically overflow, would require 2^256 slashes.
+        unchecked {
+            slashId = ++_slashCount[operatorSet.key()];
+        }
+
+        // Initialize the shares array.
+        shares = new uint256[](params.strategies.length);
+
+        // For each strategy in the operator set, slash any existing allocation
+        for (uint256 i = 0; i < params.strategies.length; i++) {
```
ypatil12, line 340: One thing we could do is have the loop be internal, but the other parts be within the `slashOperator` function

src/contracts/core/AllocationManager.sol
```diff
@@ -399,6 +320,105 @@ contract AllocationManager is
...
+    function _slashOperator(
+        SlashingParams calldata params,
+        OperatorSet memory operatorSet
+    ) internal returns (uint256 slashId, uint256[] memory shares) {
+        require(isOperatorSlashable(params.operator, operatorSet), OperatorNotSlashable());
+
+        uint256[] memory wadSlashed = new uint256[](params.strategies.length);
+
+        // Cannot realistically overflow, would require 2^256 slashes.
+        unchecked {
```
ypatil12, line 332: How much byte code is this saving? We don't really use unchecked anywhere else in the protocol (except in the SM, which IMO we should be consistent and not do it)

src/contracts/core/AllocationManagerStorage.sol
```diff
@@ -14,6 +14,9 @@ abstract contract AllocationManagerStorage is IAllocationManager {
 
     // Constants
 
+    /// @dev The delay before a burn or redistribution can occur.
+    uint32 internal constant BURN_OR_REDISTRIBUTION_DELAY = 3.5 days;
```
0xClandestine, line 18:
```suggestion
    uint32 internal constant BURN_AND_REDISTRIBUTION_DELAY = 3.5 days;
```

src/contracts/core/AllocationManager.sol
```diff
@@ -985,4 +1019,28 @@ contract AllocationManager is
...
+
+    /// @inheritdoc IAllocationManager
+    function isOperatorRedistributable(
+        address operator
+    ) external view returns (bool) {
+        EnumerableSet.Bytes32Set storage sets = allocatedSets[operator];
+        uint256 length = sets.length();
+        for (uint256 i = 0; i < length; ++i) {
+            bytes32 operatorSetKey = sets.at(i);
+            if (_redistributionRecipients[operatorSetKey] != address(0)) {
```
ypatil12, line 1040: Also want to check if the operator is registered. You can be allocated, but not registered

src/contracts/core/DelegationManager.sol
```diff
@@ -315,9 +317,13 @@ contract DelegationManager is
         // Emit event for operator shares being slashed
         emit OperatorSharesSlashed(operator, strategy, totalDepositSharesToBurn);
 
-        IShareManager shareManager = _getShareManager(strategy);
-        // NOTE: for beaconChainETHStrategy, increased burnable shares currently have no mechanism for burning
-        shareManager.increaseBurnableShares(strategy, totalDepositSharesToBurn);
+        // This conditional is needed since EigenPodManager will not support redistribution initially.
```
ypatil12, line 320: Nit: does not support redistribution or burning. Also, it's said the below conditional too?

src/contracts/core/DelegationManager.sol
```diff
@@ -282,6 +282,8 @@ contract DelegationManager is
     /// @inheritdoc IDelegationManager
     function slashOperatorShares(
```
ypatil12, line 283: Return shares?

src/contracts/core/StrategyManager.sol
```diff
@@ -141,28 +145,80 @@ contract StrategyManager is
...
         IStrategy strategy,
         uint256 addedSharesToBurn
     ) external onlyDelegationManager nonReentrant {
-        (, uint256 currentShares) = EnumerableMap.tryGet(burnableShares, address(strategy));
-        EnumerableMap.set(burnableShares, address(strategy), currentShares + addedSharesToBurn);
-        emit BurnableSharesIncreased(strategy, addedSharesToBurn);
+        EnumerableMap.AddressToUintMap storage operatorSetBurnableShares =
+            _operatorSetBurnableShares[operatorSet.key()][slashId];
+        (, uint256 currentShares) = operatorSetBurnableShares.tryGet(address(strategy));
+        operatorSetBurnableShares.set(address(strategy), currentShares + addedSharesToBurn);
```
ypatil12, line 158: If we're burning/redistributing shares on a per operatorSet/slashId basis, why is this an accumulating sum for the previous slash? Should be separate for programmatic redistro (ie. clear attribution)

src/contracts/interfaces/IEigenPodManager.sol
```diff
@@ -130,6 +130,15 @@ interface IEigenPodManager is
...
     ) external;
 
+    /**
+     * @notice Increase the amount of burnable shares for a given Strategy. This is called by the DelegationManager
+     * when an operator is slashed in EigenLayer.
+     * @param strategy The strategy to burn shares in.
+     * @param addedSharesToBurn The amount of added shares to burn.
+     * @dev This function is only called by the DelegationManager when an operator is slashed.
+     */
+    function increaseBurnableShares(IStrategy strategy, uint256 addedSharesToBurn) external;
```
ypatil12, line 140: Nit: can remove since we'll upgrade EPM

src/contracts/interfaces/IShareManager.sol
```diff
@@ -37,12 +37,12 @@ interface IShareManager {
...
 
-    /**
-     * @notice Increase the amount of burnable shares for a given Strategy. This is called by the DelegationManager
-     * when an operator is slashed in EigenLayer.
-     * @param strategy The strategy to burn shares in.
-     * @param addedSharesToBurn The amount of added shares to burn.
-     * @dev This function is only called by the DelegationManager when an operator is slashed.
-     */
-    function increaseBurnableShares(IStrategy strategy, uint256 addedSharesToBurn) external;
+    // /**
```
ypatil12, line 40: Nit: add back

src/contracts/core/StrategyManager.sol
```diff
@@ -141,28 +145,80 @@ contract StrategyManager is
...
+        OperatorSet calldata operatorSet,
+        uint256 slashId,
+        IStrategy strategy
+    ) external nonReentrant {
+        EnumerableMap.AddressToUintMap storage operatorSetBurnableShares =
+            _operatorSetBurnableShares[operatorSet.key()][slashId];
+        (, uint256 sharesToBurn) = operatorSetBurnableShares.tryGet(address(strategy));
+        operatorSetBurnableShares.remove(address(strategy));
+
+        _checkBurnOrRedistributionDelay(operatorSet, strategy, slashId);
```
ypatil12, line 193: doesn't the check only need to be on a per operatorSet/slashID basis. All strategies in the slashID will have the same time at which they can be burned/redistributed

src/contracts/core/StrategyManager.sol
```diff
@@ -141,28 +145,80 @@ contract StrategyManager is
...
+            operatorSet: operatorSet,
+            slashId: slashId,
+            strategy: strategy,
+            recipient: _getRedistributionRecipient(operatorSet),
+            sharesToBurn: sharesToBurn
+        });
+    }
+
+    /// @inheritdoc IStrategyManager
+    function burnOrDistributeShares(OperatorSet calldata operatorSet, uint256 slashId) external nonReentrant {
```
ypatil12, line 205: Can we move this function (and above) to the ALM? So, the ALM will keep track of the withdrawal queue and then it calls into the SM to burn/redistribute? Clean for the AVS to only have to worry about interacting with the ALM. Hopefully in future with larger contract size native eth redistributions can also be completed from ALM

src/contracts/core/StrategyManager.sol
```diff
@@ -141,28 +145,80 @@ contract StrategyManager is
...
+            sharesToBurn: sharesToBurn
+        });
+    }
 
-        // Burning acts like withdrawing, except that the destination is to the burn address.
-        // If we have no shares to burn, we don't need to call the strategy.
-        if (sharesToBurn != 0) {
-            strategy.withdraw(DEFAULT_BURN_ADDRESS, strategy.underlyingToken(), sharesToBurn);
+    /// @inheritdoc IStrategyManager
+    function burnOrDistributeShares(
```
ypatil12, line 183: Logic here can be combined with the below function in a single internal?

src/contracts/core/StrategyManager.sol
```diff
@@ -141,28 +145,80 @@ contract StrategyManager is
...
+        OperatorSet calldata operatorSet,
+        uint256 slashId,
+        IStrategy strategy
+    ) external nonReentrant {
+        EnumerableMap.AddressToUintMap storage operatorSetBurnableShares =
+            _operatorSetBurnableShares[operatorSet.key()][slashId];
+        (, uint256 sharesToBurn) = operatorSetBurnableShares.tryGet(address(strategy));
+        operatorSetBurnableShares.remove(address(strategy));
```
ypatil12, line 191: Check before removing storage?

src/contracts/core/StrategyManager.sol
```diff
@@ -141,28 +145,80 @@ contract StrategyManager is
...
-        EnumerableMap.remove(burnableShares, address(strategy));
-        emit BurnableSharesDecreased(strategy, sharesToBurn);
+        EnumerableMap.AddressToUintMap storage burnableShares = _burnableShares;
+        (, uint256 sharesToBurn) = burnableShares.tryGet(address(strategy));
+        burnableShares.remove(address(strategy));
+
+        // NOTE: We use max values for the operatorSet and slashId to indicate that this is a pre-distribution burn.
+        // QUESTION: Is there a cleaner way to do this?
+        _burnShares({
+            operatorSet: OperatorSet(address(this), type(uint32).max),
```
ypatil12, line 174: We also could just keep the _old_ burn function and upgrade later to remove. So we would temporarily have two diff functions to burn. Don't see as a huge problem since SM isn't close to contract limit

ypatil12 (commented):
Main feedback is moving the burn/redistro to be initiated within the ALM, which then calls the SM. This is because the storage for the delays & slashIDs already lives there

src/contracts/core/AllocationManager.sol
```diff
@@ -985,4 +1019,28 @@ contract AllocationManager is
...
         return _slashCount[operatorSet.key()];
     }
+
+    /// @inheritdoc IAllocationManager
+    function getBurnOrRedistributionTimestamp(
+        OperatorSet memory operatorSet,
+        IStrategy strategy,
+        uint256 slashId
+    ) external view returns (uint32) {
+        return _burnOrRedistributionTimestamp[operatorSet.key()][strategy][slashId];
```
ypatil12, line 1029: Let's use block number, since the withdrawal queue is also in block numbers

src/contracts/core/StrategyManagerStorage.sol
```diff
@@ -33,6 +34,8 @@ abstract contract StrategyManagerStorage is IStrategyManager {
 
     // Immutables
 
+    IAllocationManager public immutable allocationManager;
```
ypatil12, line 37: We can remove the `DEFAULT_BURN_ADDRESS` as an immutable here?

src/contracts/core/AllocationManager.sol
0xClandestine: rip, yeah left a todo

src/contracts/core/AllocationManager.sol
0xClandestine: Will see if that causes `stack-too-deep`.

src/contracts/core/StrategyManager.sol
```diff
@@ -395,22 +494,48 @@ contract StrategyManager is
...
+
+        return (strategies, shares);
+    }
+
+    /// @inheritdoc IStrategyManager
+    function getOperatorSetStrategiesWithBurnableShares(
+        OperatorSet calldata operatorSet,
+        uint256 slashId
+    ) external view returns (address[] memory, uint256[] memory) {
+        EnumerableMap.AddressToUintMap storage operatorSetBurnableShares =
```
ypatil12, line 530: The cronjob to burn/redistro currently relies on knowing which operatorSets/slashIDs to call redistribute on. Is there any way we can get that from a view function?

src/contracts/core/AllocationManager.sol
0xClandestine: `StrategyManager` uses unchecked to increment nonce, see `depositIntoStrategyWithSignature`.

src/contracts/core/StrategyManager.sol
ypatil12: Something like:

1. `getOperatorSetsToBurn`
2. `getSlashIDsToBurn(operatorSet)` 
3. `getStrategiesInSlashID(operatorSet, slashID)`

src/contracts/core/StrategyManager.sol
0xClandestine: Main issue is the event parameters change for `DecreaseBurnableShares`

src/contracts/core/StrategyManager.sol
0xClandestine: Mostly reworked original `burnShares` fn, suppose it wouldn't hurt to move 191 down a few lines.

src/contracts/core/StrategyManager.sol
ypatil12: Ahh okay, this is a fine solution then. Let's just document in the event natspec

src/contracts/core/DelegationManager.sol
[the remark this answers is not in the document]
0xClandestine: b858864

src/contracts/core/AllocationManager.sol
0xClandestine: array initialization needs to be in the internal method... moved the check

57c9a0f

src/contracts/core/AllocationManager.sol
0xClandestine: [f8d92d6](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1358/commits/f8d92d630d51fddf320c4c52bfa7a17309dc9900)

src/contracts/core/AllocationManager.sol
0xClandestine: [8f093c6](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1358/commits/8f093c6c42b475058a6e2104f2c35c5d89163156)

src/contracts/core/AllocationManager.sol
0xClandestine: [554512d](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1358/commits/554512d2da903c130c623113ab124f4c0b256337)

src/contracts/core/StrategyManager.sol
0xClandestine: [9144e4e](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1358/commits/9144e4e27f7b23f822fbf7b7639c79c268fd9532)

src/contracts/core/StrategyManager.sol
0xClandestine: We could combine the two internal methods specifically for `burnOrDistributeShares `, wouldn't provide much benefit though given it's SM.

src/contracts/core/StrategyManagerStorage.sol
0xClandestine: Assuming you mean with the logic move to ALM?

src/contracts/core/StrategyManager.sol
ypatil12: can we modify to enable this?

ypatil12: we can have a mapping from operatorSet => `slashStruct`. The `slashStruct` contains slashID, list of strategies, list of shares, list of bool (completed)

once all burn/redistribution is completed we delete

0xClandestine: We're not currently storing a list of `slashIds` per `operatorSet`.

ypatil12: actually don't even need bools, just remove once cleared

src/contracts/core/AllocationManager.sol
```diff
@@ -985,4 +1017,31 @@ contract AllocationManager is
...
+    function getBurnOrRedistributionBlock(
+        OperatorSet memory operatorSet,
+        IStrategy strategy,
+        uint256 slashId
+    ) external view returns (uint32) {
+        return _burnOrRedistributionBlock[operatorSet.key()][strategy][slashId];
+    }
+
+    /// @inheritdoc IAllocationManager
+    function isOperatorRedistributable(
```
ypatil12, line 1031: Coming back to this one after writing test plan. It might just be easiest to check if the operator is registered to a single redistributable operatorSet. Even though if the operator is not allocated, the intent is to show the staker that there is a redistributable operatorSet that will result in your funds not being burned if the operator is allocated

bowenli86:
> > where's the new event that we discussed yesterday? will it be on the next PR?
> 
> Rather than an additional event, I decided to modify `DecreaseBurnableShares` to include `recipient`.

no `recipient` is added yet?

image

0xClandestine:
> > > where's the new event that we discussed yesterday? will it be on the next PR?
> > 
> > 
> > Rather than an additional event, I decided to modify `DecreaseBurnableShares` to include `recipient`.
> 
> no `recipient` is added yet?
> 
> image

Given we opted to move logic from SM wasn't sure if we still wanted to update SM events.

src/contracts/core/AllocationManager.sol
0xClandestine: After modifier optimizations not sure we still need this

ypatil12:
Need to update the description

PR #1373, opened 2025-05-15
feat: simplify removeDepositShares in StrategyManager

bowenli86:
**Motivation:**

the _removeDepositShares() returns a bool which is not of any meaning. simplify removeDepositShares in StrategyManager by removing that returned value

**Modifications:**

simplify _removeDepositShares() by removing returned extra bool

**Result:**

_removeDepositShares() is simplified and binary size is reduced


bowenli86:
don't need to a new deployment of SM. can come with whichever SM deployment needed in the future

bowenli86:
chatted with greg, merging and will release with redistribution

PR #1374, opened 2025-05-15
release: merge ELIP5 commits in release-dev/token-observability-v2 to moocow

bowenli86:
**Motivation:**

merge ELIP5 commits in release-dev/token-observability-v2 to moocow


src/contracts/token/Eigen.sol
```diff
@@ -193,4 +201,12 @@ contract Eigen is OwnableUpgradeable, ERC20VotesUpgradeable {
...
         return "mode=timestamp";
     }
+
+    /**
+     * @notice Returns the bEIGEN token
+     * @return The bEIGEN token
+     */
+    function beigen() public view returns (IERC20) {
+        return bEIGEN;
+    }
```
wadealexc, line 211: We already have a method `Eigen.bEIGEN`

Why add another one here?

src/test/harnesses/EigenHarness.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity ^0.8.27;
 import "../../contracts/token/Eigen.sol";
 
 contract EigenHarness is Eigen {
-    constructor(IERC20 _bEIGEN) Eigen(_bEIGEN) {}
+    constructor(IERC20 _bEIGEN) Eigen(_bEIGEN, "1.6.0") {}
```
wadealexc, line 7: Hmm, I'm noticing that some of our harnesses wrap the underlying constructor by providing `version` as a constrcutor param, while others take the approach you have here - fill in a version string automatically.

I'm not sure this is a good PR to unify both of those styles (good followup work). However, I do think that we should be consistent with the style you're using here --

`DelegationManagerHarness` supplies `v9.9.9`, which intuitively makes sense to me as a "test version". Let's do that here.

(also there's the issue of whether to prepend with "v" - reviewing this, I noticed we're kinda doing this wrong in our release scripts. See Slack comment here: https://eigenlabs.slack.com/archives/C05PX94L3H9/p1747408010259309)

wadealexc (changes requested):
See 2 comments above

src/test/harnesses/EigenHarness.sol
wadealexc: (following discussion on slack, we should _not_ prepend with "v" as zeus currently does not supply versions with a leading "v")

src/contracts/token/Eigen.sol
bowenli86: right, just recall solidity compiler auto adds getters

src/test/harnesses/EigenHarness.sol
bowenli86: extracted the test version out as constant for harness classes

src/test/TestConstants.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+library TestConstants {
+    string constant TEST_VERSION = "v9.9.9";
```
wadealexc, line 5: See my previous comment - we should _not_ prepend with `v`

wadealexc (approved):
LGTM pending removing leading "v" from the test version constant

PR #1387, opened 2025-05-22
feat: escrow funds in unique clone contracts

0xClandestine:
**Motivation:**

Current implementation is broken by rebase tokens.

**Modifications:**

- Rename SWR -> `SlashEscrowFactory`.
- Add factory logic that deploys clones unique to their slash ID.

**Result:**

Funds will now be stored in clone contracts unique to their slash ID.

### DIFF
--- a/src/contracts/core/SlashEscrow.sol
+++ b/src/contracts/core/SlashEscrow.sol
@@ -10,10 +10,6 @@ contract SlashEscrow is ISlashEscrow {
     using OperatorSetLib for *;
     using SafeERC20 for IERC20;
 
-    error InvalidDeploymentParameters();
-
-    error OnlySlashEscrowFactory();
-
     /// @inheritdoc ISlashEscrow
     function burnOrRedistributeUnderlyingTokens(
         ISlashEscrowFactory slashEscrowFactory,
--- a/src/contracts/core/SlashEscrow.sol
+++ b/src/contracts/core/SlashEscrow.sol
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin-upgrades/contracts/proxy/ClonesUpgradeable.sol";
+import "../interfaces/ISlashEscrow.sol";
+import "../interfaces/ISlashEscrowFactory.sol";
+
+contract SlashEscrow is ISlashEscrow {
+    using OperatorSetLib for *;
+    using SafeERC20 for IERC20;
+
+    error InvalidDeploymentParameters();
+
+    error OnlySlashEscrowFactory();
+
+    /// @inheritdoc ISlashEscrow
+    function burnOrRedistributeUnderlyingTokens(
+        ISlashEscrowFactory slashEscrowFactory,
+        ISlashEscrow slashEscrowImplementation,
+        OperatorSet calldata operatorSet,
+        uint256 slashId,
+        address recipient,
+        IStrategy strategy
+    ) external virtual {
+        // Assert that the deployment parameters are valid by validating against the address of this proxy.
+        require(
+            verifyDeploymentParameters(slashEscrowFactory, slashEscrowImplementation, operatorSet, slashId),
+            InvalidDeploymentParameters()
+        );
+
+        // Assert that the caller is the slash escrow factory.
+        require(msg.sender == address(slashEscrowFactory), OnlySlashEscrowFactory());
+
+        // Burn or redistribute the underlying tokens.
+        IERC20 underlyingToken = strategy.underlyingToken();
+        underlyingToken.safeTransfer(recipient, underlyingToken.balanceOf(address(this)));
+    }
+
+    /// @inheritdoc ISlashEscrow
+    function verifyDeploymentParameters(
+        ISlashEscrowFactory slashEscrowFactory,
+        ISlashEscrow slashEscrowImplementation,
+        OperatorSet calldata operatorSet,
+        uint256 slashId
+    ) public view virtual returns (bool) {
+        return ClonesUpgradeable.predictDeterministicAddress(
+            address(slashEscrowImplementation),
+            keccak256(abi.encodePacked(operatorSet.key(), slashId)),
+            address(slashEscrowFactory)
+        ) == address(this);
+    }
+}
--- a/src/contracts/core/SlashEscrowFactory.sol
+++ b/src/contracts/core/SlashEscrowFactory.sol
@@ -7,15 +7,9 @@ import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
 import "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
 import "../permissions/Pausable.sol";
 import "../mixins/SemVerMixin.sol";
-import "./SlashingWithdrawalRouterStorage.sol";
+import "./SlashEscrowFactoryStorage.sol";
 
-contract SlashingWithdrawalRouter is
-    Initializable,
-    SlashingWithdrawalRouterStorage,
-    OwnableUpgradeable,
-    Pausable,
-    SemVerMixin
-{
+contract SlashEscrowFactory is Initializable, SlashEscrowFactoryStorage, OwnableUpgradeable, Pausable, SemVerMixin {
     using SafeERC20 for IERC20;
     using OperatorSetLib for *;
     using EnumerableSetUpgradeable for *;
@@ -41,7 +35,7 @@ contract SlashingWithdrawalRouter is
         _disableInitializers();
     }
 
-    /// @inheritdoc ISlashingWithdrawalRouter
+    /// @inheritdoc ISlashEscrowFactory
     function initialize(address initialOwner, uint256 initialPausedStatus) external initializer {
         _transferOwnership(initialOwner);
         _setPausedStatus(initialPausedStatus);
@@ -56,7 +50,7 @@ contract SlashingWithdrawalRouter is
      *
      */
 
-    /// @inheritdoc ISlashingWithdrawalRouter
+    /// @inheritdoc ISlashEscrowFactory
     function startBurnOrRedistributeShares(
         OperatorSet calldata operatorSet,
         uint256 slashId,
@@ -92,7 +86,7 @@ contract SlashingWithdrawalRouter is
         emit StartBurnOrRedistribution(operatorSet, slashId, strategy, underlyingAmount, uint32(block.number));
     }
 
-    /// @inheritdoc ISlashingWithdrawalRouter
+    /// @inheritdoc ISlashEscrowFactory
     function burnOrRedistributeShares(
         OperatorSet calldata operatorSet,
         uint256 slashId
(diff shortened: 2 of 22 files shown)


src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -53,12 +50,11 @@ contract SlashingWithdrawalRouter is
      *
      */
 
-    /// @inheritdoc ISlashingWithdrawalRouter
+    /// @inheritdoc ISlashEscrowFactory
     function startBurnOrRedistributeShares(
```
ypatil12, line 54: IMO we should deploy the escrow in this function

ypatil12: oh nvm, I see what you do. You start the countdown immediately

src/contracts/core/SlashEscrow.sol
```diff
@@ -0,0 +1,53 @@
...
+        // Assert that the caller is the slash escrow factory.
+        require(msg.sender == address(slashEscrowFactory), OnlySlashEscrowFactory());
+
+        // Burn or redistribute the underlying tokens.
+        IERC20 underlyingToken = strategy.underlyingToken();
+        underlyingToken.safeTransfer(recipient, underlyingToken.balanceOf(address(this)));
+    }
+
+    /// @inheritdoc ISlashEscrow
+    function verifyDeploymentParameters(
```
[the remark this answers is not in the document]
0xClandestine: The alternative is store `slashEscrowManager` and initialize it when the contract is created.

This method validates the provided args are correct, and is as safe keccak256 is (like any other address).

src/contracts/core/StrategyManager.sol
```diff
@@ -154,13 +155,41 @@ contract StrategyManager is
...
-                strategy.withdraw(address(slashingWithdrawalRouter), strategy.underlyingToken(), sharesToBurn);
+            operatorSetBurnableShares.set(address(strategy), sharesToBurn);
+
+            // Notify the `SlashEscrowFactory` contract that it received underlying tokens to burn or redistribute.
+            slashEscrowFactory.startBurnOrRedistributeShares(operatorSet, slashId, strategy);
+        }
+    }
+
+    /// @inheritdoc IStrategyManager
+    function burnShares(OperatorSet calldata operatorSet, uint256 slashId) external nonReentrant {
```
ypatil12, line 170: Nice, so slashEscrow has the storage to be able to call this

src/contracts/core/StrategyManager.sol
```diff
@@ -154,13 +155,41 @@ contract StrategyManager is
...
+            IStrategy(strategy).withdraw({
+                recipient: address(slashEscrowFactory.getSlashEscrow(operatorSet, slashId)),
+                token: IStrategy(strategy).underlyingToken(),
+                amountShares: sharesToBurn
+            });
 
-            // Notify the `SlashingWithdrawalRouter` contract that it received underlying tokens to burn or redistribute.
-            slashingWithdrawalRouter.startBurnOrRedistributeShares(operatorSet, slashId, strategy, amountOut);
+            // Emit an event to notify the that burnable shares have been decreased.
+            emit BurnableSharesDecreased(operatorSet, slashId, IStrategy(strategy), sharesToBurn);
```
ypatil12, line 192: Nit: can be burnt or redistributed, perhaps we change naming to reflect

src/contracts/core/SlashEscrowFactory.sol
0xClandestine: This reminds me, we must be careful to not allow a case where storage can be cleared before tokens have been withdrawn. This can happen if `burnOrRedistirbuteShares` is called before the cron job has called `burnShares`.

src/contracts/core/StrategyManager.sol
```diff
@@ -154,13 +155,41 @@ contract StrategyManager is
...
+    /// @inheritdoc IStrategyManager
+    function burnShares(OperatorSet calldata operatorSet, uint256 slashId) external nonReentrant {
+        EnumerableMap.AddressToUintMap storage operatorSetBurnableShares =
+            _operatorSetBurnableShares[operatorSet.key()][slashId];
+
+        // Cache length to avoid sloads.
+        uint256 length = operatorSetBurnableShares.length();
+
+        // Iterate over all strategies that have burnable shares.
+        for (uint256 i = 0; i < length; ++i) {
```
ypatil12, line 178: Note: we do not provide a protocol guarantee around independence of slashes. That is, if a token blacklists the transfer to an address, then all strategies in that given slashID will be unable to be redistributed. This is similar to our behavior around withdrawals currently. Should we get clarification from product on this? I am fine, but we should document this

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -69,27 +65,27 @@ contract SlashingWithdrawalRouter is
...
-        EnumerableMapUpgradeable.AddressToUintMap storage pendingBurnOrRedistributions =
-            _pendingBurnOrRedistributions[operatorSet.key()][slashId];
+        // Create a storage pointer to `_pendingStrategiesForSlashId`.
+        EnumerableSetUpgradeable.AddressSet storage pendingStrategiesForSlashId =
+            _pendingStrategiesForSlashId[operatorSet.key()][slashId];
 
         // Add the slash ID to the pending slash IDs set.
         pendingSlashIds.add(slashId);
 
         // Add the operator set to the pending operator sets set.
```
ypatil12, line 75: So the guarantee we're providing here is that, if you have separate delays, they can be completed earliest to latest

src/contracts/core/StrategyManager.sol
0xClandestine: Agree though, need clarification.

src/contracts/core/SlashEscrowFactory.sol
ypatil12: Delays do not block completion. Blacklists in untrusted code do.

src/contracts/core/SlashEscrowFactory.sol
0xClandestine: [5471ac5](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1387/commits/5471ac52f9e6d9b1f5be6500d0483f8cc016f40c)

src/contracts/core/StrategyManager.sol
0xClandestine: [d722d8c](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1387/commits/d722d8c10d9b3d13e4dcb745dc41001eff4f3efe)

src/contracts/core/SlashEscrow.sol
```diff
@@ -0,0 +1,49 @@
...
+        OperatorSet calldata operatorSet,
+        uint256 slashId
+    ) public view virtual returns (bool) {
+        return ClonesUpgradeable.predictDeterministicAddress(
+            address(slashEscrowImplementation),
+            keccak256(abi.encodePacked(operatorSet.key(), slashId)),
+            address(slashEscrowFactory)
+        ) == address(this);
+    }
+}
```
0xClandestine, line 49: @wadealexc Is this safe? See `ISlashEscrow` for more details.

0xClandestine: Basically instead of storing slashEscrowFactory which requires a slot, and an initializer we simply validate against the deterministic address of the contract.

wadealexc: discussed in slack, yes, as long as slash ids are unique per operator set (which it appears they are)

src/contracts/core/SlashEscrow.sol
```diff
@@ -0,0 +1,49 @@
...
+        ISlashEscrowFactory slashEscrowFactory,
+        ISlashEscrow slashEscrowImplementation,
+        OperatorSet calldata operatorSet,
+        uint256 slashId,
+        address recipient,
+        IStrategy strategy
+    ) external virtual {
+        // Assert that the deployment parameters are valid by validating against the address of this proxy.
+        require(
+            verifyDeploymentParameters(slashEscrowFactory, slashEscrowImplementation, operatorSet, slashId),
```
ypatil12, line 24: What is this verification actually doing? Is it needed

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -2,49 +2,46 @@
...
-    /// @inheritdoc ISlashingWithdrawalRouter
+    /// @inheritdoc ISlashEscrowFactory
     function initialize(address initialOwner, uint256 initialPausedStatus) external initializer {
         _transferOwnership(initialOwner);
         _setPausedStatus(initialPausedStatus);
 
-        // Set the global burn or redistribution delay to 3.5 days in blocks assuming 12 second blocks.
-        _globalBurnOrRedistributionDelayBlocks = 3.5 days / 12 seconds;
+        // Set the global burn or redistribution delay to 4 days in blocks assuming 12 second blocks.
+        _globalBurnOrRedistributionDelayBlocks = 4 days / 12 seconds;
```
ypatil12, line 44: We can't calculate this in the initializer since the delay will be different on testnet

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -106,22 +102,25 @@ contract SlashingWithdrawalRouter is
         // Assert that the escrow is not paused.
         require(!_paused[operatorSet.key()][slashId], IPausable.CurrentlyPaused());
 
+        // Assert that funds have been sent to the `SlashEscrow` before clearing storage.
+        strategyManager.decreaseBurnableShares(operatorSet, slashId);
```
ypatil12, line 106: We should note this will not transfer if funds have already been sent. Note that we don't assert, but comment says we do

src/contracts/core/SlashEscrow.sol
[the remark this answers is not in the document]
ypatil12: Any issue with doing this? Makes the calldata much cleaner. We can set in constructor upon deployment

src/contracts/core/SlashEscrow.sol
0xClandestine: Yeah would need to initialize and OZ doesn't provide clones with immutable args.

src/contracts/core/SlashEscrowFactory.sol
0xClandestine: oh fair, will pass in another param

src/contracts/core/SlashEscrowFactory.sol
0xClandestine: Good point

src/contracts/core/SlashEscrow.sol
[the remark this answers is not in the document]
0xClandestine: confirmed with sec

src/contracts/core/SlashEscrowFactory.sol
0xClandestine: [dd95040](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1387/commits/dd95040931a6492c2ac6cf4ef5720000f61073ca)

src/contracts/core/SlashEscrowFactory.sol
0xClandestine: [6d6b7b8](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1387/commits/6d6b7b88974646219a242d19a77b823c8c05f5b1)

PR #1390, opened 2025-05-22
ci: use foundry v1.2.1 instead of latest stable version to speed up CI

bowenli86:
**Motivation:**

**Modifications:**

**Result:**


bowenli86:
@0xClandestine which version of foundry did you use that produced significant shorter build time? v1.2.1 doesn't seem to be better than v1.1 according the ci numbers

PR #1399, opened 2025-05-24
fix: enumerable map overwrite

ypatil12:
**Motivation:**

Currently, there's a bug in the `SM` where if you loop through the burnable shares queue, you may not clear all due to swap and pop of an Enumerable Map. Furthermore, we also are constrained by a token transfer taking too much gas and blocking transfer out of funds. 

**Modifications:**

- Iterate backwards on `decreaseBurnOrRedistributableShares ` 
- Overloaded `decreaseBurnableShares` with a version to pass in an index. This function will escrow a single share (called by above too). Now, we do not need a max strategy per opSet requirement
- Unit tests for both `increaseBurnOrRedistributableShares` and `decreaseBurnOrRedistributableShares`
- Added the following introspection:
-- `getBurnOrRedistributableShares(operatorSet, slashId) returns (Strategy[] Strats, uint256[] shares)
-- `getBurnOrRedistributableShares(operatorSet, slashId, strategy) returns (shares)
-- `getBurnOrRedistributableCount(operaotrSet, slashed) returns (count)`

**Result:**

Correct code with unit tests

### DIFF
--- a/src/contracts/core/StrategyManager.sol
+++ b/src/contracts/core/StrategyManager.sol
@@ -179,9 +179,13 @@ contract StrategyManager is
             decreaseBurnableShares(operatorSet, slashId, i - 1);
         }
     }
-    
+
     /// @inheritdoc IStrategyManager
-    function decreaseBurnableShares(OperatorSet calldata operatorSet, uint256 slashId, uint256 index) public returns (uint256) {
+    function decreaseBurnableShares(
+        OperatorSet calldata operatorSet,
+        uint256 slashId,
+        uint256 index
+    ) public returns (uint256) {
         EnumerableMap.AddressToUintMap storage operatorSetBurnableShares =
             _operatorSetBurnableShares[operatorSet.key()][slashId];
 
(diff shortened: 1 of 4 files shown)


bowenli86 (commented):
can you add tests to cover the case?

ypatil12:
> can you add tests to cover the case?

Yes, this is pending, not merging until we have tests

src/contracts/core/StrategyManager.sol
```diff
@@ -156,44 +156,57 @@ contract StrategyManager is
         EnumerableMap.AddressToUintMap storage operatorSetBurnableShares =
             _burnOrRedistributableShares[operatorSet.key()][slashId];
 
-        if (sharesToBurn != 0) {
```
0xClandestine, line 159: Any reason this condition was removed? Pretty sure it's for event sanitization..

0xClandestine (approved):
one comment

src/contracts/core/StrategyManager.sol
0xClandestine: merging for now, will note and fix if needed

PR #1401, opened 2025-05-24
chore: use internal getters; update `isOperatorRedistributable`

ypatil12:
**Motivation:**

We want to use internal getters wherever possible for style. 

**Modifications:**

- Use `getRedistributionRecipient` in `isOperatorRedistributable` 
- Update `isOperatorRedistributable` to get all allocated/registered sets and then check if slashable and redistributable set for each
- More comprehensive unit tests

**Result:**

Cleaner code + tests passing

### DIFF
--- a/src/test/integration/IntegrationDeployer.t.sol
+++ b/src/test/integration/IntegrationDeployer.t.sol
@@ -282,7 +282,7 @@ abstract contract IntegrationDeployer is ExistingDeploymentParser {
 
         // First, deploy the new contracts as empty contracts
         emptyContract = new EmptyContract();
-        
+
         slashEscrowFactory =
             SlashEscrowFactory(address(new TransparentUpgradeableProxy(address(emptyContract), address(eigenLayerProxyAdmin), "")));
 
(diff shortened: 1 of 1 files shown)


src/contracts/core/AllocationManager.sol
```diff
@@ -1022,7 +1040,7 @@ contract AllocationManager is
     function isRedistributingOperatorSet(
         OperatorSet memory operatorSet
     ) public view returns (bool) {
-        return _redistributionRecipients[operatorSet.key()] != address(0);
+        return getRedistributionRecipient(operatorSet) != DEFAULT_BURN_ADDRESS;
```
0xClandestine, line 1043: Pretty sure you end up doing the same comparison twice where `isRedistributingOperatorSet` is used.

0xClandestine (approved):
One comment, looks good.

src/contracts/core/AllocationManager.sol
ypatil12: ya we do, but cleaner to just use internals per Alex review

PR #1406, opened 2025-05-25
feat: simplify escrow delay; add convenience functions

ypatil12:
**Motivation:**

The escrow delay currently always you to complete escrows for a portion of strategies if there exists a strategy with a larger delay. This makes our codebase more complex. We also want to have a view function for offchain cronjob that needs to be called only once. 

**Modifications:**

- Update `releaseEscrow` to obey the maximum delay across all strategies for a slash
- Add a `getBurnOrRedistributionDelay` view function
- Add a convenience view function `getPendingEscrows` for offchain burn job. This returns all pending operatorSets, and their associated redistribution status, slashIds, and completeBlocks 
- Standardize `uint32` for delay everywhere 
- Make `deploySlashEscrow` a public function

**Result:**

Simpler & correct code.


0xClandestine (approved):
lgtm

PR #1421, opened 2025-05-29
feat: KeyRegistrar

eigenmikem:
**Motivation:**

Operator key management should be brought into core and out of middleware.

**Modifications:**

New core contract, `KeyRegistrar`, that manages operator keys and supports ECDSA and BN254 signatures.

**Result:**

Simpler for both AVSs and operators.

### DIFF
--- a/src/contracts/permissions/KeyRegistrar.sol
+++ b/src/contracts/permissions/KeyRegistrar.sol
@@ -193,7 +193,7 @@ contract KeyRegistrar is KeyRegistrarStorage, PermissionControllerMixin, Signatu
         );
         bytes32 signableDigest = _calculateSignableDigest(structHash);
 
-        // Use hash-to-G1 approach like the middleware
+        // hash to g1
         BN254.G1Point memory messagePoint = BN254.hashToG1(signableDigest);
         _verifyBN254Signature(messagePoint, signature, g1Point, g2Point);
 


src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+     * @param rotationDelay Delay in blocks before key rotation takes effect
+     * @dev Only authorized callers for the AVS can configure operator sets
+     */
+    function configureOperatorSet(
+        address avs,
+        uint32 operatorSetId,
+        CurveType curveType,
+        uint256 rotationDelay
+    ) external checkCanCall(avs) {
+        if (avs == address(0)) revert ZeroAddress();
```
ypatil12, line 114: Don't think this is needed since we do `checkCanCall` ?

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+        operatorSetConfigs[avs][operatorSetId] = OperatorSetConfig({
+            curveType: curveType,
+            isActive: true,
+            rotationDelay: rotationDelay
+        });
+
+        emit OperatorSetConfigured(avs, operatorSetId, curveType, rotationDelay);
+    }
+
+    /**
```
ypatil12, line 125: For styling, let's move all these natspec to an interface and use `///@inheritdoc`

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+     * @dev Validates key format and ensures global uniqueness
+     */
+    function _registerECDSAKey(
+        address avs,
+        uint32 operatorSetId,
+        address operator,
+        bytes calldata pubkey
+    ) internal {
+        // Validate ECDSA public key format
+        if (pubkey.length != 65) revert InvalidKeyFormat();
```
ypatil12, line 181: nit: `require(x, revert());`

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+     * @param pubkey The ECDSA public key bytes
+     * @dev Validates key format and ensures global uniqueness
+     */
+    function _registerECDSAKey(
+        address avs,
+        uint32 operatorSetId,
+        address operator,
+        bytes calldata pubkey
+    ) internal {
+        // Validate ECDSA public key format
```
ypatil12, line 180: do we need to validate against the keyType set from config?

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+        // Note: Aggregate key is NOT updated here - use updateAPK callback instead
+    }
+
+    /**
+     * @notice Deregisters a cryptographic key for an operator with a specific AVS
+     * @param operator Address of the operator to deregister key for
+     * @param avs Address of the AVS
+     * @param operatorSetId ID of the operator set
+     * @dev Can be called by operator directly or by addresses they've authorized via PermissionController
+     * @dev Succeeds silently if key was not registered
```
ypatil12, line 287: Why not revert?

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+        if (!config.isActive) revert OperatorSetNotConfigured();
+        
+        KeyInfo memory keyInfo = operatorKeyInfo[avs][operatorSetId][operator];
+        
+        if (!keyInfo.isRegistered) {
+            return; // Silently succeed if key was not registered
+        }
+
+        // Note: We don't remove from globalKeyRegistry on deregistration
+        // This ensures keys can never be reused, even after deregistration
```
ypatil12, line 306: Add this to dev comment too

ypatil12: Is this needed?

ypatil12 (commented):
First pass

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+     */
+    function configureOperatorSet(
+        address avs,
+        uint32 operatorSetId,
+        CurveType curveType,
+        uint256 rotationDelay
+    ) external checkCanCall(avs) {
+        if (avs == address(0)) revert ZeroAddress();
+        
+        operatorSetConfigs[avs][operatorSetId] = OperatorSetConfig({
```
ypatil12, line 116: Shouldn't we prevent key type from being updated if already set?

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+        // Extract the G1 point from the stored key data
+        (uint256 g1X, uint256 g1Y, , ) = abi.decode(keyInfo.keyData, (uint256, uint256, uint256[2], uint256[2]));
+        BN254.G1Point memory operatorKey = BN254.G1Point(g1X, g1Y);
+
+        if (isRotation) {
+            // For rotation, we need to handle both old and new keys
+            // Since the key data is already updated to the new key, we need to get the old key differently
+            // This is a limitation of the current approach - we could store rotation history if needed
+            // For now, we'll recalculate the entire APK by iterating through all registered operators
+            _recalculateAggregateBN254Key(avs, operatorSetId);
```
ypatil12, line 521: Confused on the flow here, I think it's causing complexity because the AVS is initiating a rotation?

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+     * @param isRotation True if this is a key rotation, false if it's a new registration
+     * @dev This function should be called by the AVSRegistrar when operators register/rotate keys
+     * @dev Only authorized callers for the AVS can update the APK
+     */
+    function updateAPK(
+        address avs,
+        uint32 operatorSetId,
+        address operator,
+        bool isRotation
+    ) external checkCanCall(avs) {
```
ypatil12, line 498: I'm thinking we should _not_ use UAM here and just gate the call on if the msg.sender is the avs registrar (stored in the ALM). Reason is that we explicitly state the avs registrar should gate registration/deregistration and how we are saying any component of the AVS can update the key

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+    /**
+     * @notice Deregisters a cryptographic key for an operator with a specific AVS
+     * @param operator Address of the operator to deregister key for
+     * @param avs Address of the AVS
+     * @param operatorSetId ID of the operator set
+     * @dev Can be called by operator directly or by addresses they've authorized via PermissionController
+     * @dev Succeeds silently if key was not registered
+     * @dev Does NOT update aggregate key - use updateAPK callback for that
+     */
+    function deregisterKey(
```
ypatil12, line 290: Shouldn't the AVS have the ability to deregister a key too? Assume that the key has been compromised. Operator can deregister. AVS can deregister AND update the apk (if relevant)

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+        }
+
+        // Register key based on curve type
+        if (config.curveType == CurveType.ECDSA) {
+            _registerECDSAKey(avs, operatorSetId, operator, pubkey);
+        } 
+        else if (config.curveType == CurveType.BN254) {
+            _registerBN254Key(avs, operatorSetId, operator, pubkey, signature);
+        } 
+        else {
```
8sunyuan, line 159: else block unnecessary since only 2 enum types

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+                isZero = false;
+                break;
+            }
+        }
+        if (isZero) revert ZeroPubkey();
+
+        bytes32 keyHash = keccak256(pubkey);
+        
+        // Check global uniqueness - reject if key has ever been used
+        if (globalKeyRegistry[keyHash]) {
```
8sunyuan, line 197: So we are enforcing uniqueness of keys on a per operatorSet basis? Different keys between different AVSs sounds reasonable but not sure about different opSets for the same AVS.

src/contracts/core/KeyRegistrar.sol
8sunyuan: Agreed on this. In which case, this should just be renamed as a one-time callable `initializeOperatorSet` function  and then have a separate `setRotationDelay` setter.

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+        operatorKeyInfo[avs][operatorSetId][operator] = KeyInfo({
+            isRegistered: true,
+            lastRotationBlock: block.number,
+            keyData: pubkey
+        });
+        
+        // Update global key registry
+        globalKeyRegistry[keyHash] = true;
+        
+        // Note: Aggregate key is NOT updated here - use updateAPK callback instead
```
8sunyuan, line 278: Theres no situation where we wouldn't want to update the APK on each register/deregister call. So why would we leave this up to the AVS to implement the callback?

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,781 @@
...
+        bool isZero = true;
+        for (uint i = 1; i < 65; i++) {
+            if (pubkey[i] != 0) {
+                isZero = false;
+                break;
+            }
+        }
+        if (isZero) revert ZeroPubkey();
+
+        bytes32 keyHash = keccak256(pubkey);
```
8sunyuan, line 194: Whereever we hash the pubkeys, can we use the same hashing function for consistency? Use the `getKeyHash` view method below for example

ypatil12 (commented):
Super clean, minor comments

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,439 @@
...
+     */
+
+    /**
+     * @notice Validates and registers an ECDSA public key
+     * @param operatorSet The operator set to register the key for
+     * @param operator Address of the operator
+     * @param pubkey The ECDSA public key bytes
+     * @dev Validates key format and ensures global uniqueness
+     */
+    function _registerECDSAKey(
```
solimander, line 170: should verify a sig over a message that includes the operator's address, no? otherwise, anyone can frontrun to burn keys and cause ECDSA registrations to fail

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,439 @@
...
+        
+        OperatorSetConfig memory config = operatorSetConfigs[operatorSet.key()];
+        require(config.isActive, OperatorSetNotConfigured());
+
+        // Check if the key is already registered
+        require(!operatorKeyInfo[operatorSet.key()][operator].isRegistered, KeyAlreadyRegistered());
+
+        // Register key based on curve type
+        if (config.curveType == CurveType.ECDSA) {
+            _registerECDSAKey(operatorSet, operator, pubkey);
```
gpsanant, line 90: this should also require a signature on msg.sender to avoid frontrunning

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,439 @@
...
+    ) external checkCanCall(operatorSet.avs) {
+        
+        OperatorSetConfig memory config = operatorSetConfigs[operatorSet.key()];
+        require(config.isActive, OperatorSetNotConfigured());
+        
+        KeyInfo memory keyInfo = operatorKeyInfo[operatorSet.key()][operator];
+        
+        if (!keyInfo.isRegistered) {
+            // Revert if key was not registered
+            revert KeyNotFound(operatorSet, operator);
```
gpsanant, line 115: require

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,439 @@
...
+
+        // Validate G1 point
+        g1Point = BN254.G1Point(g1X, g1Y);
+        require(!(g1X == 0 && g1Y == 0), ZeroPubkey());
+
+        // Construct BN254 G2 point from coordinates
+        BN254.G2Point memory g2Point = BN254.G2Point(g2X, g2Y);
+
+        // Verify that G1 and G2 form a valid keypair
+        require(BN254.pairing(g1Point, BN254.negGeneratorG2(), BN254.generatorG1(), g2Point), InvalidKeypair());
```
gpsanant, line 227: you're already checking this in the signature check below

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,439 @@
...
+
+    /// @inheritdoc IKeyRegistrar
+    function getOperatorSetConfig(
+        OperatorSet memory operatorSet
+    ) external view returns (OperatorSetConfig memory) {
+        return operatorSetConfigs[operatorSet.key()];
+    }
+
+    /// @inheritdoc IKeyRegistrar
+    function getBN254Key(
```
solimander, line 376: validate operator set `curveType`. abi.decode will revert with unclear error if operator set uses ECDSA

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,439 @@
...
+    /// @inheritdoc IKeyRegistrar
+    function checkAndUpdateKey(
+        OperatorSet memory operatorSet,
+        address operator
+    ) external checkCanCall(operatorSet.avs) {
+        OperatorSetConfig memory config = operatorSetConfigs[operatorSet.key()];
+        require(config.isActive, OperatorSetNotConfigured());
+        
+        KeyInfo memory keyInfo = operatorKeyInfo[operatorSet.key()][operator];
+        if (!keyInfo.isRegistered) {
```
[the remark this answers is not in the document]
ypatil12: Interface returns bool

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,398 @@
...
+
+            // Validate G1 point
+            g1Point = BN254.G1Point(g1X, g1Y);
+            require(!(g1X == 0 && g1Y == 0), ZeroPubkey());
+
+            // Construct BN254 G2 point from coordinates
+            BN254.G2Point memory g2Point = BN254.G2Point(g2X, g2Y);
+
+            // Verify the signature to prevent rogue key attacks with domain separation
+            bytes32 messageHash = keccak256(
```
0xrajath: Should this be a EIP712 message hash like how the current middleware does it? Can just use the OZ lib directly instead of rolling our own.

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,398 @@
...
+                abi.encodePacked(
+                    "EigenLayer.KeyRegistrar.v1", // Domain separator
+                    address(this), // Contract address for additional separation
+                    operatorSet.avs,
+                    operatorSet.id,
+                    operator,
+                    keyData
+                )
```
0xrajath: Should there be a chainID here as well?

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,398 @@
...
+    function initialize(
+        address initialOwner
+    ) external initializer {
+        require(initialOwner != address(0), ZeroAddress());
+        _transferOwnership(initialOwner);
+    }
+
+    /// @inheritdoc IKeyRegistrar
+    function configureOperatorSet(OperatorSet memory operatorSet, CurveType curveType) external {
+        require(address(allocationManager.getAVSRegistrar(operatorSet.avs)) == msg.sender, "Unauthorized");
```
ypatil12: This can be UAM settable. The old key rotation had to be done by the `AVSRegistrar` but we're good now

ypatil12: `checkCanCall(operatorSet.avs)`

src/contracts/interfaces/IKeyRegistrar.sol
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /// @dev Structure to store key information
+    struct KeyInfo {
+        bool isRegistered;
+        bytes keyData; // Flexible storage for different curve types
+    }
+
+    /// @dev Configuration for each operator set
+    struct OperatorSetConfig {
```
ypatil12: We use `OperatorSetConfig` in multichain. Do we even need this structure at all? We know it's active if a curve type his set in the operatorKeyInfo. Let's remove?

src/test/unit/KeyRegistrarUnit.t.sol
```diff
@@ -0,0 +1,835 @@
...
+import "forge-std/Test.sol";
+import "src/contracts/permissions/KeyRegistrar.sol";
+import "src/contracts/interfaces/IKeyRegistrar.sol";
+import "src/contracts/libraries/BN254.sol";
+import "src/contracts/interfaces/IPermissionController.sol";
+import "src/contracts/mixins/PermissionControllerMixin.sol";
+import "src/contracts/libraries/OperatorSetLib.sol";
+import "src/test/mocks/AllocationManagerMock.sol";
+
+contract MockPermissionController is IPermissionController {
```
ypatil12: Our setup already deploys this. No need for a mock here

src/test/unit/KeyRegistrarUnit.t.sol
```diff
@@ -0,0 +1,835 @@
...
+    function getAppointees(address, address, bytes4) external pure override returns (address[] memory) {
+        return new address[](0);
+    }
+
+    function version() external pure override returns (string memory) {
+        return "1.0.0";
+    }
+}
+
+contract KeyRegistrarUnitTests is Test {
```
ypatil12: Let's use `EigenLayerUnitTestSetup` ?

src/contracts/core/KeyRegistrar.sol
ypatil12: Agreed. Let's do what the `RegistryCoordinator` does here and inherit from `EIP712Upgradeable`

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,398 @@
...
+        // Decode signature
+        BN254.G1Point memory sigPoint;
+        {
+            (uint256 sigX, uint256 sigY) = abi.decode(signature, (uint256, uint256));
+            sigPoint = BN254.G1Point(sigX, sigY);
+        }
+
+        // Generate challenge value using Fiat-Shamir transform to prevent rogue key attacks
+        uint256 gamma = uint256(
+            keccak256(
```
ypatil12, line 267: Why not use the same format as what we do in middleware repo? Hash to message to G1: 

https://github.com/Layr-Labs/eigenlayer-middleware/blob/3fb5b61076475108bd87d4e6c7352fd60b46af1c/src/SlashingRegistryCoordinator.sol#L1101-L1109

https://github.com/Layr-Labs/eigenlayer-middleware/blob/3fb5b61076475108bd87d4e6c7352fd60b46af1c/src/BLSApkRegistry.sol#L91-L106

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,408 @@
...
+
+        // Create EIP-712 compliant message hash
+        bytes32 structHash = keccak256(
+            abi.encode(
+                BN254_KEY_REGISTRATION_TYPEHASH,
+                operator,
+                operatorSet.avs,
+                operatorSet.id,
+                keccak256(keyData),
+                block.chainid
```
ypatil12: Nit: doesn't need block.chainid since `eip712Domain` has the block.chainid

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,408 @@
...
+            _registerBN254Key(operatorSet, operator, keyData, signature);
+        } else {
+            revert InvalidCurveType();
+        }
+
+        emit KeyRegistered(operatorSet, operator, curveType, keyData);
+    }
+
+    /// @inheritdoc IKeyRegistrar
+    function deregisterKey(address operator, OperatorSet memory operatorSet) external {
```
gpsanant, line 90: should we only allow rotation if they're still registered?

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,408 @@
...
+            // Validate G1 point
+            g1Point = BN254.G1Point(g1X, g1Y);
+            require(!(g1X == 0 && g1Y == 0), ZeroPubkey());
+
+            // Construct BN254 G2 point from coordinates
+            g2Point = BN254.G2Point(g2X, g2Y);
+        }
+
+        // Create EIP-712 compliant message hash
+        bytes32 structHash = keccak256(
```
gpsanant, line 191: need view functions for these for offchain integrations to work

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,408 @@
...
+            )
+        );
+        bytes32 messageHash = _hashTypedDataV4(structHash);
+
+        // Verify that the signature was created by the key address
+        address recoveredSigner = ECDSA.recover(messageHash, signature);
+        require(recoveredSigner == keyAddress, InvalidSignature());
+
+        // Store key data using internal helper
+        _storeKeyData(operatorSet, operator, keyData, keyHash);
```
gpsanant, line 157: nit: can move this to `registerKey` for both curve types

src/contracts/permissions/KeyRegistrar.sol
0xrajath: +1 Need something similar to `pubkeyRegistrationMessageHash` in `SlashingRegistryCoordinator`.

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,397 @@
...
+
+        // Clear key info
+        delete operatorKeyInfo[operatorSet.key()][operator];
+
+        emit KeyDeregistered(operatorSet, operator, curveType);
+    }
+
+    /// @inheritdoc IKeyRegistrar
+    function checkKey(OperatorSet memory operatorSet, address operator) external view returns (bool) {
+        require(address(allocationManager.getAVSRegistrar(operatorSet.avs)) == msg.sender, InvalidPermissions());
```
ypatil12: View function, doesn't need to have checks on who can call (holdover from apk updates)

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,397 @@
...
+        _disableInitializers();
+    }
+
+    /// @inheritdoc IKeyRegistrar
+    function initialize(
+        address initialOwner
+    ) external initializer {
+        require(initialOwner != address(0), ZeroAddress());
+        __EIP712_init("EigenLayer.KeyRegistrar", "v1");
+        _transferOwnership(initialOwner);
```
ypatil12: No owner, can remove the Ownable inheritance. Also, realized we can just use the `SignatureUtilsMixin` instead of the EIP712. See how we use in DM (apologies for the oversight here)

ypatil12: Signatures can also be checked via `_checkIsValidSignatureNow`. This also allows for smart contracts to be ECDSA operators, which is our current recommendation

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    /**
+     * @notice Returns the message hash for BN254 key registration
+     * @param operator The operator address
+     * @param operatorSet The operator set
+     * @param keyData The BN254 key data
+     * @return The message hash for signing
+     */
+    function getBN254KeyRegistrationMessageHash(
```
ypatil12, line 395: Let's add this to the interface too

ypatil12: (same for ecdsa)

ypatil12 (commented):
Last comment. LGTM otherwise

src/contracts/core/KeyRegistrar.sol
```diff
@@ -0,0 +1,439 @@
...
+
+    /**
+     * @notice Verifies a BN254 signature using Fiat-Shamir challenge to prevent rogue key attacks
+     * @param messageHash Hash of the message being signed
+     * @param signature The signature bytes
+     * @param pubkeyG1 The G1 component of the public key
+     * @param pubkeyG2 The G2 component of the public key
+     * @dev Uses gamma challenge value derived from message and public key components
+     */
+    function _verifyBN254Signature(
```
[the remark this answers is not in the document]
ypatil12: Fine to leave it out for now in interest of time. Can come back to this next week

src/contracts/permissions/KeyRegistrar.sol
eigenmikem: Just waiting on rajath giving the ok on the views

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,405 @@
...
+        // Check global uniqueness
+        require(!globalKeyRegistry[keyHash], KeyAlreadyRegistered());
+
+        // Create EIP-712 compliant message hash
+        bytes32 structHash = keccak256(
+            abi.encode(ECDSA_KEY_REGISTRATION_TYPEHASH, operator, operatorSet.avs, operatorSet.id, keyAddress)
+        );
+        bytes32 signableDigest = _calculateSignableDigest(structHash);
+
+        _checkIsValidSignatureNow(keyAddress, signableDigest, signature, type(uint256).max);
```
ypatil12, line 154: We should add an expiry in a future PR

ypatil12: (or now if possible if we're just waiting)

src/contracts/permissions/KeyRegistrar.sol
0xrajath: What view functions would be needed to generate the `signature` field? that needs to be passed in while registering a key?

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,405 @@
...
+     * @notice Returns the message hash for BN254 key registration
+     * @param operator The operator address
+     * @param operatorSet The operator set
+     * @param keyData The BN254 key data
+     * @return The message hash for signing
+     */
+    function getBN254KeyRegistrationMessageHash(
+        address operator,
+        OperatorSet memory operatorSet,
+        bytes calldata keyData
```
0xrajath, line 398: What is this keyData? Can you be more explicit as to what that is in the natspec? I.e how to construct it or what it is?

ypatil12 (approved):

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -0,0 +1,405 @@
...
+     * @param signature The signature bytes
+     * @param pubkeyG1 The G1 component of the public key
+     * @param pubkeyG2 The G2 component of the public key
+     */
+    function _verifyBN254Signature(
+        BN254.G1Point memory messagePoint,
+        bytes memory signature,
+        BN254.G1Point memory pubkeyG1,
+        BN254.G2Point memory pubkeyG2
+    ) public view {
```
solimander, line 257: nit: make internal and use test harness in tests, if needed

solimander (approved):

src/contracts/permissions/KeyRegistrar.sol
eigenmikem: This function is going to get refactored into a library in the next couple weeks, so I'm going to leave it for now

PR #1423, opened 2025-05-30
feat: multi chain interfaces

ypatil12:
**Motivation:**

Add all multi chain interfaces for easier development. 

**Modifications:**

Adds the following:
- `IBN254CertificateVerifier` (implements `IBaseCertificateVerifier`)
- `IECDSACertificateVerifier` (implements `IBaseCertificateVerifier`)
- `IBN254TableCalculator`
- `IECDSATableCalculator`
- `IOperatorTableUpdater`
- `ICrossChainRegistry`

**Result:**

Multichain interfaces. Pending adding of `OperatorTableCalculator`


8sunyuan:
Change uint96 values to uint256. Also are removing the operator table bytes view functions?

src/contracts/interfaces/IBaseCertificateVerifier.sol
```diff
@@ -0,0 +1,60 @@
...
+        BN254
+    }
+}
+
+interface IBaseCertificateVerifierEvents is IBaseCertificateVerifierTypes {
+    /// @notice Emitted when the owner of an operatorSet is updated
+    event OperatorSetOwnerUpdated(OperatorSet operatorSet, address owner);
+
+    /// @notice Emitted when the max staleness period of an operatorSet is updated
+    event MaxStalenessPeriodUpdated(OperatorSet operatorSet, uint32 maxStalenessPeriod);
```
0xrajath, line 24: Thoughts on splitting up the `avs` and `operatorSetId` and having them as indexed fields for easier offchain filtering through event topics?

src/contracts/interfaces/IBaseCertificateVerifier.sol
```diff
@@ -0,0 +1,60 @@
...
+    error CertificateStale();
+    /// @notice Thrown when the reference timestamp does not exist
+    error ReferenceTimestampDoesNotExist();
+    /// @notice Thrown when certificate verification fails
+    error VerificationFailed();
+}
+
+/// @notice A base interface that verifies certificates for a given operatorSet
+/// @notice This is a base interface that all curve certificate verifiers (eg. BN254, ECDSA) must implement
+/// @dev A single `CertificateVerifier` can be used for ONLY 1 operatorSet
```
0xrajath, line 44: I'm assuming this `@dev` comment is outdated? It's going to be a singleton contract right?

src/contracts/interfaces/IBN254CertificateVerifier.sol
```diff
@@ -0,0 +1,104 @@
...
+    }
+}
+
+interface IBN254CertificateVerifierEvents is IBN254CertificateVerifierTypes {
+    /// @notice Emitted when an BN254 table is updated
+    event TableUpdated(OperatorSet operatorSet, uint32 referenceTimestamp, BN254OperatorSetInfo operatorSetInfo);
+}
+
+/// @notice An interface for verifying BN254 certificates
+/// @notice This implements a base interface that all curve certificate verifiers (eg. BN254, ECDSA) must implement
```
0xrajath, line 45: `@notice` seems to be wrong here.

src/contracts/interfaces/IBN254CertificateVerifier.sol
```diff
@@ -0,0 +1,104 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "src/contracts/libraries/OperatorSetLib.sol";
+import "src/contracts/interfaces/ICrossChainRegistry.sol";
+import "src/contracts/interfaces/IBaseCertificateVerifier.sol";
+import "src/contracts/interfaces/IBN254TableCalculator.sol";
```
0xrajath, line 7: Thinking out loud here. For all these files, should these imports be relative imports? I've had trouble compiling contracts when the dependent lib (i.e in this case `eigenlayer-contracts`) has absolute imports instead of relative imports. the remappings seem to break.

src/contracts/interfaces/IECDSACertificateVerifier.sol
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+interface IECDSACertificateVerifierEvents is IECDSACertificateVerifierTypes {
+    /// @notice Emitted when an ECDSA table is updated
+    event TableUpdated(OperatorSet operatorSet, uint32 referenceTimestamp, ECDSAOperatorInfo[] operatorInfos);
+}
+
+/// @notice An interface for verifying ECDSA certificates
+/// @notice This implements a base interface that all curve certificate verifiers (eg. BN254, ECDSA) must implement
```
0xrajath, line 29: `@notice` seems to be wrong.

src/contracts/interfaces/IOperatorTableUpdater.sol
```diff
@@ -0,0 +1,121 @@
...
+     * @param globalTableRoot the root of the global table
+     */
+    event NewglobalTableRoot(uint32 referenceTimestamp, bytes32 globalTableRoot);
+}
+
+interface IOperatorTableUpdater is
+    IOperatorTableUpdaterErrors,
+    IOperatorTableUpdaterEvents,
+    IECDSACertificateVerifierTypes,
+    IBN254CertificateVerifierTypes,
```
0xrajath, line 30: Out of curiousity. Why not have separate OperatorTableUpdaters for ECDSA and BN254 instead of clubbing them together here? What if you wanted to add a new curve in the future?

All the other contracts have them separated right?

src/contracts/interfaces/ICrossChainRegistry.sol
```diff
@@ -0,0 +1,126 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "src/contracts/libraries/OperatorSetLib.sol";
+
+interface IOperatorTableCalculator {
+// TODO: implement, stub for now
+}
```
0xrajath, line 8: What is this stub?

src/contracts/interfaces/IBaseCertificateVerifier.sol
ypatil12: Will leave that to @bowenli86 , currently we use operatorSet where applicable

src/contracts/interfaces/IBaseCertificateVerifier.sol
ypatil12: Correct

src/contracts/interfaces/IBaseCertificateVerifier.sol
bowenli86: OperatorSet in log is `avs + id`, we should be good for now

src/contracts/interfaces/IBN254CertificateVerifier.sol
ypatil12: Ya, can update

src/contracts/interfaces/ICrossChainRegistry.sol
ypatil12: The contract that returns the stake weights for all operators in an operatorSet

src/contracts/interfaces/IOperatorTableUpdater.sol
ypatil12: Mainly because we want a single entity to update all types of tables

0xrajath: I'm assuming a new curve will just be added to this contract then?

ypatil12: Yup

PR #1429, opened 2025-06-02
feat: bn254 operator table contracts

8sunyuan:
**Motivation:**

Implementing the `BN254TableCalculator contracts`. Currently the calculator weights slashable stake in an operator set by simply adding all the slashable stake together with no multipliers/linear combination calculation involved

**Modifications:**

Updates and added to interfaces

**Result:**


src/contracts/multichain/BN254TableCalculator.sol
```diff
@@ -0,0 +1,126 @@
...
+                continue;
+            }
+
+            // Read the weights for the operator and encode them into the operatorInfoLeaves
+            // for all weights, add them to the total weights. The ith index returns the weights array for the ith operator
+            for (uint256 j = 0; j < subArrayLength; j++) {
+                totalWeights[j] += weights[i][j];
+            }
+            (BN254.G1Point memory g1Point,) = keyRegistrar.getBN254Key(operatorSet, operators[i]);
+            operatorInfoLeaves[i] = keccak256(abi.encode(BN254OperatorInfo({pubkey: g1Point, weights: weights[i]})));
```
ypatil12, line 163: Do we want to encode the leaves with a salt? We do this for rewards. Can be done in a separate PR

src/contracts/multichain/BN254TableCalculator.sol
```diff
@@ -0,0 +1,126 @@
...
+ * @title BN254TableCalculator
+ * @notice Contract that calculates BN254 operator tables for a given operatorSet
+ * @dev This contract uses an IOperatorWeightCalculator to get operator weights and formats them into the required table structure
+ */
+contract BN254TableCalculator is Initializable, OwnableUpgradeable, BN254TableCalculatorStorage {
+    using Merkle for bytes32[];
+    using BN254 for BN254.G1Point;
+
+    constructor(
+        IOperatorWeightCalculator _operatorWeightCalculator,
```
ypatil12: Thoughts on enabling this to have multiple operator weight calculators? The AVS can then add multiple to this "registry".

src/contracts/multichain/SlashableWeightCalculator.sol
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     * @notice Set the multipliers for each strategy. Note that the strategy doesn't necessarily need to be in the operatorSet
+     * to be able to be set here.
+     * @param operatorSet The operatorSet to set the multipliers for
+     * @param strategies The strategies to set the multipliers for
+     * @param multipliers The multipliers to set for the strategies
+     */
+    function setStrategyMultipliers(
```
ypatil12: This will end up being the "default" operator weight calculator. In such case, I'm wondering if the protocol should even be setting these multiples for others... my take is NO given the operational load. As such, would we want one that doesn't have multipliers and assumes like-assets (stablecoins, some LSTs).

src/contracts/multichain/SlashableWeightCalculatorStorage.sol
```diff
@@ -0,0 +1,38 @@
...
+
+    // Immutables
+    IAllocationManager public immutable allocationManager;
+
+    // Storage variables
+    /// @notice The multipliers for each strategy, per operatorSet
+    mapping(bytes32 operatorSetKey => mapping(IStrategy => uint256)) internal _multipliers;
+
+    /// @notice The lookahead blocks for the slashable stake calculation
+    uint256 public lookaheadBlocks = 100_800;
```
ypatil12: Why not set in initializer?

src/contracts/multichain/SlashableWeightCalculator.sol
```diff
@@ -0,0 +1,151 @@
...
+        uint256[] calldata multipliers
+    ) external onlyOwner {
+        // Validate the lengths of strategies and multipliers
+        require(strategies.length == multipliers.length, ArrayLengthMismatch());
+
+        // Get the key for the operatorSet
+        bytes32 operatorSetKey = operatorSet.key();
+
+        for (uint256 i = 0; i < strategies.length; i++) {
+            _multipliers[operatorSetKey][strategies[i]] = multipliers[i];
```
ypatil12: Would be nice to have introspection on multipliers

ypatil12 (commented):
First pass

ypatil12:
We also need to return the OpSet config on calculateOperatorTable (crossChainRegistry interface will have a global opSetConfig)

src/contracts/multichain/BN254TableCalculator.sol
8sunyuan: Is it for leaf uniqueness? is it necessary when we already enforce global key uniqueness in the keyRegistrar

src/contracts/multichain/SlashableWeightCalculator.sol
8sunyuan: I assume that this would be a "default" that AVSs would redeploy and setup themselves for their multipliers. Definitely don't want to be responsible for setting multipliers on behalf of everyone but if you are weighing LSTs for example it has to be updated occasionally. I think this is just part of the set of contracts that AVSs should configure and maintain. wdyt

ypatil12 (commented):
Can we add a base `IOperatorTableCalculator` that the ECDSA & BN254 table calcs implement?

src/contracts/multichain/BN254TableCalculator.sol
```diff
@@ -0,0 +1,126 @@
...
+        OperatorSet calldata operatorSet
+    ) public view virtual returns (BN254OperatorSetInfo memory operatorSetInfo) {
+        return _calculateOperatorTable(operatorSet);
+    }
+
+    /// @inheritdoc IBN254TableCalculator
+    function calculateOperatorTableBytes(
+        OperatorSet calldata operatorSet
+    ) external view virtual returns (bytes memory operatorTableBytes) {
+        return abi.encode(_calculateOperatorTable(operatorSet));
```
ypatil12, line 44: This should also encode the `OperatorSet.key()` and `OpsetConfig`

src/contracts/multichain/SlashableWeightCalculator.sol
ypatil12: Issue is Hourglass wants a default implementation, will chat with them tomorrow

src/contracts/multichain/BN254TableCalculator.sol
ypatil12: Correct. We should be fine then. Also need to update the bytes to return the opSet which is also unique so we good

src/contracts/multichain/BN254TableCalculator.sol
8sunyuan: I think we should try to avoid too many places for configuration as it adds overall complexity for an AVS.
From an AVS perspective.

CrossChainRegistry
- Where OperatorSetConfig is set
- Where OperatorTableCalculator is set. If they want a different kind of scheme then they can just deploy a new OperatorTableCalculator, OperatorWeightCalculator, and then call 

OperatorTableCalculator
- Does exactly as specified, calculates the OperatorTable
- Works for ECDSA, BN254 keys
- Utilizes a specific weighting scheme for the operator weights/stake

OperatorWeightCalculator
- Calculates operator weights
- Can be linear, non-linear, ceiling limited, etc

src/contracts/multichain/BN254TableCalculator.sol
8sunyuan: I assumed the `OpsetConfig` would be `OperatorSet.key()` would be included in the table generation separately.
So does this mean AVSs can put arbitrary bytes data into their table depending on how they implement this function

src/contracts/multichain/SlashableWeightCalculatorStorage.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1429/commits/965ead7469420398e35b654b35899c72d7b91453
Added for ECDSATableCalculator and BN254TableCalculator

src/contracts/multichain/SlashableWeightCalculator.sol
8sunyuan: Removed multipliers for now and will add in later impl

src/contracts/multichain/BN254TableCalculator.sol
ypatil12: Let's punt this for now

src/contracts/multichain/BN254TableCalculator.sol
ypatil12: Decided offline: the `CrossChainRegistry` will handle

src/contracts/multichain/BN254TableCalculator.sol
ypatil12: Agreed offline -> we'll have a single `OperatorTableCalculator` per operatorSet

src/contracts/multichain/BN254TableCalculator.sol
```diff
@@ -0,0 +1,190 @@
...
+     */
+    function initialize(address initialOwner, uint256 _lookaheadBlocks) external initializer {
+        lookaheadBlocks = _lookaheadBlocks;
+        _transferOwnership(initialOwner);
+    }
+
+    /// @inheritdoc IBN254TableCalculator
+    function calculateOperatorTable(
+        OperatorSet calldata operatorSet
+    ) public view virtual returns (BN254OperatorSetInfo memory operatorSetInfo) {
```
ypatil12: Doesn't need to be public?

src/contracts/multichain/BN254TableCalculator.sol
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * @notice Get the operator weights for a given operatorSet based on the slashable stake.
+     * @param operatorSet The operatorSet to get the weights for
+     * @return operators The addresses of the operators in the operatorSet
+     * @return weights The weights for each operator in the operatorSet, this is a 2D array where the first index is the operator
+     * and the second index is the type of weight. In this case its of length 1 and returns the slashable stake for the operatorSet.
+     */
+    function _getOperatorWeights(
+        OperatorSet calldata operatorSet
+    ) public view virtual returns (address[] memory operators, uint256[][] memory weights) {
```
ypatil12: Public or internal?

src/contracts/multichain/BN254TableCalculator.sol
```diff
@@ -0,0 +1,190 @@
...
+
+            (BN254.G1Point memory g1Point,) = keyRegistrar.getBN254Key(operatorSet, operators[i]);
+            operatorInfos[i] = BN254OperatorInfo({pubkey: g1Point, weights: weights[i]});
+        }
+
+        return operatorInfos;
+    }
+
+    /// @inheritdoc IBN254TableCalculator
+    function setLookaheadBlocks(
```
ypatil12: Let's use this as the default implementation, let's just cut this and then use the maximum? We can get rid of the Ownable inheritance as well

src/contracts/multichain/BN254TableCalculator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1429/commits/8f4078ce60e31bede6c4e27d28bd0fdaf904171d

src/contracts/multichain/BN254TableCalculator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1429/commits/8f4078ce60e31bede6c4e27d28bd0fdaf904171d fixed, is internal

src/contracts/multichain/BN254TableCalculator.sol
8sunyuan: sounds clean https://github.com/Layr-Labs/eigenlayer-contracts/pull/1429/commits/d7ee1d50b816abd861bfa96ca686cc823355ad2d

ypatil12:
moved ECDSA: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1443

ypatil12:
LGTM. Will fix the coverage separately

PR #1432, opened 2025-06-02
feat: rms core contract implementation draft

bdchatham:
**Motivation:**
This PR is an implementation of the requirements and recommendations outlined in [TDD:RMS ReleaseManager](https://www.notion.so/eigen-labs/TDD-RMS-ReleaseManager-20013c11c3e080a3a98bc16c261a6e72?source=copy_link). This introduces a new canonical EigenLayer contract for managing releases for AVS applications. 

## Requirements
- Single source of truth for discovering AVS releases whom are registered.
- Release events 
- Version string (ideally used for semantic versions)
- AVS-level artifacts (not operator set)
- OCI compliant 

## Open Considerations
- Support for arbitrary "release config": What are the use-cases? Should those not be packaged in the artifact itself?
- Support for OCI indexes: Do indexes hinder the "verifiability" of the contract? 

**Modifications:**
- ReleaseManager interface, implementation and storage contracts.

- Added a draft of the ReleaseManager contracts.

**Result:**
- Canonical protocol support for releases by an AVS.

- AVSs will be able to publish their artifacts to a protocol contract to later be used for verifiable SDLC.

### DIFF
--- a/src/contracts/core/ReleaseManager.sol
+++ b/src/contracts/core/ReleaseManager.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.27;
 
-import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
+import {OwnableUpgradeable} from "./OwnableUpgradeable.sol";
 import {ReentrancyGuard} from "@openzeppelin/contracts/security/ReentrancyGuard.sol";
 import {Checkpoints} from "@openzeppelin/contracts/utils/structs/Checkpoints.sol";
 
@@ -11,8 +11,8 @@ import {ReleaseManagerStorage} from "./ReleaseManagerStorage.sol";
 
 /**
  * @title ReleaseManager
- * @author EigenLabs
- * @notice Contract for managing an append-only log of artifact releases for AVS deployments.
+ * @author Your Organization
+ * @notice Contract for managing an append-only log of OCI artifact releases for AVS deployments.
  */
 contract ReleaseManager is ReentrancyGuard, OwnableUpgradeable, ReleaseManagerStorage {
     /**
@@ -63,60 +63,41 @@ contract ReleaseManager is ReentrancyGuard, OwnableUpgradeable, ReleaseManagerSt
     }
 
     /// @inheritdoc IReleaseManager
-    function publishArtifacts(
+    function publishRelease(
         address avs,
-        Artifact[] calldata _artifacts,
+        bytes32 digest,
+        string calldata registryUrl,
         string calldata version,
         uint256 deploymentDeadline
     ) external nonReentrant onlyRegistered(avs) onlyAuthorized(avs) {
         if (deploymentDeadline == 0) revert InvalidDeadline();
-        if (_artifacts.length == 0) revert ArrayLengthMismatch();
+        if (digest == bytes32(0)) revert InvalidDigest();
+        if (bytes(registryUrl).length == 0) revert InvalidDigest();
 
-        bytes32[] memory digests = new bytes32[](_artifacts.length);
+        PublishedRelease memory release = PublishedRelease({
+            digest: digest,
+            registryUrl: registryUrl,
+            version: version,
+            deploymentDeadline: deploymentDeadline,
+            publishedAt: block.timestamp
+        });
 
-        // First, store all artifacts
-        for (uint256 i = 0; i < _artifacts.length; i++) {
-            Artifact memory artifact = _artifacts[i];
-            artifact.publishedAt = block.timestamp;
+        // Add to the append-only log
+        allPublishedReleases[avs].push(release);
 
-            // Create unique key: keccak256(abi.encodePacked(avs, digest))
-            bytes32 key = _getArtifactKey(avs, artifact.digest);
+        // Update the checkpoint to point to the new index
+        uint256 newIndex = allPublishedReleases[avs].length - 1;
+        releaseIndexHistory[avs].push(block.number, newIndex);
 
-            // Store artifact
-            artifacts[key] = artifact;
-            artifactExists[key] = true;
-            digests[i] = artifact.digest;
-        }
-
-        // Create releases for each artifact
-        for (uint256 i = 0; i < _artifacts.length; i++) {
-            PublishedRelease memory release = PublishedRelease({
-                digest: _artifacts[i].digest,
-                registryUrl: _artifacts[i].registryUrl,
-                version: version,
-                deploymentDeadline: deploymentDeadline,
-                publishedAt: block.timestamp
-            });
-
-            // Add to the append-only log
-            allPublishedReleases[avs].push(release);
-
-            // Update the checkpoint to point to the new index
-            uint256 newIndex = allPublishedReleases[avs].length - 1;
-            releaseIndexHistory[avs].push(block.number, newIndex);
-        }
-
-        emit ArtifactsPublished(avs, version, deploymentDeadline, digests);
+        emit ReleasePublished(avs, version, digest, registryUrl, deploymentDeadline);
     }
 
     /// @inheritdoc IReleaseManager
-    function deprecateArtifact(
+    function deprecateRelease(
         address avs,
         bytes32 digest
     ) external nonReentrant onlyRegistered(avs) onlyAuthorized(avs) {
-        // Verify artifact exists
-        bytes32 artifactKey = _getArtifactKey(avs, digest);
-        if (!artifactExists[artifactKey]) revert ArtifactNotFound();
+        if (digest == bytes32(0)) revert InvalidDigest();
 
         // Mark as deprecated
         bytes32 deprecationKey = keccak256(abi.encodePacked(avs, digest));
@@ -125,9 +106,9 @@ contract ReleaseManager is ReentrancyGuard, OwnableUpgradeable, ReleaseManagerSt
         isDeprecated[deprecationKey] = true;
 
         // Add to deprecated list
-        deprecatedArtifacts[avs].push(digest);
+        deprecatedReleases[avs].push(digest);
 
-        emit ArtifactDeprecated(avs, digest);
+        emit ReleaseDeprecated(avs, digest);
     }
 
     /**
@@ -155,27 +136,10 @@ contract ReleaseManager is ReentrancyGuard, OwnableUpgradeable, ReleaseManagerSt
     }
 
     /**
-     * @notice Generate unique key for artifact storage
-     * @param avs The AVS address
-     * @param digest The artifact digest
-     * @return The unique key
-     */
-    function _getArtifactKey(address avs, bytes32 digest) internal pure returns (bytes32) {
-        return keccak256(abi.encodePacked(avs, digest));
-    }
-
-    /**
      *
      *                         VIEW FUNCTIONS
      *
      */
-
-    /// @inheritdoc IReleaseManager
-    function getArtifact(address avs, bytes32 digest) external view returns (Artifact memory) {
-        bytes32 key = _getArtifactKey(avs, digest);
-        if (!artifactExists[key]) revert ArtifactNotFound();
-        return artifacts[key];
-    }
 
     /// @inheritdoc IReleaseManager
     function getPublishedReleases(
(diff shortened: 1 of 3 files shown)


src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,193 @@
...
+        registeredAVS[avs] = true;
+        emit AVSRegistered(avs);
+    }
+
+    /// @inheritdoc IReleaseManager
+    function deregister(address avs) external onlyAuthorized(avs) {
+        if (!registeredAVS[avs]) revert AVSNotRegistered();
+        registeredAVS[avs] = false;
+        emit AVSDeregistered(avs);
+    }
```
0xrajath, line 63: Since this is going to be a canonical contract on L1, we actually don't need to separately register the AVSs here. It will have access to call the Core protocol and check.

To go one level deeper. I would say that we should be checking if the operator set of the AVS is valid and not the AVS directly.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,193 @@
...
+
+    /// @inheritdoc IReleaseManager
+    function publishRelease(
+        address avs,
+        bytes32 digest,
+        string calldata registryUrl,
+        string calldata version,
+        uint256 deploymentDeadline
```
0xrajath, line 71: I think we need to do this at the operator set level rather than at the AVS level. For example, in the hourglass case, different executor operator sets can have different performer logic right? And each of them will need it's own release.

so you should actually be passing in `OperatorSet operatorSet` in here instead of `address avs` IMO. You can take a look at how it's done in the Mailbox if you want inspiration.

0xrajath: Also related to https://github.com/Layr-Labs/eigenlayer-contracts/pull/1432/files#r2124216248. You don't need to separately register the operator set either. You can have a validation check against the `AllocationManager.isOperatorSet` which returns a bool saying whether it's a valid operator set or not that's registered in the core protocol (and throw an error if not)

0xrajath: You would just need to set the AllocationManager as an `immutable` through the constructor during contract creation to have access to the AllocationManager in this contract.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,193 @@
...
+
+    modifier onlyRegistered(address avs) {
+        if (!registeredAVS[avs]) revert AVSNotRegistered();
+        _;
+    }
+
+    modifier onlyAuthorized(address avs) {
+        if (!_isAuthorized(avs, msg.sender)) revert Unauthorized();
+        _;
+    }
```
0xrajath, line 32: You don't need this.

Take a look at how the other core contracts do this. They inherit a contract called `PermissionControllerMixin` which allows you to do `checkCanCall(avs)` as a modifier on the functions below.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     *
+     *                         MODIFIERS
+     *
+     */
+
+    modifier onlyRegistered(address avs) {
+        if (!registeredAVS[avs]) revert AVSNotRegistered();
+        _;
+    }
```
0xrajath, line 27: You don't need this due to https://github.com/Layr-Labs/eigenlayer-contracts/pull/1432/files#r2124216248

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,193 @@
...
+        // Check if caller is the AVS itself
+        if (avs == caller) return true;
+
+        // Check if caller is an admin via permission controller
+        if (address(permissionController) != address(0)) {
+            return permissionController.isAdmin(avs, caller);
+        }
+
+        return false;
+    }
```
0xrajath, line 136: You don't need this: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1432/files#r2124242306

src/contracts/core/ReleaseManagerStorage.sol
```diff
@@ -0,0 +1,35 @@
...
+
+    /// @notice Mapping to track deprecated releases per AVS
+    /// @dev Key is keccak256(avs, digest)
+    mapping(bytes32 => bool) public isDeprecated;
+
+    /// @notice List of deprecated release digests per AVS
+    mapping(address => bytes32[]) public deprecatedReleases;
+
+    /// @notice Storage gap for future upgrades
+    uint256[44] private __gap;
```
0xrajath, line 34: Can make this `uint256[50] private __gap;`. For context this is intended for saving storage slots for future upgrades. No need to start off at 44.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,193 @@
...
+import {IReleaseManager} from "../interfaces/IReleaseManager.sol";
+import {IPermissionController} from "../interfaces/IPermissionController.sol";
+import {ReleaseManagerStorage} from "./ReleaseManagerStorage.sol";
+
+/**
+ * @title ReleaseManager
+ * @author Your Organization
+ * @notice Contract for managing an append-only log of OCI artifact releases for AVS deployments.
+ */
+contract ReleaseManager is ReentrancyGuard, OwnableUpgradeable, ReleaseManagerStorage {
```
0xrajath, line 17: You most likely don't need `OwnableUpgradeable` since there are no functions setting any configs by the owner of this contract.

0xrajath: `ReentrancyGuard` can be `ReentrancyGuardUpgradeable` instead. It adds an addtional 50 storage slots. OpenZeppelin should have a lib for that.

src/contracts/interfaces/IReleaseManager.sol
```diff
@@ -0,0 +1,152 @@
...
+}
+
+interface IReleaseManager is IReleaseManagerErrors, IReleaseManagerEvents {
+    struct PublishedRelease {
+        bytes32 digest;
+        string registryUrl;
+        string version;
+        uint256 deploymentDeadline;
+        uint256 publishedAt;
+    }
```
0xrajath, line 53: Small nit: Move this to a `IReleaseMangerTypes` interface above and inherit that here. Take a look at how the Mailbox does it.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     *
+     *                         INITIALIZATION
+     *
+     */
+
+    function initialize(address _permissionController) public initializer {
+        __Ownable_init();
+        permissionController = IPermissionController(_permissionController);
```
0xrajath, line 42: Setting of the PermissionController should happen in the constructor and it should be set as an `immutable` so that it can't be changed later.

https://solidity-by-example.org/immutable/

0xrajath: For some context, `initialize` is a workaround in Upgradeable contracts that 'sort of' does the job of the constructor at upgrade time. But the issue is that that variable will be in storage instead which leads to the following issues:

1. It's more expensive to access since it's a storage read. Immutables are stored in the contract bytecode at creation time (very cheap to access)
2. Storage slots can be overriden or changed.

0xrajath: https://docs.openzeppelin.com/upgrades-plugins/writing-upgradeable#initializers

0xrajath: Take a look at the `constructor` and `initialize` functions of the other core contracts for inspiration.

src/contracts/core/ReleaseManager.sol
bdchatham: Re: operator set level
I originally had this mental model, but heard opinion that applications shouldn't vary by operator set. The binaries should abstract that from the protocol and be configurable.

The use-cases I see otherwise are for granular control over software rollouts to operator sets, but I got feedback this is not a useful path, given operators dictate this today. In the future when we have a "virtualization as an AVS" then control over which operators you want to deploy you software, similar to traditional wave-based or zonal deployments, could make more sense given the operators aren't slash-able for faulty AVS releases. 

Given that's not an immediate use-case, it's not a strong reason for me to push back. There are also other ways to do operator set association, like at the OCI artifact level.

src/contracts/core/ReleaseManager.sol
bdchatham: Fairly well aware of the permission controller from my work on UAM. Clearly using it wrong here. I'll address this.

src/contracts/core/ReleaseManager.sol
bdchatham: Let's chat more about operator set level in the followup review. That's a critical and contentious point.

src/contracts/core/ReleaseManagerStorage.sol
bdchatham: Got it. I had only done brief research on this so was sort-of in the dark. This is helpful info.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,193 @@
...
+            deploymentDeadline: deploymentDeadline,
+            publishedAt: block.timestamp
+        });
+
+        // Add to the append-only log
+        allPublishedReleases[avs].push(release);
+
+        // Update the checkpoint to point to the new index
+        uint256 newIndex = allPublishedReleases[avs].length - 1;
+        releaseIndexHistory[avs].push(block.number, newIndex);
```
0xrajath, line 90: Out of curiosity. Why are we checkpointing the releases? Is that a requirement?

0xrajath: If it is, I have some suggestions:
1. You can maintain a single checkpoint array of `PublishedRelease` rather than checkpoint the array index
2. There should be a timestamp based checkpoint lib avaiable (rather than block number based). That's always more preferrable unless you had a specific reason to use block number based checkpointing.

0xrajath: Checkpointing is slightly more nuanced, so happy to help out here.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,193 @@
...
+
+    /// @inheritdoc IReleaseManager
+    function deprecateRelease(
+        address avs,
+        bytes32 digest
+    ) external nonReentrant onlyRegistered(avs) onlyAuthorized(avs) {
+        if (digest == bytes32(0)) revert InvalidDigest();
+
+        // Mark as deprecated
+        bytes32 deprecationKey = keccak256(abi.encodePacked(avs, digest));
```
0xrajath, line 103: I would suggest using `abi.encode` here. It's guaranteed to have collision resistance.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,193 @@
...
+        if (isDeprecated[deprecationKey]) revert AlreadyDeprecated();
+
+        isDeprecated[deprecationKey] = true;
+
+        // Add to deprecated list
+        deprecatedReleases[avs].push(digest);
```
0xrajath, line 109: Do we need a separate array of deprecated releases? Is that a requirement?

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /// @inheritdoc IReleaseManager
+    function publishRelease(
+        address avs,
+        bytes32 digest,
+        string calldata registryUrl,
+        string calldata version,
+        uint256 deploymentDeadline
+    ) external nonReentrant onlyRegistered(avs) onlyAuthorized(avs) {
```
0xrajath, line 72: Small nit: You most likely don't need `nonReentrant` and subsequently not have this contract inherit `ReentrancyGuard`. This is because you're not making any external calls to unknown contracts. Having that `nonReentrant` modifier adds additional gas cost.

0xrajath: Same for the `deprecateRelease()` function as well.

src/contracts/core/ReleaseManager.sol
seanmcgary: @0xrajath this is wrong.
> For example, in the hourglass case, different executor operator sets can have different performer logic right?

A Performer is capable to complete ALL task types (technically) and it is up to the Operator to ensure they are joining the correct operator set for their hardware class

PR #1462, opened 2025-06-11
test(wip): basic harness for precision loss analysis

wadealexc:
[DO NOT MERGE]

WIP harness for precision loss analysis. Run with `forge t --mt test_rounding -vvvv`


src/test/integration/tests/Rounding.t.sol
```diff
@@ -11,66 +11,131 @@ contract Integration_Rounding is IntegrationCheckUtils {
...
+        attacker = new User("Attacker"); // attacker can be both operator and staker
         badAVS = new AVS("BadAVS");
         strategy = lstStrats[0];
         token = IERC20Metadata(address(strategy.underlyingToken()));
-
+        deal(address(token), address(attacker), uint256(1000000000000000000)); // TODO: make the balance a fuzzing param
+        
+        // good staker and operator setup
+        goodStaker = new User("goodStaker");
+        deal(address(token), address(goodStaker), uint256(1000000000000000000)); // TODO: make the balance a fuzzing param
```
wadealexc: Don't need to make this one a fuzzing param. The point is to have _any_ excess balance - so this should be fine. It shouldn't impact any actual math - it's just available to be withdrawn if an exploit does exist.

The attacker's balance should be a fuzzing param though, that part is correct

src/test/integration/tests/Rounding.t.sol
```diff
@@ -11,66 +11,131 @@ contract Integration_Rounding is IntegrationCheckUtils {
...
 
-    function test_rounding() public rand(0) {
-        _magnitudeManipulation();
-        _setupFinal();
-        _final();
+    // TODO: parameterize deposits
+    // TODO: consider manual fuzzing from 1 up to WAD - 1
+    function test_rounding(uint64 wadToSlash) public rand(0) { 
+        // bound wadToSlash to 1 < wadToSlash < WAD - 1
+        wadToSlash = uint64(bound(wadToSlash, 1, WAD - 1));
```
wadealexc: no real reason to exclude `WAD` itself, especially if we're looking for a "total assets entering and leaving" approach

src/test/integration/tests/Rounding.t.sol
nadir-akhtar: Good point!

src/test/integration/tests/Rounding.t.sol
nadir-akhtar: I think we need to exclude WAD as, otherwise, we try to slash 0 mags later in the final slash

src/test/integration/tests/Rounding.t.sol
```diff
@@ -11,66 +11,131 @@ contract Integration_Rounding is IntegrationCheckUtils {
...
+        }));
+        
+        _print("slash");
+        
+        // roll forward past the escrow delay
+        rollForward({blocks: slashEscrowFactory.getGlobalEscrowDelay() + 1});
+        
+        // release funds
+        vm.prank(address(attacker));
+        slashEscrowFactory.releaseSlashEscrow(rOpSet, 1); // 1 is used as it's the first slashId
```
wadealexc: nice - since the attacker is the redistribution target, this should work. we can probably pretty easily move this to a "total entering + leaving" analysis by adding a full withdrawal of the attacker's assets -- then, the `_print` method will show the attacker's total assets in the "held balance" line. if it's more than the initial balance, we have a problem.

src/test/integration/tests/Rounding.t.sol
wadealexc: oh, then these maybe just need to be separate params? my point is mainly that we want to also check the case where the exploit step slashes 100%

src/test/integration/tests/Rounding.t.sol
```diff
@@ -0,0 +1,277 @@
...
+
+        // Register for both opsets
+        attacker.registerForOperatorSet(mOpSet);
+        attacker.registerForOperatorSet(rOpSet);
+        
+        _print("setup");
+    }
+
+    // TODO: consider incremental manual fuzzing from 1 up to WAD - 1
+    function test_rounding_allMagsSlashed(uint16 slashes, uint64 initialWadToSlash, uint64 _initTokenBalance, uint24 r) public rand(r) {
```
wadealexc: IMO we should choose a lane for param generation:

1. either use `rand(r)` and the `_randUint` method to generate _all_ parameters
2. or allow forge to generate all parameters and if we need additional randomness, use `cheats.randomUint` and similar

#1 means we should do sequential runs, supplying (1, 2, 3, 4, etc)
#2 means we just let forge decide and we only control the number of runs

wadealexc: Or I guess there's a way to do both - let forge generate parameters via #2, but run tests using `forge t --fuzz-seed <1, 2, 3, ...>`

That may be most convenient?

src/test/integration/tests/Rounding.t.sol
```diff
@@ -0,0 +1,277 @@
...
+        vm.pauseGasMetering();
+        // Bound slashes to a reasonable range, with at least 1 slash.
+        // Note: Runs after 2500 hit OOG errors with default gas.
+        slashes = uint16(bound(slashes, 1, 5000));
```
wadealexc: IMO number of slashes doesn't really impact the "path" so much as it defines the depth you're testing to (and overall runtime). I think this should not be a random param, but something we define.

src/test/integration/tests/Rounding.t.sol
```diff
@@ -0,0 +1,277 @@
...
+        // Each slash requires at least 1 mag. As such, we need to bound wadToSlash to 1 <= wadToSlash <= WAD - 1.
+        initialWadToSlash = uint64(bound(initialWadToSlash, 1, WAD - 1));
+        
+        // Bound initTokenBalance to a reasonable range to avoid overflow, with at least 1 token.
+        // Using ~18.45 quintillion tokens max (should be enough for any realistic test).
+        initTokenBalance = uint64(bound(_initTokenBalance, 1, type(uint64).max));
+        deal(address(token), address(attacker), initTokenBalance);
+        
+        _magnitudeManipulation(initialWadToSlash); // Manipulate operator magnitude for a given strategy.
+        _deposit(initialWadToSlash); // Setup operator with new opSet as well as honest stake in same strategy.
```
wadealexc: Here's a point we may want to change - we're allocating all available magnitude before we use multiple slashes to "get funds out".

It kinda makes sense that this wouldn't result in any loss.

What if we don't allocate all available magnitude here, and instead allocate a portion of it (parameterized). We then use multiple slashes to get a portion of the funds out via redistribution, and then attempt to withdraw remaining funds via queue+completeWithdrawals?

src/test/integration/tests/Rounding.t.sol
```diff
@@ -0,0 +1,277 @@
...
+        // We do two slashes, the sum of which slash 1 WAD (all operator magnitude) in total. 
+        // Each slash requires at least 1 mag. As such, we need to bound wadToSlash to 1 <= wadToSlash <= WAD - 1.
+        initialWadToSlash = uint64(bound(initialWadToSlash, 1, WAD - 1));
+        
+        // Bound initTokenBalance to a reasonable range to avoid overflow, with at least 1 token.
+        // Using ~18.45 quintillion tokens max (should be enough for any realistic test).
+        initTokenBalance = uint64(bound(_initTokenBalance, 1, type(uint64).max));
+        deal(address(token), address(attacker), initTokenBalance);
+        
+        _magnitudeManipulation(initialWadToSlash); // Manipulate operator magnitude for a given strategy.
```
wadealexc: I kinda want to refactor this a bit to more explicitly set the parameter we're trying to set here -- we're using this method to "set max magnitude", not "initialWadToSlash"

src/test/integration/tests/Rounding.t.sol
wadealexc: Maybe this is captured in partialMagsSlashed? not sure.

src/test/integration/tests/Rounding.t.sol
nadir-akhtar: We currently do that with the final slash in the "allMagsSlashed" test case -- the bound just prevents us from slashing everything at once before we deposit. Lmk if you agree w/ that!

src/test/integration/tests/Rounding.t.sol
nadir-akhtar: I actually think it's valuable to randomize within a bound! I've just found a test where 500 slashes does _not_ show precision loss by the end of the test, but 1 slash _does_.

I suggest we either do one of two things:
1. Keep # of slashes as parameterized value (easy, more simple)
2. Measure more granular precision loss after every operation (hard, more granular)

We can measure it how we like, but I suggest we keep slashing as a parametrized value until we have more sophisticated intermediate measurements for precision loss in the file. So we can do option 1 until option 2 is ready.

src/test/integration/tests/Rounding.t.sol
nadir-akhtar: Indeed -- currently what we do in both the partial and total mags slashed test cases!

src/test/integration/tests/Rounding.t.sol
nadir-akhtar: I prefer option #2 but open to either. My impression was we eventually want to give Foundry some more direction (e.g. slash from initialMaxMag 1 to WAD - 1), but I believe we discussed that as a parallel, second route which we would do _in addition_ to letting Foundry fire whatever inputs it feels like

src/test/integration/tests/Rounding.t.sol
nadir-akhtar: `partialMagsSlashed` somewhat does this. It allocates all magnitude but doesn't fully slash at the end, leaving some partial residual magnitude. We can def add another test case that partially allocates.

PR #1463, opened 2025-06-11
refactor: `BN254OperatorTableCalculator`

ypatil12:
**Motivation:**

We want to refactor the table calculator so that devs only have to worry about doing 1 thing: calculating weights of the operator. Everything else should be handled by a base contract

**Modifications:**

- Create `BN254TableCalculatorBase`, an abstract contract that has all methods implemented, except `_getOperatorWeights`
- `BN254TableCalculator` now inherits from `BN254TableCalculatorBase` and implements simple addition of operator stakes
- Add unit tests for `BN254TableCalculatorBase` and a tree file
- Resizes arrays for operators who have no slashable stake or are not registered in the `KeyRegistrar`

**Result:**

Cleaner abstraction


src/contracts/multichain/BN254TableCalculatorBase.sol
```diff
@@ -0,0 +1,161 @@
...
+                continue;
+            }
+
+            // Read the weights for the operator and encode them into the operatorInfoLeaves
+            // for all weights, add them to the total weights. The ith index returns the weights array for the ith operator
+            for (uint256 j = 0; j < subArrayLength; j++) {
+                totalWeights[j] += weights[i][j];
+            }
+            (BN254.G1Point memory g1Point,) = keyRegistrar.getBN254Key(operatorSet, operators[i]);
+            operatorInfoLeaves[i] = keccak256(abi.encode(BN254OperatorInfo({pubkey: g1Point, weights: weights[i]})));
```
eigenmikem, line 147: Should we maybe "effectively" resize this array? As it stands operatorInfoLeaves will potentially have blank entries if operators aren't registered

ypatil12: Registration checked in the get operator weights function

ypatil12: Nvm. Also beed to adjust number of operators

PR #1469, opened 2025-06-12
feat: release manager

0xClandestine:
**Motivation:**

We want to enable AVSs to publish release metadata on-chain, giving operators a reliable and verifiable source of truth.

**Modifications:**

* Added a new core contract: `ReleaseManager`.
* AVSs can call `publishRelease(operatorSet, release)` to push release data.

**Result:**

AVSs can now post release info on-chain, allowing operators to query a canonical, auditable source for updates.

### DIFF
--- a/docs/core/ReleaseManager.md
+++ b/docs/core/ReleaseManager.md
@@ -30,7 +30,7 @@ An AVS in the context of `ReleaseManager` is defined as the `address` of the con
 ### Important Notes on Release Management
 
 * **Latest Release Validity**: Only the latest release for an operator set is considered valid. Previous releases become obsolete as soon as a new release is published.
-* **Upgrade Deadlines**: The `upgradeByTime` timestamp is a suggested deadline and is not enforced on-chain or off-chain. It serves as a communication mechanism for AVSs to indicate when operators should complete their upgrades.
+* **Upgrade Deadlines**: The `upgradeByTime` timestamp (in Unix time) is a suggested deadline and is not enforced on-chain or off-chain. It serves as a communication mechanism for AVSs to indicate when operators should complete their upgrades.
 * **Multiple Releases in Same Block**: If multiple releases are published in the same block with the same `upgradeByTime`, the last transaction processed in that block will determine the latest valid release.
 
 ---
@@ -40,7 +40,7 @@ An AVS in the context of `ReleaseManager` is defined as the `address` of the con
 A release represents a collection of software artifacts that operators in an operator set must run. Each release is associated with a specific operator set and contains:
 
 * **Artifacts**: An array of software artifacts, each with a digest (hash) and registry URL
-* **Upgrade By Time**: A timestamp by which operators must complete the upgrade
+* **Upgrade By Time**: A Unix timestamp by which operators should complete the upgrade
 
 The release structure is defined as:
 


src/contracts/interfaces/IReleaseManager.sol
```diff
@@ -0,0 +1,133 @@
...
+    /// @param operatorSet The operator set to publish the release for.
+    /// @param digest The unique identifier for the release content.
+    /// @param registryUrl The URL where the release artifacts can be found.
+    /// @param version The semantic version of the release.
+    /// @dev Only callable by addresses with permissioned by the operator set's AVS.
+    function publishRelease(
+        OperatorSet calldata operatorSet,
+        bytes32 digest,
+        string calldata registryUrl,
+        Version calldata version
```
0xrajath: Can we make this into a struct called `Release`? And it should be an an array of `Release`.

Every release should have an array of digests associated with it.

0xrajath: Or maybe `Release` just has `digest` and `registryUrl`. And the version stays the same across both.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,216 @@
...
+
+    /// @inheritdoc IReleaseManager
+    function publishRelease(
+        OperatorSet calldata operatorSet,
+        bytes32 releaseDigest,
+        string calldata registryUrl,
+        Version calldata version
+    ) external checkCanCall(operatorSet.avs) {
+        // Parse the next releaseId (equal to the total number of releases).
+        uint256 releaseId = getTotalOperatorSetReleases(operatorSet);
```
0xrajath: Why do we need this releaseId?

src/contracts/core/ReleaseManagerStorage.sol
```diff
@@ -0,0 +1,36 @@
...
+    /// @dev operatorSet, releaseId, releaseDigest => (uint32(deprecationTimestamp), version))
+    mapping(bytes32 operatorSetKey => EnumerableMap.Bytes32ToUintMap releaseDigests) internal _releaseDigests;
+
+    /// @notice The deployment deadline for a release digest.
+    /// @dev operatorSet => (uint224(releaseId))
+    mapping(bytes32 operatorSetKey => Snapshots.DefaultZeroHistory) internal _releaseSnapshots;
+
+    /// @notice The registry URL for a release digest.
+    mapping(bytes32 operatorSetKey => mapping(bytes32 releaseDigest => string registryUrl)) internal
+        _releaseRegistryUrls;
```
0xrajath: Why can't all of this just be replaced by a version of `https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/Checkpoints.sol` ?

Then you just have an `operatorSetKey` -> Checkpoint mapping

In each checkpoint it has

```
uint256 timestamp
Release[] release
uint256 expiry
```

And `Release` is
 ```
bytes32 digest
string registryUrl
Version version
```

File: src/contracts/interfaces/IReleaseManager.sol (hunk 1 above)
Comment (0xrajath) [reply src/contracts/interfaces/IReleaseManager.sol]:
```
Release {
  bytest digest,
  string registryUrl,
  Version version
}
```

File: src/contracts/core/ReleaseManager.sol [hunk 4]
```diff
@@ -0,0 +1,218 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts/utils/Strings.sol";
+import "../mixins/PermissionControllerMixin.sol";
+import "../mixins/SemVerMixin.sol";
+import "./ReleaseManagerStorage.sol";
+
+contract ReleaseManager is Initializable, ReleaseManagerStorage, PermissionControllerMixin, SemVerMixin {
```
0xrajath, line 9: Needs to be upgradeable.

0xClandestine: ?

0xrajath: Oh I just meant when setting it up initially.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,88 @@
...
+        return _operatorSetReleases[operatorSet.key()][index].artifacts;
+    }
+
+    function getLatestReleaseArtifacts(
+        OperatorSet memory operatorSet
+    ) external view returns (Artifact[] memory) {
+        Release[] storage releases = _operatorSetReleases[operatorSet.key()];
+
+        return releases[releases.length - 1].artifacts;
+    }
```
0xrajath: I would change these 2 to `getRelease` equivalent that return the Release object rather than just the aritifact.

0xrajath: We probably need a function to check the upgradeByTime for a given release.

0xClandestine: good catch, meant to change that

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,88 @@
...
+        Artifact[] calldata artifacts,
+        uint32 upgradeByTime
+    ) external checkCanCall(operatorSet.avs) {
+        Release[] storage releases = _operatorSetReleases[operatorSet.key()];
+
+        uint256 newTotalReleases = releases.length + 1;
+
+        assembly {
+            sstore(releases.slot, newTotalReleases)
+        }
```
0xrajath: Thinking out loud here. Is there any benefit to making the releases an `EnumerableSet` instead of an array? I'm assuming we would still need a key, which in your current case is the array Index serves the role.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,88 @@
...
+     *                         EXTERNAL FUNCTIONS
+     *
+     */
+
+    /// @inheritdoc IReleaseManager
+    function publishRelease(
+        OperatorSet calldata operatorSet,
+        Artifact[] calldata artifacts,
+        uint32 upgradeByTime
+    ) external checkCanCall(operatorSet.avs) {
```
0xrajath: Can we return the array index here at which it was inserted?

src/contracts/core/ReleaseManager.sol
0xrajath: I'm assuming your current assembly is resizing the array right?

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,88 @@
...
+            sstore(releases.slot, newTotalReleases)
+        }
+
+        Release storage release = releases[newTotalReleases - 1];
+
+        for (uint256 i = 0; i < artifacts.length; ++i) {
+            release.artifacts.push(artifacts[i]);
+        }
+
+        release.upgradeByTime = upgradeByTime;
```
0xrajath: Can we have a require check here such that we have an invariant that the `upgradeByTime` should be strictly greater than the previous entry's `upgradeByTime`?

0xrajath: This ensures that the `upgradeByTime` for each entry is monotonically increasing and is easy to reason about.

src/contracts/interfaces/IReleaseManager.sol
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    struct Release {
+        Artifact[] artifacts;
+        uint32 upgradeByTime;
+    }
+}
+
+interface IReleaseManagerEvents is IReleaseManagerTypes {
+    event ReleasePublished(OperatorSet indexed operatorSet, Artifact[] artifacts, uint32 upgradeByTime);
```
0xrajath: Can you index on a struct here? or should it be the operatorSetKey?

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,88 @@
...
+import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts/utils/Strings.sol";
+import "../mixins/PermissionControllerMixin.sol";
+import "../mixins/SemVerMixin.sol";
+import "./ReleaseManagerStorage.sol";
+
+contract ReleaseManager is Initializable, ReleaseManagerStorage, PermissionControllerMixin, SemVerMixin {
+    using EnumerableSet for EnumerableSet.Bytes32Set;
+    using OperatorSetLib for OperatorSet;
+    using Strings for uint16;
```
0xrajath: Do you still need these if you're not using them?

src/contracts/interfaces/IReleaseManager.sol
```diff
@@ -0,0 +1,47 @@
...
+
+interface IReleaseManagerTypes {
+    struct Artifact {
+        bytes32 digest;
+        string registryUrl;
+    }
+
+    struct Release {
+        Artifact[] artifacts;
+        uint32 upgradeByTime;
```
0xrajath, line 25: Any reason this can't be a `uint256`?

0xClandestine: Only using `uint32` given we use it everywhere else.

src/contracts/interfaces/IReleaseManager.sol
0xClandestine: Yeah so long as the struct has a fixed size of 32 bytes or less.

src/contracts/core/ReleaseManager.sol
0xClandestine: Not really besides extra introspection, which we decided wasn't needed.

src/contracts/core/ReleaseManager.sol
0xClandestine: [2209f1d](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/2209f1db52014f002d2d44bcb2e8ffb4b8bc9798)

src/contracts/core/ReleaseManager.sol
0xClandestine: [4ba590d](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/4ba590d256de87d49afb2490b80e00c20aaddce6)

src/contracts/core/ReleaseManager.sol
0xClandestine: [4ba590d](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/4ba590d256de87d49afb2490b80e00c20aaddce6)

src/contracts/core/ReleaseManager.sol
0xClandestine: Is this needed though? Should def require it's greater than or equal to current block.timestamp but see much benefit to the above.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,83 @@
...
+     */
+
+    /// @inheritdoc IReleaseManager
+    function publishRelease(
+        OperatorSet calldata operatorSet,
+        Release calldata release
+    ) external checkCanCall(operatorSet.avs) returns (uint256 releaseId) {
+        Release[] storage releases = _operatorSetReleases[operatorSet.key()];
+
+        require(release.upgradeByTime >= block.timestamp, UpgradeByTimeNotInFuture());
```
0xClandestine: Wondering if this should just be `>` rather than `>=`.

src/contracts/core/ReleaseManager.sol
0xrajath: I think we should have this check. Otherwise older releases can have `upgradeByTime` at a later date than a newer one.

src/contracts/core/ReleaseManager.sol
0xrajath: I think it can be `>=` if they want to do instant upgrade for whatever reason.

src/contracts/core/ReleaseManager.sol
0xClandestine: I thought the expectation was that the latest is always what should be used. Let's imagine a case where bad release is pushed, and they just wanna overwrite and move forward.

src/contracts/core/ReleaseManager.sol
0xClandestine: Should probably rename the error then, cool.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /// @inheritdoc IReleaseManager
+    function getLatestRelease(
+        OperatorSet memory operatorSet
+    ) external view returns (Release memory) {
+        Release[] storage releases = _operatorSetReleases[operatorSet.key()];
+
+        return releases[releases.length - 1];
+    }
```
0xrajath: Can we do the following getters as well?
1. A getter that gives the `upgradeByTime` for a given releaseId and operatorSetkey?
2. A getter that tells if a release is valid or not based on if there are newer releases ahead of it and if the timestamp is behind or ahead of the latest release's `upgradeByTime`.

src/contracts/core/ReleaseManager.sol
0xrajath: Ah interesting. Then we need some way to enforce that in the getters:

https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469#discussion_r2152474913

src/contracts/core/ReleaseManager.sol
0xClandestine: `Release` struct contains `upgradeByTime`. Do you still want a simpler alias specifically for upgradeByTime?

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,78 @@
...
+        releaseId = releases.length;
+        // Increment total releases for this operator set.
+        releases.push();
+        // Copy the release to storage.
+        for (uint256 i = 0; i < release.artifacts.length; ++i) {
+            releases[releaseId].artifacts.push(release.artifacts[i]);
+        }
+        releases[releaseId].upgradeByTime = release.upgradeByTime;
+
+        emit ReleasePublished(operatorSet, release);
```
0xrajath: Can we also have the `releaseId` emitted as an indexed topic for this event? Ideally the first param.

src/contracts/core/ReleaseManager.sol
0xrajath: Yeah that would be a simple helper for operators who just want to know what time to upgrade by

src/contracts/core/ReleaseManager.sol
0xClandestine: Can do prefer to have it `(operatorSet, releaseId, release)` though to maintain the same order used in storage.

0xClandestine: Will index as well.

src/contracts/core/ReleaseManager.sol
0xClandestine: [55f80d6](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/55f80d65ffecc21e020e156b5dfc80817b7bda53)

src/contracts/core/ReleaseManager.sol
0xrajath: Ok. I just realised these are the getters we need:
1. `function getTotalReleases(OperatorSet memory operatorSet) external view returns (uint256)`
2. `function getRelease(OperatorSet memory operatorSet, uint256 releaseId) external view returns (Release memory, bool isValid)`: We should also return a bool saying if the release is valid or not. Use function 5 below.
3. `function getLatestRelease(OperatorSet memory operatorSet) external view returns (Release memory, uint256 releaseId)`: We should return the `releaseId` as well here IMO.
4. `function getLatestUpgradeByTime(OperatorSet memory operatorSet) external view returns (uint32 upgradeByTime)`
5. `function isValidRelease(OperatorSet memory operatorSet, uint256 releaseId) external view returns (bool isValid)`

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     *
+     *                         INITIALIZING FUNCTIONS
+     *
+     */
+    constructor(
+        IPermissionController _permissionController,
+        string memory _version
+    ) ReleaseManagerStorage() PermissionControllerMixin(_permissionController) SemVerMixin(_version) {
```
ypatil12: Nit: no need to add `ReleaseManagerStorage` here?

src/contracts/core/ReleaseManager.sol
0xClandestine: [bd23dce](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/bd23dcebe8595d00a65bdd8044ea08a50760522a)

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,85 @@
...
+     */
+
+    /// @inheritdoc IReleaseManager
+    function publishRelease(
+        OperatorSet calldata operatorSet,
+        Release calldata release
+    ) external checkCanCall(operatorSet.avs) returns (uint256 releaseId) {
+        Release[] storage releases = _operatorSetReleases[operatorSet.key()];
+
+        require(release.upgradeByTime >= block.timestamp, InvalidUpgradeByTime());
```
ypatil12, line 37: Seems impractical to update at the current timestamp?

src/contracts/core/ReleaseManager.sol
0xClandestine: [be5e280](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/be5e280702c7ea3f4369144175da69deb015d6d5)

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,85 @@
...
+
+        return releases[releases.length - 1];
+    }
+
+    /// @inheritdoc IReleaseManager
+    function getLatestUpgradeByTime(
+        OperatorSet memory operatorSet
+    ) external view returns (uint256) {
+        return getLatestRelease(operatorSet).upgradeByTime;
+    }
```
0xrajath: Ok. I just realised these are the getters we need:
1. `function getTotalReleases(OperatorSet memory operatorSet) external view returns (uint256)`
2. `function getRelease(OperatorSet memory operatorSet, uint256 releaseId) external view returns (Release memory, bool isValid)`: We should also return a bool saying if the release is valid or not. Use function 5 below.
3. `function getLatestRelease(OperatorSet memory operatorSet) external view returns (Release memory, uint256 releaseId)`: We should return the `releaseId` as well here IMO.
4. `function getLatestUpgradeByTime(OperatorSet memory operatorSet) external view returns (uint32 upgradeByTime)`
5. `function isValidRelease(OperatorSet memory operatorSet, uint256 releaseId) external view returns (bool isValid)`

0xrajath: `isValidRelease` should work like: A getter that tells if a release is valid or not based on if there are newer releases ahead of it and if the timestamp is behind or ahead of the latest release's `upgradeByTime`.

0xClandestine: [f4ed5e7](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/f4ed5e79036ea1ef7d4e8e7cc3006203cdfe93b2)

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,90 @@
...
+
+    /// @inheritdoc IReleaseManager
+    function isLatestRelease(OperatorSet memory operatorSet, uint256 releaseId) external view returns (bool) {
```
0xrajath: We should be explicit about what this function is for. It should be `isValidRelease` IMO.

0xrajath: And we should use this in `getRelease` as well and return a bool too (In addition to `Release`)

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,90 @@
...
+
+    /// @inheritdoc IReleaseManager
+    function getRelease(OperatorSet memory operatorSet, uint256 releaseId) external view returns (Release memory) {
+        return _operatorSetReleases[operatorSet.key()][releaseId];
+    }
+
+    /// @inheritdoc IReleaseManager
+    function getLatestRelease(
+        OperatorSet memory operatorSet
+    ) public view returns (Release memory) {
```
0xrajath: We should return the `releaseId` too here IMO.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,90 @@
...
+
+    /// @inheritdoc IReleaseManager
+    function getTotalReleases(
+        OperatorSet memory operatorSet
+    ) public view returns (uint256) {
+        return _operatorSetReleases[operatorSet.key()].length;
+    }
+
+    /// @inheritdoc IReleaseManager
+    function getRelease(OperatorSet memory operatorSet, uint256 releaseId) external view returns (Release memory) {
```
0xrajath, line 66: We should call `isValidRelease` here and also return a bool telling if the specific release is valid or not.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,90 @@
...
+    ) public view returns (Release memory) {
+        Release[] storage releases = _operatorSetReleases[operatorSet.key()];
+
+        return releases[releases.length - 1];
+    }
+
+    /// @inheritdoc IReleaseManager
+    function getLatestUpgradeByTime(
+        OperatorSet memory operatorSet
+    ) external view returns (uint256) {
```
0xrajath: Should we return `uint32` vs `uint256` here?

src/contracts/core/ReleaseManager.sol
0xClandestine: [5aa2b22](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/5aa2b22ff522b6e3ff3a436bc3ba17de9e420771)

src/contracts/core/ReleaseManager.sol
0xClandestine: 0e15d38

src/contracts/core/ReleaseManager.sol
0xClandestine: 0e15d38

src/contracts/core/ReleaseManager.sol
0xClandestine: [4ff2213](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/4ff2213a9a35a64a30b5e8029ddf4832adbf2f6e)

src/contracts/core/ReleaseManager.sol
0xrajath: We need: `function getRelease(OperatorSet memory operatorSet, uint256 releaseId) external view returns (Release memory, bool)`

src/contracts/core/ReleaseManagerStorage.sol
```diff
@@ -0,0 +1,14 @@
...
+import "../interfaces/IReleaseManager.sol";
+import "../interfaces/IAllocationManager.sol";
+
+abstract contract ReleaseManagerStorage is IReleaseManager {
+    // Mutables
+
+    /// @notice Returns an array of releases for a given operator set.
+    mapping(bytes32 operatorSetKey => Release[]) internal _operatorSetReleases;
+
+    uint256[49] private __gap;
```
ypatil12, line 18: Can we add the OZ link that we have in other contracts for the GAP reason?

src/contracts/core/ReleaseManager.sol
0xClandestine: Not needed imo, can easily deduce this via other getters. Also returning tuples for everything is really annoying syntax imo.

ypatil12 (commented):
Also think we should move this to a `permissions` or `avs` folder.

src/contracts/core/ReleaseManager.sol
0xClandestine: Setting `updateByTime` to the current block would communicate urgency to operators.

src/contracts/core/ReleaseManagerStorage.sol
0xClandestine: [e2e661b](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/e2e661b4b3935313c7d1edd543a9ce707102aff9)

src/contracts/core/ReleaseManager.sol
0xClandestine: [e2e661b](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/e2e661b4b3935313c7d1edd543a9ce707102aff9)

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,92 @@
...
+     *
+     */
+
+    /// @inheritdoc IReleaseManager
+    function publishRelease(
+        OperatorSet calldata operatorSet,
+        Release calldata release
+    ) external checkCanCall(operatorSet.avs) returns (uint256 releaseId) {
+        Release[] storage releases = _operatorSetReleases[operatorSet.key()];
+
```
ypatil12, line 36: What happens if multiple releases are published in the same block with the same upgradeByTime? We just use latest tx there?

src/contracts/core/ReleaseManager.sol
0xrajath: What if they wanted to do an instant upgrade for whatever reason?

src/contracts/core/ReleaseManager.sol
0xClandestine: Yeah, the idea is that this contract simply communicates the latest release to operators with a suggested time to upgrade to it by. Upgrade times aren't enforced at all, on or off-chain.

docs/core/ReleaseManager.md
```diff
@@ -0,0 +1,235 @@
...
+
+| File | Notes |
+| -------- | -------- |
+| [`PermissionControllerMixin.sol`](../../src/contracts/mixins/PermissionControllerMixin.sol) | account delegation |
+| [`SemVerMixin.sol`](../../src/contracts/mixins/SemVerMixin.sol) | semantic versioning |
+| [`OperatorSetLib.sol`](../../src/contracts/libraries/OperatorSetLib.sol) | encode/decode operator sets |
+
+## Overview
+
+The `ReleaseManager` manages software releases for AVS operator sets. It provides a standardized way for AVSs to publish software artifacts (binaries, docker images, etc.) that operators in their operator sets should upgrade to by specified deadlines. This ensures operators run compatible and up-to-date software versions required by the AVS.
```
bdchatham, line 19: Could you add image index to be first in the list of examples? That's the preferred type so think it's worth calling out.

docs/core/ReleaseManager.md
```diff
@@ -0,0 +1,235 @@
...
+    Release calldata release
+) 
+    external 
+    checkCanCall(operatorSet.avs) 
+    returns (uint256 releaseId)
+```
+
+_Note: this method can be called directly by an AVS, or by a caller authorized by the AVS. See [`PermissionController.md`](../permissions/PermissionController.md) for details._
+
+AVSs use this method to publish new software releases for their operator sets. Each release contains one or more artifacts that represent the software components operators must run (e.g., validator clients, network monitors, etc.). The AVS specifies a deadline (`upgradeByTime`) by which all operators in the operator set must upgrade to the new release.
```
bdchatham, line 108: Can we explicit call out the units of upgradeByTime. I know it's implied by being a smart contract but just for completeness and making it abundantly clear.

src/contracts/core/ReleaseManager.sol
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
+import "../mixins/PermissionControllerMixin.sol";
+import "../mixins/SemVerMixin.sol";
```
bdchatham, line 6: For my understanding, what is this doing? Versioning for the contract?

src/contracts/core/ReleaseManager.sol
bdchatham: Without the bool, I assume nil is returned if none exists?

src/contracts/core/ReleaseManager.sol
ypatil12: Can we add this to docs to clarify?

0xClandestine: [dbb9eb8](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/dbb9eb801031e794f6a3ce87373b775b7e048508)

src/contracts/core/ReleaseManager.sol
0xClandestine: Yeah we use it in all core contracts to track the contract versions, unrelated to RMS.

docs/core/ReleaseManager.md
0xClandestine: [6e2f21e](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/6e2f21e7051205f32efac43083284171b1901c84)

docs/core/ReleaseManager.md
0xClandestine: [6e2f21e](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1469/commits/6e2f21e7051205f32efac43083284171b1901c84)

ypatil12 (approved):

bdchatham (approved):

PR #1470, opened 2025-06-16
feat: ecdsa cert verifier

eigenmikem:
**Motivation:**

Singleton Certificate Verifier is needed for ECDSA signatures.

**Modifications:**

New core multichain contract, `ECDSACertificateVerifier`.

**Result:**

Certificate Verifiers for initial release complete.


src/contracts/multichain/ECDSACertificateVerifierStorage.sol
```diff
@@ -0,0 +1,46 @@
...
+    mapping(bytes32 => address) internal _operatorSetOwners;
+
+    /// @dev Mapping from operatorSet key to maximum staleness period
+    mapping(bytes32 => uint32) internal _maxStalenessPeriods;
+
+    /// @dev Mapping from operatorSet key to latest reference timestamp
+    mapping(bytes32 => uint32) internal _latestReferenceTimestamps;
+
+    /// @dev Mapping from referenceTimestamp to the number of operators
+    mapping(bytes32 operatorSetKey => mapping(uint32 referenceTimestamp => uint256 numOperators)) internal _numOperators;
```
ypatil12, line 41: Forget why we need this?

src/contracts/multichain/ECDSACertificateVerifier.sol
```diff
@@ -0,0 +1,316 @@
...
+/**
+ * @title ECDSACertificateVerifier
+ * @notice Verifies ECDSA certificates across multiple operator sets
+ * @dev Implements ECDSA signature verification with operator information caching
+ */
+contract ECDSACertificateVerifier is Initializable, ECDSACertificateVerifierStorage, SignatureUtilsMixin {
+    using ECDSA for bytes32;
+
+    // EIP-712 type hash for certificate verification
+    bytes32 public constant ECDSA_CERTIFICATE_TYPEHASH =
```
ypatil12: Nit: keep in storage contract

src/contracts/multichain/ECDSACertificateVerifier.sol
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    /**
+     * @notice Constructor for the certificate verifier
+     * @dev Disables initializers to prevent implementation initialization
+     * @param _operatorTableUpdater Address authorized to update operator tables
+     */
+    constructor(
+        IOperatorTableUpdater _operatorTableUpdater
+    ) ECDSACertificateVerifierStorage(_operatorTableUpdater) SignatureUtilsMixin("1.0.0") {
```
ypatil12: Pass in version for `SignatureUtilsMixin` as a string parameter

src/contracts/multichain/ECDSACertificateVerifier.sol
```diff
@@ -0,0 +1,316 @@
...
+        emit TableUpdated(operatorSet, referenceTimestamp, operatorInfos);
+    }
+
+    ///@inheritdoc IECDSACertificateVerifier
+    function verifyCertificate(
+        OperatorSet calldata operatorSet,
+        ECDSACertificate calldata cert
+    ) external view returns (uint256[] memory) {
+        uint96[] memory signedStakes96 = _verifyECDSACertificate(operatorSet, cert);
+        uint256[] memory signedStakes = new uint256[](signedStakes96.length);
```
ypatil12: Lets used uint256 for signedStakes, instead of uint96

ypatil12: Makes it much easier to work with across the contract

src/contracts/multichain/ECDSACertificateVerifier.sol
```diff
@@ -0,0 +1,316 @@
...
+            uint256 threshold = (totalStakes[i] * totalStakeProportionThresholds[i]) / 10_000;
+            if (signedStakes[i] < threshold) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    ///@inheritdoc IECDSACertificateVerifier
+    function verifyCertificateNominal(
```
ypatil12: Do you think we should put the _math_ for nominal/proportional verification in a library? That way we don't have to repeat the code between ECDSA and BN254

src/contracts/multichain/ECDSACertificateVerifier.sol
```diff
@@ -0,0 +1,316 @@
...
+
+        require(validSignatures, VerificationFailed());
+
+        // Process each operator to check if they signed
+        uint256 operatorCount = _numOperators[operatorSetKey][cert.referenceTimestamp];
+        for (uint256 i = 0; i < operatorCount; i++) {
+            // Check if this operator is in the signers list
+            bool isSigner = false;
+            for (uint256 j = 0; j < signers.length; j++) {
+                if (_operatorInfos[operatorSetKey][cert.referenceTimestamp][uint32(i)].pubkey == signers[j]) {
```
ypatil12: Should we use an enumerable set for the pubkeys here? So we don't have to iterate

src/contracts/multichain/ECDSACertificateVerifier.sol
```diff
@@ -0,0 +1,316 @@
...
+                uint256[] storage weights = _operatorInfos[operatorSetKey][cert.referenceTimestamp][uint32(i)].weights;
+                for (uint256 j = 0; j < weights.length && j < signedStakes.length; j++) {
+                    signedStakes[j] += uint96(weights[j]);
+                }
+            }
+        }
+
+        // After processing, check if all signed stakes are zero
+        bool anySigned = false;
+        for (uint256 i = 0; i < signedStakes.length; i++) {
```
ypatil12, line 195: Why is this check needed?

src/contracts/multichain/ECDSACertificateVerifier.sol
```diff
@@ -0,0 +1,316 @@
...
+
+        // Assert that the reference timestamp exists
+        require(_latestReferenceTimestamps[operatorSetKey] == cert.referenceTimestamp, ReferenceTimestampDoesNotExist());
+
+        // Get the total stakes
+        uint96[] memory totalStakes = _getTotalStakes(operatorSet, cert.referenceTimestamp);
+        uint96[] memory signedStakes = new uint96[](totalStakes.length);
+
+        // Compute the EIP-712 digest for signature recovery
+        bytes32 structHash =
```
ypatil12: Let's just use a public view helper function for this? So offchain can also calculate

src/contracts/multichain/ECDSACertificateVerifier.sol
```diff
@@ -0,0 +1,306 @@
...
+            keccak256(abi.encode(ECDSA_CERTIFICATE_TYPEHASH, cert.referenceTimestamp, cert.messageHash));
+        bytes32 signableDigest = keccak256(abi.encodePacked("\x19\x01", domainSeparator(), structHash));
```
ypatil12: We may not be able to use this `domainSeparator`, since it has a chainID. Need to figure out if hourglass wants to have different certificates for the same task on different chains

src/contracts/multichain/ECDSACertificateVerifier.sol
```diff
@@ -0,0 +1,306 @@
...
+
+            // Recover the signer
+            address signer = messageHash.recover(signature);
+
+            // If any signature is invalid (returns address(0)), the whole certificate is invalid
+            if (signer == address(0)) {
+                return (signers, false);
+            }
+
+            // Check for duplicate signers
```
ypatil12: Can we just require that the signatures are ordered based on the signer address? Then don't need another iteration

src/contracts/multichain/ECDSACertificateVerifier.sol
```diff
@@ -0,0 +1,306 @@
...
+        signers = new address[](signatureCount);
+
+        for (uint256 i = 0; i < signatureCount; i++) {
+            bytes memory signature = new bytes(65);
+            for (uint256 j = 0; j < 65; j++) {
+                signature[j] = signatures[i * 65 + j];
+            }
+
+            // Recover the signer
+            address signer = messageHash.recover(signature);
```
ypatil12: We should add a comment that we do NOT support smart contract based signatures for multichain

src/contracts/multichain/ECDSACertificateVerifier.sol
ypatil12: Let's have a different domainSeparator here, just for this contract. Can also remove `SigantureUtilsMixin` inheritance

src/contracts/multichain/ECDSACertificateVerifier.sol
ypatil12: Nvm, don't need with new domain separator pathway

src/contracts/multichain/ECDSACertificateVerifier.sol
ypatil12: Or... we could just support it?

src/contracts/multichain/ECDSACertificateVerifierStorage.sol
eigenmikem: Copying arrays of custom types into storage causes lots of issues, easier to just store in indexed mapping but then need this

src/contracts/multichain/ECDSACertificateVerifier.sol
eigenmikem: Can do that - amount of repeated code is pretty minimal though

ypatil12 (approved):
LGTM

src/contracts/multichain/ECDSACertificateVerifier.sol
ypatil12: Can do in a different PR

PR #1473, opened 2025-06-18
feat: ecdsa table calculator

eigenmikem:
**Motivation:**

Multichain needs a table calculator base implementation for ECDSA signtures.

**Modifications:**

New contracts, `ECDSATableCalculator` and `ECDSATableCalculatorBase`.

**Result:**

Multichain contracts complete

### DIFF
--- a/src/contracts/multichain/ECDSATableCalculatorBase.sol
+++ b/src/contracts/multichain/ECDSATableCalculatorBase.sol
@@ -93,22 +93,9 @@ abstract contract ECDSATableCalculatorBase is IECDSATableCalculator {
             return new ECDSAOperatorInfo[](0);
         }
 
-        // Count operators with registered ECDSA keys
+        // Create the operator infos array with maximum possible size
+        operatorInfos = new ECDSAOperatorInfo[](operators.length);
         uint256 operatorCount = 0;
-        for (uint256 i = 0; i < operators.length; i++) {
-            if (keyRegistrar.isRegistered(operatorSet, operators[i])) {
-                operatorCount++;
-            }
-        }
-
-        // If no operators have registered keys, return empty array
-        if (operatorCount == 0) {
-            return new ECDSAOperatorInfo[](0);
-        }
-
-        // Create the operator infos array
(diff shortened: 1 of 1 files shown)


src/contracts/multichain/ECDSATableCalculatorBase.sol
```diff
@@ -0,0 +1,130 @@
...
+
+        // If there are no weights, return an empty array
+        if (weights.length == 0) {
+            return new ECDSAOperatorInfo[](0);
+        }
+
+        // Count operators with registered ECDSA keys
+        uint256 operatorCount = 0;
+        for (uint256 i = 0; i < operators.length; i++) {
+            if (keyRegistrar.isRegistered(operatorSet, operators[i])) {
```
ypatil12: Instead of having two calls to `keyRegistrar`, we can just do what we do in `Bn254TableCalculatorBase`? where we increment in the main loop and then resize?

ypatil12: Generally want to be consistent between both

src/contracts/multichain/ECDSATableCalculatorBase.sol
```diff
@@ -0,0 +1,130 @@
...
+        uint256 infoIndex = 0;
+
+        for (uint256 i = 0; i < operators.length; i++) {
+            // Skip if the operator has not registered their ECDSA key
+            if (!keyRegistrar.isRegistered(operatorSet, operators[i])) {
+                continue;
+            }
+
+            // Get the ECDSA address (public key) for the operator
+            address ecdsaAddress = keyRegistrar.getECDSAAddress(operatorSet, operators[i]);
```
ypatil12, line 107: Don't need the above call right? If ecdsaAddress is 0, we know operator hasn't registered key

ypatil12: I guess we do the same in Bn254... should be consistent too

ypatil12 (approved):

PR #1474, opened 2025-06-18
feat: multichain deployments

0xClandestine:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/script/utils/CrosschainDeployLib.sol
+++ b/script/utils/CrosschainDeployLib.sol
@@ -54,7 +54,7 @@ library CrosschainDeployLib {
     /// @return The encoded salt.
     /// Deployer EOA (20 bytes) | Cross-chain flag (1 byte) | Entropy (11 bytes)
     /// 0xbebebebebebebebebebebebebebebebebebebebe|ff|1212121212121212121212
-    function computeSalt(address eoa, bytes11 salt) internal view returns (bytes32) {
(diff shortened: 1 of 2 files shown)


script/utils/CrosschainDeployLib.sol
```diff
@@ -0,0 +1,141 @@
...
+        return createx.computeCreate2Address(computeProtectedSalt(deployer, salt), initCodeHash);
+    }
+
+    /*
+     * @notice Returns the predicted address of a `TransparentUpgradeableProxy` deployed with CreateX.
+     */
+    function computeCrosschainUpgradeableProxyAddress(
+        address implementation,
+        address admin,
+        bytes11 salt
```
ypatil12, line 147: TODO: add deterministic salt per contract name. Discussed using FFI offline

script/utils/CrosschainDeployLib.sol
```diff
@@ -0,0 +1,141 @@
...
+    function deployCreate2(bytes32 salt, bytes memory initCode) external payable returns (address newContract);
+    function computeCreate2Address(
+        bytes32 salt,
+        bytes32 initCodeHash
+    ) external view returns (address computedAddress);
+}
+
+bytes11 constant EMPTY_CONTRACT_SALT = bytes11(uint88(0xffffffffffffffffffffff));
+
+library CrosschainDeployLib {
```
ypatil12, line 20: Let's move this to `script/releases`

PR #1480, opened 2025-06-23
refactor: operators can deregister keys if not slashable

eigenmikem:
**Motivation:**

Operators should be able to deregister their keys in the `KeyRegistrar` if they aren't slashable

**Modifications:**

Changed permissions on `deregisterKey`

**Result:**

Operators can deregister their keys in the `KeyRegistrar` if they aren't slashable

### DIFF
--- a/src/contracts/permissions/KeyRegistrar.sol
+++ b/src/contracts/permissions/KeyRegistrar.sol
@@ -89,10 +89,7 @@ contract KeyRegistrar is KeyRegistrarStorage, PermissionControllerMixin, Signatu
     }
 
     /// @inheritdoc IKeyRegistrar
-    function deregisterKey(address operator, OperatorSet memory operatorSet) external {
-        // Only the operator or authorized addresses can deregister keys
-        require(_checkCanCall(operator), InvalidPermissions());
-
+    function deregisterKey(address operator, OperatorSet memory operatorSet) external checkCanCall(operator) {
         // Operators can only deregister if they are not slashable for this operator set
         require(
             !allocationManager.isOperatorSlashable(operator, operatorSet), OperatorStillSlashable(operatorSet, operator)


src/contracts/interfaces/IKeyRegistrar.sol
```diff
@@ -69,8 +70,10 @@ interface IKeyRegistrar is IKeyRegistrarErrors, IKeyRegistrarEvents, ISemVerMixi
      * @notice Deregisters a cryptographic key for an operator with a specific operator set
      * @param operator Address of the operator to deregister key for
      * @param operatorSet The operator set to deregister the key from
-     * @dev Can be called by avs directly or by addresses they've authorized via PermissionController
+     * @dev Can be called by the operator directly or by addresses they've authorized via PermissionController
+     * @dev Can be called by the AVS directly or by addresses they've authorized via PermissionController
```
ypatil12: Let's not allow the AVS to call the function

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -89,7 +90,19 @@ contract KeyRegistrar is KeyRegistrarStorage, PermissionControllerMixin, Signatu
 
     /// @inheritdoc IKeyRegistrar
     function deregisterKey(address operator, OperatorSet memory operatorSet) external {
-        require(address(allocationManager.getAVSRegistrar(operatorSet.avs)) == msg.sender, InvalidPermissions());
+        // Check if the caller is the AVS registrar (AVS-initiated deregistration)
+        bool isAVSCall = address(allocationManager.getAVSRegistrar(operatorSet.avs)) == msg.sender;
```
ypatil12: No need for AVS to call, only operator should be able to deregister a key

ypatil12: AVS can already force eject an operator in the `AllocationManager`

src/contracts/permissions/KeyRegistrar.sol
```diff
@@ -89,7 +90,13 @@ contract KeyRegistrar is KeyRegistrarStorage, PermissionControllerMixin, Signatu
 
     /// @inheritdoc IKeyRegistrar
     function deregisterKey(address operator, OperatorSet memory operatorSet) external {
-        require(address(allocationManager.getAVSRegistrar(operatorSet.avs)) == msg.sender, InvalidPermissions());
+        // Only the operator or authorized addresses can deregister keys
+        require(_checkCanCall(operator), InvalidPermissions());
```
ypatil12: Can use modifier, like in above function?

PR #1483, opened 2025-06-24
refactor: remove redistribution delay

0xClandestine:
**Motivation:**

**Modifications:**

**Result:**


src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -13,7 +12,6 @@ import "./SlashEscrowFactoryStorage.sol";
 contract SlashEscrowFactory is
     Initializable,
     SlashEscrowFactoryStorage,
-    OwnableUpgradeable,
```
ypatil12, line 16: This will break testnet/preprod storage?

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -188,25 +168,6 @@ contract SlashEscrowFactory is
...
-     */
-
-    /// @inheritdoc ISlashEscrowFactory
-    function setGlobalEscrowDelay(
-        uint32 delay
-    ) external onlyOwner {
-        _setGlobalEscrowDelay(delay);
-    }
-
-    /// @inheritdoc ISlashEscrowFactory
```
ypatil12, line 204: Don't we still need a strategy delay for EIGEN?

src/contracts/core/SlashEscrowFactory.sol
```diff
@@ -89,20 +83,6 @@ contract SlashEscrowFactory is
         // Add the strategy to the pending strategies for the slash ID.
```
ypatil12, line 83: Do we still need a slash escrow clone contract? App developers should be able to transfer assets out via one call to the `StrategyManager` now?

0xClandestine:
Closed in favor of #1485

PR #1492, opened 2025-06-27
feat: add `publishMetadataURI`

0xClandestine:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/src/contracts/core/ReleaseManager.sol
+++ b/src/contracts/core/ReleaseManager.sol
@@ -49,6 +49,14 @@ contract ReleaseManager is Initializable, ReleaseManagerStorage, PermissionContr
         emit ReleasePublished(operatorSet, releaseId, release);
     }
 
+    /// @inheritdoc IReleaseManager
+    function publishMetadataURI(
+        OperatorSet calldata operatorSet,
+        string calldata metadataURI
+    ) external checkCanCall(operatorSet.avs) {
+        emit MetadataURIPublished(operatorSet, metadataURI);
+    }
+
     /**
      *
      *                         VIEW FUNCTIONS
--- a/src/contracts/core/ReleaseManager.sol
+++ b/src/contracts/core/ReleaseManager.sol
@@ -54,6 +54,7 @@ contract ReleaseManager is Initializable, ReleaseManagerStorage, PermissionContr
         OperatorSet calldata operatorSet,
         string calldata metadataURI
     ) external checkCanCall(operatorSet.avs) {
+        _operatorSetMetadataURI[operatorSet.key()] = metadataURI;
         emit MetadataURIPublished(operatorSet, metadataURI);
     }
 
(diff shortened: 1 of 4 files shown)


src/contracts/core/ReleaseManager.sol
```diff
@@ -49,6 +49,14 @@ contract ReleaseManager is Initializable, ReleaseManagerStorage, PermissionContr
         emit ReleasePublished(operatorSet, releaseId, release);
     }
 
+    /// @inheritdoc IReleaseManager
+    function publishMetadataURI(
+        OperatorSet calldata operatorSet,
+        string calldata metadataURI
+    ) external checkCanCall(operatorSet.avs) {
+        emit MetadataURIPublished(operatorSet, metadataURI);
```
0xrajath, line 60: @bdchatham @0xClandestine We should store the MetadataURI in storage as well right? And have a getter for it? Or is the intention that it will only be picked up by an indexer?

0xrajath: I think we should have it in storage too rather than just emitting an event.

0xClandestine: Worth noting we already do this elsewhere in core, emitting a string and not storing. AVSD iirc.

bdchatham: I agree, I think we should store it

src/contracts/core/ReleaseManager.sol
```diff
@@ -49,6 +49,14 @@ contract ReleaseManager is Initializable, ReleaseManagerStorage, PermissionContr
         emit ReleasePublished(operatorSet, releaseId, release);
     }
 
+    /// @inheritdoc IReleaseManager
+    function publishMetadataURI(
```
0xrajath, line 54: Also I think we should a have a require check that the metadata has been set for the OperatorSet before an AVS can publishRelease for that OperatorSet (i.e require check in `publishRelease`)

0xrajath: It'll force AVSs to release some metadata that the operators can reference. Otherwise AVS's will just ignore this step.

src/contracts/core/ReleaseManager.sol
0xClandestine: [ef1ca6f](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1492/commits/ef1ca6f81aebd9d7f3ce1227614607ae9e3b7773)

src/contracts/core/ReleaseManager.sol
0xClandestine: [d1ddece](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1492/commits/d1ddece703732c08311483ca386810152ff0ec2d)

PR #1501, opened 2025-06-30
feat: operator table updater pauser

ypatil12:
**Motivation:**

Currently, `disableRoot` is called by the `owner` of the `OperatorTableCalculator`. We want to update such that it can be called by the `pauser`. This is to enable cleaner operations. 

**Modifications:**

- Add `Pausable` to `operatorTableCalculator`
- Guard `disableGlobalTableRoot` with `onlyPauser`
- Guard `updateGlobalTableRoot` and `updateOperatorTable` with `whenNotPaused(index)`

**Result:**

Cleaner disable mechanics.


src/contracts/multichain/OperatorTableUpdater.sol
```diff
@@ -5,10 +5,17 @@ import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
 import "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
 
 import "../libraries/Merkle.sol";
+import "../permissions/Pausable.sol";
```
nadir-akhtar, line 8: It's a bit strange to be using this library when it comes with irrelevant functionality -- would prefer a way with less "cruft", e.g. defining a specific role in storage and using a `require` statement to ensure only that role can call `disableRoot()`

That said, it may make sense to also add pausing guards across more functions (e.g. on `confirmGlobalTableRoot`) -- in which case having the `Pausable` dependency would make sense

ypatil12: Yeah, we should add it to `confirmGlobalTableRoot` too

ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1501/commits/d8e2f1a4b08481e0d2bb6f8186890bad6a46633d

PR #1502, opened 2025-06-30
feat: substitute calcSlashedAmount for scaleForBurning

nadir-akhtar:
**Motivation:**

**Modifications:**

**Result:**


src/contracts/libraries/SlashingLib.sol
```diff
@@ -178,6 +164,10 @@ library SlashingLib {
         uint256 prevMaxMagnitude,
         uint256 newMaxMagnitude
     ) internal pure returns (uint256) {
+        if (prevMaxMagnitude == 0) {
+            // TODO: consider throwing an error instead
+            return 0;
+        }
```
wadealexc: Copying my note from slack:

1. Why did we need to add this, given the existing code does not check for this? Is there something unique to combining the methods that I'm missing?
2. If we want to validate this, IMO we should keep the branch out of this library, and check explicitly in `DM.slashOperatorShares`. Where possible, this library's methods should be pure math expressions to make it easier to abstract away when reviewing.

Can add a comment to the library that it expects `prevMaxMag != 0` (and any other invariants we expect, like that `newMaxMag < prevMaxMag`, `operatorShares != 0`, etc)

src/contracts/core/DelegationManager.sol
```diff
@@ -793,8 +793,8 @@ contract DelegationManager is
         // less than or equal to MIN_WITHDRAWAL_DELAY_BLOCKS ago. These shares are still slashable.
         uint256 scaledSharesAdded = curQueuedScaledShares - prevQueuedScaledShares;
 
-        return SlashingLib.scaleForBurning({
-            scaledShares: scaledSharesAdded,
+        return SlashingLib.calcSlashedAmount({
+            operatorShares: scaledSharesAdded.mulWad(prevMaxMagnitude),
```
wadealexc, line 820: Since we're here, the ideal "most readable" change would be to move the call to `calcSlashedAmount` up to `slashOperatorShares`. Maybe something like this:

```solidity
    function slashOperatorShares(
        address operator,
        OperatorSet calldata operatorSet,
        uint256 slashId,
        IStrategy strategy,
        uint64 prevMaxMagnitude,
        uint64 newMaxMagnitude
    ) external onlyAllocationManager nonReentrant returns (uint256 totalDepositSharesToSlash) {
        uint256 operatorSharesSlashed = SlashingLib.calcSlashedAmount({
            operatorShares: operatorShares[operator][strategy],
            prevMaxMagnitude: prevMaxMagnitude,
            newMaxMagnitude: newMaxMagnitude
        });

        // Comment explaining conversion between scaled shares in queue and operator shares
        uint256 slashableQueuedShares = _getSlashableSharesInQueue(operator, strategy);
        uint256 queuedSharesSlashed = SlashingLib.calcSlashedAmount({
            operatorShares: slashableQueuedShares.mulWad(prevMaxMagnitude),
            prevMaxMagnitude: prevMaxMagnitude,
            newMaxMagnitude: newMaxMagnitude
        });
```

src/contracts/libraries/SlashingLib.sol
ypatil12: Agreed. Confused on why we have to add this?

src/contracts/core/DelegationManager.sol
```diff
@@ -793,8 +793,8 @@ contract DelegationManager is
         // less than or equal to MIN_WITHDRAWAL_DELAY_BLOCKS ago. These shares are still slashable.
         uint256 scaledSharesAdded = curQueuedScaledShares - prevQueuedScaledShares;
 
-        return SlashingLib.scaleForBurning({
-            scaledShares: scaledSharesAdded,
+        return SlashingLib.calcSlashedAmount({
```
ypatil12, line 819: Math behind this. Might be worth adding a dev comment? 

**Variables** 
- $s_n$ - The amount of shares in the storage of the {EP,S}M at time n 
- $k_n$ - The “staker deposit scaling factor” at time n. This is initialized to 1. 
- $m_n$ - The operator magnitude at time n 
- $op_{n}$ - The shares of the operator $op_{n}=s_{n}k_{n}m_{n}$ 
- $ss_{n}$ - The scaled shares of the operator in storage. This is stored upon queueing of a withdrawal $ss_{n}=s_{n}k_{n}$. See `scaleForQueueWithdrawal` in `SlashingLib`

**Scale For Burning**
$ss_{n}(m_{n}-m_{n+1})$ 
= $s_nk_n(m_{n}-m_{n+1})$ 
= $s_nk_nm_{n}- s_nk_nm_{n+1}$ 
= $op_n - s_nk_nm_{n+1}$

**Calc Slashed Amount**

$op_{n} - op_{n}*\frac{m_{n+1}}{m_n}$ 

$op_{n} - s_nk_nm_n*\frac{m_{n+1}}{m_n}$ 

$op_{n} - s_nk_n{m_{n+1}}$

src/contracts/core/DelegationManager.sol
ypatil12: We probably should call this something different? These really just are `operatorSharedSlashedFromQueue`

ypatil12 (commented):
Can we change the base to `release-dev/slashing-ux-improvements`? Also need to run `make bindings` after.

src/contracts/libraries/SlashingLib.sol
nadir-akhtar: In `scaleForBurning`, magnitudes are subtracted from each other, and their difference is multiplied against scaled shares. If `prevMaxMagnitude` is 0, `newMaxMagnitude` is implied also to be 0 (since it must be <= `prevMaxMagnitude` since magnitude decreases with slashes), so this would lead to multiplying by 0.

When using `calcSlashedAmount`, since we now divide by `prevMaxMagnitude`, if `prevMaxMagnitude` is 0 then this leads to a divide-by-zero error, which caused a couple of tests to fail, motivating this branch. I think it's reasonable to include this statement here, but we can also check for it where the function is being called in the DM.

nadir-akhtar: Tests in question are:
* `test_sharesBurnableAtMinDelayBlocks`
* `test_slashableSharesInQueue`

If you make the change in the DelegationManager above without adding this statement, both these tests will fail with `[FAIL: panic: division or modulo by zero (0x12)]`. Adding this branching statement resolves that.

src/contracts/core/DelegationManager.sol
nadir-akhtar: Good point! Will make that change

src/contracts/core/DelegationManager.sol
nadir-akhtar: Appreciate the math. Agreed

src/contracts/libraries/SlashingLib.sol
ypatil12: Gotcha. This makes sense. Should be fine to not throw given that magnitude can eventually be slashed down to 0.

src/contracts/core/DelegationManager.sol
nadir-akhtar: [8a664fc](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1502/commits/8a664fce340702a91db0bab2336e789414e568fa)

src/contracts/core/DelegationManager.sol
nadir-akhtar: [8a664fc](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1502/commits/8a664fce340702a91db0bab2336e789414e568fa)

ypatil12 (approved):

PR #1518, opened 2025-07-07
refactor: generator table no longer special cased

eigenmikem:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/docs/multichain/destination/OperatorTableUpdater.md
+++ b/docs/multichain/destination/OperatorTableUpdater.md
@@ -25,7 +25,7 @@ The following values are set upon initialization:
 
 * `generator` is an EigenLabs-run entity that signs off on `globalTableRoots`. The operatorSet is of size 1. 
 * `globalRootConfirmationThreshold`: 10000. The threshold in basis points required for global root confirmation. Since the operatorSet is of size 1 a single signature is needed.
-* `referenceTimestamp`: A past timestamp at which the `generator` is set. We hardcode this value to 1 upon initialization. This value is also the `latestReferenceTimestamp`. Once roots are updated, the `latestReferenceTimestamp` will increase. *Note: the reference timestamp for the `generator`, given by `operatorTableUpdater.getGeneratorReferenceTimestamp` will remain at the initialization value unless the generator's operator table is updated through the standard flow*.
+* `referenceTimestamp`: A past timestamp at which the `generator` is set. This value is also the `latestReferenceTimestamp`. Once roots are updated, the `latestReferenceTimestamp` will increase. *Note: the reference timestamp for the `generator`, given by `operatorTableUpdater.getGeneratorReferenceTimestamp` will remain at the initialization value unless the generator's operator table is updated through the standard flow*.
 * `generatorInfo`: The key material needed to verify certificates of the `generator`
 * `operatorSetConfig`: A configuration for the `generator` 
     * `maxStalenessPeriod`: 0. Set to zero to confirm `globalTableRoots` without updating the `generator` operatorSet. See [`CertificateVerifier`](./CertificateVerifier.md#overview) for specifics`OperatorTableUpdater`. It is the same across all destination chains, even for destination chains that are supported after the initial deployment. 


src/contracts/multichain/OperatorTableUpdater.sol
```diff
@@ -38,42 +38,51 @@ contract OperatorTableUpdater is
...
+        // Create merkle tree with generator as first leaf and dummy data for additional leaves
+        bytes32[] memory leaves = new bytes32[](2);
+        leaves[0] = generatorLeafHash;
+        leaves[1] = bytes32(uint256(1));
+        bytes32 initialGlobalTableRoot = Merkle.merkleizeKeccak(leaves);
+
+        // Set the initial global table root first, before updating the generator
+        _globalTableRoots[referenceTimestamp] = initialGlobalTableRoot;
+        _isRootValid[initialGlobalTableRoot] = true;
         _referenceBlockNumbers[referenceTimestamp] = uint32(block.number);
```
ypatil12, line 75: I think we can remove setting the ref block number and timestamp, right?

src/contracts/multichain/OperatorTableUpdater.sol
```diff
@@ -38,42 +38,51 @@ contract OperatorTableUpdater is
...
         _referenceTimestamps[uint32(block.number)] = referenceTimestamp;
 
         // Set the latest reference timestamp
         _latestReferenceTimestamp = referenceTimestamp;
 
+        // Now update the generator's operator table
+        _updateGenerator(referenceTimestamp, generatorInfo, generatorConfig);
```
ypatil12, line 82: Do we need an internal func if it's only called in one place?

src/contracts/multichain/OperatorTableUpdater.sol
```diff
@@ -38,42 +38,51 @@ contract OperatorTableUpdater is
...
-     * @dev We also update the operator table for the global root confirmer set, to begin signing off on global roots
-     * @dev Uses INITIAL_GLOBAL_TABLE_ROOT constant to break circular dependency for certificate verification
      */
     function initialize(
         address owner,
         uint256 initialPausedStatus,
-        OperatorSet calldata _generator,
+        OperatorSet calldata generator,
         uint16 _globalRootConfirmationThreshold,
         uint32 referenceTimestamp,
```
ypatil12, line 52: Possible to not pass in a reference timestamp here? Just use 0?

ypatil12 (commented):
First pass

src/contracts/multichain/OperatorTableUpdater.sol
eigenmikem: Just trying to be conservative, but technically they are unused

src/contracts/multichain/OperatorTableUpdater.sol
eigenmikem: Consistency some other internals but happy to remove

src/contracts/multichain/OperatorTableUpdater.sol
eigenmikem: Might as well pass in a real timestamp imo unless it is annoying for deployment

src/contracts/multichain/OperatorTableUpdater.sol
ypatil12: Yeah I think we should remove the block number unless we actually calculate it off chain and pass it in, then its fine

src/contracts/multichain/OperatorTableUpdater.sol
ypatil12: Yeah I think we should remove the block number unless we actually calculate it off chain and pass it in, then its fine

src/contracts/multichain/OperatorTableUpdater.sol
ypatil12: Fair, then we should also do a real block number imo

ypatil12:
Can we also update the documentation, specifically around ghost operator set

docs/multichain/destination/OperatorTableUpdater.md
```diff
@@ -19,13 +19,13 @@ The `OperatorTableUpdater` is responsible for updating the `GlobalTableRoot` and
...
 
 * `generator` is an EigenLabs-run entity that signs off on `globalTableRoots`. The operatorSet is of size 1. 
 * `globalRootConfirmationThreshold`: 10000. The threshold in basis points required for global root confirmation. Since the operatorSet is of size 1 a single signature is needed.
-* `referenceTimestamp`: A past timestamp at which the `generator` is set. We hardcode this value to 1 upon initialization. This value is also the `latestReferenceTimestamp`. Once roots are updated, the `latestReferenceTimestamp` will increase. *Note: the reference timestamp for the `generator`, given by `operatorTableUpdater.getGeneratorReferenceTimestamp` will remain 1 unless [`updateGenerator`](#updategenerator) is called*.
+* `referenceTimestamp`: A past timestamp at which the `generator` is set. We hardcode this value to 1 upon initialization. This value is also the `latestReferenceTimestamp`. Once roots are updated, the `latestReferenceTimestamp` will increase. *Note: the reference timestamp for the `generator`, given by `operatorTableUpdater.getGeneratorReferenceTimestamp` will remain at the initialization value unless the generator's operator table is updated through the standard flow*.
```
ypatil12: Aren't we using a real timestamp instead of 1 now?

eigenmikem: good catch

docs/multichain/destination/OperatorTableUpdater.md
```diff
@@ -19,13 +19,13 @@ The `OperatorTableUpdater` is responsible for updating the `GlobalTableRoot` and
 The contract supports both BN254 and ECDSA operator tables and routes updates to the appropriate certificate verifier based on the curve type.
 
 ## Parameterization
-Upon initialization, the `generator` is updated. The `generator` is represented in storage as an operatorSet. The `generator` should be considered a ghost-operatorSet` since it does not exist in the core protocol, does not have stake backing it, and is not transported to other chains via the multichain protocol. It can only be updated upon initialization or by a [privileged role](#updategenerator). This entity is the same across all destination chains. 
+Upon initialization, the `generator` is set and an initial global table root is calculated. The `generator` is represented in storage as an operatorSet. The `generator` should be considered a "ghost-operatorSet" since it does not exist in the core protocol, does not have stake backing it, and is not transported to other chains via the multichain protocol. It can only be set upon initialization and cannot be changed afterward. This entity is the same across all destination chains.
```
ypatil12, line 22: Technically the generator is now transported to other chains via the multichain protocol

src/contracts/interfaces/IOperatorTableUpdater.sol
```diff
@@ -82,16 +76,6 @@ interface IOperatorTableUpdater is
         uint32 referenceBlockNumber
     ) external;
 
-    /**
-     * @notice Set the operatorSet which certifies against global roots
-     * @param operatorSet the operatorSet which certifies against global roots
-     * @dev The `operatorSet` is used to verify the certificate of the global table root
-     * @dev Only callable by the owner of the contract
-     */
-    function setGenerator(
```
ypatil12, line 91: Without this function, what would the recourse be if the `Generator` BLS keys are compromised? 

Would we just update the `BLSKey` in the generators table calculator?

src/contracts/multichain/OperatorTableUpdater.sol
```diff
@@ -38,42 +38,51 @@ contract OperatorTableUpdater is
...
+        _generator = generator;
         _setGlobalRootConfirmationThreshold(_globalRootConfirmationThreshold);
-        _updateGenerator(referenceTimestamp, generatorInfo, generatorConfig);
 
-        /// @dev The first global table root is the `INITIAL_GLOBAL_TABLE_ROOT`
-        /// @dev This is used to enable the first call to `confirmGlobalTableRoot` to pass since it expects
-        /// @dev the `Generator` to have a valid initial global table root
-        _globalTableRoots[referenceTimestamp] = INITIAL_GLOBAL_TABLE_ROOT;
-        _isRootValid[INITIAL_GLOBAL_TABLE_ROOT] = true;
+        // Calculate the initial global table root containing the generator as the only leaf
```
ypatil12, line 61: Can we put all of the below logic in the internal _setGenerator function?

PR #1549, opened 2025-07-16
fix: Multichain (Pt 2) fixes

nadir-akhtar:
**Motivation:**

Results of internal review of the contracts primarily involved with offchain/onchain interaction

**Modifications:**

* Add reentrancy guard to `OperatorTableUpdater`
* Add miscellaneous documentation surrounding design decisions
* Misc fixes

**Result:**

Cleaner and safer code


src/contracts/multichain/ECDSACertificateVerifier.sol
```diff
@@ -324,16 +329,22 @@ contract ECDSACertificateVerifier is Initializable, ECDSACertificateVerifierStor
     ) public view returns (uint256[] memory) {
         bytes32 operatorSetKey = operatorSet.key();
         require(_latestReferenceTimestamps[operatorSetKey] == referenceTimestamp, ReferenceTimestampDoesNotExist());
+
         uint256 operatorCount = _numOperators[operatorSetKey][referenceTimestamp];
-        require(operatorCount > 0, ReferenceTimestampDoesNotExist());
+        require(operatorCount > 0, OperatorCountZero());
```
ypatil12, line 334: We can actually remove this check given that we already have this check in line 331, still need to cache the count in line 333 though

ypatil12 (commented):
ONe comment

src/contracts/multichain/ECDSACertificateVerifier.sol
nadir-akhtar: One of the tests fails without the check for this so leaving it in -- updated that test accordingly to look for the correct revert error

src/contracts/interfaces/IBaseCertificateVerifier.sol
```diff
@@ -27,6 +27,8 @@ interface IBaseCertificateVerifierErrors {
     error VerificationFailed();
     /// @notice Thrown when the global table root is disabled
     error RootDisabled();
+    /// @notice Thrown when the operator count is zero
+    error OperatorCountZero();
```
ypatil12: This should be in `IECDSACerticficateVerifierErrors`, since it's not use in the `BN254CertificateVerifier`

ypatil12 (approved):

PR #1557, opened 2025-07-18
fix: submitResult checks

0xrajath:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/src/contracts/avs/task/TaskMailbox.sol
+++ b/src/contracts/avs/task/TaskMailbox.sol
@@ -194,6 +194,7 @@ contract TaskMailbox is
             task.executorOperatorSetTaskConfig.curveType,
             task.executorOperatorSetTaskConfig.consensus,
             executorOperatorSet,
+            task.creationTime,
             executorCert
         );
         require(isCertificateValid, CertificateVerificationFailed());
@@ -366,6 +367,7 @@ contract TaskMailbox is
      * @param curveType The curve type used for signature verification
      * @param consensus The consensus configuration
      * @param executorOperatorSet The executor operator set
+     * @param creationTime The creation time of the task
      * @param executorCert The executor certificate to verify
      * @return isCertificateValid Whether the certificate is valid
      */
@@ -373,6 +375,7 @@ contract TaskMailbox is
         IKeyRegistrarTypes.CurveType curveType,
         Consensus memory consensus,
         OperatorSet memory executorOperatorSet,
+        uint96 creationTime,
         bytes memory executorCert
     ) internal returns (bool isCertificateValid) {
         if (consensus.consensusType == ConsensusType.STAKE_PROPORTION_THRESHOLD) {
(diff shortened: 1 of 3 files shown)


src/contracts/avs/task/TaskMailbox.sol
```diff
@@ -387,7 +393,9 @@ contract TaskMailbox is
                 IBN254CertificateVerifierTypes.BN254Certificate memory bn254Cert =
                     abi.decode(executorCert, (IBN254CertificateVerifierTypes.BN254Certificate));
 
-                // Validate that the certificate has a non-empty signature
+                // Validate the certificate
+                require(bn254Cert.referenceTimestamp == creationTime.toUint32(), InvalidReferenceTimestamp());
```
ypatil12: Why do we need creationTime and referenceTimestamp?

src/contracts/avs/task/TaskMailbox.sol
```diff
@@ -387,7 +409,9 @@ contract TaskMailbox is
                 IBN254CertificateVerifierTypes.BN254Certificate memory bn254Cert =
                     abi.decode(executorCert, (IBN254CertificateVerifierTypes.BN254Certificate));
 
-                // Validate that the certificate has a non-empty signature
+                // Validate the certificate
+                require(bn254Cert.referenceTimestamp == operatorTableReferenceTimestamp, InvalidReferenceTimestamp());
```
ypatil12, line 413: Shouldn't we just get this from storage of the taskHash?

src/contracts/avs/task/TaskMailbox.sol
```diff
@@ -398,7 +422,15 @@ contract TaskMailbox is
                 IECDSACertificateVerifierTypes.ECDSACertificate memory ecdsaCert =
                     abi.decode(executorCert, (IECDSACertificateVerifierTypes.ECDSACertificate));
 
-                // Validate that the certificate has a non-empty signature
+                // Validate the certificate
+                require(ecdsaCert.referenceTimestamp == operatorTableReferenceTimestamp, InvalidReferenceTimestamp());
+                require(
+                    ecdsaCert.messageHash
+                        == IECDSACertificateVerifier(ECDSA_CERTIFICATE_VERIFIER).calculateCertificateDigest(
+                            ecdsaCert.referenceTimestamp, keccak256(result)
```
ypatil12: Why is `result` passed in as bytes and not a hash?

src/contracts/avs/task/TaskMailbox.sol
0xrajath: It is from storage. I've passed it into the internal function which is called from `submitResult`

src/contracts/avs/task/TaskMailbox.sol
0xrajath: Very good point.

0xrajath: Done in [31ac398](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1557/commits/31ac3985d880b4da59a03b43468846dcd49a930b)

src/contracts/avs/task/TaskMailbox.sol
0xrajath: Oh do you mean why are we passing in the referenceTimestamp as part of the cert from the aggregator? That's because the aggregator is passing in the entire cert (referenceTimestamp, messageHash, Sig).

ypatil12: Gotcha, makes sense

ypatil12 (approved):

PR #1569, opened 2025-07-28
fix(h4): add pagination

ypatil12:
**Motivation:**

`getActiveGenerationReservation` can have an OOG error if there are many gen reservations

**Modifications:**

Paginate the function by adding two view funcs:

- `getActiveGenerationReservationsByRange`
- `getActiveGenerationReservationCount`

**Result:**

DOS-resistant API


src/contracts/multichain/CrossChainRegistry.sol
```diff
@@ -288,6 +288,27 @@ contract CrossChainRegistry is
         return operatorSets;
     }
 
+    /// @inheritdoc ICrossChainRegistry
+    function getActiveGenerationReservationsByRange(
+        uint256 startIndex,
+        uint256 endIndex
+    ) external view returns (OperatorSet[] memory) {
+        require(startIndex <= endIndex, InvalidRange());
```
nadir-akhtar, line 299: I think we should make this a strict inequality (i.e. `startIndex < endIndex`). If these are equal, then `length` below will be 0, and the `for` loop will return nothing

nadir-akhtar: Update: will keep as-is and return nothing when indices are equal, as reverts can cause offchain issues

nadir-akhtar (approved):

PR #1571, opened 2025-07-28
fix: Release Manager internal review fixes

nadir-akhtar:
**Motivation:**

After an internal security review, a few helpful fixes were determined for code clarity and reduced complexity.

**Modifications:**

* Reduced complexity for publishing release
* More appropriate function visibility

**Result:**

Cleaner, safer code

### DIFF
--- a/src/contracts/core/ReleaseManager.sol
+++ b/src/contracts/core/ReleaseManager.sol
@@ -78,6 +78,8 @@ contract ReleaseManager is Initializable, ReleaseManagerStorage, PermissionContr
         OperatorSet memory operatorSet
     ) external view returns (uint256, Release memory) {
         Release[] storage releases = _operatorSetReleases[operatorSet.key()];
+        require(releases.length > 0, NoReleases());
+
         uint256 latestReleaseId = releases.length - 1;
         return (latestReleaseId, releases[latestReleaseId]);
     }
--- a/src/contracts/core/ReleaseManager.sol
+++ b/src/contracts/core/ReleaseManager.sol
@@ -40,12 +40,7 @@ contract ReleaseManager is Initializable, ReleaseManagerStorage, PermissionContr
         // New release id is the length of the array before this call.
         releaseId = releases.length;
         // Increment total releases for this operator set.
-        releases.push();
-        // Copy the release to storage.
-        for (uint256 i = 0; i < release.artifacts.length; ++i) {
-            releases[releaseId].artifacts.push(release.artifacts[i]);
-        }
-        releases[releaseId].upgradeByTime = release.upgradeByTime;
+        releases.push(release);
 
         emit ReleasePublished(operatorSet, releaseId, release);
     }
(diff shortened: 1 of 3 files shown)


src/contracts/core/ReleaseManager.sol
```diff
@@ -1,12 +1,11 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.27;
 
-import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
 import "../mixins/PermissionControllerMixin.sol";
 import "../mixins/SemVerMixin.sol";
 import "./ReleaseManagerStorage.sol";
 
-contract ReleaseManager is Initializable, ReleaseManagerStorage, PermissionControllerMixin, SemVerMixin {
```
0xrajath: Why are we removing the `Initializable` ? It's meant to be an upgradeable contract right?

nadir-akhtar: Reason for removing was due to its lack of use. Since this is intended to be a simple contract I anticipated we wouldn't need it going forward. That said, if we think we may need it in the future, we can keep it in

0xrajath: Yes it's meant to be upgradeable. We will have more complex features in the future.

src/contracts/core/ReleaseManager.sol
```diff
@@ -97,7 +91,7 @@ contract ReleaseManager is Initializable, ReleaseManagerStorage, PermissionContr
     }
 
     /// @inheritdoc IReleaseManager
-    function isValidRelease(OperatorSet memory operatorSet, uint256 releaseId) external view returns (bool) {
+    function isLatestReleaseId(OperatorSet memory operatorSet, uint256 releaseId) external view returns (bool) {
```
0xrajath: I would say keep it as `isValidRelease`, because we're using this function as a way to see if a given releaseId is valid. As of now it's 'valid' if it's the latest release (but could change in the future).

src/contracts/core/ReleaseManager.sol
```diff
@@ -40,12 +37,7 @@ contract ReleaseManager is Initializable, ReleaseManagerStorage, PermissionContr
         // New release id is the length of the array before this call.
         releaseId = releases.length;
         // Increment total releases for this operator set.
-        releases.push();
-        // Copy the release to storage.
-        for (uint256 i = 0; i < release.artifacts.length; ++i) {
-            releases[releaseId].artifacts.push(release.artifacts[i]);
-        }
-        releases[releaseId].upgradeByTime = release.upgradeByTime;
+        releases.push(release);
```
0xrajath, line 43: Just wanted to double check that this has the intended functionality as the previous code.

0xrajath: i.e adding the element to the end of the list.

nadir-akhtar: To my knowledge, as well as according to tests like `testFuzz_publishMultipleReleases` and `test_getRelease_afterMultiplePublishes` (which perform multiple releases and check each individual member of storage for correctness), functionality is identical.

src/contracts/core/ReleaseManager.sol
nadir-akhtar: Removed commit

src/contracts/core/ReleaseManager.sol
nadir-akhtar: Removed commit

PR #1575, opened 2025-07-29
fix(h-04): race condition

ypatil12:
**Motivation:**

- There is an offchain race condition where updating the table can cause the entire service to panic if it is frontrun by another tx 

**Modifications:**

- Return if the table has already been updated
- Clean up integration testing lib 

**Result:**

- Clear off chain responses


nadir-akhtar (commented):
Logic looks good, though unit tests need to be updated and there's an interface commit that would merit removal

nadir-akhtar (approved):

PR #1589, opened 2025-08-05
fix(L-03): add `hasActiveGenerationReservation`

0xClandestine:
**Motivation:**

Certora L-03: `CrossChainRegistry` getters do not revert if there's not an active generation reservation for the given operator set.

**Modifications:**

- Added `hasActiveGenerationReservation`.
- Added comments in other view methods to first call `hasActiveGenerationReservation`.

**Result:**

Integrators can validate view before calling.

### DIFF
--- a/src/contracts/interfaces/ICrossChainRegistry.sol
+++ b/src/contracts/interfaces/ICrossChainRegistry.sol
@@ -280,8 +280,8 @@ interface ICrossChainRegistry is ICrossChainRegistryErrors, ICrossChainRegistryE
 
     /**
      * @notice Gets the operatorTableCalculator for a given operatorSet
-     * @dev You may want to query `hasActiveGenerationReservation` before calling this method
      * @param operatorSet the operatorSet to get the operatorTableCalculator for
+     * @dev You should check if an operatorSet has an active generation reservation prior to calling this method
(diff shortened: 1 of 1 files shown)


src/contracts/interfaces/ICrossChainRegistry.sol
```diff
@@ -280,6 +290,7 @@ interface ICrossChainRegistry is ICrossChainRegistryErrors, ICrossChainRegistryE
 
     /**
      * @notice Gets the operatorSetConfig for a given operatorSet
+     * @dev You may want to query `hasActiveGenerationReservation` before calling this method.
```
ypatil12: Dev comment should come after params. Also, let's not use may, let's say "you should check if an operatorSet has an activeGenerationReservation prior to calling this method"

ypatil12 (commented):
1 comment

src/contracts/interfaces/ICrossChainRegistry.sol
0xClandestine: [a2e2fcd](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1589/commits/a2e2fcd8361784f0cc1044e5082c8193b7a3f365)

PR #1593, opened 2025-08-06
feat: compute registry

0xrajath:
**Motivation:**

We need a ComputeRegistry for AVSs to register their operator set to  EigenCompute so that the EigenCompute operator can register that to that AVS operator set and pull the latest release from ReleaseManager.

Additionally the AVS needs to sign a Terms of Service before they can register for EigenCompute

https://lucid.app/lucidchart/26f35348-cd23-47d4-b848-0f4f9335c719/edit?invitationId=inv_8dc26eb0-772d-4018-884f-f54d46d3ee0f&page=0_0#

**Modifications:**

* ComputeRegistry core contract, interface and storage. 
* Terms of Service is signed through a EIP-712 signature.
* Unit tests.
* Updated Bindings

**Result:**

ComputeRegistry

### DIFF
--- a/src/contracts/cloud/ComputeRegistry.sol
+++ b/src/contracts/cloud/ComputeRegistry.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.27;
 
 import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
+import "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
 
 import "../mixins/PermissionControllerMixin.sol";
 import "../mixins/SignatureUtilsMixin.sol";
@@ -14,7 +15,13 @@ import "./ComputeRegistryStorage.sol";
  * @notice This contract handles permissionless (de)registration of AVS operator sets to the EigenCompute Operator.
  * It enables AVSs to easily access managed operator infrastructure as part of EigenCloud for quick bootstrapping.
  */
-contract ComputeRegistry is Initializable, ComputeRegistryStorage, PermissionControllerMixin, SignatureUtilsMixin {
+contract ComputeRegistry is
+    Initializable,
+    OwnableUpgradeable,
+    ComputeRegistryStorage,
+    PermissionControllerMixin,
+    SignatureUtilsMixin
+{
     using OperatorSetLib for OperatorSet;
 
     /**
@@ -62,12 +69,13 @@ contract ComputeRegistry is Initializable, ComputeRegistryStorage, PermissionCon
 
     /**
      * @notice Initializes the contract
+     * @param _owner The owner of the contract
      * @param _tosHash The hash of the Terms of Service that AVS operators must sign
      */
-    function initialize(
-        bytes32 _tosHash
-    ) external initializer {
-        tosHash = _tosHash;
+    function initialize(address _owner, bytes32 _tosHash) external initializer {
+        __Ownable_init();
+        transferOwnership(_owner);
+        _setTosHash(_tosHash);
     }
 
     /**
@@ -119,6 +127,34 @@ contract ComputeRegistry is Initializable, ComputeRegistryStorage, PermissionCon
         isOperatorSetRegistered[operatorSetKey] = false;
 
         emit OperatorSetDeregistered(operatorSet);
+    }
+
+    /**
+     * @notice Updates the Terms of Service hash
+     * @param _tosHash The new Terms of Service hash
+     * @dev Only callable by the contract owner
+     */
+    function setTosHash(
+        bytes32 _tosHash
+    ) external onlyOwner {
+        _setTosHash(_tosHash);
+    }
+
+    /**
+     *
+     *                         INTERNAL FUNCTIONS
+     *
+     */
+
+    /**
+     * @dev Internal function to set the Terms of Service hash
+     * @param _tosHash The new Terms of Service hash
+     */
+    function _setTosHash(
+        bytes32 _tosHash
+    ) internal {
+        tosHash = _tosHash;
+        emit TosHashSet(_tosHash);
     }
 
     /**
--- a/src/contracts/cloud/ComputeRegistry.sol
+++ b/src/contracts/cloud/ComputeRegistry.sol
@@ -1,0 +1,103 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+
+import "../mixins/PermissionControllerMixin.sol";
+import "../mixins/SemVerMixin.sol";
+import "./ComputeRegistryStorage.sol";
+
+/**
+ * @title ComputeRegistry
+ * @author Layr Labs, Inc.
+ * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
+ * @notice This contract handles permissionless (de)registration of AVS operator sets to the EigenCompute Operator.
+ * It enables AVSs to easily access managed operator infrastructure as part of EigenCloud for quick bootstrapping.
+ */
+contract ComputeRegistry is Initializable, ComputeRegistryStorage, PermissionControllerMixin, SemVerMixin {
+    using OperatorSetLib for OperatorSet;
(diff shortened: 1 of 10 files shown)


src/contracts/cloud/ComputeRegistry.sol
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * @inheritdoc IComputeRegistry
+     */
+    function registerForCompute(
+        OperatorSet memory operatorSet,
+        bytes memory signature
+    ) external checkCanCall(operatorSet.avs) isValidOperatorSet(operatorSet) {
+        // Check if there is at least one release for the operator set
+        // The ReleaseManager will revert with `NoReleases()` if there are no releases for the operator set
+        RELEASE_MANAGER.getLatestRelease(operatorSet);
```
0xrajath, line 93: Thinking out loud if we need a `try-catch` here. I personally think we don't and this is cleaner with better tradeoffs around reverts and gas

src/contracts/cloud/ComputeRegistry.sol
```diff
@@ -0,0 +1,200 @@
...
+
+        // Verify the signature
+        _checkIsValidSignatureNow({
+            signer: msg.sender,
+            signableDigest: calculateTOSAgreementDigest(operatorSet, msg.sender),
+            signature: signature,
+            expiry: MAX_EXPIRY
```
0xrajath, line 100: Currently expiry is hardcoded for the signature to max expiry. There's no material need for the TOS signature to expire. But thinking out loud to check my assumptions.

src/contracts/cloud/ComputeRegistry.sol
bdchatham: What are the specific tradeoffs with that? Not clear to me why you would catch. 

As it stands, I would think we want to revert regardless of the reason if the getLatestRelease reverts/returns an error. Is it a concern of coupling in the case of upgrades to one might require changes to the other?

ypatil12: Agree, try-catch is cursed

src/contracts/cloud/ComputeRegistry.sol
```diff
@@ -0,0 +1,200 @@
...
+        });
+
+        // Check if already registered
+        bytes32 operatorSetKey = operatorSet.key();
+        require(!isOperatorSetRegistered[operatorSetKey], OperatorSetAlreadyRegistered());
+
+        // Register the operator set
+        isOperatorSetRegistered[operatorSetKey] = true;
+        _operatorSetTosSignature[operatorSetKey] =
```
ypatil12, line 105: What's the purpose of storing the sig?

src/contracts/cloud/ComputeRegistry.sol
```diff
@@ -0,0 +1,200 @@
...
+            TOSSignature({signer: msg.sender, tosHash: tosHash, signature: signature});
+
+        emit OperatorSetRegistered(operatorSet, msg.sender, tosHash, signature);
```
ypatil12: Do we also need to check that a key type has been set?

src/contracts/interfaces/IComputeRegistry.sol
```diff
@@ -0,0 +1,150 @@
...
+     *
+     *                         WRITE FUNCTIONS
+     *
+     */
+
+    /**
+     * @notice Registers an operator set for compute services
+     * @param operatorSet The operator set to register
+     * @param signature The EIP-712 signature of the Terms of Service
+     * @dev Requires the caller to have permission to call on behalf of the operatorSet.avs
```
ypatil12: We have a new format for natspec: essentially adding all reverts/events/documentation related items for clarity:

Eg: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1584

src/contracts/cloud/ComputeRegistry.sol
0xrajath: So that it's accessible later in storage. Just a signature for the Terms of Service.

ypatil12: What happens if the TOS change? Do they need to re-sign?

ypatil12: Only way to do so rn is deregister and register

bdchatham: Agree - seems like we're able to mutate the tos, but their signature over it isn't possible

src/contracts/cloud/ComputeRegistry.sol
```diff
@@ -0,0 +1,200 @@
...
+        _checkIsValidSignatureNow({
+            signer: msg.sender,
+            signableDigest: calculateTOSAgreementDigest(operatorSet, msg.sender),
+            signature: signature,
+            expiry: MAX_EXPIRY
+        });
+
+        // Check if already registered
+        bytes32 operatorSetKey = operatorSet.key();
+        require(!isOperatorSetRegistered[operatorSetKey], OperatorSetAlreadyRegistered());
```
bdchatham: nit: should we make this the very first operation of the method to fail immediately if they're already registered?

src/contracts/cloud/ComputeRegistry.sol
0xrajath: Well depends on the intended functionality here. Do we want to deregister everyone when the TOS are changed and force them to re-register?

src/contracts/cloud/ComputeRegistry.sol
0xrajath: @bdchatham It was more of Solidity specific question directed at Yash. Generally try-catch in Solidity is horrible compared to typical languages and it has weird behavior with regards to how gas is used due to the possible flow of control here. 

So generally we try to avoid it if we can. In this case I think it's fine to directly just make the external call to RMS and let it fail in that call if there are no releases. Just wanted to double check my thinking.

src/contracts/cloud/ComputeRegistry.sol
0xrajath: Resolved in [cf7a504](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1593/commits/cf7a504b868c2dd6fd206bd0e62c50e841d27cd6)

As discussed offline, the TOS is no longer mutable at the moment for the MVP. If we support mutability of the TOS in the future, we'll figure out a way for the AVS to resign the TOS.

src/contracts/cloud/ComputeRegistry.sol
0xrajath: Done in [79ff664](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1593/commits/79ff66481a669b2cdc9b6aedb846277362fc18b2)

src/contracts/cloud/ComputeRegistry.sol
0xrajath: Done in [355f24d](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1593/commits/355f24d3e979a4c7a4177435305ca6e2fbb376d0)

0xrajath: Also checking if there's a `CrossChainRegistry.generationReservation`

src/contracts/interfaces/IComputeRegistry.sol
0xrajath: Done in [1157799](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1593/commits/11577995a577b53776b20977f9d42b06715da941)

src/contracts/interfaces/IComputeRegistry.sol
```diff
@@ -0,0 +1,181 @@
...
+interface IComputeRegistryTypes {
+    /// @notice The Terms of Service signature
+    struct TOSSignature {
+        address signer;
+        bytes32 tosHash;
+        bytes signature;
+    }
+}
+
+interface IComputeRegistryErrors {
```
ypatil12, line 19: Let’s add the 4 byte error code too

ypatil12: See KeyRegistrar for eg.

0xrajath: Done in [a6539d2](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1593/commits/a6539d2577ae1eb8db3d547c5e45fe5867c750ac)

PR #1595, opened 2025-08-06
fix(L-02): use default zero hash in place valid 0 leaves

nadir-akhtar:
**Motivation:**

0 leaves are currently considered valid within the Merkle library. In the event of an uninitialized struct, this can lead to proofs unintentionally being valid. As such, a default non-zero hash is used to represent these placeholder leaves for padding.

**Modifications:**

* Added a new "MERKLE_ZERO_HASH" constant derived from `keccak256("MERKLE_ZERO_HASH")`
* Created a new internal helper function to consolidate logic

**Result:**

* 0 leaves no longer being valid for constructing proofs.

_Note: this change may impact offchain services that generate roots + proofs. This change should only be merged if offchain services are ready to integrate with this change to root & proof calculations. Closing this PR is a possible outcome as this is a non-essential fix._

### DIFF
--- a/docs/core/libraries/Merkle.md
+++ b/docs/core/libraries/Merkle.md
@@ -1,0 +1,313 @@
+## Merkle
+
+| File | Notes |
+| -------- | -------- |
+| [`Merkle.sol`](../../../src/contracts/libraries/Merkle.sol) | Core Merkle tree library |
+
+## Overview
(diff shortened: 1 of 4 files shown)


elhajin:
@claude  review that PR , and is it still relevant ?

PR #1597, opened 2025-08-06
fix(audit): Merkle library infos

nadir-akhtar:
**Motivation:**

As part of an audit for the Merkle library, several low and informational findings were found. This PR consolidates all informational findings to fix, as well as additional commits for findings discovered along the way, with a specific commit for each finding.

All Lows are separated into individual PRs into the [release-dev/merkle-audit-fixes](https://github.com/Layr-Labs/eigenlayer-contracts/tree/release-dev/merkle-audit-fixes) branch

**Modifications:**

* Completely uplifted natspec, matching current EigenLayer quality and standards
* Four new error codes:
  * `InvalidIndex` for indices outside of the Merkle tree's max index
  * `LeavesNotPowerOfTwo` for leaves (particularly for SHA256 functions) that don't match the expected length condition
  * `NoLeaves` for an empty `leaves` array
  * `NotEnoughLeaves` for leaves (particularly for SHA256 functions) that are not 2 or greater
* Explicit return in `processInclusionProofKeccak` for 0 length proofs to return the leaf
* Requirement in `processInclusionProof(Keccak|Sha256)` that the index is 0, i.e. the index was for a leaf within the tree
* Refactoring in `merkleize(Sha256|Keccak)` and `getProofKeccak` of logic for readability and performance
* Newly added `getProofSha256` to mirror `getProofKeccak` to reduce burden of producing roots offchain
* Newly added `isPowerOfTwo` helper function
* Newly added documentation file for the Merkle library

**Result:**

* Improved documentation/natspec
* Four new error codes
* Refactored and simplified code
* Newly added `getProofSha256` and `isPowerOfTwo` functions
* Updated and added unit tests

### DIFF
--- a/src/contracts/libraries/Merkle.sol
+++ b/src/contracts/libraries/Merkle.sol
@@ -26,7 +26,7 @@ library Merkle {
      * hash matches the root of the tree. The tree is built assuming `leaf` is
      * the 0 indexed `index`'th leaf from the bottom left of the tree.
      *
-     * Note this is for a Merkle tree using the keccak/sha3 hash function
+     * Note this is for a Merkle tree using the keccak256 hash function
      */
     function verifyInclusionKeccak(
         bytes memory proof,
@@ -46,7 +46,7 @@ library Merkle {
      *
      * _Available since v4.4._
      *
-     * Note this is for a Merkle tree using the keccak/sha3 hash function
+     * Note this is for a Merkle tree using the keccak256 hash function
      */
     function processInclusionProofKeccak(
         bytes memory proof,
--- a/src/contracts/libraries/Merkle.sol
+++ b/src/contracts/libraries/Merkle.sol
@@ -47,6 +47,7 @@ library Merkle {
      * @dev If the proof length is 0 then the leaf hash is returned.
      * @dev Reverts for:
      *      - InvalidProofLength: proof.length is not a multiple of 32.
+     *      - InvalidIndex: index is not 0 at conclusion of computation.
      *
      * Note this is for a Merkle tree using the keccak256 hash function
      */
@@ -81,6 +82,10 @@ library Merkle {
                 }
             }
         }
+
+        // Confirm proof was fully consumed by end of computation
+        require(index == 0, InvalidIndex());
+
         return computedHash;
     }
 
@@ -106,6 +111,9 @@ library Merkle {
      * from `leaf` using `proof`. A `proof` is valid if and only if the rebuilt
      * hash matches the root of the tree. The tree is built assuming `leaf` is
      * the 0 indexed `index`'th leaf from the bottom left of the tree.
+     * @dev Reverts for:
+     *      - InvalidProofLength: proof.length is not a multiple of 32.
+     *      - InvalidIndex: index is not 0 at conclusion of computation.
      *
      * Note this is for a Merkle tree using the sha256 hash function
      */
@@ -135,6 +143,10 @@ library Merkle {
                 }
             }
         }
+
+        // Confirm proof was fully consumed by end of computation
+        require(index == 0, InvalidIndex());
+
         return computedHash[0];
     }
 
--- a/src/contracts/libraries/Merkle.sol
+++ b/src/contracts/libraries/Merkle.sol
@@ -25,6 +25,10 @@ library Merkle {
     // @dev Error code: 0xf6558f51
     error LeavesNotPowerOfTwo();
 
+    // @notice Thrown when the provided leaves' length was 0.
+    // @dev Error code: 0xbaec3d9a
+    error NoLeaves();
+
     /**
      * @notice Verifies that a given leaf is included in a Merkle tree
      * @param proof The proof of inclusion for the leaf
@@ -167,11 +171,13 @@ library Merkle {
      * @param leaves the leaves of the Merkle tree
      * @return The computed Merkle root of the tree.
      * @dev Reverts for:
+     *      - NoLeaves: leaves.length is 0.
      *      - LeavesNotPowerOfTwo: leaves.length is not a power of two.
      */
     function merkleizeSha256(
         bytes32[] memory leaves
     ) internal pure returns (bytes32) {
+        require(leaves.length > 0, NoLeaves());
         require(isPowerOfTwo(leaves.length), LeavesNotPowerOfTwo());
 
         //there are half as many nodes in the layer above the leaves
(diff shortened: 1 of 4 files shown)


src/contracts/libraries/Merkle.sol
```diff
@@ -209,12 +209,10 @@ library Merkle {
         //while we haven't computed the root
         while (numNodesInLayer != 1) {
             //overwrite the first numNodesInLayer nodes in layer with the pairwise hashes of their children
-            for (uint256 i = 0; i < layer.length; i++) {
-                if (i == index) {
-                    uint256 siblingIndex = i + 1 - 2 * (i % 2);
-                    proof = abi.encodePacked(proof, layer[siblingIndex]);
-                    index /= 2;
-                }
+            if (index < layer.length) {
```
ypatil12: Note: the final implementation doesn't have this if statement, even though commit does

src/contracts/libraries/Merkle.sol
```diff
@@ -206,14 +207,14 @@ library Merkle {
...
+
         //while we haven't computed the root
         while (numNodesInLayer != 1) {
             //overwrite the first numNodesInLayer nodes in layer with the pairwise hashes of their children
-            if (index < layer.length) {
-                uint256 siblingIndex = index + 1 - 2 * (index % 2);
-                proof = abi.encodePacked(proof, layer[siblingIndex]);
-                index /= 2;
-            }
+            uint256 siblingIndex = index + 1 - 2 * (index % 2);
```
ypatil12: Okay, I see why we removed the if statement now, since we're guaranteed that index < layer.length for every loop iteration

src/contracts/libraries/Merkle.sol
```diff
@@ -132,13 +134,16 @@ library Merkle {
...
-     *  @dev A pre-condition to this function is that leaves.length is a power of two.  If not, the function will merkleize the inputs incorrectly.
+     * @param leaves the leaves of the merkle tree
+     * @return The computed Merkle root of the tree.
+     * @dev Reverts for:
+     *      - LeavesNotPowerOfTwo: leaves.length is not a power of two.
      */
     function merkleizeSha256(
         bytes32[] memory leaves
     ) internal pure returns (bytes32) {
+        require(isPowerOfTwo(leaves.length), LeavesNotPowerOfTwo());
```
ypatil12, line 187: Why do we revert here, but not in the keccak function?

ypatil12, line 187: We use this function in beacon chain library. Is the `VALIDATOR_FIELDS_LENGTH` always guaranteed to be a power of 2?

src/contracts/libraries/Merkle.sol
```diff
@@ -54,7 +55,12 @@ library Merkle {
         bytes32 leaf,
         uint256 index
     ) internal pure returns (bytes32) {
+        if (proof.length == 0) {
```
ypatil12, line 75: Why not add this to the merkle library too?

src/contracts/libraries/Merkle.sol
ypatil12: I-01

src/contracts/libraries/Merkle.sol
ypatil12: I-02

src/contracts/libraries/Merkle.sol
ypatil12: I-03

ypatil12: I-03

src/contracts/libraries/Merkle.sol
ypatil12: I-04

ypatil12 (commented):
I-09 LGTM

src/contracts/libraries/Merkle.sol
```diff
@@ -167,11 +171,13 @@ library Merkle {
...
      * @return The computed Merkle root of the tree.
      * @dev Reverts for:
+     *      - NoLeaves: leaves.length is 0.
      *      - LeavesNotPowerOfTwo: leaves.length is not a power of two.
      */
     function merkleizeSha256(
         bytes32[] memory leaves
     ) internal pure returns (bytes32) {
+        require(leaves.length > 0, NoLeaves());
         require(isPowerOfTwo(leaves.length), LeavesNotPowerOfTwo());
```
ypatil12, line 187: I-13 LGTM. Noting that we also cover `merkleizeSha256` panicking on  leaves.length == 1 given the power of 2 requirement

src/contracts/libraries/Merkle.sol
```diff
@@ -167,17 +172,18 @@ library Merkle {
...
-     *      - NoLeaves: leaves.length is 0.
+     *      - NotEnoughLeaves: leaves.length is less than 2.
      *      - LeavesNotPowerOfTwo: leaves.length is not a power of two.
+     * @dev Unlike the Keccak version, this function does not allow a single-leaf tree.
      */
     function merkleizeSha256(
         bytes32[] memory leaves
     ) internal pure returns (bytes32) {
-        require(leaves.length > 0, NoLeaves());
+        require(leaves.length > 1, NotEnoughLeaves());
```
ypatil12, line 186: Maybe we should just keep the old `>0` since we have the power of 2 requirement?

ypatil12 (commented):
I-14

ypatil12:
For each function should we add where (if at all) we use it elsewhere in the protocol. 

Example:

```solidity
@dev Used by `RewardsCoordinator.xyz`
```

ypatil12 (commented):
Isn't this L-03 too?

src/contracts/libraries/Merkle.sol
nadir-akhtar: Good eye! Correct, it was included for I-01 but removed down the line

src/contracts/libraries/Merkle.sol
nadir-akhtar: This was our existing behavior. Larger conversation here about whether we change the behavior of either function -- and what potential downstream impacts it'll have. 

Defaulting to maintaining the existing behavior before we verify that changes to this behavior are safely handled by other contracts + offchain services

nadir-akhtar: Prior, there was a comment saying that a precondition to this function is the assumption that the inputted leaves will be a power of 2, presumably pre-padded. As such, this requirement should be compatible with our existing use -- integration / unit tests seem to assert this as well.

src/contracts/libraries/Merkle.sol
nadir-akhtar: The Merkle library, or to the SHA256 variant?

In the SHA256 function, we explicitly disallow proof.length of 0 (since historical behavior has been to disallow single-leaf trees). Hence the discrepancy.

If we want to resolve this discrepancy, this would be part of the "larger conversation" for which we'd want another PR with dedicated conversation. For the infos, the intent is clarifying / refactoring existing behavior.

src/contracts/libraries/Merkle.sol
nadir-akhtar: The "power of 2" requirement would pass for `leaves.length == 1`, which (as it stands) we actively want to prevent. As such, a special conditional / error code is necessary.

src/contracts/libraries/Merkle.sol
ypatil12: Makes sense

src/contracts/libraries/Merkle.sol
ypatil12: Makes sense

ypatil12 (approved):
LGTM

PR #1598, opened 2025-08-08
fix: sp multichain audit fixes

ypatil12:
**Motivation:**

SP multichain audit fixes

**Modifications:**

- Middleware ref typo
- KeyRegistrar return doc

**Result:**

Correct(er) code


0xClandestine (approved):
Looks good, only nit would be to use a permalink for Layr-Labs/eigenlayer-middleware.

PR #1600, opened 2025-08-09
fix: Add empty array check in StrategyManager strategy removal

crStiv:
Prevents potential underflow when accessing array elements by checking for empty strategy lists.


ypatil12:
no tests

PR #1603, opened 2025-08-11
test: improve `Merkle` coverage

0xClandestine:
**Motivation:**

Recent audits haven't instilled confidence in `Merkle` and we want to fix that.

**Modifications:**

- Differential test against `Murky` (the most commonly used test implementation) for root and inclusion proofs.
- TODO: Differential test against `OZ` for inclusion proofs.

**Result:**

More confidence, more coverage.


src/test/unit/libraries/Merkle.t.sol
```diff
@@ -0,0 +1,90 @@
...
+    bytes[] proofs;
+
+    function test_verifyInclusion_ValidProof() public {
+        assertValidProofs();
+    }
+
+    function test_verifyInclusion_EmptyProofs() public {
+        proofs = new bytes[](proofs.length);
+        assertInvalidProofs();
+    }
```
0xClandestine: NOTE: Adding tests to `MerkleBaseTest` will automatically test both sha256 and keccak tree variants.

src/contracts/libraries/Merkle.sol
```diff
@@ -53,7 +53,7 @@ library Merkle {
         bytes32 leaf,
         uint256 index
     ) internal pure returns (bytes32) {
-        require(proof.length % 32 == 0, InvalidProofLength());
+        require(proof.length != 0 && proof.length % 32 == 0, InvalidProofLength());
```
nadir-akhtar: Let's not introduce breaking changes to the Merkle library at this time -- as mentioned before, will reexamine any breaking changes after a more thorough review of both onchain and offchain interactions

src/test/unit/libraries/Merkle.t.sol
```diff
@@ -0,0 +1,200 @@
...
+import "src/contracts/libraries/Merkle.sol";
+import "src/test/utils/Murky.sol";
+
+abstract contract MerkleBaseTest is Test, MurkyBase {
+    bytes32[] leaves; // The contents of the merkle tree (unsorted).
+    bytes32 root; // The root of the merkle tree.
+    bytes[] proofs; // The proofs for each leaf in the tree.
+
+    function setUp() public {
+        leaves = _genLeaves(vm.randomBool() ? 9 : 10);
```
nadir-akhtar: Instead of non-deterministically generated leaves for the whole test suite, let's pass this in as a fuzz value with locally instantiated leaves / proofs / roots per test. Perhaps we can have specific tests for testing even vs odd 

Moreover, let's use a random UINT from a larger range of values -- perhaps from 0 to ~65 (just above a power of 2) -- that way we can cover more territory

src/test/unit/libraries/Merkle.t.sol
```diff
@@ -0,0 +1,214 @@
...
+    /// @dev Takes in `seed` which should be provided as a fuzz input.
+    /// Ensures that RNG is deterministic, and tests are easily reproducible.
+    modifier rng(uint seed) {
+        _rng(seed);
+        _;
+    }
+
+    function _rng(uint seed) internal {
+        vm.setSeed(seed);
+        leaves = _genLeaves(vm.randomBool() ? MIN_LEAVES : MAX_LEAVES);
```
nadir-akhtar: Rather than randomly choosing between the min and the max exclusively, can we choose from a range of these values?

foundry.toml
```diff
@@ -80,6 +80,9 @@
     no_match_path = "script/**/*.sol"
     fs_permissions = [{ access = "read-write", path = "./"}]
 
+    # If enabled, allows internal expectRevert calls to be used in tests.
+    allow_internal_expect_revert = true
```
0xClandestine, line 84: This allows us to test library reverts without needing a harness.

src/contracts/libraries/Merkle.sol
0xClandestine: [e28b08a](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1603/commits/e28b08a40053c0b08148f46704b325e88fddbfea)

src/test/unit/libraries/Merkle.t.sol
0xClandestine: Changed leaf generation to be per test, odd and even still tested in a single test.

nadir-akhtar (commented):
Overall looks good -- just wanna get more intermediate coverage as well

PR #1615, opened 2025-08-26
fix: enforce ordering of nonsigners in bn254CV

eigenmikem:
**Motivation:**

Currently there is no ordering check for non-signers in the BN254CV, which can present some edge cases with duplicate non-signers.

**Modifications:**

Ordering check for non-signers.

**Result:**

Addresses minor issue brought up in audit.

### DIFF
--- a/src/contracts/interfaces/IBN254CertificateVerifier.sol
+++ b/src/contracts/interfaces/IBN254CertificateVerifier.sol
@@ -31,6 +31,7 @@ interface IBN254CertificateVerifierTypes is IOperatorTableCalculatorTypes {
      * @param signature the G1 signature of the message. The signature is over the signable digest, which is calculated by `calculateCertificateDigest`
      * @param apk the G2 aggregate public key
      * @param nonSignerWitnesses an array of witnesses of non-signing operators
+     * @dev Non-signer witnesses MUST be strictly increasing by `operatorIndex`
      * @dev The `referenceTimestamp` is used to key into the operatorSet's stake weights. It is NOT the timestamp at which the certificate was generated off-chain
      */
     struct BN254Certificate {
--- a/docs/multichain/destination/CertificateVerifier.md
+++ b/docs/multichain/destination/CertificateVerifier.md
@@ -427,6 +427,7 @@ The contract supports 3 verification patterns:
  * @param signature the G1 signature of the message. The signature is over the signable digest, which is calculated by `calculateCertificateDigest`
  * @param apk the G2 aggregate public key
  * @param nonSignerWitnesses an array of witnesses of non-signing operators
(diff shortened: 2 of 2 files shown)


src/contracts/interfaces/IBN254CertificateVerifier.sol
```diff
@@ -52,6 +52,11 @@ interface IBN254CertificateVerifierErrors {
     /// @dev Error code: 0x03f4a78e
     /// @dev We enforce that operator indices are within valid bounds to prevent out-of-bounds access in the merkle tree verification
     error InvalidOperatorIndex();
+
+    /// @notice thrown when the non-signer witnesses are not strictly increasing by operatorIndex
+    /// @dev Error code: 0xec6268b8
+    /// @dev We enforce strictly increasing order to prevent duplicates and ensure deterministic processing
+    error NonSignerIndicesNotSorted();
```
ypatil12, line 60: Let's add this to error reason in the natspec below and the docs

src/contracts/multichain/BN254CertificateVerifier.sol
```diff
@@ -202,10 +202,16 @@ contract BN254CertificateVerifier is
...
     ) internal returns (BN254.G1Point memory nonSignerApk) {
         nonSignerApk = BN254.G1Point(0, 0);
+        uint32 previousOperatorIndex = 0;
 
         for (uint256 i = 0; i < cert.nonSignerWitnesses.length; i++) {
             BN254OperatorInfoWitness memory witness = cert.nonSignerWitnesses[i];
 
+            if (i > 0) {
+                // Enforce strictly increasing order of non-signer operator indices
+                require(witness.operatorIndex > previousOperatorIndex, NonSignerIndicesNotSorted());
```
ypatil12, line 212: Looks good to me.

This assumes that the operatorSet that is transported does not have duplicate operators, which is a fine assumption

src/contracts/interfaces/IBN254CertificateVerifier.sol
```diff
@@ -52,6 +52,11 @@ interface IBN254CertificateVerifierErrors {
     /// @dev Error code: 0x03f4a78e
     /// @dev We enforce that operator indices are within valid bounds to prevent out-of-bounds access in the merkle tree verification
     error InvalidOperatorIndex();
+
+    /// @notice thrown when the non-signer witnesses are not strictly increasing by operatorIndex
+    /// @dev Error code: 0xec6268b8
+    /// @dev We enforce strictly increasing order to prevent duplicates and ensure deterministic processing
```
ypatil12, line 59: Can we also note that the nonsigners need to be sorted where we define the bn254 certificate struct. See ecdsa certificate struct where we explain

nadir-akhtar (approved):

PR #1624, opened 2025-09-03
feat(draft): preview `slashOperator` shares + underlying slashed

0xClandestine:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/src/contracts/core/AllocationManager.sol
+++ b/src/contracts/core/AllocationManager.sol
@@ -1032,6 +1032,7 @@ contract AllocationManager is
             _isOperatorRedistributable(operator, registeredSets) || _isOperatorRedistributable(operator, allocatedSets);
     }
 
+    /// @inheritdoc IAllocationManager
     function previewSlashOperatorShares(
         address operator,
         OperatorSet memory operatorSet,
@@ -1047,7 +1048,7 @@ contract AllocationManager is
         uint64 maxMagnitude = getMaxMagnitude(operator, strategy);
         uint64 slashedMagnitude = uint64(uint256(allocation.currentMagnitude).mulWadRoundUp(wadToSlash));
         uint64 newMaxMagnitude = maxMagnitude - slashedMagnitude;
-        
+
         shares = SlashingLib.calcSlashedAmount({
             operatorShares: delegation.operatorShares(operator, strategy),
             prevMaxMagnitude: maxMagnitude,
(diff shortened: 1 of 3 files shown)


src/contracts/core/AllocationManager.sol
```diff
@@ -1031,4 +1031,42 @@ contract AllocationManager is
         return
             _isOperatorRedistributable(operator, registeredSets) || _isOperatorRedistributable(operator, allocatedSets);
     }
+
+    /// @inheritdoc IAllocationManager
```
ypatil12, line 1035: How much space do we have left?

src/contracts/core/AllocationManager.sol
```diff
@@ -1031,4 +1031,42 @@ contract AllocationManager is
...
+    function previewSlashOperatorShares(
+        address operator,
+        OperatorSet memory operatorSet,
+        IStrategy strategy,
+        uint256 wadToSlash
+    ) public view returns (uint256 shares) {
+        // Fetch operator's allocation details.
+        Allocation memory allocation = getAllocation(operator, operatorSet, strategy);
```
ypatil12, line 1043: We should make an internal function that does this logic and is called by this function as well as `slashOperator`

src/contracts/core/AllocationManager.sol
```diff
@@ -1031,4 +1031,42 @@ contract AllocationManager is
...
+
+        // Skip if no slashable allocation.
+        if (allocation.currentMagnitude == 0) return 0;
+
+        uint64 maxMagnitude = getMaxMagnitude(operator, strategy);
+        uint64 slashedMagnitude = uint64(uint256(allocation.currentMagnitude).mulWadRoundUp(wadToSlash));
+        uint64 newMaxMagnitude = maxMagnitude - slashedMagnitude;
+
```
ypatil12, line 1051: Needs to also include shares that are being deallocated

src/contracts/core/AllocationManager.sol
0xClandestine: Agree, will sort that out once my fn is sound.

src/contracts/core/AllocationManager.sol
0xClandestine: -180 bytes rip

0xClandestine: will sort this out as well

PR #1629, opened 2025-09-17
fix: task replay directed at same operator set

0xrajath:
**Motivation:**

Fixing a bug where a random actor could permissionlessly call `TaskMailbox::submitResult` with an old cert+result combination (as long as the cert is not stale) but for a completely different task directed at the same operator set in the TaskMailbox. This could lead to DoS of the taskCreator. 

**Modifications:**

The `messageHash` of the cert now is `keccak256(abi.encode(taskHash, result))` instead of just `keccak256(result)`. This ensures that the messageHash is applicable only for a specific task and can't be replayed across tasks.

**Result:**

Fixes `SigmaPrime: ELHG-02` finding.

### DIFF
--- a/src/contracts/avs/task/TaskMailbox.sol
+++ b/src/contracts/avs/task/TaskMailbox.sol
@@ -208,7 +208,7 @@ contract TaskMailbox is
             task.executorOperatorSetTaskConfig.consensus,
             executorOperatorSet,
             task.operatorTableReferenceTimestamp,
-            keccak256(result),
+            getMessageHash(taskHash, result),
             executorCert
         );
 
@@ -406,15 +406,15 @@ contract TaskMailbox is
      * @notice Validates a BN254 certificate's basic requirements
      * @param cert The BN254 certificate to validate
      * @param operatorTableReferenceTimestamp The expected reference timestamp
-     * @param resultHash The expected message hash
+     * @param messageHash The expected message hash
      */
     function _validateBN254Certificate(
         IBN254CertificateVerifierTypes.BN254Certificate memory cert,
         uint32 operatorTableReferenceTimestamp,
-        bytes32 resultHash
+        bytes32 messageHash
     ) internal pure {
         require(cert.referenceTimestamp == operatorTableReferenceTimestamp, InvalidReferenceTimestamp());
-        require(cert.messageHash == resultHash, InvalidMessageHash());
+        require(cert.messageHash == messageHash, InvalidMessageHash());
         require(!(cert.signature.X == 0 && cert.signature.Y == 0), EmptyCertificateSignature());
     }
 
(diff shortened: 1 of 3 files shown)


src/test/unit/TaskMailboxUnit.t.sol
```diff
@@ -1578,6 +1584,82 @@ contract TaskMailboxUnitTests_submitResult is TaskMailboxUnitTests {
...
+        TaskParams memory taskParams2 = _createValidTaskParams();
+        vm.prank(creator);
+        bytes32 taskHash2 = taskMailbox.createTask(taskParams2);
+
+        vm.warp(block.timestamp + 1);
+
+        // Try to use the certificate from task1 for task2
+        // This creates a certificate with messageHash for task1 but tries to use it for task2
+        IBN254CertificateVerifierTypes.BN254Certificate memory replayedCert =
+            _createValidBN254CertificateForResult(taskHash, result, _getTaskReferenceTimestamp(taskHash2));
```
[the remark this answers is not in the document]
0xrajath: This is infact the actual attack. The attacker would use the reference timestamp of the new task to get past the referenceTimestamp checks.

src/test/unit/TaskMailboxUnit.t.sol
```diff
@@ -1578,6 +1584,82 @@ contract TaskMailboxUnitTests_submitResult is TaskMailboxUnitTests {
...
+        TaskParams memory taskParams2 = _createValidTaskParams();
+        vm.prank(creator);
+        bytes32 taskHash2 = taskMailbox.createTask(taskParams2);
+
+        vm.warp(block.timestamp + 1);
+
+        // Try to use certificate from task1 for task2
+        // This creates a certificate with messageHash for task1 but tries to use it for task2
+        IECDSACertificateVerifierTypes.ECDSACertificate memory replayedCert =
+            _createValidECDSACertificateForResult(taskHash1, result, _getTaskReferenceTimestamp(taskHash2));
```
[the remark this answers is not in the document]
0xrajath: This is infact the actual attack. The attacker would use the reference timestamp of the new task to get past the referenceTimestamp checks.

src/contracts/interfaces/ITaskMailbox.sol
```diff
@@ -427,6 +427,14 @@ interface ITaskMailbox is ITaskMailboxErrors, ITaskMailboxEvents {
         IECDSACertificateVerifierTypes.ECDSACertificate memory cert
     ) external pure returns (bytes memory);
 
+    /**
+     * @notice Gets the message hash for a task that needs to be signed by operators
+     * @param taskHash Unique identifier of the task
+     * @param result Task execution result data
+     * @return The message hash for the task that needs to be signed by operators
+     */
+    function getMessageHash(bytes32 taskHash, bytes memory result) external pure returns (bytes32);
```
bdchatham, line 436: Is there significance to the term `message`? Something like `getTaskResultHash` strikes me as more specific and clear

0xrajath: Yeah that's what the CertVerifier calls it. So keeping the same terminology.

src/contracts/avs/task/TaskMailbox.sol
```diff
@@ -586,4 +586,9 @@ contract TaskMailbox is
     ) external pure returns (bytes memory) {
         return abi.encode(cert);
     }
+
+    /// @inheritdoc ITaskMailbox
+    function getMessageHash(bytes32 taskHash, bytes memory result) public pure returns (bytes32) {
+        return keccak256(abi.encode(taskHash, result));
```
ypatil12, line 592: Have we validated we can replicate `abi.encode` in the offchain aggregator? We've used `abi.encodePacked` everywhere for leaves on trees. This should be fine, though, if the offchain code is just calling `getMessageHash`.

ypatil12: Also, the operators also sign over the `referenceTimestamp`- see `calculateCertificateDigest` in the certVerifier.

bdchatham: The current plan is to just have the off-chain code call getMessageHash instead of its current step to keccak the result. This is then passed into the bn254/ecdsa digest generation methods on respective CertVerifiers as the messageHash. 

I don't see any hidden complexity or issues with this but all ears if there are concerns I'm missing.

ypatil12: SGTM

PR #1638, opened 2025-09-25
Certora formal verification specifications and CI

aehyvari:
**Motivation:**

Add certora formal verification specifications and CI integration

**Modifications:**

This PR introduces Certora formal verification specifications and a suggestion for integrating the Certora prover to the CI using Certora's run action for github.  The CI integration follows the pre-existing configuration.

**Result:**

The new specifications will be run against the code as part of the CI.


.github/workflows/certora-prover.yml
```diff
@@ -49,93 +55,29 @@ jobs:
           # Use dev branch for scheduled runs, otherwise use the branch that triggered the workflow
           ref: ${{ github.event_name == 'schedule' && 'dev' || github.ref }}
```
elhajin, line 56: this repo doesn’t have a dev branch , we should use 'main' instead

PR #1643, opened 2025-09-30
draft: split `AllocationManager`

0xClandestine:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -22,7 +22,7 @@
         "forge-std/=lib/forge-std/src/"
     ]
     # Specifies the exact version of Solidity to use, overriding auto-detection.
-    solc_version = '0.8.27'
+    solc_version = '0.8.30'
     # If enabled, treats Solidity compiler warnings as errors, preventing artifact generation if warnings are present.
     deny_warnings = true 
     # If set to true, changes compilation pipeline to go through the new IR optimizer.
--- a/script/deploy/devnet/deploy_from_scratch.s.sol
+++ b/script/deploy/devnet/deploy_from_scratch.s.sol
@@ -235,7 +235,8 @@ contract DeployFromScratch is Script, Test {
             SEMVER
         );
 
-        strategyManagerImplementation = new StrategyManager(IAllocationManager(address(allocationManager)), delegation, eigenLayerPauserReg, SEMVER);
+        strategyManagerImplementation =
+            new StrategyManager(IAllocationManager(address(allocationManager)), delegation, eigenLayerPauserReg, SEMVER);
         avsDirectoryImplementation = new AVSDirectory(delegation, eigenLayerPauserReg, SEMVER);
         eigenPodManagerImplementation =
             new EigenPodManager(ethPOSDeposit, eigenPodBeacon, delegation, eigenLayerPauserReg, SEMVER);
--- a/script/deploy/devnet/deploy_from_scratch.s.sol
+++ b/script/deploy/devnet/deploy_from_scratch.s.sol
@@ -228,14 +228,14 @@ contract DeployFromScratch is Script, Test {
         delegationImplementation = new DelegationManager(
             strategyManager,
             eigenPodManager,
-            allocationManager,
+            IAllocationManager(address(allocationManager)),
             eigenLayerPauserReg,
             permissionController,
             MIN_WITHDRAWAL_DELAY,
             SEMVER
         );
 
-        strategyManagerImplementation = new StrategyManager(allocationManager, delegation, eigenLayerPauserReg, SEMVER);
+        strategyManagerImplementation = new StrategyManager(IAllocationManager(address(allocationManager)), delegation, eigenLayerPauserReg, SEMVER);
         avsDirectoryImplementation = new AVSDirectory(delegation, eigenLayerPauserReg, SEMVER);
         eigenPodManagerImplementation =
             new EigenPodManager(ethPOSDeposit, eigenPodBeacon, delegation, eigenLayerPauserReg, SEMVER);
@@ -243,7 +243,7 @@ contract DeployFromScratch is Script, Test {
             IRewardsCoordinatorTypes.RewardsCoordinatorConstructorParams(
                 delegation,
                 strategyManager,
-                allocationManager,
+                IAllocationManager(address(allocationManager)),
                 eigenLayerPauserReg,
                 permissionController,
                 REWARDS_COORDINATOR_CALCULATION_INTERVAL_SECONDS,
@@ -480,7 +480,7 @@ contract DeployFromScratch is Script, Test {
             "rewardsCoordinator: strategyManager address not set correctly"
         );
         require(
-            delegationContract.allocationManager() == allocationManager,
+            delegationContract.allocationManager() == IAllocationManager(address(allocationManager)),
             "delegationManager: allocationManager address not set correctly"
         );
         require(
--- a/script/deploy/local/deploy_from_scratch.slashing.s.sol
+++ b/script/deploy/local/deploy_from_scratch.slashing.s.sol
@@ -238,13 +238,13 @@ contract DeployFromScratch is Script, Test {
         delegationImplementation = new DelegationManager(
             strategyManager,
             eigenPodManager,
-            allocationManager,
+            IAllocationManager(address(allocationManager)),
             eigenLayerPauserReg,
             permissionController,
             MIN_WITHDRAWAL_DELAY,
             SEMVER
         );
-        strategyManagerImplementation = new StrategyManager(allocationManager, delegation, eigenLayerPauserReg, SEMVER);
+        strategyManagerImplementation = new StrategyManager(IAllocationManager(address(allocationManager)), delegation, eigenLayerPauserReg, SEMVER);
         avsDirectoryImplementation = new AVSDirectory(delegation, eigenLayerPauserReg, SEMVER);
         eigenPodManagerImplementation =
             new EigenPodManager(ethPOSDeposit, eigenPodBeacon, delegation, eigenLayerPauserReg, SEMVER);
@@ -252,7 +252,7 @@ contract DeployFromScratch is Script, Test {
             IRewardsCoordinatorTypes.RewardsCoordinatorConstructorParams(
                 delegation,
                 strategyManager,
-                allocationManager,
+                IAllocationManager(address(allocationManager)),
                 eigenLayerPauserReg,
                 permissionController,
                 REWARDS_COORDINATOR_CALCULATION_INTERVAL_SECONDS,
--- a/script/deploy/local/deploy_from_scratch.slashing.s.sol
+++ b/script/deploy/local/deploy_from_scratch.slashing.s.sol
@@ -244,7 +244,8 @@ contract DeployFromScratch is Script, Test {
             MIN_WITHDRAWAL_DELAY,
             SEMVER
         );
-        strategyManagerImplementation = new StrategyManager(IAllocationManager(address(allocationManager)), delegation, eigenLayerPauserReg, SEMVER);
+        strategyManagerImplementation =
+            new StrategyManager(IAllocationManager(address(allocationManager)), delegation, eigenLayerPauserReg, SEMVER);
         avsDirectoryImplementation = new AVSDirectory(delegation, eigenLayerPauserReg, SEMVER);
         eigenPodManagerImplementation =
             new EigenPodManager(ethPOSDeposit, eigenPodBeacon, delegation, eigenLayerPauserReg, SEMVER);
--- a/script/releases/Env.sol
+++ b/script/releases/Env.sol
@@ -206,14 +206,14 @@ library Env {
      */
     function allocationManager(
         DeployedProxy
-    ) internal view returns (AllocationManager) {
-        return AllocationManager(_deployedProxy(type(AllocationManager).name));
+    ) internal view returns (IAllocationManager) {
+        return IAllocationManager(_deployedProxy(type(AllocationManager).name));
     }
 
     function allocationManager(
         DeployedImpl
-    ) internal view returns (AllocationManager) {
-        return AllocationManager(_deployedImpl(type(AllocationManager).name));
+    ) internal view returns (IAllocationManager) {
+        return IAllocationManager(_deployedImpl(type(AllocationManager).name));
     }
 
     function avsDirectory(
--- a/script/tasks/complete_withdrawal_from_strategy.s.sol
+++ b/script/tasks/complete_withdrawal_from_strategy.s.sol
@@ -75,7 +75,7 @@ contract CompleteWithdrawFromStrategy is Script, Test {
         DepositScalingFactor memory dsf = DepositScalingFactor(dm.depositScalingFactor(msg.sender, strategies[0]));
 
         // Get TM for Operator in strategies
-        uint64[] memory maxMagnitudes = am.getMaxMagnitudesAtBlock(msg.sender, strategies, startBlock);
+        uint64[] memory maxMagnitudes = IAllocationManager(address(am)).getMaxMagnitudesAtBlock(msg.sender, strategies, startBlock);
         uint256 slashingFactor = _getSlashingFactor(em, msg.sender, strategies[0], maxMagnitudes[0]);
         uint256 sharesToWithdraw = dsf.calcWithdrawable(amount, slashingFactor);
 
--- a/script/tasks/complete_withdrawal_from_strategy.s.sol
+++ b/script/tasks/complete_withdrawal_from_strategy.s.sol
@@ -75,7 +75,8 @@ contract CompleteWithdrawFromStrategy is Script, Test {
         DepositScalingFactor memory dsf = DepositScalingFactor(dm.depositScalingFactor(msg.sender, strategies[0]));
 
         // Get TM for Operator in strategies
-        uint64[] memory maxMagnitudes = IAllocationManager(address(am)).getMaxMagnitudesAtBlock(msg.sender, strategies, startBlock);
+        uint64[] memory maxMagnitudes =
+            IAllocationManager(address(am)).getMaxMagnitudesAtBlock(msg.sender, strategies, startBlock);
         uint256 slashingFactor = _getSlashingFactor(em, msg.sender, strategies[0], maxMagnitudes[0]);
         uint256 sharesToWithdraw = dsf.calcWithdrawable(amount, slashingFactor);
 
--- a/script/utils/ExistingDeploymentParser.sol
+++ b/script/utils/ExistingDeploymentParser.sol
@@ -104,8 +104,8 @@ contract ExistingDeploymentParser is Script, Logger {
     UpgradeableBeacon public strategyBeacon;
 
     /// @dev AllocationManager
-    AllocationManager public allocationManager;
-    AllocationManager public allocationManagerImplementation;
+    IAllocationManager public allocationManager;
+    IAllocationManager public allocationManagerImplementation;
 
     /// @dev AVSDirectory
     AVSDirectory public avsDirectory;
@@ -230,9 +230,9 @@ contract ExistingDeploymentParser is Script, Logger {
         );
 
         // AllocationManager
-        allocationManager = AllocationManager(json.readAddress(".addresses.allocationManager"));
+        allocationManager = IAllocationManager(json.readAddress(".addresses.allocationManager"));
         allocationManagerImplementation =
-            AllocationManager(json.readAddress(".addresses.allocationManagerImplementation"));
+            IAllocationManager(json.readAddress(".addresses.allocationManagerImplementation"));
 
         // AVSDirectory
         avsDirectory = AVSDirectory(json.readAddress(".addresses.avsDirectory"));
--- a/src/contracts/core/AllocationManager.sol
+++ b/src/contracts/core/AllocationManager.sol
@@ -54,14 +54,14 @@ contract AllocationManager is
         _disableInitializers();
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function initialize(
         uint256 initialPausedStatus
     ) external initializer {
         _setPausedStatus(initialPausedStatus);
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function slashOperator(
         address avs,
         SlashingParams calldata params
@@ -75,7 +75,7 @@ contract AllocationManager is
         return _slashOperator(params, operatorSet);
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function modifyAllocations(
         address operator,
         AllocateParams[] memory params
@@ -167,7 +167,7 @@ contract AllocationManager is
         }
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function clearDeallocationQueue(
         address operator,
         IStrategy[] calldata strategies,
@@ -179,7 +179,7 @@ contract AllocationManager is
         }
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function registerForOperatorSets(
         address operator,
         RegisterParams calldata params
@@ -206,7 +206,7 @@ contract AllocationManager is
         getAVSRegistrar(params.avs).registerOperator(operator, params.avs, params.operatorSetIds, params.data);
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function deregisterFromOperatorSets(
         DeregisterParams calldata params
     ) external onlyWhenNotPaused(PAUSED_OPERATOR_SET_REGISTRATION_AND_DEREGISTRATION) {
@@ -236,7 +236,7 @@ contract AllocationManager is
         getAVSRegistrar(params.avs).deregisterOperator(params.operator, params.avs, params.operatorSetIds);
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function setAllocationDelay(address operator, uint32 delay) external {
         if (msg.sender != address(delegation)) {
             require(_checkCanCall(operator), InvalidCaller());
@@ -245,7 +245,7 @@ contract AllocationManager is
         _setAllocationDelay(operator, delay);
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function setAVSRegistrar(address avs, IAVSRegistrar registrar) external checkCanCall(avs) {
         // Check that the registrar is correctly configured to prevent an AVSRegistrar contract
         // from being used with the wrong AVS
@@ -254,13 +254,13 @@ contract AllocationManager is
         emit AVSRegistrarSet(avs, getAVSRegistrar(avs));
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function updateAVSMetadataURI(address avs, string calldata metadataURI) external checkCanCall(avs) {
         if (!_avsRegisteredMetadata[avs]) _avsRegisteredMetadata[avs] = true;
         emit AVSMetadataURIUpdated(avs, metadataURI);
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function createOperatorSets(address avs, CreateSetParams[] calldata params) external checkCanCall(avs) {
         require(_avsRegisteredMetadata[avs], NonexistentAVSMetadata());
         for (uint256 i = 0; i < params.length; i++) {
@@ -268,7 +268,7 @@ contract AllocationManager is
         }
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function createRedistributingOperatorSets(
         address avs,
         CreateSetParams[] calldata params,
@@ -284,7 +284,7 @@ contract AllocationManager is
         }
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function addStrategiesToOperatorSet(
         address avs,
         uint32 operatorSetId,
@@ -300,7 +300,7 @@ contract AllocationManager is
         }
     }
 
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerActions
     function removeStrategiesFromOperatorSet(
         address avs,
         uint32 operatorSetId,
@@ -699,7 +699,9 @@ contract AllocationManager is
      *
      */
 
-    /// @inheritdoc IAllocationManager
+    // TODO: Inherit doc
+    // TODO: Make commonly used getters internal methods that can be easily shared between Actions and View contracts.
+
     function getAllocation(
         address operator,
         OperatorSet memory operatorSet,
@@ -710,7 +712,6 @@ contract AllocationManager is
         return allocation;
     }
 
-    /// @inheritdoc IAllocationManager
     function getAllocationDelay(
         address operator
     ) public view returns (bool, uint32) {
@@ -728,7 +729,6 @@ contract AllocationManager is
         return (isSet, delay);
     }
 
-    /// @inheritdoc IAllocationManager
     function getAVSRegistrar(
         address avs
     ) public view returns (IAVSRegistrar) {
@@ -737,7 +737,6 @@ contract AllocationManager is
         return address(registrar) == address(0) ? IAVSRegistrar(avs) : registrar;
     }
 
-    /// @inheritdoc IAllocationManager
     function isOperatorSlashable(address operator, OperatorSet memory operatorSet) public view returns (bool) {
         RegistrationStatus memory status = registrationStatus[operator][operatorSet.key()];
 
@@ -746,7 +745,6 @@ contract AllocationManager is
         return status.registered || block.number <= status.slashableUntil;
     }
 
-    /// @inheritdoc IAllocationManager
     function getRedistributionRecipient(
         OperatorSet memory operatorSet
     ) public view returns (address) {
@@ -755,7 +753,6 @@ contract AllocationManager is
         return redistributionRecipient == address(0) ? DEFAULT_BURN_ADDRESS : redistributionRecipient;
     }
 
-    /// @inheritdoc IAllocationManager
     function isRedistributingOperatorSet(
         OperatorSet memory operatorSet
     ) public view returns (bool) {
--- a/src/contracts/core/AllocationManager.sol
+++ b/src/contracts/core/AllocationManager.sol
@@ -18,7 +18,8 @@ contract AllocationManager is
     AllocationManagerStorage,
     ReentrancyGuardUpgradeable,
     PermissionControllerMixin,
-    SemVerMixin
+    SemVerMixin,
+    IAllocationManagerActions
 {
     using DoubleEndedQueue for DoubleEndedQueue.Bytes32Deque;
     using Snapshots for Snapshots.DefaultWadHistory;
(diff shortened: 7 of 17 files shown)


src/contracts/interfaces/IExtsload.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/// @notice Interface for functions to access any storage slot in a contract
+interface IExtsload {
```
ypatil12: How exactly are we using extsload?

0xClandestine: left incase I ended up using, can remove

src/contracts/core/AllocationManager.sol
```diff
@@ -691,205 +649,110 @@ contract AllocationManager is
...
-    }
+
+    // TODO: If/Else instead of ternary
+    // TODO: Inherit doc
+    // TODO: Make commonly used getters internal methods that can be easily shared between Actions and View contracts.
 
     /**
      *
      *                         VIEW FUNCTIONS
      *
```
ypatil12, line 654: Maybe we have a separate section for all external view functions and internal view functions?

src/contracts/interfaces/IAllocationManager.sol
```diff
@@ -223,7 +225,37 @@ interface IAllocationManagerEvents is IAllocationManagerTypes {
     event StrategyRemovedFromOperatorSet(OperatorSet operatorSet, IStrategy strategy);
 }
 
-interface IAllocationManager is IAllocationManagerErrors, IAllocationManagerEvents, ISemVerMixin {
+interface IAllocationManagerStorage {
+    /**
+     * @notice The DelegationManager contract for EigenLayer
```
ypatil12, line 230: So immutables have to be redeclared. I guess that makes sense since storage contract isn't inherited by the implementation contract

src/contracts/interfaces/IExtsload.sol
```diff
@@ -0,0 +1,25 @@
...
+    /// @notice Called by external contracts to access granular pool state
+    /// @param slot Key of slot to sload
+    /// @return value The value of the slot as bytes32
+    function extsload(
+        bytes32 slot
```
ypatil12: We can remove this now, right?

src/contracts/mixins/SplitContractMixin.sol
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+abstract contract SplitContractMixin {
+    /// @notice The address of the second half of the contract.
+    address internal immutable secondHalf;
```
ypatil12: Wonder if we call this lens or view contract?

src/contracts/mixins/SplitContractMixin.sol
```diff
@@ -0,0 +1,54 @@
...
+            returndatacopy(0, 0, returndatasize())
+
+            switch result
+            // delegatecall returns 0 on error.
+            case 0 { revert(0, returndatasize()) }
+            default { return(0, returndatasize()) }
+        }
+    }
+
+    function _delegateView(
```
ypatil12, line 56: Can we add some comments to this function explaining how it works - just so others can refer to it. Kinda like the `_delegate` func

foundry.toml
```diff
@@ -55,7 +55,8 @@
     ]
     # An array of file paths from which warnings should be ignored during compilation.
     ignored_warnings_from = [
-        "src/test"
+        "src/test",
+        "src/contracts/core/AllocationManager.sol" # TODO: Remove
```
ypatil12: Reminder on TODO

foundry.toml
```diff
@@ -22,7 +22,7 @@
         "forge-std/=lib/forge-std/src/"
     ]
     # Specifies the exact version of Solidity to use, overriding auto-detection.
-    solc_version = '0.8.27'
+    solc_version = '0.8.30'
```
ypatil12, line 25: Where are we using namespaced storage for 0.8.30?

src/contracts/interfaces/IExtsload.sol
0xClandestine: removed in the other pr, will do tho

src/contracts/interfaces/IAllocationManager.sol
0xClandestine: yeah just needs the immutables really, call it "storage" since other contracts may have public variables

src/contracts/mixins/SplitContractMixin.sol
0xClandestine: `SplitViewMixin` and `viewImplementation` maybe, after adding the static-delegatecall changes it's name would be fitting.

foundry.toml
0xClandestine: See `AllocationManagerView` contract definition

src/contracts/mixins/SplitContractMixin.sol
0xClandestine: [a20f83a](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1643/commits/a20f83a4b3d62617f4ce20abf84ec17346548681)

foundry.toml
0xClandestine: [641ba47](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1643/commits/641ba475d439f9991131a58e02b1976edc256ebd)

src/contracts/mixins/SplitContractMixin.sol
0xClandestine: [99855a2](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1643/commits/99855a203f6b692d31e27a4f3b1226c47d626239)

src/contracts/core/AllocationManagerView.sol
```diff
@@ -0,0 +1,486 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import "./storage/AllocationManagerStorage.sol";
+
+/// @notice Non-state mutating view functions, (static) called by the `AllocationManager`.
+contract AllocationManagerView is // `AllocationManagerStorage` starts at slot 151.
+    AllocationManagerStorage, IAllocationManagerView layout at 151 {
```
ypatil12: How do we get 151 as where the layout starts?

src/test/DevnetLifecycle.t.sol
```diff
@@ -46,7 +46,7 @@ contract Devnet_Lifecycle_Test is Test, IAllocationManagerTypes {
         delegationManager = DelegationManager(0x3391eBafDD4b2e84Eeecf1711Ff9FC06EF9Ed182);
         strategyManager = StrategyManager(0x70f8bC2Da145b434de66114ac539c9756eF64fb3);
         avsDirectory = AVSDirectory(0xCa839541648D3e23137457b1Fd4A06bccEADD33a);
-        allocationManager = AllocationManager(0xAbD5Dd30CaEF8598d4EadFE7D45Fd582EDEade15);
+        allocationManager = IAllocationManager(0xAbD5Dd30CaEF8598d4EadFE7D45Fd582EDEade15);
         wethStrategy = StrategyBase(0x4f812633943022fA97cb0881683aAf9f318D5Caa);
         weth = IERC20(0x94373a4919B3240D86eA41593D5eBa789FEF3848);
```
ypatil12, line 51: For sanity, can we add an upgrade test that does the following:

1. Make a bunch of writes to the ALM
- Create AVS 
- Register opSet (add strategies)
- Slash operator
2. Do reads - registrar, strategies in opSets, getMembers, getAllocationDelay, allocatedMagnitude, encumbered magnitude, minimumSlashableStake, slashCount etc etc
3. Upgrade
4. Ensure that reads from 2 match reads post upgrade

src/contracts/mixins/SplitContractMixin.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+abstract contract SplitContractMixin {
```
ypatil12, line 4: Can we add a unit test file for this, just so we have coverage

src/contracts/core/AllocationManagerView.sol
ypatil12: Looks like thats where the ALM storage starts. We should comment this, and make note of it when writing docs

src/contracts/mixins/SplitContractMixin.sol
0xClandestine: [b5fa797](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1643/commits/b5fa797a521c3eb6829762ff43f3a0b6397985be)

src/contracts/core/AllocationManagerView.sol
0xClandestine: [ec8e978](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1643/commits/ec8e978ca828238e062acc3fbfa1b15154f3a88b)

src/test/DevnetLifecycle.t.sol
0xClandestine: added a basic test, also confirmed via forge inspect storageLayout

[684c758](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1643/commits/684c7586c53c6efda2993ba4b7c7377c89669154)

src/contracts/core/storage/AllocationManagerStorage.sol
```diff
@@ -64,7 +63,7 @@ abstract contract AllocationManagerStorage is IAllocationManager {
     /// OPERATOR => OPERATOR SET (REGISTRATION/DEREGISTRATION)
 
     /// @notice Returns the allocation delay info for each `operator`; the delay and whether or not it's previously been set.
-    mapping(address operator => AllocationDelayInfo) internal _allocationDelayInfo;
+    mapping(address operator => IAllocationManagerTypes.AllocationDelayInfo) internal _allocationDelayInfo;
```
eigenmikem: Can we just have `IAllocationManagerStorage` inherit from `IAllocationManagerTypes` in the interface file?

0xClandestine: 00e046e

src/test/integration/tests/ALM_Multi.t.sol
```diff
@@ -67,26 +67,27 @@ contract Integration_ALM_Multi is IntegrationCheckUtils {
...
-            _dispatchAllocated(i);
-            _dispatchFullyAllocated(i);
-            _dispatchRegAlloc(i);
-            _dispatchRegFullyAlloc(i);
-
-            // Ensure all pending actions are completed for the next iteration
-            _rollForward_DeallocationDelay();
-        }
-    }
+    // function test_Multi(uint24 _r) public rand(_r) {
```
ypatil12: What is the error here? Does it pass if we decrease number of iters?

src/test/integration/tests/upgrade/AllocationManagerUpgrade.t.sol
```diff
@@ -0,0 +1,48 @@
...
+        // Perform upgrade
+        console.log("Performing upgrade...".green());
+        _upgradeEigenLayerContracts();
+        console.log("Upgrade completed".green());
+
+        // Test basic functionality after upgrade
+        console.log("Post-upgrade: Testing basic functionality...".green());
+
+        // Verify the same values are still there
+        uint32 deallocationDelayAfter = allocationManager.DEALLOCATION_DELAY();
```
ypatil12: Can we also check actual values in storage in this test, not just immutable. So,

1. Create AVS, register, allocate, slash
2. Get registrar, members, allocated sets, magnitude
3. UPgrade
4. Do same calls in 2, validate

0xClandestine: oh rip deleted a bit too much

src/test/integration/tests/ALM_Multi.t.sol
0xClandestine: it fails first run every time

ypatil12: What's the error message? Is it due to something we've changed?

docs/core/AllocationManager.md
```diff
@@ -29,11 +30,51 @@ The `AllocationManager` manages AVS metadata registration, registration and dere
...
+
+**Main Contract (`AllocationManager`):**
+- Contains all state-mutating functions (actions)
+- Inherits from `SplitContractMixin` which provides delegation capabilities
+- Delegates all view function calls to the separate view contract
+- Maintains the same external interface as a monolithic contract
+
+**View Contract (`AllocationManagerView`):**
+- Contains all read-only view functions
+- Shares the same storage layout as the main contract using `layout at 151` directive
```
0xrajath, line 60: What's `layout at 151` directive?

script/deploy/devnet/deploy_from_scratch.s.sol
```diff
@@ -60,6 +61,8 @@ contract DeployFromScratch is Script, Test {
     StrategyBase public baseStrategyImplementation;
     AllocationManager public allocationManagerImplementation;
     AllocationManager public allocationManager;
+    AllocationManagerView public allocationManagerView;
+    AllocationManagerView public allocationManagerViewImplementation;
```
0xrajath: Is the view contract also upgradeable?

script/releases/Env.sol
```diff
@@ -206,14 +208,20 @@ library Env {
...
-    ) internal view returns (AllocationManager) {
-        return AllocationManager(_deployedImpl(type(AllocationManager).name));
+    ) internal view returns (IAllocationManager) {
+        return IAllocationManager(_deployedImpl(type(AllocationManager).name));
+    }
+
+    function allocationManagerView(
+        DeployedImpl
+    ) internal view returns (IAllocationManagerView) {
+        return IAllocationManagerView(_deployedImpl(type(AllocationManagerView).name));
```
0xrajath, line 224: I noticed there was a proxy for the view contract as well. Don't we need that in the Zeus env file as well?

script/releases/Env.sol
```diff
@@ -206,14 +208,20 @@ library Env {
      */
     function allocationManager(
         DeployedProxy
-    ) internal view returns (AllocationManager) {
-        return AllocationManager(_deployedProxy(type(AllocationManager).name));
+    ) internal view returns (IAllocationManager) {
```
0xrajath, line 211: Out of curiosity, is there any reason only these functions are returning `IAllocationManager` vs `AllocationManager`. All the other zeus env functions seem to be returning the contract itself. 

Doing this also affects the zeus script -> I can see you've had to cast appropriately in the zeus scripts.

script/tasks/complete_withdrawal_from_strategy.s.sol
```diff
@@ -75,7 +75,8 @@ contract CompleteWithdrawFromStrategy is Script, Test {
         DepositScalingFactor memory dsf = DepositScalingFactor(dm.depositScalingFactor(msg.sender, strategies[0]));
 
         // Get TM for Operator in strategies
-        uint64[] memory maxMagnitudes = am.getMaxMagnitudesAtBlock(msg.sender, strategies, startBlock);
+        uint64[] memory maxMagnitudes =
+            IAllocationManager(address(am)).getMaxMagnitudesAtBlock(msg.sender, strategies, startBlock);
```
0xrajath, line 79: Same here. What's the reason for the casting here?

script/utils/ExistingDeploymentParser.sol
```diff
@@ -230,9 +234,16 @@ contract ExistingDeploymentParser is Script, Logger {
...
         // AllocationManager
-        allocationManager = AllocationManager(json.readAddress(".addresses.allocationManager"));
+        allocationManager = IAllocationManager(json.readAddress(".addresses.allocationManager"));
         allocationManagerImplementation =
-            AllocationManager(json.readAddress(".addresses.allocationManagerImplementation"));
+            IAllocationManager(json.readAddress(".addresses.allocationManagerImplementation"));
+
+        // allocationManagerView = IAllocationManagerView(json.readAddress(".addresses.allocationManagerView"));
+
+        // FIXME: hotfix - remove later...
```
0xrajath, line 241: What's this?

src/contracts/core/storage/AllocationManagerStorage.sol
```diff
@@ -4,11 +4,10 @@ pragma solidity ^0.8.27;
...
 import "@openzeppelin/contracts/utils/structs/DoubleEndedQueue.sol";
 
-import "../interfaces/IAllocationManager.sol";
-import "../interfaces/IDelegationManager.sol";
-import {Snapshots} from "../libraries/Snapshots.sol";
+import "../../interfaces/IAllocationManager.sol";
+import {Snapshots} from "../../libraries/Snapshots.sol";
 
-abstract contract AllocationManagerStorage is IAllocationManager {
+abstract contract AllocationManagerStorage is IAllocationManagerStorage {
```
0xrajath, line 10: Why is it inherting `IAllocationManagerStorage` ?

src/contracts/mixins/SplitContractMixin.sol
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * @dev Delegates the current call to `implementation`.
+     *
+     * This function does not return to its internal call site, it will return directly to the external caller.
+     *
+     * Copied from OpenZeppelin Contracts v4.9.0 (proxy/Proxy.sol).
+     */
+    function _delegate(
```
0xrajath, line 21: Instead of us defining this here, should we use this from the OZ library directly (Mainly to prevent any mistakes on our end)? Thoughts?

I can see it's being used in `_delegateView` below

src/contracts/mixins/SplitContractMixin.sol
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * @dev Performs a delegate call to `implementation` in the context of a view function.
+     *
+     * This function typecasts the non-view `_delegate` function to a view function in order to
+     * allow its invocation from a view context. This is required because the EVM itself does not
+     * enforce view/pure mutability, and using inline assembly, it is possible to cast a function
+     * pointer to a view (read-only) signature. This pattern is sometimes used for readonly proxies,
+     * but it should be used cautiously since any state-modifying logic in the underlying delegate
```
0xrajath, line 51: If I understand this correctly, it will still allow state-modifying logic? Or am I misunderstanding it?

src/contracts/interfaces/IAllocationManager.sol
```diff
@@ -223,7 +225,37 @@ interface IAllocationManagerEvents is IAllocationManagerTypes {
     event StrategyRemovedFromOperatorSet(OperatorSet operatorSet, IStrategy strategy);
 }
 
-interface IAllocationManager is IAllocationManagerErrors, IAllocationManagerEvents, ISemVerMixin {
+interface IAllocationManagerStorage is IAllocationManagerEvents {
```
0xrajath, line 228: Why does this need to inherit `IAllocationManagerEvents`? Shouldn't types be sufficient?

src/contracts/core/AllocationManagerView.sol
```diff
@@ -0,0 +1,497 @@
...
+///
+///      1. Initializable
+///      2. Deprecated_OwnableUpgradeable
+///      3. Pausable
+///      4. AllocationManagerStorage
+///
+///      Since `AllocationManagerView` only needs access to the storage variables from
+///      `AllocationManagerStorage` (without the other mixins), it uses `layout at 151` to
+///      align its storage layout with the main `AllocationManager` contract.
+contract AllocationManagerView is IAllocationManagerView, AllocationManagerStorage layout at 151 {
```
0xrajath, line 19: Can we get some test cases to validate that the storage layout starts at 151. Also can we get some documentation around this in the core contract docs (i.e how you got the 151 number and what happens during an upgrade and if additional slots are used up)

src/contracts/interfaces/IAllocationManager.sol
0xrajath, line 1: Can we update the AllocationManager contract docs to include the inheritance chain diagram? I feel like i'm getting lost with the deep nested inheritance.

src/contracts/core/AllocationManager.sol
```diff
@@ -4,21 +4,24 @@ pragma solidity ^0.8.27;
...
-import "./AllocationManagerStorage.sol";
+import "./storage/AllocationManagerStorage.sol";
 
 contract AllocationManager is
     Initializable,
     Deprecated_OwnableUpgradeable,
     Pausable,
     AllocationManagerStorage,
     ReentrancyGuardUpgradeable,
+    SplitContractMixin,
```
0xrajath: Thoughts on adding this to the end of the inheritance to not mess with the inheritance chain? I know it has no storage slots - but still...

src/contracts/core/AllocationManager.sol
```diff
@@ -37,6 +40,7 @@ contract AllocationManager is
      * @dev Initializes the DelegationManager address, the deallocation delay, and the allocation configuration delay.
      */
     constructor(
+        IAllocationManagerView _allocationManagerView,
```
0xrajath, line 43: I noticed in the scripts that there was a proxy and implementation contract for the view functions. Is that the case? If yes, which value is passed here?

Also why do we need the view contract to be upgradeable?

src/contracts/core/AllocationManager.sol
```diff
@@ -690,206 +648,108 @@ contract AllocationManager is
...
-    /// @inheritdoc IAllocationManager
+    /// @inheritdoc IAllocationManagerView
     function getOperatorSetCount(
-        address avs
-    ) external view returns (uint256) {
-        return _operatorSets[avs].length();
+        address
+    ) external view returns (uint256 count) {
+        _delegateView(viewImplementation);
+        count;
```
0xrajath, line 662: Might be a dumb question, but how is that value being assigned to `count` here?

0xrajath: Or should it be `count = _delegateView(viewImplementation);` for clarity?

src/contracts/core/AllocationManagerView.sol
0xrajath: Also is `AllocationManagerStorage layout at 151` something that's enabled by Solidity 0.8.30 ? Seeing it for the first time.

script/deploy/local/deploy_from_scratch.slashing.s.sol
```diff
@@ -65,6 +66,8 @@ contract DeployFromScratch is Script, Test {
     StrategyBase public baseStrategyImplementation;
     AllocationManager public allocationManagerImplementation;
     AllocationManager public allocationManager;
+    AllocationManagerView public allocationManagerView;
+    AllocationManagerView public allocationManagerViewImplementation;
```
0xrajath: Why do we have a proxy and implementation for the view contract? My understanding was that it's going to be a non-upgradeable contract.

src/test/integration/users/AVS.t.sol
```diff
@@ -80,7 +80,7 @@ contract AVS is Logger, IAllocationManagerTypes, IAVSRegistrar {
         print.method("updateAVSMetadataURI");
 
         console.log("Setting AVS metadata URI to: %s", uri);
-        _tryPrankAppointee_AllocationManager(IAllocationManager.updateAVSMetadataURI.selector);
+        _tryPrankAppointee_AllocationManager(IAllocationManagerActions.updateAVSMetadataURI.selector);
```
0xrajath, line 83: Wouldn't `IAllocationManager` work here since it inherits `IAllocationManagerActions` ?

src/test/unit/SplitContractMixin.t.sol
0xrajath, line 1: Can we get some tests here to ensure that only view functions can be delegate called? And revert otherwise.

src/contracts/core/AllocationManager.sol
0xrajath: If `count = _delegateView(viewImplementation);` works, then we should do that style for all the other functions below. It's a lot more clear.

src/test/integration/tests/ALM_Multi.t.sol
0xClandestine: fixed

docs/core/AllocationManager.md
0xClandestine: Added in 0.8.30, Solidity typically assigns variables starting from 0, in the case of view we don't need `Initializable`, `Ownable`, or `Pausable`. Since each import takes 50 slots (gap) and pausable has an off-by-one error we've chosen to ignore in the past we start at 150. This is documented in `AllocationManager.md` if you wanna read more.

script/deploy/devnet/deploy_from_scratch.s.sol
0xClandestine: no removed, f79a513

script/releases/Env.sol
0xClandestine: See, above non-upgradeable.

script/releases/Env.sol
0xClandestine: `IAllocationManager` contains all ALM methods both logic + view; `AllocationManager` no longer contains view so throws a compile error.

script/tasks/complete_withdrawal_from_strategy.s.sol
0xClandestine: see above

script/utils/ExistingDeploymentParser.sol
0xClandestine: This is how we introduce new contracts into forktests (since they don't already exist onchain).

src/contracts/core/storage/AllocationManagerStorage.sol
0xClandestine: Since storage is inherited by both logic and view, but the remaining methods from `IAllocationManager` are not.

src/contracts/mixins/SplitContractMixin.sol
0xClandestine: `Proxy` contains logic we do not want, particularly the fallback/receive methods; Also I'm confident in correctness.

src/contracts/mixins/SplitContractMixin.sol
0xClandestine: Yes, opted against static-delegatecall after putting more thought into it and attempting to implement.

1) Created infinite loop (broke nested delegatecalls).
2) Has the same issues as enforcing view keyword (must enforce _staticDelegatecall).
3) Exposes a delegatecall method that **should** only be called the contract itself (opening other attack vectors).

src/contracts/core/AllocationManagerView.sol
0xClandestine: 1) Added integration fork test outlined by @ypatil12 in an above comment. Additionally, existing fork tests and the ci storage workflow cover this (ci explicitly compares local layout to mainnet). Oh and you can double-double check with `forge inspect AllocationManager storageLayout`.

0xClandestine: > Also is `AllocationManagerStorage layout at 151` something that's enabled by Solidity 0.8.30 ? Seeing it for the first time.

Yes, see above comment or `AllocationManager.md`.

src/contracts/core/AllocationManager.sol
0xClandestine: No preference here, doesn't matter imo. By definition our `mixins` should never have storage.

src/contracts/core/AllocationManager.sol
0xClandestine: View isn't upgradeable, see above response.

src/contracts/core/AllocationManager.sol
0xClandestine: When using return in a higher level (solidity) context the compiler intelligently decides whether to use a jump (to an internal or public method) or the return opcode. When using the return opcode in assembly it always bubbles up the returndata and since our methods defined in view and logic share the same signatures the bubbled up returndata sets `count` as expected.

0xClandestine: While `count = _delegateView(viewImplementation);` would be nice it would require a custom `_delegateView` implementation for each type we want to support and be error prone.

script/deploy/local/deploy_from_scratch.slashing.s.sol
0xClandestine: Removed, see above.

src/test/integration/users/AVS.t.sol
0xClandestine: When using `.function.selector` syntax you must use the interface the method is defined in (inheritance doesn't count for some reason).

src/test/unit/SplitContractMixin.t.sol
0xClandestine: Bit of an odd thing to test, since the compiler explicitly does this.

src/test/integration/tests/upgrade/AllocationManagerUpgrade.t.sol
```diff
@@ -0,0 +1,298 @@
...
+        console.log("Testing AllocationManager storage state capture...".green());
+
+        // Test that we can capture basic storage state
+        console.log("Capturing basic storage state...".green());
+        AllocationManagerState memory state = _captureBasicAllocationManagerState();
+
+        // Verify we can read the immutable variables
+        assertTrue(address(state.delegation) != address(0), "DelegationManager should be set");
+        // Note: eigenStrategy might be zero address on some networks/block numbers
+        console.log("EigenStrategy address:", address(state.eigenStrategy));
```
ypatil12: shouldn't we be calling `_upgradeEigenLayerContracts` somewhere?

0xClandestine: rip claude

src/contracts/core/AllocationManagerView.sol
nadir-akhtar: It appears to have been introduced with Solidity v0.8.29: https://docs.soliditylang.org/en/v0.8.29/contracts.html#custom-storage-layout

With that, we should update the version pragma of the contract to be `pragma solidity ^0.8.29` or greater.

docs/core/AllocationManager.md
```diff
@@ -29,11 +30,51 @@ The `AllocationManager` manages AVS metadata registration, registration and dere
...
+**Backwards Compatibility:**
+- The external interface remains identical to a monolithic contract
+- All existing integrations continue to work without modification
+- View functions are transparently delegated using `_delegateView()`
+- No breaking changes to the ABI or function signatures
+
+**Implementation Details:**
+- View functions in the main contract use `_delegateView(viewImplementation)` to delegate calls
+- The `viewImplementation` address is set during construction and stored as an immutable variable
+- The `_delegateView()` function uses assembly to safely cast the non-view `_delegate()` function to a view context
```
nadir-akhtar: I don't believe "safely" to be true -- having the `_delegateView` function is handy for stating which functions are _intended_ to be view functions, but it wouldn't actually enforce anything. It's more of a convenience method for informing users of _expected_, rather than guaranteed, behavior.

src/test/integration/tests/ALM_Multi.t.sol
```diff
@@ -67,6 +67,7 @@ contract Integration_ALM_Multi is IntegrationCheckUtils {
         }
     }
 
+    // TODO: FIX THIS OR REPLACE (SUPER FLAKEY)
```
nadir-akhtar, line 70: How has this test failed post-ALM changes?

src/contracts/core/AllocationManagerView.sol
```diff
@@ -0,0 +1,497 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
```
nadir-akhtar: The custom storage layout syntax below appears to have been introduced with Solidity v0.8.29: https://docs.soliditylang.org/en/v0.8.29/contracts.html#custom-storage-layout

With that, we should update the version pragma of the contract to the following, or greater:
```suggestion
pragma solidity ^0.8.29;
```

src/test/unit/AllocationManagerUnit.t.sol
```diff
@@ -91,13 +92,18 @@ contract AllocationManagerUnitTests is EigenLayerUnitTestSetup, IAllocationManag
 
     function _initializeAllocationManager(IPauserRegistry _pauserRegistry, uint _initialPausedStatus)
         internal
-        returns (AllocationManagerHarness)
+        returns (IAllocationManager)
     {
-        return allocationManager = AllocationManagerHarness(
+        // TODO: improve
```
nadir-akhtar: Is this TODO still relevant?

src/test/unit/SplitContractMixin.t.sol
nadir-akhtar: Tests for this would actually fail. If the `delegate` implements a non-view function, it can be called with `delegateView` successfully. `_delegateView` does not enforce that delegate calls do not modify state; it's more of a _suggestion_ that the underlying function doesn't modify state.

If you're curious, try plugging the below code block into `SplitContractMixin.t.sol` locally, and run `forge t SplitContractMixin.t.sol`. You'll see that `test_setValue` passes, indicating that `setValue` (which is a view function and uses `delegateView`) is able to execute a state-modifying operation.

```solidity
    // rest of SplitContractMixinTest above

    function setValue(uint) public view returns (uint result) {
        _delegateView(address(delegate));
        result;
    }

    function test_setValue() public {
        (bool success, bytes memory data) = address(this).call(abi.encodeWithSelector(this.setValue.selector, 100));
        assertTrue(success);

        (bool success, bytes memory data) = address(this).call(abi.encodeWithSelector(this.getValue.selector));
        assertTrue(success);
        uint result = abi.decode(data, (uint));
        assertEq(result, 100);
    }
}

// Mock contract to test delegation
contract Delegate is Test {
    uint value;

    function getValue() public view returns (uint result) {
        return value;
    }

    function setValue(uint _value) public {
        value = _value;
    }
}
```

src/test/utils/Logger.t.sol
```diff
@@ -52,25 +52,25 @@ abstract contract Logger is Test {
     bytes32 constant LOG_STATE_SLOT = bytes32(0);
 
     modifier noTracing() {
```
nadir-akhtar, line 54: What was the need to comment out the below lines?

nadir-akhtar (commented):
Split looks like a solid start! Especially with the `layout at <storage slot>` syntax making this super convenient.

A few notes on lingering TODOs, as well as some room for clarification in our docs on `_delegateView` allowing for delegatecall effects that may change state if the delegate contract implements such a function.

src/test/unit/SplitContractMixin.t.sol
0xClandestine: @nadir-akhtar We're talking about different things here, @0xrajath was suggesting adding tests to confirm that if we send calldata with a selector that's unknown to the view contract it reverts. I'm saying it's standard solidity behavior for a contract to revert if the functions unrecognized.

src/test/unit/AllocationManagerUnit.t.sol
0xClandestine: no will remove

docs/core/AllocationManager.md
0xClandestine: [b8bd499](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1643/commits/b8bd499df0923e2270db53912f186ecaf7457221)

src/contracts/core/AllocationManagerView.sol
0xClandestine: Nice catch [8037f8a](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1643/commits/8037f8a92b393d73ee6cde1d2b50e44171e40b56)

src/test/integration/tests/ALM_Multi.t.sol
0xClandestine: Blind revert on setup sometimes, happens outside of this pr as well

0xClandestine: Leaving comment for future improvement, test currently passing.

src/test/unit/AllocationManagerUnit.t.sol
0xClandestine: [272182a](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1643/commits/272182ac94eeab22a03b9ac2376c1ef8e9836184)

src/test/utils/Logger.t.sol
0xClandestine: They need reworking, was removed in my integration pr that never made it to main.

src/contracts/core/AllocationManager.sol
0xClandestine: [69f2688](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1643/commits/69f26880a8e4aeb71854c5a3dd8dffe1e84cc276)

src/contracts/interfaces/IAllocationManager.sol
0xClandestine: [4c161f0](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1643/commits/4c161f0df4c74011714ac0c9aedf3872a0b737da)

nadir-akhtar (approved):
Looks good! Much needed improvement to our contract codesize 💪

src/test/integration/tests/upgrade/AllocationManagerUpgrade.t.sol
```diff
@@ -0,0 +1,245 @@
...
+        operator.registerForOperatorSet(operatorSet);
+        check_Registration_State_NoAllocation(operator, operatorSet, allStrats);
+
+        // 6. Operator allocates to the operator set
+        allocateParams = _genAllocation_AllAvailable(operator, operatorSet);
+        operator.modifyAllocations(allocateParams);
+        check_IncrAlloc_State_Slashable(operator, allocateParams);
+
+        // 7. Roll blocks to complete allocation
+        _rollBlocksForCompleteAllocation(operator, operatorSet, strategies);
```
ypatil12, line 87: Can we also slash here just for completeness to write to every (I think?) slot

0xClandestine: will do in semver removal pr for sake of avoiding collecting approvals again

0xrajath: +1

PR #1645, opened 2025-10-01
feat: slashing commitments

ypatil12:
**Motivation:**

We want to solve the following use cases:
- As an AVS I want to make a commitment on the address that can slash a given operatorSet
- As an AVS I want to allow *one* address to slash an 
- As an AVS I DO NOT want to allow the admin to slash the operatorSet via `PermissionController` actions

**Modifications:**

- `updateSlasher`: can set a new slasher on a `DEALLOCATION_DELAY`
- `migrateSlasher`: migrates the slasher address from the `PermissionController` to the `AllocationManager`. Can be migrated only once for a given operatorSet
- `createOperatorSet`: modified to take in a slasher address, which can immediately slash an operatorSet

We are currently 1.3kB over. 

**Result:**

Stronger commitments.

### DIFF
--- a/src/contracts/core/AllocationManager.sol
+++ b/src/contracts/core/AllocationManager.sol
@@ -317,6 +317,44 @@ contract AllocationManager is
         }
     }
 
+    /// @inheritdoc IAllocationManager
+    function updateSlasher(OperatorSet memory operatorSet, address slasher) external checkCanCall(operatorSet.avs) {
+        require(_operatorSets[operatorSet.avs].contains(operatorSet.id), InvalidOperatorSet());
+        uint32 effectBlock = uint32(block.number) + DEALLOCATION_DELAY + 1;
+        _updateSlasher(operatorSet, slasher, effectBlock);
+    }
+
+    /// @inheritdoc IAllocationManager
+    function migrateSlasher(
+        OperatorSet[] memory operatorSets
+    ) external {
+        for (uint256 i = 0; i < operatorSets.length; i++) {
+            // Check that the operatorSet exists
+            require(_operatorSets[operatorSets[i].avs].contains(operatorSets[i].id), InvalidOperatorSet());
+
+            // Check that the operatorSet is not already migrated
+            require(!_slashers[operatorSets[i].key()].isMigrated, OperatorSetAlreadyMigrated());
+
+            // Get the slasher from the permission controller.
+            address[] memory slashers =
+                permissionController.getAppointees(operatorSets[i].avs, address(this), this.slashOperator.selector);
+
+            address slasher;
+            // If there are no slashers, set the slasher to the AVS
+            if (slashers.length == 0) {
+                slasher = operatorSets[i].avs;
+            } else {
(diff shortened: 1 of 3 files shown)


src/contracts/core/AllocationManager.sol
```diff
@@ -314,6 +317,44 @@ contract AllocationManager is
         }
     }
 
+    /// @inheritdoc IAllocationManager
+    function updateSlasher(OperatorSet memory operatorSet, address slasher) external checkCanCall(operatorSet.avs) {
+        require(_operatorSets[operatorSet.avs].contains(operatorSet.id), InvalidOperatorSet());
+        uint32 effectBlock = uint32(block.number) + DEALLOCATION_DELAY + 1;
```
nadir-akhtar: As mentioned on Slack: let's use the `ALLOCATION_CONFIGURATION_DELAY` of 17.5 days so that stakers have the opportunity (3.5 days) to react to a slasher change. Else, if the slasher can be updated within a `DEALLOCATION_DELAY`, then stakers will have no chance to reach, as even the earliest deallocation will be subjected to the new rules.

ypatil12: Updated: https://github.com/Layr-Labs/eigenlayer-contracts/blob/7df1c373b58462c576906b21196c4716a9bfb755/src/contracts/core/AllocationManager.sol#L773-L779

eigenmikem (approved):

docs/core/AllocationManager.md
```diff
@@ -18,6 +18,16 @@ Libraries and Mixins:
 | [`OperatorSetLib.sol`](../../src/contracts/libraries/OperatorSetLib.sol) | encode/decode operator sets |
 | [`Snapshots.sol`](../../src/contracts/libraries/Snapshots.sol) | historical state |
 
+### Function Deprecation Notice Q2 2026
+
+The `v1.9.0` slashing upgrade introduces a single slasher per operatorSet that is stored  in the `AllocationManager`. The below functions will be deprecated in Q2 2026.
```
nadir-akhtar:
```suggestion
The `v1.9.0` slashing upgrade introduces a single slasher per operatorSet that is stored in the `AllocationManager`. The functions in the leftmost column will be deprecated in Q2 2026 in favor of the newly specified function (if applicable). 
```

src/contracts/core/AllocationManager.sol
```diff
@@ -325,6 +344,48 @@ contract AllocationManager is
...
+                continue;
+            }
+
+            // Get the slasher from the permission controller.
+            address[] memory slashers =
+                permissionController.getAppointees(operatorSets[i].avs, address(this), this.slashOperator.selector);
+
+            address slasher;
+            // If there are no slashers or the first slasher is the 0 address, set the slasher to the AVS
+            if (slashers.length == 0 || slashers[0] == address(0)) {
```
nadir-akhtar, line 373: Good choice -- making sure that the address for the slasher can never be zero, if an AVS hypothetically set their slasher appointee to the zero address.

PR #1646, opened 2025-10-01
feat: instant alloc delay from dm

ypatil12:
**Motivation:**

Newly created operators in core have to wait 17.5 days to begin allocating.  This is a UX painpoint for new AVSs to onboard their operators. 

**Modifications:**

Update the `AllocationManager` so that the allocation delay for a newly created operator is active immediately block. This allows operators to make an allocation at the very next block. 

**Result:**

Easier onboarding


src/contracts/core/AllocationManager.sol
```diff
@@ -503,8 +507,9 @@ contract AllocationManager is
      * allocating magnitude to an operator set, and the magnitude becoming slashable.
      * @param operator The operator to set the delay on behalf of.
      * @param delay The allocation delay in blocks.
+     * @param allocationConfigurationDelay The delay after which the allocation delay takes effect. 0 if the operator is newly registered, otherwise ALLOCATION_CONFIGURATION_DELAY.
      */
-    function _setAllocationDelay(address operator, uint32 delay) internal {
+    function _setAllocationDelay(address operator, uint32 delay, uint32 allocationConfigurationDelay) internal {
```
nadir-akhtar: Rather than passing in an arbitrary `uint32`, which we expect to be constrained to either 0 or `ALLOCATION_CONFIGURATION_DELAY` based on whether this is an operator's first registration or subsequent registration respectively, I recommend instead using existing storage within this function to determine what `effectBlock` to set.

`info.effectBlock` is `0` before registration, and non-zero after registration. By checking within this function if `info.effectBlock == 0`, we can then set the new `effectBlock` accordingly: `uint32(block.number) + 1` if `true`, `uint32(block.number) + ALLOCATION_CONFIGURATION_DELAY + 1` if `false`. This constrains the function inputs to as few as possible, and allows the internal function to figure out the correct delay for the operator.

Side question: is having the delay activate on the block following this call a consequence of the implementation, or is there a safety or other concern meriting it?

src/test/integration/tests/upgrade/ZeroRegistrationDelay.t.sol
```diff
@@ -0,0 +1,48 @@
...
+        // The delay should be set since we are now with a 1 block set allocation delay
+        assertTrue(isSetOperator2, "isSet should be true");
+
+        // The delay for operator1 should still not be set
+        (isSetOperator1,) = allocationManager.getAllocationDelay(address(operator));
+        assertFalse(isSetOperator1, "isSet should be false");
+
+        // 4. Assert that operator1 is set after the allocation delay
+        // We subtract 1 since we've already rolled forward 2 blocks and the original delay is `config delay + 1`
+        rollForward({blocks: ALLOCATION_CONFIGURATION_DELAY - 1});
```
nadir-akhtar: For my understanding, where does the two-block forward roll happen? Is it within `_upgradeEigenLayerContracts()` + `_newRandomOperator_NoAssets()`?

src/contracts/core/AllocationManager.sol
ypatil12: Good points. I think we cannot do the `effectBlock == 0` check because we only want newly registered operators (in core) to be able to allocate after a single block. However, if we checked `effectBlock==0`, then both newly registered operators and operators that have previously registered, but not used slashing, could allocate instantly, which we don't want. Thoughts on passing in a bool instead? 

I kept +1 for consistency with the old implementation. Though, it remains to be seen if we want an immediate allocation (same block)

src/test/integration/tests/upgrade/ZeroRegistrationDelay.t.sol
ypatil12: Yeah, we have two calls to `_newRandomOperator_NoAssets()`. Which have the following code: https://github.com/Layr-Labs/eigenlayer-contracts/blob/e067fded718f349cbc739d9870e272d7cec35253/src/test/integration/IntegrationBase.t.sol#L104

docs/core/AllocationManager.md
```diff
@@ -847,15 +849,16 @@ function setAllocationDelay(
...
+This function sets an operator's allocation delay, in blocks. This delay can be updated by the operator once set. The initial setting of this value by a newly created operator via [`DelegationManager.registerAsOperator`](./DelegationManager.md#registerasoperator) will take 1 block to take effect. The setting of this value for an operator who has already been registered in core or further updates _take `ALLOCATION_CONFIGURATION_DELAY` blocks_ to take effect. Because having a delay is a requirement to allocating slashable stake, this effectively means that once the slashing release goes live, no already-created operators will be able to allocate slashable stake for at least `ALLOCATION_CONFIGURATION_DELAY` blocks.
```
eigenmikem, line 853: Might be worth adding a line or two about this into `DelegationManager.md` in case people are left with the impression that changing their allocation delay is instantaneous based on registering

src/contracts/core/AllocationManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1646/commits/d96438ba8af1325cd017aa739f0f074a00b348b1

docs/core/AllocationManager.md
ypatil12: Good callout. Updated: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1646/commits/ecc8376ce2eb421f125e5f41e604d4ab0b5645c1

src/contracts/core/AllocationManager.sol
nadir-akhtar: Good callout on already existing operators requiring a different flow! The boolean approach makes total sense.

There's a slight optimization you can do for codesize by removing the `newlyRegistered = false` statement (since it'll already be false by default). But other than that, looks solid.

ypatil12: True, will remove

ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1646/commits/3503884d86a8fe61adbfc59ce96bc5f52c31ae71

nadir-akhtar: Looked over it again, here's what I think the cleanest implementation is, for most clarity and least codesize:
```solidity
    function setAllocationDelay(address operator, uint32 delay) external {
        /// If the caller is the delegationManager, the operator is newly registered
        /// This results in *newly-registered* operators in the core protocol to have their allocation delay effective after 1 block
        bool newlyRegistered = (msg.sender != address(delegation));

        if (!newlyRegistered) {
            require(_checkCanCall(operator), InvalidCaller());
            require(delegation.isOperator(operator), InvalidOperator());
        }

        _setAllocationDelay(operator, delay, newlyRegistered);
    }
```

ypatil12: Cleaner! Updated: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1646/commits/3850113ce547de33aecff94311ad6cb906b346ca

src/contracts/core/AllocationManager.sol
```diff
@@ -514,7 +521,16 @@ contract AllocationManager is
         }
 
         info.pendingDelay = delay;
-        info.effectBlock = uint32(block.number) + ALLOCATION_CONFIGURATION_DELAY + 1;
+
+        /// If the caller is the delegationManager, the operator is newly registered
+        /// This results in *newly-registered* operators in the core protocol to have their allocation delay effective after 1 block
```
nadir-akhtar: Let's remove "after 1 block" or change it to "effective immediately"

src/contracts/core/AllocationManager.sol
```diff
@@ -237,11 +237,17 @@ contract AllocationManager is
 
     /// @inheritdoc IAllocationManager
     function setAllocationDelay(address operator, uint32 delay) external {
-        if (msg.sender != address(delegation)) {
+        /// If the caller is the delegationManager, the operator is newly registered
+        /// This results in *newly-registered* operators in the core protocol to have their allocation delay effective after 1 block
```
nadir-akhtar: Same here as the other comment about removing/changing "after 1 block"

src/contracts/core/AllocationManager.sol
```diff
@@ -237,11 +237,17 @@ contract AllocationManager is
...
+        bool newlyRegistered = (msg.sender == address(delegation));
```
nadir-akhtar, line 242: Was going to remark I had this boolean inverted in my comment -- good stuff fixing it 👍🏽

src/contracts/core/AllocationManager.sol
ypatil12: Ah, I knew I was missing something

src/contracts/core/AllocationManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1646/commits/fd12ef93ff15b8c9e14a164986e574d56deb2415

src/contracts/core/AllocationManager.sol
ypatil12: https://github.com/Layr-Labs/eigenlayer-contracts/pull/1646/commits/fd12ef93ff15b8c9e14a164986e574d56deb2415

PR #1750, opened 2026-04-28
feat: add slash resolution delay 

elhajin:
**Motivation:**

Burn and redistribution of slashed shares previously settled instantly once an AVS triggered a clear. This left no window to react before funds moved, which is risky for both stakers and AVSs. We want a built-in safety buffer plus a way to halt these flows in an emergency.

**Modifications:**

- Introduced a 7-day (`50,400` blocks) slash resolution delay before slashed shares can be burned or redistributed.
- Added a dedicated pause flag (`PAUSED_BURN_OR_REDISTRIBUTABLE_SHARES`) so burn/redistribution can be paused independently of other flows.
- Grandfathered pre-upgrade slashes: existing pending slashes have a zero resolution block and remain immediately clearable after the upgrade.
- Added unit, integration, and upgrade tests covering the new behavior.

**Result:**

- Slashed shares can no longer be burned or redistributed until the 7-day delay has elapsed.
- Pauser can pause burn/redistribution specifically without affecting unrelated functionality.


src/contracts/core/storage/StrategyManagerStorage.sol
nadir-akhtar: This gap needs to be decremented as we're adding a new state variable (the mapping `_slashResolutionBlock`) -- 32 -> 31

src/contracts/core/StrategyManager.sol
```diff
@@ -416,6 +421,7 @@ contract StrategyManager is
         if (remainingStrategies == 0) {
             // Remove the slash id from the pending slash ids.
```
nadir-akhtar, line 426: Let's either add a new comment or update this comment to specify that we're also clearing the `_slashResolutionBlock` since no strategies remain for this slash ID.

src/contracts/core/StrategyManager.sol
```diff
@@ -389,7 +392,9 @@ contract StrategyManager is
         uint256 slashId,
         IStrategy strategy,
         address recipient
-    ) internal returns (uint256) {
+    ) internal onlyWhenNotPaused(PAUSED_BURN_OR_REDISTRIBUTABLE_SHARES) returns (uint256) {
```
nadir-akhtar: To make this more explicit, let's try renaming this to `PAUSED_BURNING_AND_REDISTRIBUTION`, or `PAUSED_CLEARING_SHARES`?

nadir-akhtar (commented):
Small implementation overall! Tests complement the feature holistically from what I can tell.

Few small remarks. Also, bindings check seems to be failing. When comments are resolved, should be good to go.

nadir-akhtar:
@claude Want to provide any thoughts?

src/contracts/core/storage/StrategyManagerStorage.sol
elhajin: [1aa5720](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1750/changes/1aa5720ec226837c6cc1a0de92f5c4ba58039556)

src/contracts/core/StrategyManager.sol
elhajin: [1aa5720](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1750/changes/1aa5720ec226837c6cc1a0de92f5c4ba58039556)

src/contracts/core/StrategyManager.sol
elhajin: [1aa5720](https://github.com/Layr-Labs/eigenlayer-contracts/pull/1750/changes/1aa5720ec226837c6cc1a0de92f5c4ba58039556)

nadir-akhtar (dismissed):

